PR #46, opened 2019-11-26
Ban use of defparam

### LINKED ISSUE #12: Ban use of defparam
Our to-be-published style guide explicitly bans the use of the `defparam` construct.
This feature request aims to enforce just that.

**Assignment:**
Implement this check as a rule inside here:
https://github.com/google/verible/tree/master/verilog/analysis/checkers

Include unit-tests, positive and negative.

Add a TODO to enable by default here:
https://github.com/google/verible/blob/master/verilog/analysis/default_rules.h

Add an integration test here:
https://github.com/google/verible/blob/master/verilog/tools/lint/BUILD

**Suggestion:**
Use a syntax tree matcher, starts with 'Nodek', defined in https://github.com/google/verible/blob/master/verilog/CST/verilog_matchers.h

Questions? ask in verible-dev@groups.google.com!
https://groups.google.com/forum/#!forum/verible-dev

tgorochowik:
Fixes #12

### DIFF
--- a/verilog/analysis/default_rules.h
+++ b/verilog/analysis/default_rules.h
@@ -52,6 +52,7 @@ constexpr const char* kDefaultRuleSet[] = {
     // TODO(b/117554159): "case-missing-default",
     // TODO(b/117131903): "proper-parameter-declaration",
     // TODO(b/131637160): "signal-name-style",
+    // TODO(b/138351837): "forbid-defparam",
     // "endif-comment",
 };
 }  // namespace analysis
--- a/verilog/tools/lint/testdata/defparam_usage.sv
+++ b/verilog/tools/lint/testdata/defparam_usage.sv
@@ -1,0 +1,8 @@
+module defparam_usage #(parameter int MY_PARAM = 0);
+endmodule;
+
+module foo;
+  // defparam shouldn't be used, this should trigger the forbid-defparam rule
+  defparam p0.MY_PARAM = 1;
+  defparam_usage p0();
+endmodule
--- a/verilog/analysis/checkers/BUILD
+++ b/verilog/analysis/checkers/BUILD
@@ -22,6 +22,7 @@ cc_library(
         ":explicit_function_task_parameter_type_rule",
         ":explicit_parameter_storage_type_rule",
         ":explicit_task_lifetime_rule",
+        ":forbid_defparam_rule",
         ":forbidden_macro_rule",
         ":forbidden_symbol_rule",
         ":generate_label_rule",
@@ -176,6 +177,27 @@ cc_test(
         "//verilog/parser:verilog_token_enum",
         "@com_google_googletest//:gtest_main",
     ],
+)
+
+cc_library(
+    name = "forbid_defparam_rule",
+    srcs = ["forbid_defparam_rule.cc"],
+    hdrs = ["forbid_defparam_rule.h"],
+    deps = [
+        "//common/analysis:citation",
+        "//common/analysis:lint_rule_status",
+        "//common/analysis:syntax_tree_lint_rule",
+        "//common/analysis/matcher",
+        "//common/analysis/matcher:bound_symbol_manager",
+        "//common/analysis/matcher:matcher_builders",
+        "//common/text:symbol",
+        "//common/text:syntax_tree_context",
+        "//verilog/CST:verilog_matchers",  # fixdeps: keep
+        "//verilog/analysis:descriptions",
+        "//verilog/analysis:lint_rule_registry",
+        "@com_google_absl//absl/strings",
+    ],
+    alwayslink = 1,
 )
 
 cc_library(
--- a/verilog/analysis/checkers/BUILD
+++ b/verilog/analysis/checkers/BUILD
@@ -200,6 +200,20 @@ cc_library(
     alwayslink = 1,
 )
 
+cc_test(
+    name = "forbid_defparam_rule_test",
+    srcs = ["forbid_defparam_rule_test.cc"],
+    deps = [
+        ":forbid_defparam_rule",
+        "//common/analysis:linter_test_utils",
+        "//common/analysis:syntax_tree_linter_test_utils",
+        "//common/text:symbol",
+        "//verilog/CST:verilog_nonterminals",
+        "//verilog/analysis:verilog_analyzer",
+        "@com_google_googletest//:gtest_main",
+    ],
+)
+
 cc_library(
     name = "module_begin_block_rule",
     srcs = ["module_begin_block_rule.cc"],
--- a/verilog/analysis/checkers/forbid_defparam_rule.cc
+++ b/verilog/analysis/checkers/forbid_defparam_rule.cc
@@ -1,0 +1,68 @@
+// Copyright 2017-2019 The Verible Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "verilog/analysis/checkers/forbid_defparam_rule.h"
+
+#include <string>
+#include <vector>
+
+#include "absl/strings/str_cat.h"
+#include "common/analysis/citation.h"
+#include "common/analysis/lint_rule_status.h"
+#include "common/analysis/matcher/bound_symbol_manager.h"
+#include "common/analysis/matcher/matcher.h"
+#include "common/text/symbol.h"
+#include "common/text/syntax_tree_context.h"
+#include "verilog/analysis/descriptions.h"
+#include "verilog/analysis/lint_rule_registry.h"
+
+namespace verilog {
+namespace analysis {
+
+using verible::GetStyleGuideCitation;
+
+// Register the lint rule
+VERILOG_REGISTER_LINT_RULE(ForbidDefparamRule);
+
+absl::string_view ForbidDefparamRule::Name() { return "forbid-defparam"; }
+const char ForbidDefparamRule::kTopic[] = "defparam";
+const char ForbidDefparamRule::kMessage[] =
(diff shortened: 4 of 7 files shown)


verilog/analysis/checkers/defparam_rule.cc
```diff
@@ -0,0 +1,64 @@
...
+
+namespace verilog {
+namespace analysis {
+
+using verible::GetStyleGuideCitation;
+
+// Register the lint rule
+VERILOG_REGISTER_LINT_RULE(DefparamRule);
+
+absl::string_view DefparamRule::Name() { return "defparam"; }
```
fangism: Let's register this ::Name() as "forbid-defparam".

verilog/analysis/checkers/defparam_rule.cc
```diff
@@ -0,0 +1,64 @@
...
+  return absl::StrCat(
+      "Do not use defparam. See:",
+      GetStyleGuideCitation(kTopic), ".");
+}
+
+void DefparamRule::HandleSymbol(
+    const verible::Symbol& symbol, const verible::SyntaxTreeContext& context) {
+  verible::matcher::BoundSymbolManager manager;
+  if (matcher_.Matches(symbol, &manager)) {
+    violations_.push_back(verible::LintViolation(symbol, kMessage, context));
```
fangism: I recommend narrowing the range to that of the 'defparam' token itself, which should be in child position [0].

verilog/analysis/checkers/defparam_rule_test.cc
```diff
@@ -0,0 +1,51 @@
...
+namespace {
+
+using verible::LintTestCase;
+using verible::RunLintTestCases;
+
+TEST(DefparamRuleTests, Various) {
+  constexpr int kToken = SymbolIdentifier;
+  const std::initializer_list<LintTestCase> kTestCases = {
+      {""},
+      {"module parametrized; parameter int MY_PARAM = 0; endmodule; "
```
fangism: 'parameterized' (spelling)

verilog/analysis/checkers/defparam_rule_test.cc
```diff
@@ -0,0 +1,51 @@
...
+
+namespace verilog {
+namespace analysis {
+namespace {
+
+using verible::LintTestCase;
+using verible::RunLintTestCases;
+
+TEST(DefparamRuleTests, Various) {
+  constexpr int kToken = SymbolIdentifier;
```
fangism: With my suggestion to narrow down to the keyword, this should be the token enum TK_defparam.

verilog/analysis/checkers/defparam_rule_test.cc
```diff
@@ -0,0 +1,51 @@
...
+  const std::initializer_list<LintTestCase> kTestCases = {
+      {""},
+      {"module parametrized; parameter int MY_PARAM = 0; endmodule; "
+       "module foo; parametrized p0(); endmodule"},
+
+      {"module parametrized; parameter int MY_PARAM = 0; endmodule; "
+       "module foo; ",
+       "defparam ", {kToken, "p0"}, ".MY_PARAM = 1; "
```
fangism: With suggested change, this should expect to point to {kToken, "defparam"}.

verilog/analysis/default_rules.h
```diff
@@ -52,6 +52,7 @@ constexpr const char* kDefaultRuleSet[] = {
     // TODO(b/117554159): "case-missing-default",
     // TODO(b/117131903): "proper-parameter-declaration",
     // TODO(b/131637160): "signal-name-style",
+    // TODO(b/138351837): "defparam",
```
fangism: same as before "forbid-defparam"

verilog/analysis/checkers/defparam_rule.cc
tgorochowik: Thank you for the review!

Could you please elaborate on that? As far as I can see some other checkers try to use the BoundMap from the manager to extract more detailed info, but it is empty in this case. Whose child node should be used here, should it be extracted from the context? And what is the proper way to access it?

tgorochowik: `violations_.push_back(verible::LintViolation( *(context.top().children()[0].get()), kMessage, context));`

Seems to work, is that what you meant?

tgorochowik:
@googlebot I signed it!

verilog/analysis/checkers/defparam_rule.cc
tgorochowik: Sure! updated

verilog/analysis/checkers/defparam_rule.cc
tgorochowik: Changed the implementation to what I pasted above (+ some sanity checks), please let me know if that's what you had in mind

verilog/analysis/checkers/defparam_rule_test.cc
tgorochowik: Right, thanks!

verilog/analysis/checkers/defparam_rule_test.cc
tgorochowik: Updated

verilog/analysis/checkers/defparam_rule_test.cc
tgorochowik: Updated

verilog/analysis/default_rules.h
tgorochowik: Updated

tgorochowik:
@fangism thank you very much for your review! I've added the integration tests and addressed your inline comments.

One additional question: should the class (and files) for the rule also be renamed to ForbidDefparamRule?

verilog/tools/lint/testdata/defparam_usage.sv
```diff
@@ -0,0 +1,7 @@
+module defparam_usage #(parameter int MY_PARAM = 0);
+endmodule;
+
+module foo;
+  defparam p0.MY_PARAM = 1;
```
fangism: Just add a comment marking this line as the one expected to trigger a the lint rule "forbid-defparam".

verilog/analysis/checkers/defparam_rule.h
```diff
@@ -0,0 +1,65 @@
...
+  static std::string GetDescription(DescriptionType);
+
+  void HandleSymbol(const verible::Symbol& symbol,
+                    const verible::SyntaxTreeContext& context) override;
+
+  verible::LintRuleStatus Report() const override;
+
+ private:
+  // Matches the defparam construct.
+  const verible::matcher::Matcher matcher_ = NodekDefParamAssignList();
```
fangism: I overlooked this before, I would recommend matching on the CST node created here:

https://github.com/google/verible/blob/f74e900dbdd40c48fcaf1ff68db0c7f5bd342658/verilog/parser/verilog.y#L5286

so the node tagged `NodeEnum::kParameterOverride`.

verilog/analysis/checkers/defparam_rule.cc
```diff
@@ -0,0 +1,68 @@
...
+      "Do not use defparam. See:",
+      GetStyleGuideCitation(kTopic), ".");
+}
+
+void DefparamRule::HandleSymbol(
+    const verible::Symbol& symbol, const verible::SyntaxTreeContext& context) {
+  verible::matcher::BoundSymbolManager manager;
+  if (matcher_.Matches(symbol, &manager)) {
+    if (!context.empty() && !context.top().children().empty()) {
+      auto defparam = context.top().children()[0].get();
```
fangism: (See comment in .h file first on suggested node matcher.)
Take a look at GetSubtreeAsLeaf:
https://github.com/google/verible/blob/f74e900dbdd40c48fcaf1ff68db0c7f5bd342658/common/text/tree_utils.h#L102
With the update to the matcher, symbol will match at the node tagged `kParameterOverride`, whose child[0] will point to the `defparam` keyword leaf.  Then you don't need to manipulate the `context`.

```
const auto& defparam_token = GetSubtreeAsLeaf(symbol, NodeEnum::kParameterOverride, 0).get();
CHECK_EQ(defparam_token.token_enum, TK_defparam);
violations_.push_back(verible::LintViolation(defparam_token, kMessage, context));
```

verilog/analysis/checkers/defparam_rule_test.cc
```diff
@@ -0,0 +1,51 @@
...
+TEST(DefparamRuleTests, Various) {
+  constexpr int kToken = TK_defparam;
+  const std::initializer_list<LintTestCase> kTestCases = {
+      {""},
+      {"module parameterized #(parameter int MY_PARAM = 0); endmodule; "
+       "module foo; parameterized p0(); endmodule"},
+
+      {"module parameterized #(parameter int MY_PARAM = 0); endmodule; "
+       "module foo; ",
+       {kToken, "defparam"}, " p0.MY_PARAM = 1; "
```
fangism: With the above changes, I expect this test will now match the expected `{kToken, "defparam"}` violation.

verilog/analysis/checkers/defparam_rule_test.cc
```diff
@@ -0,0 +1,51 @@
...
+       "parameterized p0(); endmodule"},
+  };
```
fangism: Some suggestions for additional test cases:
* a test case with two violations (could be in the same module, or in different modules)
* a test case where the "defparam" is a substring of a (non-keyword) identifier, e.g. `wire not_defparam;`
* a test case where "defparam" is buried inside a /* comment */ or // comment.

verilog/tools/lint/testdata/defparam_usage.sv
tgorochowik: Sure! Added a comment above, like other tests do

verilog/analysis/checkers/defparam_rule.h
tgorochowik: Ok, sure, that makes sense

verilog/analysis/checkers/defparam_rule.cc
tgorochowik: Right, that looks much cleaner this way, thanks for the detailed explanation!

verilog/analysis/checkers/defparam_rule_test.cc
tgorochowik: Sure, added the suggested test cases

tgorochowik:
@fangism Thank you again for the review. I just pushed an update which renames the files and addresses all your comments

PR #59, opened 2019-11-28
Need space between : and ; (case null statement)

### LINKED ISSUE #18: Need space between ':' and ';' (default case null statement)
The following input:

```
function f;
case (x)
default :;  // this line, note spacing
endcase
endfunction
```

should really be formatted to:

```
function f;
  case (x)
    default: ;  // this line, note spacing
  endcase
endfunction
```

There are two spacing issues, this one is about spacing between ':' and ';'.

tgorochowik:
#18 says that the space is needed only for the default statement, but I guess it should be applied for each null statement in case, is that correct?

So with this change the following:
```
function f;
  case (x)
    1:;
    default :;
  endcase
endfunction
```

Is formatted to:

```
function f;
  case (x)
    1: ; // <- the space is added here too
    default : ;
  endcase
endfunction
```

Fixes #18

### DIFF
--- a/verilog/formatting/token_annotator.cc
+++ b/verilog/formatting/token_annotator.cc
@@ -113,6 +113,10 @@ static WithReason<int> SpacesRequiredBetween(const PreFormatToken& left,
   if (right.TokenEnum() == ',') return {0, "No space before comma"};
   if (left.TokenEnum() == ',') return {1, "Require space after comma"};
 
+  if (left.TokenEnum() == ':' && right.TokenEnum() == ';') {
+    return {1, "Require space after null statement in case statement"};
+  }
+
   if (right.TokenEnum() == ';') return {0, "No space before semicolon"};
   if (left.TokenEnum() == ';') return {1, "Require space after semicolon"};
 


fangism:
Correct.  Sorry, I forgot to mark this as having already been assigned/fixed, but the commits haven't been pushed yet.  I'll review and update other issues.

verilog/formatting/token_annotator.cc
```diff
@@ -113,6 +113,10 @@ static WithReason<int> SpacesRequiredBetween(const PreFormatToken& left,
   if (right.TokenEnum() == ',') return {0, "No space before comma"};
   if (left.TokenEnum() == ',') return {1, "Require space after comma"};
 
+  if (left.TokenEnum() == ':' && right.TokenEnum() == ';') {
+    return {1, "Require space after null statement in case statement"};
+  }
+
```
fangism: Exactly.  The fix I made encoded this a little differently, but this would be perfectly acceptable.

fangism:
Fixed by 
https://github.com/google/verible/commit/507a49a53b62d8dd797f3579d5f39424bff3c2b3

PR #60, opened 2019-11-29
& binary operator needs space on both sides

### LINKED ISSUE #37: [space] & binary operator needs space on both sides
should:
```systemverilog
a = b & ~(c << d);   
```
is:
```systemverilog
a = b &~(c << d);    // missing space after &
```

tgorochowik:
Fixes #37.

I've also removed the message about not accidentally joining `-` and `-` into `--` as it has been solved by the same change.

I am not sure about `-` and `=` though - could you please give an example of where this would be legal (other than the actual `-=`);

### DIFF
--- a/verilog/formatting/formatter_test.cc
+++ b/verilog/formatting/formatter_test.cc
@@ -370,6 +370,44 @@ static const std::initializer_list<FormatterTestCase> kFormatterTestCases = {
         "  parameter int a=--b- --c;",
         "parameter int a = --b - --c;\n"
     },
+    // ^~ and ~^ are bitwise nor, but ^ ~ isn't
+    {
+        "  parameter int a=b^~(c<<d);",
+        "parameter int a = b ^~ (c << d);\n"
+    },
+    {
+        "  parameter int a=b~^(c<<d);",
+        "parameter int a = b ~^ (c << d);\n"
+    },
+    {
+        "  parameter int a=b^ ~ (c<<d);",
+        "parameter int a = b ^ ~(c << d);\n"
+    },
+    {
+        "  parameter int a=b ^ ~(c<<d);",
+        "parameter int a = b ^ ~(c << d);\n"
+    },
+    // ~| is unary reduction NOR, |~ and | ~ aren't
+    {
+        "  parameter int a=b| ~(c<<d);",
+        "parameter int a = b | ~(c << d);\n"
+    },
+    {
+        "  parameter int a=b|~(c<<d);",
+        "parameter int a = b | ~(c << d);\n"
+    },
+    {
+        "  parameter int a=b| ~| ( c<<d);",
+        "parameter int a = b | ~|(c << d);\n"
+    },
+    {
+        "  parameter int a=b| ~| ~| ( c<<d);",
+        "parameter int a = b | ~|~|(c << d);\n"
+    },
+    {
+        "  parameter int a=b| ~~~( c<<d);",
+        "parameter int a = b | ~~~(c << d);\n"
+    },
     {
         "  parameter  int   foo=- - 1 ;",   // double negative
         "parameter int foo = - -1;\n",
--- a/verilog/formatting/formatter_test.cc
+++ b/verilog/formatting/formatter_test.cc
@@ -342,11 +342,38 @@ static const std::initializer_list<FormatterTestCase> kFormatterTestCases = {
         "  parameter  int   foo=$bar(-  z,- y ) ;",
         "parameter int foo = $bar(-z, -y);\n",
     },
-    /* TODO(b/143739545): prevent token joining
-    {"  parameter  int   foo=- - 1 ;",   // double negative
-    "parameter int foo = - -1;\n",
+    {
+        "  parameter int a=b&~(c<<d);",
+        "parameter int a = b & ~(c << d);\n"
     },
-    */
+    {
+        "  parameter int a=~~~~b;",
+        "parameter int a = ~~~~b;\n"
+    },
+    {
+        "  parameter int a = ~ ~ ~ ~ b;",
+        "parameter int a = ~~~~b;\n"
+    },
+    {
+        "  parameter int a   =   ~--b;",
+        "parameter int a = ~--b;\n"
+    },
+    {
+        "  parameter int a   =   ~ --b;",
+        "parameter int a = ~--b;\n"
+    },
+    {
+        "  parameter int a = ~ ++ b;",
+        "parameter int a = ~++b;\n"
+    },
+    {
+        "  parameter int a=--b- --c;",
+        "parameter int a = --b - --c;\n"
+    },
+    {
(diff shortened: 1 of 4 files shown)


verilog/formatting/token_annotator.cc
```diff
@@ -93,9 +93,9 @@ static WithReason<int> SpacesRequiredBetween(const PreFormatToken& left,
   }
 
   // Unary operators (context-sensitive)
-  if (IsUnaryPrefixExpressionOperand(left, context)) {
+  if (IsUnaryPrefixExpressionOperand(left, context) &&
+      !IsUnaryOperator(static_cast<yytokentype>(right.TokenEnum()))) {
```
fangism: Will this preclude `--++x` and `~--x` and `~~x` (as weird as they may seem)?
I actually have a plan to use the lexer itself to pre-compute all illegal 0-space combinations of tokens (internal feature request, not yet published).

I suggest filing a separate issue for this (I'll establish the link) leaving this specific matter to be addressed separately.  It's ok to have a known bug+TODO with fixing spacing around `&`.

fangism: And I can't think of an example where `-` before `=` is legal.  It is possible to prove as such with a static analysis of the bison-generated automata.  They even provide some summaries of where tokens can appear at the start/end of nonterminals.

tgorochowik: Thank you for the review!

You are right that some unary combination were broken with my previous patch. I've updated it to check whether the currently processed token can in fact be both binary and unary, if so, and the right token is unary, it is treated as a binary token (this handles the `& ~` case.

Otherwise both tokens are treated as unary, which should resolve the remaining cases: `~--x` and `~~x` work - I don't think `--++x` is legal though.

I also added some integration and unit tests. Please let me know what you think about the updates. Thanks!

verilog/formatting/token_annotator_test.cc
```diff
@@ -1345,6 +1345,33 @@ TEST(TokenAnnotatorTest, AnnotateFormattingWithContextTest) {
           {0, SpacingOptions::MustAppend},
       },
 
+      // Handle '&' as binary
+      {
+          DefaultStyle,
+          {'&', "&"},
+          {'~', "~"},
+          {},
```
fangism: add a comment inside here: /* unspecified context */

verilog/formatting/token_annotator_test.cc
```diff
@@ -1345,6 +1345,33 @@ TEST(TokenAnnotatorTest, AnnotateFormattingWithContextTest) {
...
+
+      // Two unary operators
+      {
+          DefaultStyle,
+          {'~', "~"},
+          {'~', "~"},
+          {NodeEnum::kUnaryPrefixExpression},
+          {0, SpacingOptions::MustAppend},
+      },
+
```
fangism: few more test case suggestions:
* '~' '('
* '~' SymbolIdentifier

verilog/formatting/formatter_test.cc
```diff
@@ -1198,6 +1198,20 @@ static const std::initializer_list<FormatterTestCase> kFormatterTestCases = {
...
+    {"a = ~ ~ ~ ~ b;",
+     "a = ~~~~b;\n"},
+    {"a   =   ~--b;",
+     "a = ~--b;\n"},
+    {"a   =   ~ --b;",
+     "a = ~--b;\n"},
+    {"a = ~ ++ b;",
+     "a = ~++b;\n"},
+    {"a=--b- --c;",
+     "a = --b - --c;\n"}
```
fangism: Wasn't sure if these test would pass as is, but if it complains about parse failure because the context is unknown, I recommend making these look like: `parameter int a = ...;`.  See similar test cases near the comment `// parameter test cases`.  Please group these tests there.

verilog/formatting/verilog_token.cc
```diff
@@ -563,6 +563,7 @@ FTT GetFormatTokenType(yytokentype e) {
       case '+':
       case '*':
       case '-':
+      case '&':
```
fangism: You can tell this group of cases is missing other operators like `%`, `^`, `/`, `|`, ...
If you like, feel free to mark these as TODO in comments, open new issue, so they can be handled later.
If you feel up for fixing these and adding test cases in this PR, I'd accept that as well.

verilog/formatting/token_annotator.cc
```diff
@@ -93,10 +93,9 @@ static WithReason<int> SpacesRequiredBetween(const PreFormatToken& left,
   }
 
   // Unary operators (context-sensitive)
-  if (IsUnaryPrefixExpressionOperand(left, context)) {
-    // TODO(b/143739545): Don't accidentally join into valid tokens, e.g.:
-    //   '-' and '-' could form '--' (decrement).
-    //   '-' and '=' could form '-=' (subtract-assign).
+  if (IsUnaryPrefixExpressionOperand(left, context) &&
+      (left.format_token_enum != FormatTokenType::binary_operator ||
+       !IsUnaryOperator(static_cast<yytokentype>(right.TokenEnum())))) {
```
fangism: Safe for now, but leave a TODO to allow *some* unary operators on the right to be formatted with 0-space.

verilog/formatting/token_annotator_test.cc
tgorochowik: Sure, that makes sense, Added the  test for both ~(, ~Identifier, and &(, &identifier

verilog/formatting/formatter_test.cc
tgorochowik: These tests did pass, but it still makes sense to group them with the parameter tests, so I did just that

verilog/formatting/token_annotator_test.cc
tgorochowik: Added

verilog/formatting/verilog_token.cc
tgorochowik: I've added the operators you listed as well as some bitwise binary operators (`|` and `^`).

In addition to that I started adding more tests and noticed that `~|` and `~&` were defined as binary, but these operators are unary in all context so I added a patch for that too.

verilog/formatting/token_annotator.cc
tgorochowik: I've added a TODO, along with an example that `a = & ~b;` could be formatted as `a = &~a;`. I hope that's the kind of situation you had in mind.

BTW There is also an issue that the opposite to my example (`a = ~ &b`) shouldn't really be formatted as 0-space, as `~&` is a separate operator (although I think the results of `~` and `&` would technically be the same as of `~&`, so I'm not sure if we should be bothered about this).

fangism: Thanks for the TODO test cases.  Conservatively, even if the joined operator is the same, the formatter currently runs a check for lexical equivalence on the formatted output -- it would be easier to verify token-for-token if they remained lexed separately.

PR #75, opened 2019-12-03
travis: Small fixes

### LINKED ISSUE #73: CI: use --noshow_progress on bazel builds
bazel outputs a continuous stream of percentage of how far it got in the build-process. In the CI case, this means that the build log is pretty large and not super-useful.

So for all CI invocations, we should probably give the `--noshow_progress` option to blaze:
```
bazel test --noshow_progress --cxxopt='-std=c++17' //...
```

mithro:
* Don't output progress - Fixes #73 
 * Include log files in the output on failure.


hzeller:
LGTM

PR #79, opened 2019-12-04
Check struct and union names

### LINKED ISSUE #55: struct/union typedef names should be lower_snake_case_t
Check that the typedef names of structs and unions are lower_snake_case style and end with '_t'.
Checking for use in typedef (non-anonymous) is covered by #54 .
Report a violation on the offending identifier of the type declaration.

Some helper functions on naming style can be found at:
https://github.com/google/verible/blob/master/common/strings/naming_utils.h

rkapuscik:
Should:
* be lower_snake_case
* end with '_t'

Fixes #55.

Signed-off-by: Rafal Kapuscik <rkapuscik@antmicro.com>

### DIFF
--- a/verilog/CST/type.cc
+++ b/verilog/CST/type.cc
@@ -19,9 +19,11 @@
 #include "common/analysis/matcher/matcher.h"
 #include "common/analysis/matcher/matcher_builders.h"
 #include "common/analysis/syntax_tree_search.h"
+#include "common/text/concrete_syntax_leaf.h"
 #include "common/text/concrete_syntax_tree.h"
 #include "common/text/symbol.h"
 #include "common/text/tree_utils.h"
+#include "verilog/CST/identifier.h"
 #include "verilog/CST/verilog_matchers.h"  // pragma IWYU: keep
 
 namespace verilog {
@@ -31,10 +33,33 @@ std::vector<verible::TreeSearchMatch> FindAllDataTypeDeclarations(
   return verible::SearchSyntaxTree(root, NodekDataType());
 }
 
+std::vector<verible::TreeSearchMatch> FindAllTypeDeclarations(
+    const verible::Symbol& root) {
+  return verible::SearchSyntaxTree(root, NodekTypeDeclaration());
+}
+
+std::vector<verible::TreeSearchMatch> FindAllStructDataTypeDeclarations(
+    const verible::Symbol& root) {
+  return verible::SearchSyntaxTree(root, NodekStructDataType());
+}
+
+std::vector<verible::TreeSearchMatch> FindAllUnionDataTypeDeclarations(
+    const verible::Symbol& root) {
+  return verible::SearchSyntaxTree(root, NodekUnionDataType());
+}
+
 bool IsStorageTypeOfDataTypeSpecified(const verible::Symbol& symbol) {
   const auto* storage =
       verible::GetSubtreeAsSymbol(symbol, NodeEnum::kDataType, 0);
   return (storage != nullptr);
+}
+
+const verible::SyntaxTreeLeaf* GetIdentifierFromTypeDeclaration(
+    const verible::Symbol& symbol) {
+  // For enum, struct and union identifier is found at the same position
+  const auto* identifier_symbol =
+      verible::GetSubtreeAsSymbol(symbol, NodeEnum::kTypeDeclaration, 3);
+  return AutoUnwrapIdentifier(*ABSL_DIE_IF_NULL(identifier_symbol));
 }
 
 }  // namespace verilog
--- a/verilog/analysis/checkers/struct_union_name_style_rule.cc
+++ b/verilog/analysis/checkers/struct_union_name_style_rule.cc
@@ -63,6 +63,8 @@ void StructUnionNameStyleRule::HandleSymbol(const verible::Symbol& symbol,
   verible::matcher::BoundSymbolManager manager;
   if (matcher_typedef_.Matches(symbol, &manager)) {
     const char* msg;
+    //TODO: This can be changed to checking type of child (by index) when we
+    // have consistent shape for all kTypeDeclaration nodes.
     if (!FindAllStructDataTypeDeclarations(symbol).empty()) {
       msg = kMessageStruct;
     } else if (!FindAllUnionDataTypeDeclarations(symbol).empty()) {
--- a/verilog/analysis/checkers/struct_union_name_style_rule.cc
+++ b/verilog/analysis/checkers/struct_union_name_style_rule.cc
@@ -18,6 +18,7 @@
 #include <vector>
 
 #include "absl/strings/str_cat.h"
+#include "absl/strings/match.h"
 #include "common/analysis/citation.h"
 #include "common/analysis/lint_rule_status.h"
 #include "common/analysis/matcher/bound_symbol_manager.h"
@@ -63,9 +64,9 @@ void StructUnionNameStyleRule::HandleSymbol(const verible::Symbol& symbol,
   if (matcher_typedef_.Matches(symbol, &manager)) {
     const char* msg;
     if (!FindAllStructDataTypeDeclarations(symbol).empty()) {
-      msg = StructUnionNameStyleRule::kMessageStruct;
+      msg = kMessageStruct;
     } else if (!FindAllUnionDataTypeDeclarations(symbol).empty()) {
-      msg = StructUnionNameStyleRule::kMessageUnion;
+      msg = kMessageUnion;
     } else {
       // Neither a struct nor union definition
       return;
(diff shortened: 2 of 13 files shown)


verilog/CST/type.cc
```diff
@@ -31,10 +33,33 @@ std::vector<verible::TreeSearchMatch> FindAllDataTypeDeclarations(
...
+std::vector<verible::TreeSearchMatch> FindAllStructDataTypeDeclarations(
+    const verible::Symbol& root) {
+  return verible::SearchSyntaxTree(root, NodekStructDataType());
+}
+
+std::vector<verible::TreeSearchMatch> FindAllUnionDataTypeDeclarations(
+    const verible::Symbol& root) {
+  return verible::SearchSyntaxTree(root, NodekUnionDataType());
+}
+
```
fangism: Take a look at the cases of `kTypeDeclaration` in `verilog.y`:

https://github.com/google/verible/blob/d34420461f422d1cdbf4585a376b5a3210b9f828/verilog/parser/verilog.y#L3384

There are some cases where forward declarations are not covered, such as `typedef struct foo_t;`

If we choose to ignore those for the moment, make a TODO note, or file a follow-up issue.

If you wish to handle those now, consider this suggestion:
Make the forward declarations use the same form as their full declarations, like:

https://github.com/google/verible/blob/d34420461f422d1cdbf4585a376b5a3210b9f828/verilog/parser/verilog.y#L3461

but assign `nullptr` to the position where the brace-enclosed definition would normally go, for the sake of shape consistency.

and here:

https://github.com/google/verible/blob/d34420461f422d1cdbf4585a376b5a3210b9f828/verilog/parser/verilog.y#L3396

replace `$2` with a tagged node for `kStructDataType` that encloses `$2`.  (Same idea for unions)

I'm also amenable to other ideas as well, such as distinguishing forward-declarations from definitions with a new node enum.

verilog/analysis/checkers/struct_union_name_style_rule.cc
```diff
@@ -0,0 +1,89 @@
...
+      " names use lower_snake_case naming convention and end with '_t'. See ",
+      GetStyleGuideCitation(kTopic), ".");
+}
+
+void StructUnionNameStyleRule::HandleSymbol(const verible::Symbol& symbol,
+                                       const SyntaxTreeContext& context) {
+  verible::matcher::BoundSymbolManager manager;
+  if (matcher_typedef_.Matches(symbol, &manager)) {
+    const char* msg;
+    if (!FindAllStructDataTypeDeclarations(symbol).empty()) {
```
fangism: Alternative to searching inside a type declaration for the underlying struct/union type, if you shape nodes of `kTypeDeclaration` to have consistent structure (children positions), you could directly reference a child by index using:
https://github.com/google/verible/blob/d34420461f422d1cdbf4585a376b5a3210b9f828/common/text/tree_utils.h#L69

verilog/analysis/checkers/struct_union_name_style_rule.cc
```diff
@@ -0,0 +1,89 @@
...
+      msg = StructUnionNameStyleRule::kMessageUnion;
+    } else {
+      // Neither a struct nor union definition
+      return;
+    }
+    const auto* identifier_leaf =
+              GetIdentifierFromTypeDeclaration(symbol);
+    const auto name = ABSL_DIE_IF_NULL(identifier_leaf)->get().text;
+    if (!verible::IsLowerSnakeCaseWithDigits(name)
+        || (name.compare(name.size()-2, 2, "_t") != 0)) {
```
fangism: Use `EndsWith`:

https://github.com/abseil/abseil-cpp/blob/a4b757b5d42694306a9de853cee0a5fba9c7bbe9/absl/strings/match.h#L61

verilog/analysis/checkers/struct_union_name_style_rule_test.cc
```diff
@@ -0,0 +1,91 @@
...
+      {"typedef struct {logic foo; logic bar;} ", {kToken, "HelloWorld"}, ";"},
+      {"typedef struct {logic foo; logic bar;} ", {kToken, "_baz"}, ";"},
+      {"typedef struct {logic foo; logic bar;} ", {kToken, "Bad_name"}, ";"},
+      {"typedef struct {logic foo; logic bar;} ", {kToken, "bad_Name"}, ";"},
+      {"typedef struct {logic foo; logic bar;} ", {kToken, "Bad2"}, ";"},
+      {"typedef struct {logic foo; logic bar;} ", {kToken, "very_Bad_name"},
+       ";"},
+      {"typedef struct {logic foo; logic bar;} ", {kToken, "wrong_ending"},
+       ";"},
+  };
```
fangism: suggested test cases:
* name = `_t` (weird, but valid by definition)
* name = `t` (invalid, and shorter than the suffix we are looking for)
* name ending with `_`

if you choose to handle forward declarations:
* `typedef struct foo_t;` (using valid and invalid name)

hzeller:
@mithro is there a way to re-trigger travis now that the unrelated `StrContains()` problem has been fixed in https://github.com/google/verible/commit/7a5c35840d1e2ff41837eb7ac70d185c9d906afb ?

mithro:
@hzeller - Rebase your change on top or origin/master then force push over this existing branch.

verilog/analysis/checkers/struct_union_name_style_rule.cc
```diff
@@ -0,0 +1,89 @@
...
+      GetStyleGuideCitation(kTopic), ".");
+}
+
+void StructUnionNameStyleRule::HandleSymbol(const verible::Symbol& symbol,
+                                       const SyntaxTreeContext& context) {
+  verible::matcher::BoundSymbolManager manager;
+  if (matcher_typedef_.Matches(symbol, &manager)) {
+    const char* msg;
+    if (!FindAllStructDataTypeDeclarations(symbol).empty()) {
+      msg = StructUnionNameStyleRule::kMessageStruct;
```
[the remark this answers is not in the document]
rkapuscik: Fixed, thanks.

verilog/analysis/checkers/struct_union_name_style_rule.cc
rkapuscik: Thanks for the tip, changed.

verilog/analysis/checkers/struct_union_name_style_rule_test.cc
rkapuscik: I added test cases for all of those but `_t` - according to the [name-checking function](https://github.com/google/verible/blob/master/common/strings/naming_utils.cc#L61) names with leading underscore are illegal. Are you sure they should be allowed here?

verilog/CST/type.cc
rkapuscik: I see what you mean, I think it would be best to have forward declarations checked here as well.
I have changed the forward declarations to have the same structure as full declarations, but it looks quite fragile to me and I'm unsure if this is acceptable. I think the class and interface declarations would use a bit different structure anyway.

verilog/analysis/checkers/struct_union_name_style_rule.cc
rkapuscik: It could be done, but I think a helper function would be needed anyway. It would certainly be faster, though.

fangism: Leave a TODO then.

verilog/analysis/checkers/struct_union_name_style_rule_test.cc
fangism: You are correct.  In that case, mark `_t` as invalid.

verilog/parser/verilog.y
```diff
@@ -3390,12 +3390,26 @@ type_declaration
...
     { $$ = MakeTaggedNode(N::kTypeDeclaration, $1, $2, nullptr, $3, $4, nullptr); }
+  /* TODO: Figure out how to make the braced members list optional
+     to make these more robust */
   | TK_typedef TK_enum GenericIdentifier ';'
-    { $$ = MakeTaggedNode(N::kTypeDeclaration, $1, $2, nullptr, $3, $4, nullptr); }
+    { $$ = MakeTaggedNode(N::kTypeDeclaration, $1,
+                          MakeTaggedNode(N::kDataTypePrimitive,
+                                         MakeTaggedNode(N::kEnumDataType,
+                                                        $2, nullptr, nullptr)),
+                          nullptr, $3, $4, nullptr); }
```
fangism: This is good.  Later we can introduce helper functions to constrain the construction of some of these re-used nodes in CST/*.  No further change needed.

verilog/analysis/checkers/struct_union_name_style_rule_test.cc
rkapuscik: I added negative test case for both types.

verilog/analysis/checkers/struct_union_name_style_rule.cc
rkapuscik: Done.

Issue #26, opened 2019-11-25
[token partitioning] else statements need to start own partition


msfschaffner:
should:
```systemverilog
      if (trans_collected.read_write == WRITE)  item_collected_port.write(trans_collected); 
      else                                      collect_data_queue.put(trans_collected);
```
is:
```systemverilog
      if (trans_collected.read_write == WRITE)
        item_collected_port.write(trans_collected); else collect_data_queue.put(trans_collected);
```

msfschaffner:
b/144605105

ldalek-antmicro:
It looks like this commit https://github.com/google/verible/commit/36865ed868da4f141a2a8f8992eebc22824335f5 fixes the issue. @fangism can you confirm that?

fangism:
Doesn't look fixed yet.  Here's a reduced test case:

```systemverilog
task t; if (r == t) a.b(c); else d.e(f); endtask
```

I still get this output:

```systemverilog
task t;
  if (r == t) a.b(c); else d.e(f);
endtask
```

but we want:

```systemverilog
task t;
  if (r == t) a.b(c);
  else d.e(f);
endtask
```

fangism:
`verilog_format --show_token_partition_tree` actually looks correct:

```
Full token partition tree:
{ ([<auto>]) @{}
  { ([<auto>]) @{0}
    { ([task t ;]) }
  }
  { (  [<auto>]) @{1}
    { (  [<auto>]) @{1,0}
      { (  [if ( r == t )]) }
      { (    [<auto>]) @{1,0,1}
        { (    [a . b ( c ) ;]) }
      }
      { (  [else]) }
      { (    [<auto>]) @{1,0,3}
        { (    [d . e ( f ) ;]) }
      }
    }
  }
  { ([endtask]) }
}
Halting for diagnostic operation.
```

But what I think is happening in
https://github.com/google/verible/issues/26#issuecomment-561328376
is that it is choosing to fit on one line if possible.
We have to update the heuristic so that `else` forces the partition to be evaluated in pieces, rather than as a whole.

Here is the if-it-fits policy:

https://github.com/google/verible/blob/d34420461f422d1cdbf4585a376b5a3210b9f828/verilog/formatting/tree_unwrapper.cc#L624

Fit determination is here:

https://github.com/google/verible/blob/d34420461f422d1cdbf4585a376b5a3210b9f828/common/formatting/line_wrap_searcher.cc#L180

[...]

PR #80, opened 2019-12-05
[token partitioning] else statements need to start own partition #26

ldalek-antmicro:
Fixes https://github.com/google/verible/issues/26 and applies rule from https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md#begin--end


verilog/formatting/token_annotator.cc
```diff
@@ -496,6 +496,17 @@ static WithReason<SpacingOptions> BreakDecisionBetween(
...
+    if (left.TokenEnum() != TK_end)
+      return {SpacingOptions::MustWrap, "'else' token should start its own line unless preceded by 'end' without label."};
+    else
+      return {SpacingOptions::MustAppend, "'end'-'else' tokens should be together on one line."};
+  }
+
+  if ( (left.TokenEnum() == TK_else) && (right.TokenEnum() == TK_begin) ) {
+    return {SpacingOptions::MustAppend, "'else'-'begin' tokens should be together on one line."};
+  }
+
```
fangism: Just for your information, these decisions here are bound in that it doesn't let the wrap optimizer decide how to wrap.  Perhaps this is a good thing in that it constrains the search.
No change needed.

verilog/formatting/token_annotator_test.cc
```diff
@@ -1224,6 +1224,127 @@ TEST(TokenAnnotatorTest, AnnotateFormattingInfoTest) {
                {yytokentype::TK_endmodule, "endmodule"},
            }},
 
+          // if (r == t) a.b(c);
+          // else d.e(f);
```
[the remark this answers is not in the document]
ldalek-antmicro: Okay. Thanks.

verilog/formatting/token_annotator.cc
ldalek-antmicro: Okay. Thanks for info.

fangism:
Thinking about this a little more, I might approach this one differently, using the tree_wrapper.cc to shape the formatting partitions, however, as long as the end-to-end formatter_test.cc tests look good, I'll accept this implementation and revisit later.  The tests are more important than the code.

Issue #85, opened 2019-12-09
Find label mismatches between begin/end pairs


fangism:
Task: find mismatches between labels of begin/end of sequential blocks:

In this example:

```
begin : foo
  ...
end : bar
```

task is to report a violation at `bar`.

If there is no `end` label, ignore.
If there is no `begin` label, but there is an `end` label, report on the `begin` keyword that the matching label is missing.

This initial task can just focus on `begin` and `end` pairs.  We generally want this in all label-able blocks and declarations, but let those be addressed as separate issues.

Suggested rule name: "mismatched-labels"

fangism:
b/135782747

PR #88, opened 2019-12-13
Find label mismatches between begin/end pairs

tgorochowik:
Fixes: #85

### DIFF
--- a/verilog/CST/seq_block.cc
+++ b/verilog/CST/seq_block.cc
@@ -1,0 +1,64 @@
+// Copyright 2017-2020 The Verible Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "verilog/CST/seq_block.h"
+
+#include <vector>
+
+#include "common/analysis/matcher/matcher.h"
+#include "common/analysis/matcher/matcher_builders.h"
+#include "common/analysis/syntax_tree_search.h"
+#include "common/text/concrete_syntax_leaf.h"
+#include "common/text/concrete_syntax_tree.h"
+#include "common/text/symbol.h"
+#include "common/text/syntax_tree_context.h"
+#include "common/text/tree_utils.h"
+#include "verilog/CST/identifier.h"
+#include "verilog/CST/verilog_matchers.h"  // IWYU pragma: keep
+
+namespace verilog {
+
+using verible::Symbol;
+using verible::SyntaxTreeContext;
+using verible::TokenInfo;
+
+static const TokenInfo* GetSeqBlockLabelTokenInfo(const Symbol& symbol,
+                                                  NodeEnum parent) {
+  const auto *s = GetSubtreeAsSymbol(symbol, parent, 1);
+
+  if (s == nullptr) {
+    return nullptr;
+  }
+
+  const auto& node = SymbolCastToNode(*s);
+  return &AutoUnwrapIdentifier(*node.children()[1].get())->get();
+}
+
+const TokenInfo* GetBeginLabelTokenInfo(const Symbol& symbol) {
+  CHECK_EQ(NodeEnum(symbol.Tag().tag), NodeEnum::kBegin);
+  return GetSeqBlockLabelTokenInfo(symbol, NodeEnum::kBegin);
+}
+
+const TokenInfo* GetEndLabelTokenInfo(const Symbol& symbol) {
+  CHECK_EQ(NodeEnum(symbol.Tag().tag), NodeEnum::kEnd);
+  return GetSeqBlockLabelTokenInfo(symbol, NodeEnum::kEnd);
+}
+
+const Symbol* GetMatchingEnd(const Symbol& symbol,
+                             const SyntaxTreeContext& context) {
+  CHECK_EQ(NodeEnum(symbol.Tag().tag), NodeEnum::kBegin);
+  return context.top().children().back().get();
+}
+
+}  // namespace verilog
--- a/verilog/CST/seq_block.h
+++ b/verilog/CST/seq_block.h
@@ -1,0 +1,45 @@
+// Copyright 2017-2020 The Verible Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// This unit provides helper functions that pertain to SystemVerilog
+// net declaration nodes in the parser-generated concrete syntax tree.
+
+#ifndef VERIBLE_VERILOG_CST_SEQ_BLOCK_H_
+#define VERIBLE_VERILOG_CST_SEQ_BLOCK_H_
+
+#include <vector>
+
+#include "common/text/concrete_syntax_leaf.h"
+#include "common/text/concrete_syntax_tree.h"
+#include "common/text/symbol.h"
+#include "common/text/syntax_tree_context.h"
+
+namespace verilog {
+
+// Get TokenInfo of a label for a given kBegin symbol if exists, else nullptr
+const verible::TokenInfo* GetBeginLabelTokenInfo(
+    const verible::Symbol& symbol);
+
+// Get TokenInfo of a label for a given kEnd symbol if exists, else nullptr
+const verible::TokenInfo* GetEndLabelTokenInfo(
+    const verible::Symbol& symbol);
+
+// Find and return a pointer to a kEnd symbol corresponding to a given kBegin
+const verible::Symbol* GetMatchingEnd(
+    const verible::Symbol& symbol, const verible::SyntaxTreeContext& context);
+
+}  // namespace verilog
+
+#endif  // VERIBLE_VERILOG_CST_SEQ_BLOCK_H_
+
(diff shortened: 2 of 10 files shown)


verilog/analysis/checkers/mismatched_labels_rule_test.cc
```diff
@@ -0,0 +1,194 @@
...
+       "    end : ", {SymbolIdentifier, "inv_second_label"}, "\n"
+       "  end : ", {SymbolIdentifier, "inv_first_label"}, "\n"
+       "endmodule"},
+      {"module foo;\n"
+       "  initial begin : first_label\n"
+       "    for(int i=0; i<5; ++i)\n"
+       "    begin : second_label\n"
+       "    end : ", {SymbolIdentifier, "first_label"}, "\n"
+       "  end : ", {SymbolIdentifier, "second_label"}, "\n"
+       "endmodule"},
```
tgorochowik: These two tests fail with the current approach. The reason is currently the lint rule matches the begin nodes and then tries to find the corresponding end nodes. In that case, when the error is reported on the end node label, the violations are reported in the wrong order (that is, in the order of the begin nodes, not in the order of the end nodes).

I have created this draft PR early, so most of the code can be already reviewed, but this still has to be addressed.

fangism: These test cases look great.

One thing we can do to address violation ordering is to update the RunLintTestCases driver to sort the violations by starting position.  

It may even be a good idea to sort the violations some time before they are eventually reported to stdout; it is reasonable for users to expect violations to be reported in location order.  Multiple ways this could be done, e.g. overrideable virtual function hook in LintRule to perform arbitrary actions on-finish (which would address unit-testing), or collect all violations after all checks are run and do a one-time final sort before printing.

verilog/tools/lint/testdata/mismatched_labels.sv
```diff
@@ -0,0 +1,9 @@
+module mismatched_labels(
+  input clk_i
+);
+
+always_ff @(posedge clk_i)
+  begin : foo
+  end : bar
```
fangism: comment that this is where you expect the violation to be reported

tgorochowik:
Just to note: this issue depends on #91 and will be continued after it is accepted.

verilog/analysis/checkers/mismatched_labels_rule_test.cc
tgorochowik: Addressed in #91.

verilog/analysis/checkers/mismatched_labels_rule.h
```diff
@@ -0,0 +1,66 @@
...
+  static std::string GetDescription(DescriptionType);
+
+  void HandleSymbol(const verible::Symbol& symbol,
+                    const verible::SyntaxTreeContext& context) override;
+
+  verible::LintRuleStatus Report() const override;
+
+ private:
+  // Matches the begin node.
+  const verible::matcher::Matcher matcher_ = NodekBegin();
```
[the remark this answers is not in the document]
tgorochowik: I initially did that, but I wasn't sure if the begin/end nodes will always appear in kSeqBlock, thus I switched that to kBegin. The tests pass with the current implementation

verilog/tools/lint/testdata/mismatched_labels.sv
tgorochowik: Sure, added, thanks!

verilog/analysis/checkers/mismatched_labels_rule.cc
```diff
@@ -0,0 +1,119 @@
...
+  verible::matcher::BoundSymbolManager manager;
+
+  if (matcher_.Matches(symbol, &manager)) {
+
+    // Find the corresponding kEnd node.
+    // Note that this may find more than one kEnd node if the blocks
+    // are nested, in this case using the very last result ensures
+    // we use the kEnd corresponding to the given kBegin
+    std::vector<verible::TreeSearchMatch> end =
+      SearchSyntaxTree(context.top(), NodekEnd());
```
fangism: Since you're matching on `kBegin`, and going "up" the context's stack, could you just directly access the last node elements of the context, so you're guaranteed to be accessing (the last) sibling of `kBegin`?  You might be able to leverage some of the accessor functions in tree_utils.h.  (Might even modify then to support negative indexing to access elements from the back.)

If you want to go the extra distance, this would be a nice function to add to the `CST` library, given a `kBegin` node and context, `FindMatchingEnd()`.

Otherwise, leave some TODOs for me to revisit.

verilog/analysis/checkers/mismatched_labels_rule.cc
```diff
@@ -0,0 +1,119 @@
...
+
+    // This should never really happen with correct syntax
+    if (end.size() == 0) {
+      return;
+    }
+
+    // Find the label of the end statement
+    std::vector<verible::TreeSearchMatch> end_label_node =
+      SearchSyntaxTree(*end[end.size() - 1].match, NodekLabel());
```
fangism: Could this search be simplified to a direct access, given the way it is constructed in verilog.y?  Could also be a re-usable CST function, like `GetEndLabel()` (may return nullptr).

verilog/analysis/checkers/mismatched_labels_rule.cc
```diff
@@ -0,0 +1,119 @@
...
+
+    // Ignore if there is no end label
+    if (end_label_node.size() != 1) {
+      return;
+    }
+
+    // Find the label of the begin statement
+    std::vector<verible::TreeSearchMatch> begin_label_node =
+      SearchSyntaxTree(symbol, NodekLabel());
```
fangism: Same idea, would direct access work better?

verilog/analysis/checkers/mismatched_labels_rule.cc
tgorochowik: That makes perfect sense, thanks! I just pushed a new implementation with the suggested changes.

verilog/analysis/checkers/mismatched_labels_rule.cc
tgorochowik: Sure, that makes sense, thanks! I implemented this as `GetEndLabelTokenInfo` as I need both the actual label as well as its TokenInfo for the linter rule implementation

verilog/analysis/checkers/mismatched_labels_rule.cc
tgorochowik: sure, yes!

verilog/CST/seq_block.cc
```diff
@@ -0,0 +1,64 @@
...
+
+namespace verilog {
+
+using verible::Symbol;
+using verible::SyntaxTreeContext;
+using verible::TokenInfo;
+
+static const TokenInfo* GetSeqBlockLabelTokenInfo(const Symbol& symbol,
+                                                  NodeEnum parent) {
+  const auto s = GetSubtreeAsSymbol(symbol, parent, 1);
```
fangism: `const auto* s` (pointer)

verilog/CST/seq_block.cc
```diff
@@ -0,0 +1,64 @@
...
+
+  if (s == NULL) {
+    return NULL;
+  }
```
fangism: `nullptr` instead of `NULL`

verilog/CST/seq_block.cc
```diff
@@ -0,0 +1,64 @@
...
+
+const TokenInfo* GetEndLabelTokenInfo(const Symbol& symbol) {
+  CHECK_EQ(NodeEnum(symbol.Tag().tag), NodeEnum::kEnd);
+  return GetSeqBlockLabelTokenInfo(symbol, NodeEnum::kEnd);
+}
+
+const Symbol* GetMatchingEnd(const Symbol& symbol,
+                             const SyntaxTreeContext& context) {
+  CHECK_EQ(NodeEnum(symbol.Tag().tag), NodeEnum::kBegin);
+  return context.top().children()[context.top().children().size() - 1].get();
```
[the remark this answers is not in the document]
tgorochowik: Sure, that makes perfect sense, thanks!

verilog/CST/seq_block_test.cc
```diff
@@ -0,0 +1,191 @@
...
+
+  ASSERT_OK(analyzer.Analyze());
+  const auto& root = analyzer.Data().SyntaxTree();
+
+  const auto begin_statements = FindAllBeginStatements(*root);
+  ASSERT_EQ(begin_statements.size(), 1);
+
+  const auto matchingEnd =
+    GetMatchingEnd(*begin_statements[0].match, begin_statements[0].context);
+  ASSERT_EQ(NodeEnum(matchingEnd->Tag().tag), NodeEnum::kEnd);
```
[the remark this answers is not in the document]
tgorochowik: Sure, changed it where relevant, thanks!

verilog/CST/seq_block_test.cc
```diff
@@ -0,0 +1,191 @@
...
+
+  ASSERT_OK(analyzer.Analyze());
+  const auto& root = analyzer.Data().SyntaxTree();
+
+  const auto begin_statements = FindAllBeginStatements(*root);
+  ASSERT_EQ(begin_statements.size(), 1);
+
+  const auto* beginToken = GetBeginLabelTokenInfo(*begin_statements[0].match);
+  ASSERT_NE(beginToken, nullptr);
+  ASSERT_EQ(beginToken->text, "begin_label");
```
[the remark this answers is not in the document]
tgorochowik: Thanks! Used one of your suggestions about possible approaches with `ABSL_DIE_IF_NULL`. Thanks!

verilog/analysis/checkers/mismatched_labels_rule.cc
```diff
@@ -0,0 +1,95 @@
...
+
+void MismatchedLabelsRule::HandleSymbol(
+    const verible::Symbol& symbol, const verible::SyntaxTreeContext& context) {
+  verible::matcher::BoundSymbolManager manager;
+
+  if (matcher_.Matches(symbol, &manager)) {
+    const auto& matchingEnd = GetMatchingEnd(symbol, context);
+
+    const auto begin_label = GetBeginLabelTokenInfo(symbol);
+    const auto end_label = GetEndLabelTokenInfo(*matchingEnd);
```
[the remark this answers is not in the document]
tgorochowik: Changed in all the relevant places, thanks!

verilog/analysis/checkers/mismatched_labels_rule.cc
```diff
@@ -0,0 +1,95 @@
...
+
+    // Don't check anything if there is no end label
+    if (end_label == NULL) {
```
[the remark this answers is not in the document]
tgorochowik: Changed it in all the relevant places, thanks!

verilog/CST/seq_block.h
```diff
@@ -0,0 +1,45 @@
...
+#include <vector>
+
+#include "common/text/concrete_syntax_leaf.h"
+#include "common/text/concrete_syntax_tree.h"
+#include "common/text/symbol.h"
+#include "common/text/syntax_tree_context.h"
+
+namespace verilog {
+
+// Get TokenInfo of a label for a given kBegin symbol
```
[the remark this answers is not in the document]
tgorochowik: Sure, added that to the comment, thanks!

PR #91, opened 2019-12-16
Sort lint violations

tgorochowik:
This PR addresses an issue described here: https://github.com/google/verible/pull/88#discussion_r357854849

Copy for reference:

> It may even be a good idea to sort the violations some time before they are eventually reported to stdout; it is reasonable for users to expect violations to be reported in location order. Multiple ways this could be done, e.g. overrideable virtual function hook in LintRule to perform arbitrary actions on-finish (which would address unit-testing), or collect all violations after all checks are run and do a one-time final sort before printing. I suggest filing this (total ordering) as a separate issue, to keep this PR focused.

There are two patches, one solves the issue which can be seen in the test cases (see #88).
The other sorts the lint violations before printing them, for example, for:
```
module top;
  wire UglyWire;

  initial begin : first_label
    for(int i=0; i<5; ++i)
    begin : second_label
    end : i_second_label
  end : i_first_label

  wire A_sfds_asda;
endmodule
```

We get:
```
top.sv:2:8: Signal names must use lower_snake_case naming convention. [Style: signal-conventions] [signal-name-style]
top.sv:7:11: Begin/end block labels must match. [Style: mismatched-labels] [mismatched-labels]
top.sv:8:9: Begin/end block labels must match. [Style: mismatched-labels] [mismatched-labels]
top.sv:10:8: Signal names must use lower_snake_case naming convention. [Style: signal-conventions] [signal-name-style]
```
Instead of:
```                 
top.sv:10:8: Signal names must use lower_snake_case naming convention. [Style: signal-conventions] [signal-name-style]
top.sv:2:8: Signal names must use lower_snake_case naming convention. [Style: signal-conventions] [signal-name-style]
top.sv:7:11: Begin/end block labels must match. [Style: mismatched-labels] [mismatched-labels]                       
top.sv:8:9: Begin/end block labels must match. [Style: mismatched-labels] [mismatched-labels
```

Note that the printed violations are no longer grouped by lint rules and correctly sorted.

The test cases are not added to this PR yet, as I think there may be a better way to approach the sorting. If so, please feel free to criticize and point me in the right directions. Thanks!

### DIFF
--- a/common/analysis/lint_rule_status.cc
+++ b/common/analysis/lint_rule_status.cc
@@ -75,10 +75,11 @@ void LintStatusFormatter::FormatViolation(std::ostream* stream,
 
 void LintRuleStatus::WaiveViolations(
     std::function<bool(const LintViolation&)>&& is_waived) {
-  std::vector<LintViolation> filtered_violations;
-  filtered_violations.reserve(violations.size());
+  std::set<LintViolation> filtered_violations;
   std::remove_copy_if(violations.begin(), violations.end(),
-                      std::back_inserter(filtered_violations), is_waived);
+                      std::inserter(filtered_violations,
+                                    filtered_violations.begin()),
+                      is_waived);
   violations.swap(filtered_violations);
 }
 
--- a/common/analysis/lint_rule_status.cc
+++ b/common/analysis/lint_rule_status.cc
@@ -58,6 +58,44 @@ void LintStatusFormatter::FormatLintRuleStatus(std::ostream* stream,
   }
 }
 
+struct LintViolationWithStatus {
+  const LintViolation *violation;
+  const LintRuleStatus *status;
+
+  LintViolationWithStatus(const LintViolation *v, const LintRuleStatus *s)
+      : violation(v), status(s) {}
+
+  bool operator<(const LintViolationWithStatus& r) const {
+    // compares addresses which correspond to locations within the same string
+    return violation->token.text.data() < r.violation->token.text.data();
+  };
+};
+
+void LintStatusFormatter::FormatLintRuleStatuses(
+    std::ostream* stream,
+    const std::vector<LintRuleStatus> &statuses,
+    absl::string_view base,
+    absl::string_view path) const {
+
+  std::set<LintViolationWithStatus> violations;
+
+  // TODO(fangism): rewrite as a linear time merge of pre-ordered sub-sequences
+  for (auto &status : statuses) {
+    for (auto &violation : status.violations) {
+      violations.insert(LintViolationWithStatus(&violation, &status));
+    }
+  }
+
+  for (auto violation : violations) {
+    FormatViolation(stream,
+                    *violation.violation,
+                    base, path,
+                    violation.status->url,
+                    violation.status->lint_rule_name);
+    *stream << std::endl;
+  }
+}
+
 // Formats and outputs violation on stream
 // Path is file path of original file and url is a link to violated rule
 void LintStatusFormatter::FormatViolation(std::ostream* stream,
--- a/common/analysis/lint_rule_status.h
+++ b/common/analysis/lint_rule_status.h
@@ -21,6 +21,7 @@
 #include <iosfwd>
 #include <string>
 #include <vector>
+#include <set>
 
 #include "absl/strings/string_view.h"
 #include "common/text/line_column_map.h"
@@ -63,17 +64,23 @@ struct LintViolation {
   // The context (list of ancestors) of the offending token.
   // For non-syntax-tree analyses, leave this blank.
   const SyntaxTreeContext context;
+
+  bool operator<(const LintViolation& r) const {
+    // compares addresses of violations, which correspond to substring
+    // locations
+    return token.text.data() < r.token.text.data();
+  };
 };
 
 // LintRuleStatus represents the result of running a single lint rule.
 struct LintRuleStatus {
   LintRuleStatus() : violations() {}
 
-  LintRuleStatus(const std::vector<LintViolation>& vs,
+  LintRuleStatus(const std::set<LintViolation>& vs,
                  absl::string_view rule_name, const std::string& url)
       : lint_rule_name(rule_name), url(url), violations(vs) {}
 
-  explicit LintRuleStatus(const std::vector<LintViolation>& vs)
+  explicit LintRuleStatus(const std::set<LintViolation>& vs)
       : violations(vs) {}
 
   bool isOk() const { return violations.empty(); }
@@ -89,7 +96,7 @@ struct LintRuleStatus {
   std::string url;
 
   // Contains all violations of the LintRule
-  std::vector<LintViolation> violations;
+  std::set<LintViolation> violations;
 };
 
 // LintStatusFormatter is a class for printing LintRuleStatus's and
--- a/common/analysis/lint_rule_status.h
+++ b/common/analysis/lint_rule_status.h
@@ -123,6 +123,18 @@ class LintStatusFormatter {
                             absl::string_view base,
                             absl::string_view path) const;
 
+  // Formats, sorts and outputs status to stream.
+  // The violations contained in the statuses are sorted by their occurrence
+  // in the code and are not grouped by the status object.
+  // Path is the file path of original file. This is needed because it is not
+  // contained in status.
+  // Base is the string_view of the entire contents, used only for byte offset
+  // calculation.
+  void FormatLintRuleStatuses(std::ostream* stream,
+                              const std::vector<LintRuleStatus> &statuses,
+                              absl::string_view base,
+                              absl::string_view path) const;
+
   // Formats and outputs violation on stream.
   // Path is file path of original file and url is a link to the ratified rule
   // that is being violated.
--- a/common/analysis/linter_test_utils.h
+++ b/common/analysis/linter_test_utils.h
@@ -46,7 +46,7 @@ struct LintTestCase : public SynthesizedLexerTestData {
   // Returns true if every element is an exact match to the expected set.
   // TODO(b/141875806): Take a symbol translator function to produce a
   // human-readable, language-specific enum name.
-  bool ExactMatchFindings(const std::vector<LintViolation>& found_violations,
+  bool ExactMatchFindings(const std::set<LintViolation>& found_violations,
                           absl::string_view base,
                           std::ostream* diffstream) const;
 };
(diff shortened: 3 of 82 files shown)


common/analysis/lint_rule_status.cc
```diff
@@ -58,6 +58,39 @@ void LintStatusFormatter::FormatLintRuleStatus(std::ostream* stream,
   }
 }
 
+void LintStatusFormatter::FormatLintRuleStatuses(
+    std::ostream* stream,
+    const std::vector<LintRuleStatus> statuses,
```
fangism: `const&`?

common/analysis/lint_rule_status.cc
```diff
@@ -58,6 +58,39 @@ void LintStatusFormatter::FormatLintRuleStatus(std::ostream* stream,
   }
 }
 
+void LintStatusFormatter::FormatLintRuleStatuses(
```
fangism: Recommend factoring this into two functions, one that sort, the other that format the sorted result.

common/analysis/lint_rule_status.h
```diff
@@ -116,6 +116,16 @@ class LintStatusFormatter {
                             absl::string_view base,
                             absl::string_view path) const;
 
+  // Formats and outputs status to stream.
```
fangism: Comment should mention sorting.

common/analysis/linter_test_utils.h
```diff
@@ -72,10 +72,27 @@ void RunLintTestCase(const LintTestCase& test,
   // Instantiate a linter that runs a single rule to analyze text.
   LintRunner<RuleType> lint_runner(make_rule());
   const LintRuleStatus rule_status = lint_runner.Run(analyzer.Data(), filename);
-  const auto& violations(rule_status.violations);
```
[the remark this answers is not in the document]
tgorochowik: I started modifying the `LintRuleStatus` class to use a `set` instead of the `vector` to store the violations.

In general it works quite well and the code needed to solve what this PR is trying to solve is indeed much prettier, but there is one issue: It looks as though the tree can be created without any TokenInfos. This makes sorting by the text pointers useless - moreover having a set there doesn't really work, as with  a `operator<` definition that is based on the token info text pointers the items are not inserted correctly.

As far as I can see, such trees are only created in the tests, and some tests fail for that reason, for example: https://github.com/google/verible/blob/37eec341f5390744aeee70fe91417c19ceb8f10b/common/analysis/syntax_tree_linter_test.cc#L243

How should this be approached?
* Should I assume that this API will only ever be used to write tests and therefore assume it is safe to modify those tests?
* Should I keep collecting the violations as a vector but add something like `getViolationsAsSet`, and use it for linter tests + printing linter violations?

Do you have any other suggestions?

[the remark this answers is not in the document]
fangism: Ah yes, those synthetic trees in those tests will need to be populated with slightly more real leaf tokens.

Here are some examples where I needed tokens to point to `substr` ranges inside a contiguous buffer (string_view):

https://github.com/google/verible/blob/81d39b1e13688850fb918cef9b5c47f135a962dd/common/text/tree_utils_test.cc#L250

https://github.com/google/verible/blob/81d39b1e13688850fb918cef9b5c47f135a962dd/common/text/text_structure_test.cc#L434

You may certainly update those tests.

common/analysis/lint_rule_status.cc
```diff
@@ -58,6 +58,39 @@ void LintStatusFormatter::FormatLintRuleStatus(std::ostream* stream,
...
+      violations.push_back({&violation, &status});
+    }
+  }
+
+  sort(violations.begin(), violations.end(),
+       [&base](std::tuple<const LintViolation*, const LintRuleStatus*> v1,
+               std::tuple<const LintViolation*, const LintRuleStatus*> v2)
+  {
+    return
+      std::get<0>(v1)->token.left(base) < std::get<0>(v2)->token.left(base);
```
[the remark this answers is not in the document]
tgorochowik: Thanks! Updated

common/analysis/lint_rule_status.cc
```diff
@@ -58,6 +58,39 @@ void LintStatusFormatter::FormatLintRuleStatus(std::ostream* stream,
   }
 }
 
+void LintStatusFormatter::FormatLintRuleStatuses(
+    std::ostream* stream,
+    const std::vector<LintRuleStatus> statuses,
+    absl::string_view base,
+    absl::string_view path) const {
+
+  std::vector<std::tuple<const LintViolation*, const LintRuleStatus*>>
```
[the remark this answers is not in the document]
tgorochowik: Thanks! Changed this

common/analysis/lint_rule_status.cc
```diff
@@ -58,6 +58,39 @@ void LintStatusFormatter::FormatLintRuleStatus(std::ostream* stream,
...
+      violations;
+
+  for (const auto& status : statuses) {
```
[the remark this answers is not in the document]
tgorochowik: Changed the implementation to sets so it is no longer relevant.

common/analysis/linter_test_utils.h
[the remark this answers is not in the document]
tgorochowik: Thanks! Updated those tests based on your suggestion.

common/analysis/lint_rule_status.cc
tgorochowik: No longer relevant since switching to sets

common/analysis/lint_rule_status.cc
tgorochowik: updated, thanks!

common/analysis/lint_rule_status.h
tgorochowik: Updated the comment, thanks!

common/analysis/lint_rule_status.cc
```diff
@@ -58,6 +58,42 @@ void LintStatusFormatter::FormatLintRuleStatus(std::ostream* stream,
...
+
+void LintStatusFormatter::FormatLintRuleStatuses(
+    std::ostream* stream,
+    const std::vector<LintRuleStatus> &statuses,
+    absl::string_view base,
+    absl::string_view path) const {
+
+  std::set<LintViolationWithStatus> violations;
+
+  for (auto &status : statuses) {
```
fangism: // TODO(fangism): rewrite as a linear time merge of pre-ordered sub-sequences

common/analysis/lint_rule_status.h
```diff
@@ -63,17 +64,21 @@ struct LintViolation {
   // The context (list of ancestors) of the offending token.
   // For non-syntax-tree analyses, leave this blank.
   const SyntaxTreeContext context;
+
+  bool operator<(const LintViolation& r) const {
+    return (token.text.data() < r.token.text.data());
```
fangism: can remove extra ()

common/analysis/lint_rule_status.cc
```diff
@@ -58,6 +58,42 @@ void LintStatusFormatter::FormatLintRuleStatus(std::ostream* stream,
...
 
+struct LintViolationWithStatus {
+  const LintViolation *violation;
+  const LintRuleStatus *status;
+
+  LintViolationWithStatus(const LintViolation *v, const LintRuleStatus *s)
+      : violation(v), status(s) {}
+
+  bool operator<(const LintViolationWithStatus& r) const {
+    return (violation->token.text.data() < r.violation->token.text.data());
```
fangism: can remove extra ()

fangism: // compares addresses which correspond to locations within the same string

common/analysis/lint_rule_status.h
fangism: // compares addresses of violations, which correspond to substring locations

common/analysis/lint_rule_status.cc
```diff
@@ -58,6 +58,42 @@ void LintStatusFormatter::FormatLintRuleStatus(std::ostream* stream,
...
+    }
+  }
+
+  for (auto violation : violations) {
+    FormatViolation(stream,
+                    *violation.violation,
+                    base, path,
+                    violation.status->url,
+                    violation.status->lint_rule_name);
+    (*stream) << std::endl;
```
fangism: can remove extra ()

verilog/analysis/checkers/always_comb_blocking_rule.h
```diff
@@ -56,7 +56,7 @@ class AlwaysCombBlockingRule : public verible::SyntaxTreeLintRule {
   const Matcher always_comb_matcher_ =
       NodekAlwaysStatement(AlwaysCombKeyword());
 
-  std::vector<verible::LintViolation> violations_;
+  std::set<verible::LintViolation> violations_;
```
fangism: Replace `#include <vector>` with `#include <set>` in all of these rules' .h files.

verilog/analysis/checkers/always_comb_blocking_rule.cc
```diff
@@ -68,7 +68,7 @@ void AlwaysCombBlockingRule::HandleSymbol(const verible::Symbol& symbol,
           *node, NodeEnum::kNonblockingAssignmentStatement, 1);
 
       if (leaf.get().token_enum == TK_LE)
-        violations_.push_back(LintViolation(leaf, kMessage, match.context));
+        violations_.insert(LintViolation(leaf, kMessage, match.context));
```
fangism: Replace `#include <vector>` with `#include <set>` in these .cc files.

common/analysis/lint_rule_status.cc
tgorochowik: Sure, added

common/analysis/lint_rule_status.h
tgorochowik: Thanks, removed

common/analysis/lint_rule_status.cc
tgorochowik: Thanks, removed

tgorochowik: sure, added the comment

common/analysis/lint_rule_status.h
tgorochowik: sure, added the comment

common/analysis/lint_rule_status.cc
tgorochowik: removed

verilog/analysis/checkers/always_comb_blocking_rule.h
tgorochowik: Replaced

verilog/analysis/checkers/always_comb_blocking_rule.cc
tgorochowik: Sure, thanks, replaced in most of the files and added both where vectors were still used for internal logic

PR #109, opened 2019-12-18
lint: Ignore blocking statement in loop header

### LINKED ISSUE #107: always-ff-non-blocking rule wrongly flags for-loop-step assignment
On this file:

```systemverilog
module m;
 for(int i=0; i<Num_irq; i=i+1) begin    // linter flags the "i=i+1" assignment
 end
 end
 endmodule
```

```
$ verilog_lint --ruleset=none --rules=always-ff-non-blocking ~/testing/verilog/always_ff_for.sv
~/testing/verilog/always_ff_for.sv:3:27: Use only non-blocking assignments inside 'always_ff' sequential blocks. go/verilog-style#sequential-logic [always-ff-non-blocking]
```

Assignments inside for-loop-headers should be exempt from this check.

https://github.com/google/verible/blob/73936a870a7948663cc1009cd8b36a40f3ebe69d/verilog/analysis/checkers/always_ff_non_blocking_rule.cc#L56

In this function, one could examine the context, to see whether or not it contains `kForStepList` or `kLoopHeader`.

ldalek-antmicro:
Fix #107

### DIFF
--- a/verilog/analysis/checkers/always_ff_non_blocking_rule.cc
+++ b/verilog/analysis/checkers/always_ff_non_blocking_rule.cc
@@ -60,6 +60,10 @@ void AlwaysFFNonBlockingRule::HandleSymbol(const verible::Symbol &symbol,
   if (always_ff_matcher_.Matches(symbol, &manager)) {
     for (const auto &match :
          SearchSyntaxTree(symbol, NodekAssignmentStatement())) {
+
+      // this is intended to ignore assignments in for loop step statements like i=i+1
+      if (match.context.IsInside(NodeEnum::kLoopHeader)) continue ;
+
       auto *node = dynamic_cast<const verible::SyntaxTreeNode *>(match.match);
 
       if (node == nullptr) continue;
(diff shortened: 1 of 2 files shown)


verilog/analysis/checkers/always_ff_non_blocking_rule.cc
```diff
@@ -60,6 +60,9 @@ void AlwaysFFNonBlockingRule::HandleSymbol(const verible::Symbol &symbol,
   if (always_ff_matcher_.Matches(symbol, &manager)) {
     for (const auto &match :
          SearchSyntaxTree(symbol, NodekAssignmentStatement())) {
+
+      if (match.context.IsInside(NodeEnum::kLoopHeader)) continue ;
```
fangism: comment here: // this is intended to ignore assignments in for loop step statements like i=i+1

ldalek-antmicro: Added comment. Thanks

PR #111, opened 2019-12-19
Format modport contents

### LINKED ISSUE #84: Format modport ports
Desirable formatting:

```systemverilog
interface foo_if;
  modport producer(
      input ready,
      output data, valid, user, strobe, keep, last,
      import producer_reset, producer_tick, produce_blocked, produce
  );
  modport consumer(
      input data, valid, user, strobe, keep, last,
      output ready,
      import consumer_reset, consumer_tick, consume
  );
endinterface
```

Currently get:

```systemverilog
interface foo_if;
  modport producer(input ready, output data, valid, user, strobe, keep, last, import
                   producer_reset, producer_tick, produce_blocked, produce);
  modport consumer(input data, valid, user, strobe, keep, last, output ready, import
                   consumer_reset, consumer_tick, consume);
endinterface
```

Primarily, needs partitioning around `kModportSimplePortsDeclaration` nodes, with proper indentation (treated as hanging indentation 'wrapping').

rkapuscik:
Fixes #84.

### DIFF
--- a/verilog/formatting/tree_unwrapper.cc
+++ b/verilog/formatting/tree_unwrapper.cc
@@ -890,19 +890,9 @@ void TreeUnwrapper::Visit(const verible::SyntaxTreeLeaf& leaf) {
     StartNewUnwrappedLine();
   }
 
-  if (tag == ',' && CurrentUnwrappedLine().IsEmpty()) {
-    // Partition would begin with a comma,
-    // instead add this token to previous partition
-    AddTokenToCurrentUnwrappedLine();
-    MergeLastTwoPartitions();
-    // Put next token on new partition.
-    // If that token starts a new partition as well, it will just be reused.
-    StartNewUnwrappedLine();
-  } else {
-    // Advances NextUnfilteredToken(), and extends CurrentUnwrappedLine().
-    // Should be equivalent to AdvanceNextUnfilteredToken().
-    AddTokenToCurrentUnwrappedLine();
-  }
+  // Advances NextUnfilteredToken(), and extends CurrentUnwrappedLine().
+  // Should be equivalent to AdvanceNextUnfilteredToken().
+  AddTokenToCurrentUnwrappedLine();
 
   // Look-ahead to any trailing comments that are associated with this leaf,
   // up to a newline.
@@ -920,6 +910,13 @@ void TreeUnwrapper::Visit(const verible::SyntaxTreeLeaf& leaf) {
                                                            // kRegisterVariable
           })) {
         MergeLastTwoPartitions();
+      } else if (CurrentUnwrappedLine().Size() == 1) {
+        // Partition would begin with a comma,
+        // instead add this token to previous partition
+        MergeLastTwoPartitions();
+        // Put next token on new partition.
+        // If that token starts a new partition as well, it will just be reused.
+        //StartNewUnwrappedLine();
       }
       break;
     }
--- a/verilog/formatting/tree_unwrapper.cc
+++ b/verilog/formatting/tree_unwrapper.cc
@@ -348,7 +348,7 @@ void TreeUnwrapper::AppendTokenToPreviousUnwrappedLine() {
 
     ABSL_DIE_IF_NULL(target_leaf);
     auto* current_leaf = CurrentTokenPartition();
-    auto* current_leaf_parent = CurrentTokenPartition()->Parent();
+    auto* current_leaf_parent = current_leaf->Parent();
 
     // Verify continuity of token ranges between current and target leaves.
     CHECK(target_leaf->Value().TokensRange().end() ==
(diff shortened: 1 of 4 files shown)


verilog/formatting/tree_unwrapper.cc
```diff
@@ -340,6 +340,33 @@ verible::TokenWithContext TreeUnwrapper::VerboseToken(
...
 }
 
+void TreeUnwrapper::AppendTokenToPreviousUnwrappedLine() {
+    AddTokenToCurrentUnwrappedLine();
+    auto* target_leaf =
+      CurrentTokenPartition()->PreviousSibling()->RightmostDescendant();
+
+    ABSL_DIE_IF_NULL(target_leaf);
+    auto* current_leaf = CurrentTokenPartition();
+    auto* current_leaf_parent = CurrentTokenPartition()->Parent();
```
fangism: minor: `... = current_leaf->Parent();`

verilog/formatting/tree_unwrapper.cc
```diff
@@ -817,9 +850,15 @@ void TreeUnwrapper::Visit(const verible::SyntaxTreeLeaf& leaf) {
     StartNewUnwrappedLine();
   }
 
-  // Advances NextUnfilteredToken(), and extends CurrentUnwrappedLine().
-  // Should be equivalent to AdvanceNextUnfilteredToken().
-  AddTokenToCurrentUnwrappedLine();
+  if (tag == ',' && CurrentUnwrappedLine().IsEmpty()) {
+    // Partition would begin with a comma,
+    // instead add this token to previous partition
+    AppendTokenToPreviousUnwrappedLine();
```
fangism: I *think* this is equivalent to adding tokens to the current partition followed by merging the last two partitions, like here:
https://github.com/google/verible/blob/81d39b1e13688850fb918cef9b5c47f135a962dd/common/formatting/token_partition_tree.h#L78

However, I'm a little concerned that if a `/* comment */` appears before `,`, that tokens might get added to the wrong partition, in the wrong order.  Can we test that out?  It's possible I didn't handle this correctly with `;` elsewhere.

It may be safer as a policy only add to the current partition, and perform adjustments (like moving token span boundaries) after the partitions are finalized.

verilog/formatting/tree_unwrapper_test.cc
```diff
@@ -2150,10 +2150,31 @@ const TreeUnwrapperTestData kUnwrapInterfaceTestCases[] = {
...
+                          L(1,{"modport", "mp2", "("}),
+                          N(3, L(3, {"output", "c", ","})),
+                          N(3, L(3, {"input", "d"})),
+                          L(1, {")", ";"}))),
+        L(0, {"endinterface"}),
+    },
+    {
+        "interface with modport TF ports",
+        "interface foo_if;"
+        "modport mp1 (output a, input b, import c);"
```
fangism: Can you add a test case with `output a1, a2, input b1, b2, import c1, c2`?

rkapuscik: Added, along with test cases with comments.

verilog/formatting/tree_unwrapper.cc
rkapuscik: Fixed, thanks for spotting that.

verilog/formatting/tree_unwrapper.cc
rkapuscik: I changed the code to just use `MergeLastTwoPartitions`, I'm not sure why that wasn't working for me before.
I tested that on some examples with comments and it looks like it is moved correctly. I added some test cases to cover that.

> It may be safer as a policy only add to the current partition, and perform adjustments (like moving token span boundaries) after the partitions are finalized.

It is now done similarly to the `;` handling in https://github.com/google/verible/blob/master/verilog/formatting/tree_unwrapper.cc#L776. Since the commas can occur in the middle of a node, we would need to traverse the partition tree again looking for such lines. I think it could be done later, maybe in [MakeUnwrappedLinesWorklist](https://github.com/google/verible/blob/master/verilog/formatting/formatter.cc#L102)?

verilog/formatting/tree_unwrapper.cc
```diff
@@ -837,9 +839,19 @@ void TreeUnwrapper::Visit(const verible::SyntaxTreeLeaf& leaf) {
...
-  // Should be equivalent to AdvanceNextUnfilteredToken().
-  AddTokenToCurrentUnwrappedLine();
+  if (tag == ',' && CurrentUnwrappedLine().IsEmpty()) {
+    // Partition would begin with a comma,
+    // instead add this token to previous partition
+    AddTokenToCurrentUnwrappedLine();
+    MergeLastTwoPartitions();
+    // Put next token on new partition.
+    // If that token starts a new partition as well, it will just be reused.
+    StartNewUnwrappedLine();
```
fangism: The logic looks fine.  However, can we group in the switch statement below inside `case ','` (line 861)?  For this leaf-visiting function I prefer to keep the flow that shows distinct phases: pre-token (catch-up), token-handling (just `AddTokenToCurrentUnwrappedLine()` alone), post-token (look-ahead), partition tree adjustments (where the next switch statement is).

rkapuscik: Done, I moved it to the existing `case`.

PR #114, opened 2020-01-02
Check naming style of typedef'd enums

### LINKED ISSUE #108: Enforce naming style of typedef'd enums: lower_snake_case_t or lower_snake_case_e
Task: enforce that typedef'd enums are named lower_snake_case style and end with "_t" or "_e".

The section anchor in our style guide is `#enumerations`.

Note: this is slightly different from
https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md#enumerations
which permits only ending with "_t".

rkapuscik:
Fixes #108.


verilog/analysis/checkers/enum_name_style_rule_test.cc
```diff
@@ -0,0 +1,101 @@
...
+      {"typedef enum {foo, bar} ", {kToken, "e"}, ";"},
+      {"typedef enum {foo, bar} ", {kToken, "_T"}, ";"},
+      {"typedef enum {foo, bar} ", {kToken, "T"}, ";"},
+      {"typedef enum {foo, bar} ", {kToken, "_E"}, ";"},
+      {"typedef enum {foo, bar} ", {kToken, "E"}, ";"},
+      {"typedef enum {foo, bar} ", {kToken, "_"}, ";"},
+      {"typedef enum {foo, bar} ", {kToken, "foo_"}, ";"},
+  };
+  RunLintTestCases<VerilogAnalyzer, EnumNameStyleRule>(kTestCases);
+}
```
fangism: suggest a few more test cases:
* anonymous enum data declaration (no name to check at all)
* typedef of struct/union (should not be checked)
* enum typedef inside a class/module (some valid, invalid names)

verilog/CST/type.h
```diff
@@ -33,6 +33,11 @@ std::vector<verible::TreeSearchMatch> FindAllDataTypeDeclarations(
 std::vector<verible::TreeSearchMatch> FindAllTypeDeclarations(
     const verible::Symbol&);
 
+// Finds all node kEnumDataType declarations. Used for testing if the type
+// declaration is an enum.
+std::vector<verible::TreeSearchMatch> FindAllEnumDataTypeDeclarations(
```
fangism: Technically, not all appearances of the enum construct are type declarations, as some are anonymous in data declarations.  (It just happens that you're only using it inside type declarations at this time.)
Maybe name this `FindAllEnumDeclarations`?

verilog/analysis/checkers/enum_name_style_rule.cc
```diff
@@ -0,0 +1,84 @@
...
+
+using verible::GetStyleGuideCitation;
+using verible::LintRuleStatus;
+using verible::LintViolation;
+using verible::SyntaxTreeContext;
+
+absl::string_view EnumNameStyleRule::Name() {
+  return "enum-name-style";
+}
+const char EnumNameStyleRule::kTopic[] = "enum-conventions";
```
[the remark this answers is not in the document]
rkapuscik: Done.

verilog/analysis/checkers/enum_name_style_rule_test.cc
rkapuscik: Added, thanks for the suggestions!

verilog/CST/type.h
rkapuscik: You're right that they can be anonymous, but here I was going by node name (`kEnumDataType`) in `verilog_nonterminals.h`.
I'm open to renaming this, but then I suppose the `FindAllStructDataTypeDeclarations` and `FindAllUnionDataTypeDeclarations` should be renamed as well?

fangism: Also good point.  I agree that these cases of Struct/Union/Enum should be consistent with each other as well.  If it's not too much trouble, could we change those as well?  `FindAllStructDeclarations` and `FindAllEnumDeclarations`.  This way, we can use `TypeDeclarations` to explicitly refer to `typedef` constructs.

rkapuscik: Following your proposition, I renamed them to `kEnumType`, `kStructType`, `kUnionType`.

PR #115, opened 2020-01-03
Start new partition when else clause doesn't follow end keyword

### LINKED ISSUE #95: [line break/wrap] Erroneous line break in if/else without begin/else blocks
is:
```systemverilog
    if (enable) data |= interrupts;
        else data &= ~interrupts;
```
should:
```systemverilog
    if (enable) data |= interrupts;
    else data &= ~interrupts;
```

ldalek-antmicro:
Fix #95


ldalek-antmicro:
Fixed in https://github.com/google/verible/commit/e2e28a86b2300570fc88e8a095779235e1d8b018

PR #120, opened 2020-01-09
[line breaks] One-line if/else statements without begin/end blocks

### LINKED ISSUE #103: [line breaks] One-line if/else statements without begin/end blocks
splitting out from #98.
 
is:
```systemverilog
if (cfg.if_mode == Host)
  driver = HOST_DRIVER_T::type_id::create("driver", this);
else
  driver = DEVICE_DRIVER_T::type_id::create("driver", this);
```
should:
```systemverilog
if (cfg.if_mode == Host)  driver = HOST_DRIVER_T::type_id::create("driver", this);
else                      driver = DEVICE_DRIVER_T::type_id::create("driver", this);
```

ldalek-antmicro:
Fix #103


verilog/formatting/tree_unwrapper.cc
```diff
@@ -636,6 +636,22 @@ void TreeUnwrapper::Visit(const verible::SyntaxTreeNode& node) {
         break;
       }
 
+    case NodeEnum::kIfClause:
+    case NodeEnum::kElseClause: {
+      bool gotseq = false;
+      for (const auto& itr : node.children()) {
+        if (static_cast<NodeEnum>(itr->Tag().tag) == NodeEnum::kSeqBlock)
```
fangism: `.tag` could be either a node enumeration or leaf (token) enumeration depending on the type of the symbol, so to be safe, I would recommend explicitly checking whether the child is a tree-node first before using its tag.  

Also, do you need to scan every child?  or can you get away with examining specific children, positionally, using:

https://github.com/google/verible/blob/088be5950415a14f4036b366a826a7e2d9b4f679/common/text/tree_utils.h#L80

I think by construction in your updated verilog.y, you can check just the last child of these nodes for the statement block.  These are the kind of functions that best belong in the CST (statements?) library, like `GetIfClauseBody` and `GetElseClauseBody`.

common/formatting/unwrapped_line.h
```diff
@@ -74,6 +74,9 @@ class UnwrappedLine {
   // Extends PreFormatToken range spanned by this UnwrappedLine by one token.
   void SpanNextToken() { tokens_.extend_back(); }
 
+  // Extends PreFormatToken range spanned by this UnwrappedLine by one token.
```
fangism: Distinguish this comment from the one from `SpanNextToken`.
e.g. "... by one token on at the front" vs. "at the back".

common/text/tree_utils.h
```diff
@@ -106,6 +106,14 @@ const SyntaxTreeLeaf& GetSubtreeAsLeaf(const S& symbol,
       GetSubtreeAsSymbol(symbol, parent_must_be_node_enum, child_position)));
 }
 
+template <class S, class E>
+const E GetSubtreeNodeEnum(const S& symbol,
```
fangism: by-value return types should never have `const`, so just `E GetSubtreeNodeEnum`

verilog/formatting/formatter_test.cc
```diff
@@ -2589,23 +2582,18 @@ static const std::initializer_list<FormatterTestCase>
...
          "        a = y;\n"
-         "        if (a)\n"
-         "          b = c;\n"
-         "        else\n"
-         "          d = e;\n"
+         "        if (a) b = c;\n"
+         "        else d = e;\n"
          "      end\n"
          "    end\n"
          "  end\n"
```
fangism: one more suggested test case:

```
if () ...
else if () ...
else ...
```

(uses both else-if and final-else)

verilog/formatting/tree_unwrapper.cc
```diff
@@ -632,6 +632,64 @@ void TreeUnwrapper::Visit(const verible::SyntaxTreeNode& node) {
...
+      switch (subtag) {
+      case NodeEnum::kSeqBlock:
+      case NodeEnum::kConditionalStatement:
+        // Extend current token parition by begin keyword and if keyword
+        // Plus keep else-if-else... structure flat
+        TraverseChildren(node);
+        break;
+
+      default:
+        // Otherwise create ndented section by ourselves.
```
fangism: "indented"

verilog/formatting/tree_unwrapper.cc
```diff
@@ -632,6 +632,64 @@ void TreeUnwrapper::Visit(const verible::SyntaxTreeNode& node) {
...
+      if (tag == NodeEnum::kElseClause)
+        subnode = &GetSubtreeAsNode(node, NodeEnum::kElseClause, 1);
+      else
+        subnode = &node;
+
+      const auto subtag = GetSubtreeNodeEnum(*subnode, NodeEnum::kElseBody, 0);
+      switch (subtag) {
+      case NodeEnum::kSeqBlock:
+      case NodeEnum::kConditionalStatement:
+        // Extend current token parition by begin keyword and if keyword
```
fangism: "partition"

verilog/formatting/tree_unwrapper.cc
```diff
@@ -892,6 +950,31 @@ void TreeUnwrapper::Visit(const verible::SyntaxTreeNode& node) {
       // else close-out current token partition?
       break;
     }
+    case NodeEnum::kIfClause: {
+      if (Context().DirectParentsAre(
+              {NodeEnum::kConditionalStatement, NodeEnum::kElseBody}) &&
+          (GetSubtreeNodeEnum(GetSubtreeAsNode(node, NodeEnum::kIfClause, 1),
+            NodeEnum::kIfBody, 0) != NodeEnum::kSeqBlock)) {
+
+        // Merge [else] token into [if ( )] parition
```
fangism: "partition" (elsewhere too)

verilog/formatting/tree_unwrapper.cc
```diff
@@ -892,6 +950,31 @@ void TreeUnwrapper::Visit(const verible::SyntaxTreeNode& node) {
       // else close-out current token partition?
       break;
     }
+    case NodeEnum::kIfClause: {
+      if (Context().DirectParentsAre(
+              {NodeEnum::kConditionalStatement, NodeEnum::kElseBody}) &&
```
fangism: Would it make more sense to perform these transformations as a case of `kElseClause` and `kConditionalStatement` (looking downward instead of up the context stack)?

verilog/formatting/tree_unwrapper_test.cc
```diff
@@ -3398,10 +3407,12 @@ const TreeUnwrapperTestData kUnwrapFunctionTestCases[] = {
...
-                                  L(1, {"if", "(", "zz", ")"}),
-                                  NL(2, {"return", "0", ";"}),
-                                  L(1, {"else", "if", "(", "yy", ")"}),
-                                  NL(2, {"return", "1", ";"}))),
+                                  N(1,  //
+                                    L(1, {"if", "(", "zz", ")"}),
+                                    NL(2, {"return", "0", ";"})),
+                                  N(1,  //
+                                    L(1, {"else", "if", "(", "yy", ")"}),
+                                    NL(2, {"return", "1", ";"})))),
```
fangism: suggested test case:
* two else if clauses
* else if with trailing else clause

In some of the test case comments, point out the flatness that is achieved.

common/formatting/unwrapped_line.h
ldalek-antmicro: Done

common/text/tree_utils.h
ldalek-antmicro: Done

verilog/formatting/formatter_test.cc
ldalek-antmicro: Done

verilog/formatting/tree_unwrapper.cc
ldalek-antmicro: Fixed

verilog/formatting/tree_unwrapper.cc
ldalek-antmicro: Fixed

verilog/formatting/tree_unwrapper.cc
ldalek-antmicro: Fixed

verilog/formatting/tree_unwrapper.cc
ldalek-antmicro: I've rewritten it.

verilog/formatting/tree_unwrapper_test.cc
ldalek-antmicro: Added more tests and comments.

verilog/formatting/tree_unwrapper.cc
```diff
@@ -892,6 +950,21 @@ void TreeUnwrapper::Visit(const verible::SyntaxTreeNode& node) {
       // else close-out current token partition?
       break;
     }
+    case NodeEnum::kConditionalStatement: {
+      CurrentTokenPartition()->Parent()->ApplyPreOrder([](verible::VectorTree<UnwrappedLine>& node) {
+        const auto& range = node.Value().TokensRange();
+        if ((range.back().TokenEnum() == yytokentype::TK_else) &&
+            (range.end()->TokenEnum() == yytokentype::TK_if)) {
```
fangism: Dereferencing an `end()` iterator is usually suspicious, as it's usually one position past the last valid element, but I understand you're using the fact that these are iterators into a longer sequence.
No change requested for now.

PR #124, opened 2020-01-14
Adding missing unique0 case

AaronKel:
Trivial change as per 12.5.3 of 1800-2017 

unique_priority ::= unique | unique0 | priority

case_statement ::=
[ unique_priority ] case_keyword ( case_expression )

### DIFF
--- a/verilog/parser/verilog.y
+++ b/verilog/parser/verilog.y
@@ -6446,6 +6446,8 @@ case_modifier_opt
     { $$ = move($1); }
   | TK_unique
     { $$ = move($1); }
+  | TK_unique0
+    { $$ = move($1); } 
   | /* empty */
     { $$ = nullptr; }
   ;


AaronKel:
@googlebot I signed it!

verilog/parser/verilog.y
```diff
@@ -6446,6 +6446,8 @@ case_modifier_opt
     { $$ = move($1); }
   | TK_unique
     { $$ = move($1); }
+  | TK_unique0
+    { $$ = move($1); }
```
fangism: Having reviewed the LRM grammar, `case_modifier_opt` should be no different than `unique_priority_opt` on line 6707, so let's just replace the former with the latter everywhere in this file.  In the `/* empty */` case for `unique_priority_opt` please add the missing `{ $$ = nullptr; }`.

PR #139, opened 2020-01-20
Check for multiple module declarations in file

### LINKED ISSUE #130: Linter does not report more than one module per file
As per first bullet in [Basic Style Elements](https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md#basic-style-elements) section of the style guide:
* Use SystemVerilog-2012 conventions, files named as module.sv, **one file per module**

There is no corresponding linter rule.

Example:
```verilog
module first;
endmodule
module second;
endmodule
```

rkapuscik:
Fixes #130.

### DIFF
--- a/verilog/analysis/checkers/module_filename_rule.cc
+++ b/verilog/analysis/checkers/module_filename_rule.cc
@@ -74,13 +74,19 @@ void ModuleFilenameRule::Lint(const TextStructureView& text_structure,
   // Find all module declarations.
   auto module_matches = FindAllModuleDeclarations(*tree);
 
+  // If there are no modules in this source unit, suppress finding.
+  if (module_matches.empty()) return;
+
   // Remove nested module declarations
-  module_matches.erase(std::remove_if(module_matches.begin(),
-              module_matches.end(),
+  std::vector<verible::TreeSearchMatch> module_cleaned;
+  module_cleaned.reserve(module_matches.size());
+  std::back_insert_iterator<std::vector<verible::TreeSearchMatch>> back_it(module_cleaned);
+  std::remove_copy_if(module_matches.begin(),
+                      module_matches.end(),
+                      back_it,
               [](verible::TreeSearchMatch& m) {
                   return m.context.IsInside(NodeEnum::kModuleDeclaration);
-              }),
-          module_matches.end());
+              });
 
   // See if any names match the stem of the filename.
   const absl::string_view basename = verible::file::Basename(filename);
@@ -90,19 +96,17 @@ void ModuleFilenameRule::Lint(const TextStructureView& text_structure,
   if (unitname.empty()) return;
 
   auto matching_module_iter =
-      std::find_if(module_matches.begin(), module_matches.end(),
+      std::find_if(module_cleaned.begin(), module_cleaned.end(),
                    [=](const verible::TreeSearchMatch& m) {
                      return ModuleNameMatches(*m.match, unitname);
                    });
 
-  // If there are no modules in this source unit, suppress finding.
-  if (module_matches.empty()) return;
 
   // If there is at least one module with a matching name, suppress finding.
-  if (matching_module_iter != module_matches.end()) return;
+  if (matching_module_iter != module_cleaned.end()) return;
 
   // Only report a violation on the last module declaration.
-  const auto& last_module_id = GetModuleNameToken(*module_matches.back().match);
+  const auto& last_module_id = GetModuleNameToken(*module_cleaned.back().match);
   violations_.insert(verible::LintViolation(
       last_module_id, absl::StrCat(kMessage, "\"", unitname, "\"")));
 }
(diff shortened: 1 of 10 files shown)


verilog/tools/lint/testdata/defparam_usage.sv
```diff
@@ -1,3 +1,4 @@
+/// verilog_lint: waive-start one-module-per-file
 module defparam_usage #(parameter int MY_PARAM = 0);
```
fangism: Instead of waiving, we could just delete the first definition.  This module definition is not actually needed to trigger the rule.

verilog/analysis/checkers/one_module_per_file_rule_test.cc
```diff
@@ -0,0 +1,74 @@
...
+       "module ", {kToken, "moo"}, "; endmodule"},
+      {"class c; endclass\n"
+       "module foo; endmodule\n"
+       "module ", {kToken, "moo"}, "; endmodule"},
+      {"package q; endpackage\n"
+       "module foo; endmodule\n"
+       "module ", {kToken, "moo"}, "; endmodule\n"},
+  };
+  RunLintTestCases<VerilogAnalyzer, OneModulePerFileRule>(kTestCases, "");
+}
```
fangism: According to the LRM, it looks like module_declaration is allowed as a 'non_port_module_item', which suggests that modules may be nested.  I think we may want to restrict the findings to only outermost module declarations.
Suggested test case:

```systemverilog
module outer;
module inner;
endmodule : inner
endmodule : outer
```

The `module-filename` rule probably needs to be updated to account for this as well, but that can be a separate PR if you like.

verilog/tools/lint/testdata/defparam_usage.sv
rkapuscik: Changed.

verilog/analysis/checkers/one_module_per_file_rule_test.cc
rkapuscik: Good find, thanks! I've adapted both rules to account for that.

verilog/analysis/checkers/module_filename_rule.cc
```diff
@@ -73,6 +74,14 @@ void ModuleFilenameRule::Lint(const TextStructureView& text_structure,
...
   auto module_matches = FindAllModuleDeclarations(*tree);
 
+  // Remove nested module declarations
+  module_matches.erase(std::remove_if(module_matches.begin(),
+              module_matches.end(),
+              [](verible::TreeSearchMatch& m) {
+                  return m.context.IsInside(NodeEnum::kModuleDeclaration);
+              }),
+          module_matches.end());
+
```
fangism: Avoid vector::erase() in this case for its potential O(N^2) cost.
Rewrite as `std::copy_if` or `std::remove_copy_if` using a temporary vector.

verilog/analysis/checkers/one_module_per_file_rule.cc
```diff
@@ -0,0 +1,91 @@
...
+
+  auto module_matches = FindAllModuleDeclarations(*tree);
+
+  // Nested module declarations are allowed, remove those
+  module_matches.erase(std::remove_if(module_matches.begin(),
+              module_matches.end(),
+              [](verible::TreeSearchMatch& m) {
+                  return m.context.IsInside(NodeEnum::kModuleDeclaration);
+              }),
+          module_matches.end());
```
fangism: Same as other file.

verilog/analysis/checkers/module_filename_rule.cc
rkapuscik: Done.

verilog/analysis/checkers/one_module_per_file_rule.cc
rkapuscik: Done.

verilog/analysis/checkers/module_filename_rule.cc
```diff
@@ -73,6 +74,20 @@ void ModuleFilenameRule::Lint(const TextStructureView& text_structure,
...
+  // If there are no modules in this source unit, suppress finding.
+  if (module_matches.empty()) return;
+
+  // Remove nested module declarations
+  std::vector<verible::TreeSearchMatch> module_cleaned;
+  module_cleaned.reserve(module_matches.size());
+  std::back_insert_iterator<std::vector<verible::TreeSearchMatch>> back_it(module_cleaned);
+  std::remove_copy_if(module_matches.begin(),
+                      module_matches.end(),
+                      back_it,
```
fangism: FYI, can replace `back_it` with `std::back_inserter(module_cleaned)` (easier to write/read).

verilog/analysis/checkers/one_module_per_file_rule.cc
```diff
@@ -0,0 +1,97 @@
...
+      return;
+  }
+
+  // Nested module declarations are allowed, remove those
+  std::vector<verible::TreeSearchMatch> module_cleaned;
+  module_cleaned.reserve(module_matches.size());
+  std::back_insert_iterator<std::vector<verible::TreeSearchMatch>> back_it(module_cleaned);
+  std::remove_copy_if(module_matches.begin(),
+                      module_matches.end(),
+                      back_it,
```
fangism: Same suggestion.

PR #140, opened 2020-01-20
Forbid consecutive null statements

### LINKED ISSUE #132: Detect consecutive null statements like `;;`
A null statement is a lone `;`, often found in procedural timing constraint statements.
We can easily detect redundant null-statements like "@(posedge foo);;".

Note that "for (;;)" are not considered null-statements, and will not be considered.

Look for `kNullStatement`s in the CST.

ldalek-antmicro:
Fixes #132

### DIFF
--- a/verilog/analysis/checkers/forbid_consecutive_null_statements_rule_test.cc
+++ b/verilog/analysis/checkers/forbid_consecutive_null_statements_rule_test.cc
@@ -1,0 +1,148 @@
+// Copyright 2017-2020 The Verible Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "verilog/analysis/checkers/forbid_consecutive_null_statements_rule.h"
+
+#include <initializer_list>
+
+#include "gtest/gtest.h"
+#include "common/analysis/linter_test_utils.h"
+#include "common/analysis/syntax_tree_linter_test_utils.h"
+#include "common/text/symbol.h"
+#include "verilog/CST/verilog_nonterminals.h"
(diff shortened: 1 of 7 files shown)


verilog/analysis/checkers/forbid_consecutive_null_statements_rule_test.cc
```diff
@@ -0,0 +1,54 @@
...
+using verible::RunLintTestCases;
+
+TEST(ForbidConsecutiveNullStatementsRule, FunctionFailures) {
+  auto kToken = ';';
+  const std::initializer_list<LintTestCase> kTestCases = {
+      {""},
+      {"module m;\nendmodule\n"},
+      {"module m;\nalways_ff @(posedge foo) ; endmodule\n"},
+      {"module m;\nalways_ff @(posedge foo) ;", {kToken, ";"}, " endmodule\n"},
+      {"module m;\nalways_ff @(posedge foo) ;", {kToken, ";"}, {kToken, ";"}, " endmodule\n"},
```
fangism: Do you think we should only report the second `;` in a string of `;`s as a violation?  Feels uncommon enough to not matter.  I leave this up to you.

verilog/analysis/checkers/forbid_consecutive_null_statements_rule_test.cc
```diff
@@ -0,0 +1,54 @@
...
+      {"module m;\ninitial begin end\nendmodule"},
+      {"module m;\ninitial begin for (;;) ; end\nendmodule"},
+      {"module m;\ninitial begin for (;;) ;", {kToken, ";"}, " end\nendmodule"},
+      {"module m;\ninitial begin for (;;) ;", {kToken, ";"}, {kToken, ";"}, " end\nendmodule"},
+  };
```
fangism: suggested test cases:
* Add a comment between two `;`s, like `; /* comment */ ;`
* Consecutive inside begin/end blocks.
* Inside function (including for loops)
* Inside task (including for loops)
* One nested for-loop example.

verilog/analysis/checkers/forbid_consecutive_null_statements_rule_test.cc
ldalek-antmicro: IMHO reporting only second and later null statements looks better. The problem is in redundant statements not in whole string.

verilog/analysis/checkers/forbid_consecutive_null_statements_rule_test.cc
ldalek-antmicro: Done

Issue #154, opened 2020-01-29
end-else-if-begin should be on same line (partition)


fangism:
Input:

```systemverilog
module m;
if (x) begin:z1
assign x=y;
end
else
if (y) begin:z2
assign z=y;
end
endmodule
```

Currently formats to:

```systemverilog
module m;
  if (x) begin : z1
    assign x = y;
  end else
  if (y) begin : z2
    assign z = y;
  end
endmodule
```

but should be:

```systemverilog
module m;
  if (x) begin : z1
    assign x = y;
  end else if (y) begin : z2
    assign z = y;
  end
endmodule
```

With end-else-if-begin on the same line (partition).

fangism:
b/148505401

ldalek-antmicro:
Hey @fangism,
@hzeller added commit https://github.com/google/verible/commit/ceda448ae637b7665fbd1be6f0672ea172b1f4f0 that fixes this issue:

<pre>
module m;
if (x) begin:z1
assign x=y;
end
else
if (y) begin:z2
assign z=y;
end
endmodule
</pre>

token partition tree:
<pre>
Full token partition tree:
{ ([<auto>]) @{}, policy: always-expand
  { ([<auto>]) @{0}, policy: fit-else-expand
    { ([module m ;]) }
  }
  { (>>[<auto>]) @{1}, policy: always-expand
    { (>>[<auto>]) @{1,0}, policy: fit-else-expand
      { (>>[if ( x ) begin : z1]) }
      { (>>>>[<auto>]) @{1,0,1}, policy: always-expand
        { (>>>>[<auto>]) @{1,0,1,0}, policy: fit-else-expand
          { (>>>>[assign x = y ;]) }
        }
      }
      { (>>[end else if ( y ) begin : z2]) }
      { (>>>>[<auto>]) @{1,0,3}, policy: always-expand
        { (>>>>[<auto>]) @{1,0,3,0}, policy: fit-else-expand
          { (>>>>[assign z = y ;]) }
        }
      }
      { (>>[end]) }
    }
  }
  { ([endmodule]) }
}
</pre>

output:
<pre>
module m;
  if (x) begin : z1
    assign x = y;
  end else if (y) begin : z2
    assign z = y;
  end
endmodule
</pre>

Can you confirm?

ldalek-antmicro:
More examples:
----
Input:
<pre>
module zc;
initial begin
if (x) begin
end : l1
else begin
end
end
endmodule
</pre>
Output:
[...]

PR #158, opened 2020-01-29
Fix end-[label]-else-if-begin statements

### LINKED ISSUE #155: end-label-else-begin has excess wrapping indentation
Input:

```systemverilog
module m;
if (x) begin
end : l1
else begin
end
endmodule
```

formats to:

```systemverilog
module m;
  if (x) begin
  end : l1
      else begin
  end
endmodule
```

but should be:

```systemverilog
module m;
  if (x) begin
  end : l1
  else begin
  end
endmodule
```

The `else` should not be wrap-indented.

ldalek-antmicro:
Fixes #155 and #154


verilog/formatting/tree_unwrapper.cc
```diff
@@ -701,6 +711,74 @@ void TreeUnwrapper::Visit(const verible::SyntaxTreeNode& node) {
       }
       break;
     }
+    case NodeEnum::kGenerateIfClause:
+    case NodeEnum::kGenerateIfBody: {
+      const verible::SyntaxTreeNode* subnode;
+
+      if (tag == NodeEnum::kGenerateIfClause)
+        subnode = &GetSubtreeAsNode(node, NodeEnum::kGenerateIfClause, 1);
```
fangism: In general, I'd like to start moving these subtree accesses into CST helper functions (one-liners) to isolate some of the structural details.  Perhaps this could be done as a follow-up clean-up.

verilog/parser/verilog.y
```diff
@@ -5431,11 +5431,20 @@ loop_generate_construct
...
+               MakeTaggedNode(N::kGenerateIfBody, $2)),
+           MakeTaggedNode(N::kGenerateElseClause,
+               $3,
+               MakeTaggedNode(N::kGenerateElseBody, $4))); }
   | generate_if generate_item %prec less_than_TK_else
-    { $$ = MakeTaggedNode(N::kConditionalGenerateConstruct, $1, $2); }
+    { $$ = MakeTaggedNode(N::kConditionalGenerateConstruct,
+             MakeTaggedNode(N::kGenerateIfClause,
+               MakeTaggedNode(N::kGenerateIfHeader, $1),
+               MakeTaggedNode(N::kGenerateIfBody, $2))); }
```
fangism: add a nullptr for where the else-clause would go (for structural consistency)

verilog/CST/verilog_nonterminals.h
```diff
@@ -296,7 +296,13 @@ enum class NodeEnum {
   kContinuousAssign,
   kLoopGenerateConstruct,
   kConditionalGenerateConstruct,
+  kConditionalGenerateConstructCase,
```
fangism: Suggest naming this: `kCaseGenerateConstruct` to resemble other things starting with `kCase`

verilog/parser/verilog.y
ldalek-antmicro: Shouldn't we do such thing with `kConditionalStatement` https://github.com/google/verible/blob/master/verilog/parser/verilog.y#L6525?

verilog/CST/verilog_nonterminals.h
ldalek-antmicro: Done

verilog/parser/verilog.y
fangism: Yes, in general, for all uses of the same `kNodeEnum`.  Some of the `MakeTaggedNode` calls get wrapped into specific functions that expect a fixed number of positional arguments to help with consistency.

hzeller:
coverage-run

I ran some coverage on the change in question and it looks like not all branches have been covered in tests (see the red/green markings over the line-numbers on the right).

Could you add some more testing to make sure we verified all code-paths ?

Thanks!

ldalek-antmicro:
Hey @hzeller,
I've added more tests which should cover untested code.

btw. You've past nice looking coverage report. Can you tell how did you generate it?

hzeller:
Looks much better

coverage2

The coverage output comes from some internal tool. I want to see if we can bring up something for our CI

Issue #31, opened 2019-11-25
 [token partitioning] clocking declarations


msfschaffner:
should:
```systemverilog
clocking cb @(posedge clk); 
endclocking 

clocking cb2 @(posedge clk); 
endclocking
```
is:
```systemverilog
clocking cb @(posedge clk); endclocking clocking cb2 @(posedge clk); endclocking
```

msfschaffner:
b/145157871

msfschaffner:
Fixed.

PR #171, opened 2020-02-03
Implemented clocking block with ports formatting

### LINKED ISSUE #161: Incorrect line-break when formatting long clocking block
Running verible_format on:
```
  clocking cb @(posedge clk);
    input req;
    input rwn;
    input addr;
    input datain;
    input ack;
    input dataout;
    input error;
  endclocking
```
outputs this
```
  clocking cb @(posedge clk); input req; input rwn; input addr; input datain; input ack; input dataout
      ; input error;
  endclocking
```
I assume the Google style guide states that this should be a one-liner?

Three questions/problems:
- Can I can tune this behavior (as this block has a clear 'endblocking' statement I don't see the benefit of having it all on one line)? Or disable it?
- When this is processed by verible_lint, it complains about the max line length being exceeded (103 characters line length).
- Even when following the one-liner approach, I would expect this as output:
```
  clocking cb @(posedge clk); input req; input rwn; input addr; input datain; input ack;
      input dataout; input error;
  endclocking
```

mvanrompuy:
Fixes #161

### DIFF
--- a/verilog/formatting/formatter_test.cc
+++ b/verilog/formatting/formatter_test.cc
@@ -887,16 +887,17 @@ static const std::initializer_list<FormatterTestCase> kFormatterTestCases = {
         "  end\n"
         "endmodule\n",
     },
-
     {
         // clocking declarations in modules
         " module mcd ; "
         "clocking   cb @( posedge clk);\t\tendclocking "
-        "clocking cb2   @ (posedge  clk\n); endclocking endmodule",
+        "clocking cb2   @ (posedge  clk\n); input a; output b; endclocking endmodule",
         "module mcd;\n"
         "  clocking cb @(posedge clk);\n"
         "  endclocking\n"
         "  clocking cb2 @(posedge clk);\n"
+        "    input a;\n"
+        "    output b;\n"
         "  endclocking\n"
         "endmodule\n",
     },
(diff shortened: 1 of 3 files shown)


verilog/formatting/formatter_test.cc
```diff
@@ -887,28 +887,31 @@ static const std::initializer_list<FormatterTestCase> kFormatterTestCases = {
...
-        "clocking cb2   @ (posedge  clk\n); endclocking endmodule",
+        "clocking cb2   @ (posedge  clk\n); input a; output b; endclocking endmodule",
         "module mcd;\n"
         "  clocking cb @(posedge clk);\n"
         "  endclocking\n"
         "  clocking cb2 @(posedge clk);\n"
+        "    input a;\n"
+        "    output b;\n"
         "  endclocking\n"
         "endmodule\n",
```
fangism: Let's keep both versions of this test, empty clocking declaration (still syntactically valid), and with the new ports you added.

verilog/formatting/tree_unwrapper_test.cc
```diff
@@ -1170,15 +1170,24 @@ const TreeUnwrapperTestData kUnwrapModuleTestCases[] = {
...
-        "clocking cb @(posedge clk); endclocking "
-        "clocking cb2 @(posedge clk); endclocking endmodule",
+        "module mcd;\n"
+        "  clocking cb @(posedge clk);\n"
+        "  endclocking\n"
+        "  clocking cb2 @(posedge clk);\n"
+        "    input a;\n"
+        "    output b;\n"
+        "  endclocking\n"
+        "endmodule",
```
fangism: Again, keep the original test, but copy-modify the the new one you have.

verilog/formatting/formatter_test.cc
mvanrompuy: Doesn't the first one ("cb") still matches the empty clocking block case?

I noticed that the original test seemed to have two clocking blocks to handle #31 (incorrect handling of two clocking blocks). This still seemed to be covered by my adjusted version, which is why I adjusted the second block instead of adding a third case.

But I'll add a "cb3" and restore the original test if that is preferred.

fangism: Yes, even though the larger test 'covers' multiple aspects, we would still prefer to have these tests as small as possible so that a failing test case is likely to narrow down the problem.  For example, two empty clocking declarations tests specifically for the interaction between two consecutive declarations.  If you like, the 'new' test case can just have one clocking declaration, because what you are fixing is internal to a single declaration.

mvanrompuy:
Split the test into separate cases (apologies for the mess-up on the initial push).

fangism:
> Split the test into separate cases (apologies for the mess-up on the initial push).

No apologies necessary, your fix is appreciated!

Issue #118, opened 2020-01-09
DataDeclaration with QualifierList breaks formatter


rkapuscik:
When unwrapping a `DataDeclaration` partition declaring a single variable with qualifier (`const`/`static`/`var`), verible formatter crashes during post-traversal adjustments.
Example:
```verilog
module foo;
initial begin
    const int a = 0;
end
endmodule
```
Gives:
```bash
$ ./bazel-bin/verilog/tools/formatter/verilog_format failing3.sv
F verilog/formatting/tree_unwrapper.cc:856] Check failed: left_uwline->TokensRange().end() == right_uwline.TokensRange().begin()
[1]    2589 abort (core dumped)  ./bazel-bin/verilog/tools/formatter/verilog_format failing3.sv     
```

This is caused by partition flattening in https://github.com/google/verible/blob/master/verilog/formatting/tree_unwrapper.cc#L846, which happens despite the fact that the `QualifierList` partition does not have any children.

Encountered during testing #117.

fangism:
b/147427426

PR #176, opened 2020-02-04
Format bind directives

### LINKED ISSUE #145: bind directives should format similarly to module instantiations
Input:

```systemverilog
bind blaaaah fooooo #(.MaxCount(MaxCount), .MaxDelta(MaxDelta)) bar (.clk(clk), .rst(rst), .value(value));
```

Formats to:

```systemverilog
bind blaaaah fooooo #(
    .MaxCount(MaxCount), .MaxDelta(MaxDelta)
)
    bar (.clk(clk), .rst(rst), .value(value));
```

but should more closely resemble the single-instance special case of a module instantiation statement (in this case):

```systemverilog
bind blaaaah fooooo #(
    .MaxCount(MaxCount), .MaxDelta(MaxDelta)
) bar (
    .clk(clk), .rst(rst), .value(value)
);
```

(In this example, parameters and ports happen to fit on one line.)

With this grammar construct for bind_directive:

```
  'bind' bind_target_instance bind_instantiation ';'
```

'bind' and bind_target_instance ("blaaaah" in the above example) should just be attached to the same partition as the type "fooooo" of the bind_instantiation (which is a module_instantiation).

See how `kDataDeclaration` handles this in the fix-up phase in `tree_unwrapper.cc`. I suspect most or all of that case can be re-used.

https://github.com/google/verible/blob/f03aeeb314a393d8df34729e328109f1ef480eba/verilog/formatting/tree_unwrapper.cc#L888

rkapuscik:
Fixes #145.

### DIFF
--- a/common/formatting/unwrapped_line.h
+++ b/common/formatting/unwrapped_line.h
@@ -82,6 +82,9 @@ class UnwrappedLine {
   // Extends PreFormatToken range up to the given token (exclusive).
   void SpanUpToToken(token_iterator iter) { tokens_.set_end(iter); }
 
+  // Extends PreFormatToken range to start from the given token (inclusive).
+  void SpanFromToken(token_iterator iter) { tokens_.set_begin(iter); }
+
   int IndentationSpaces() const { return indentation_spaces_; }
 
   void SetIndentationSpaces(int spaces) { indentation_spaces_ = spaces; }
--- a/verilog/parser/verilog.y
+++ b/verilog/parser/verilog.y
@@ -7037,9 +7037,13 @@ bind_target_instance
   ;
 bind_instantiation
   /* similar to block_item_decl instantiation */
-  : class_id gate_instance_or_register_variable_list
+  : bind_class_id gate_instance_or_register_variable_list
     { $$ = MakeTaggedNode(N::kBindTargetInstance, $1, $2); }
     /* covers: program, module, interface, checker instantiation */
+  ;
+bind_class_id
+  : class_id
+    { $$ = MakeTaggedNode(N::kInstantiationType, $1); }
   ;
 
 clocking_declaration
--- a/verilog/formatting/tree_unwrapper.cc
+++ b/verilog/formatting/tree_unwrapper.cc
@@ -984,18 +984,6 @@ void TreeUnwrapper::Visit(const verible::SyntaxTreeNode& node) {
 
       auto& instance_list_partition = children.back();
       const auto& instance_type_partition = children.back().PreviousSibling();
-      //if (tag==NodeEnum::kBindDirective) {
-      //  // append target instance to the bind directive partition
-      //  instance_type_partition->FlattenOnce();
-      //    data_declaration_partition.MergeConsecutiveSiblings(
-      //        0, [](UnwrappedLine* left_uwline,
-      //                          const UnwrappedLine& right_uwline) {
-      //          CHECK(left_uwline->TokensRange().end() ==
-      //                right_uwline.TokensRange().begin());
-      //          left_uwline->SpanUpToToken(right_uwline.TokensRange().end());
-      //        });
-      //  // update references
-      //}
 
       if (instance_list_partition.Children().size() == 1) {
         // Flatten partition tree by one level.
@@ -1037,6 +1025,19 @@ void TreeUnwrapper::Visit(const verible::SyntaxTreeNode& node) {
           // just flatten it. Manually flatten subpartitions.
           instance_type_partition->HoistOnlyChild();
           instance_list_partition.HoistOnlyChild();
+          if (instance_list_partition.Children().empty()) {
+            //TODO: rename this, we just move a leaf to its sibling here
+            AttachTrailingSemicolonToPreviousPartition();
+          }
+          if (instance_type_partition->Children().empty()) {
+            data_declaration_partition.MergeConsecutiveSiblings(
+                0, [](UnwrappedLine* left_uwline,
+                                  const UnwrappedLine& right_uwline) {
+                  CHECK(left_uwline->TokensRange().end() ==
+                        right_uwline.TokensRange().begin());
+                  left_uwline->SpanUpToToken(right_uwline.TokensRange().end());
+                });
+          }
         }
       }
       break;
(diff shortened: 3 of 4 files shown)


rkapuscik:
@fangism could you take a look if this is what you had in mind? The fix-up was indeed mostly similar to the module instances. Are there any corner cases to look out for?

fangism:
Kind of similar.  Here's how I was thinking about it:

Consider this code:

```systemverilog
module m;
bind blaaaah fooooo bar;  // bind with module instantiation
fooooo bar();  // module instantiation
endmodule
```

Their CSTs for the bind and the plain module instantiation are very similar (respectively):

```
      Node (tag: kBindDirective) {
        (#"bind" @124-128: "bind")
        Node (tag: kReference) {
          Node (tag: kLocalRoot) {
            Node (tag: kUnqualifiedId) {
              (#SymbolIdentifier @129-136: "blaaaah")
            }
          }
        }
        Node (tag: kBindTargetInstance) {
          Node (tag: kUnqualifiedId) {
            (#SymbolIdentifier @137-143: "fooooo")
          }
          Node (tag: kGateInstanceRegisterVariableList) {
            Node (tag: kRegisterVariable) {
              (#SymbolIdentifier @144-147: "bar")
              Node (tag: kUnpackedDimensions) {
              }
            }
          }
        }
        (#';' @147-148: ";")
      }
```


```
      Node (tag: kDataDeclaration) {
        Node (tag: kInstantiationBase) {
          Node (tag: kInstantiationType) {
            Node (tag: kReferenceCallBase) {
              Node (tag: kReference) {
                Node (tag: kLocalRoot) {
                  Node (tag: kUnqualifiedId) {
                    (#SymbolIdentifier @150-156: "fooooo")
                  }
                }
              }
            }
          }
          Node (tag: kGateInstanceRegisterVariableList) {
            Node (tag: kGateInstance) {
              (#SymbolIdentifier @157-160: "bar")
              Node (tag: kUnpackedDimensions) {
              }
              Node (tag: kParenGroup) {
                (#'(' @160-161: "(")
                (#')' @161-162: ")")
              }
            }
          }
        }
        (#';' @162-163: ";")
      }
```
 
The `kBindTargetInstance` is similar to `kInstantiationBase`, with one difference being that the latter has a level of `kInstantiationType`, but the former doesn't.  Consider making these more similar in verilog.y.

Notice that `kDataDeclaration` is just one level of wrapping around `kInstantiationBase`.  In cases like this, if you happen to make additional levels of token partitioning (0-indented sections) during CST node traversal, it is easy to flatten unwanted levels using `HoistOnlyChild` in the post-traversal fix-up phase.

If you create a partition around a `kBindDirective` node, and a partition around `kBindTargetInstance` (made to shape like a `kInstantiationBase`), you have a partition structure that looks like:

```
[bind blaaaah
  [fooooo
    [bar]
  ]
]
```

Flattening the `kBindTargetInstance` would give you:

```
[bind blaaaah
 fooooo
  [bar]
]
```

This looks like a module instance with an extra partition of `bind blaaaah` in front of it.
You can fuse the two together using `MergeConsecutiveSiblings`, and get:

```
[bind blaaaah fooooo
  [bar]
]
```

Which now resembles a module instance, where the first subtree just has extra elements `bind blaaaah` in front of the type `fooooo`.
From here, you could reuse the same re-shaping logic done with module instances, such as handling of single instances.  Might take a little code refactoring to re-use and avoid copying code.

If you ever find yourself manipulating token partition trees directly, think if your transformation can be formulated using one of the higher-level operations such as `HoistOnlyChild`, `FlatttenOnce`, `MergeConsecutiveSiblings` (or maybe invent one if needed).  These are fairly lightweight operations, so I wouldn't be too concerned with creating too much partitioning/hierarchy at any level only to partially flatten it.

Does this help?

rkapuscik:
> Does this help?

Yes, after modifying the grammar it all clicked. I merged the handling of `kBindDirective` and `kDataDeclaration` nodes. Thanks!

I've also revisited the fix for qualifier list - this has exactly the same structure as a bind directive with   different partition content. While fixing #118 I opted to leave the partitions untouched, but actually I don't see a reason not to merge them here.

With those changes following code:
```verilog
bind blaaaah fooooo bar ();
foo  baz();
static int bar;            
```
is partitioned as follows:
```
{ ([<auto>]) @{}, policy: always-expand
  { ([// verilog_syntax: parse-as-module-body]) }
  { ([<auto>]) @{1}, policy: fit-else-expand
    { ([bind blaaaah fooooo bar ( ) ;]) }
  }
  { ([<auto>]) @{2}, policy: fit-else-expand
    { ([foo baz ( ) ;]) }
  }
  { ([<auto>]) @{3}, policy: fit-else-expand
    { ([static int bar ;]) }
  }
}
```

fangism:
```
  { ([<auto>]) @{1}, policy: fit-else-expand
    { ([bind blaaaah fooooo bar ( ) ;]) }
  }
```

The partitioning looks good, but also compare what happens if the bind directive looks like:

```systemverilog
bind blah foo bar(), baz();
```

(which looks syntactically valid, but I wouldn't know how to interpret this!)
Does this fall back to the multi-instance handling similar to just: `foo bar(), baz();`?

And yes to your observation that declarations with leading qualifiers/keywords should probably be handled with a similar strategy of transformations.

verilog/parser/verilog.y
```diff
@@ -7037,10 +7037,14 @@ bind_target_instance
   ;
 bind_instantiation
   /* similar to block_item_decl instantiation */
-  : class_id gate_instance_or_register_variable_list
+  : bind_class_id gate_instance_or_register_variable_list
     { $$ = MakeTaggedNode(N::kBindTargetInstance, $1, $2); }
```
fangism: Alternatively, you could just write:

```
{ $$ = MakeTaggedNode(N::kBindTargetInstance,
                      MakeTaggedNode(N::kInstantiationType, $1),
                      $2);
}
```

without adding a single-use nonterminal.

common/formatting/unwrapped_line.h
```diff
@@ -82,6 +82,9 @@ class UnwrappedLine {
   // Extends PreFormatToken range up to the given token (exclusive).
   void SpanUpToToken(token_iterator iter) { tokens_.set_end(iter); }
 
+  // Extends PreFormatToken range to start from the given token (inclusive).
+  void SpanFromToken(token_iterator iter) { tokens_.set_begin(iter); }
```
fangism: Is this still needed?  If not, can remove.

rkapuscik:
Superseded by #302.

PR #186, opened 2020-02-06
Add Makefile and refactor scripts to use it.

mithro:
- `make clean`
 - `make build`
 - `make docs`
 - `make tools`
 - `make install`

Signed-off-by: Tim 'mithro' Ansell <me@mith.ro>


Makefile
```diff
@@ -0,0 +1,72 @@
+clean:
+	rm -rf bazel-*
```
hzeller: This does not actually remove thigns (the bazel-* things are symbolic links to its internal cache).
But `bazel clean` does.

Makefile
```diff
@@ -0,0 +1,72 @@
...
+bazel-bin/verilog/tools/lint/verilog_lint:
+	bazel build //verilog/tools/lint:verilog_lint
+
+bazel-bin/verilog/tools/syntax/verilog_syntax:
+	bazel build //verilog/tools/syntax:verilog_syntax
+
+bazel-bin/verilog/tools/formatter/verilog_format:
+	bazel build //verilog/tools/formatter:verilog_format
+
+tools: bazel-bin/verilog/tools/lint/verilog_lint bazel-bin/verilog/tools/syntax/verilog_syntax bazel-bin/verilog/tools/formatter/verilog_format
```
hzeller: I'd probably put this as the very first target. So that running `make` just results in these tools.

Makefile
```diff
@@ -0,0 +1,72 @@
...
+
+.PHONY: man
+
+DOCS_DIR=docs
+docs: docs.sh *.md
+	./docs.sh $(DOCS_DIR)
+
+.PHONY: docs
+
+PREFIX=/usr/local
```
hzeller: Maybe make `PREFIX?=/usr/local` ? That way, it can be overridden by an environment variable.

I'd probably put all these variable definitions at the top of the file.

Makefile
```diff
@@ -0,0 +1,72 @@
...
+	bazel build //verilog/tools/syntax:verilog_syntax
+
+bazel-bin/verilog/tools/formatter/verilog_format:
+	bazel build //verilog/tools/formatter:verilog_format
+
+tools: bazel-bin/verilog/tools/lint/verilog_lint bazel-bin/verilog/tools/syntax/verilog_syntax bazel-bin/verilog/tools/formatter/verilog_format
+	@true
+
+# Generate man pages from the tools
+man/verilog_lint.1: bazel-bin/verilog/tools/lint/verilog_lint
```
hzeller: It would be cool if we could describe these just in a single generic rule (`man/%.1 : ...`), but unfortunately, these tools have `/lint/`, `/syntax/` and `/formatter/` in their path, which makes them escape the common pattern matching rules in make. So I guess no dice.

Makefile
```diff
@@ -0,0 +1,72 @@
...
+	mkdir -p man
+	gflags2man --help_flag="--helpfull" --dest_dir man bazel-bin/verilog/tools/lint/verilog_lint
+
+man/verilog_syntax.1: bazel-bin/verilog/tools/syntax/verilog_syntax
+	mkdir -p man
+	gflags2man --help_flag="--helpfull" --dest_dir man bazel-bin/verilog/tools/syntax/verilog_syntax
+
+man/verilog_format.1: bazel-bin/verilog/tools/formatter/verilog_format
+	mkdir -p man
+	gflags2man  --help_flag="--helpfull" --dest_dir man bazel-bin/verilog/tools/formatter/verilog_format
```
hzeller: If we submit this, we should then also update the documentation that `gflags2man` is a requirement for `make install`

docs.sh
```diff
@@ -0,0 +1,165 @@
+#!/bin/bash
```
hzeller: Given that this generates docs, I'd rename it go `generate-docs.sh`

Makefile
fangism: want to use `?=` to let user override? or is this not intended for users?

docs.sh
```diff
@@ -0,0 +1,165 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# GitHub Pages
+# ---------------
+# Generate the GitHub pages to deploy
+
+TRAVIS_REPO_SLUG=${1:-google/verible}
```
hzeller: This is a confusingly named variable. For the perspective of the user of this script, who has this in their toplevel checked out repo, everything runs locally and nothing has to to with travis.
The travis repo is just the github path ? So this variable should probably be name `GITHUB_PATH`.

Also looking at all the uses below, it is always assembled together with the full URL, so couldn't we just have
```
  GITHUB_BASE_URL=https://github.com/google/verible
```
and use that below ? (but I suspect you want to set something in travis that utilizes this variable in a different way)

hzeller:
Any update ? This would be great to have!

mithro:
I got distracted by https://github.com/google/verible/pull/194

I'm unsure if I will get back to this anytime soon.

@hzeller - I was also wondering if these should actually be bazel rules rather than in the `Makefile`?

Makefile
```diff
@@ -0,0 +1,76 @@
...
+bazel-bin/verilog/tools/formatter/verilog_format:
+	bazel build //verilog/tools/formatter:verilog_format
+
+tools: bazel-bin/verilog/tools/lint/verilog_lint bazel-bin/verilog/tools/syntax/verilog_syntax bazel-bin/verilog/tools/formatter/verilog_format
+	@true
+
+# Generate man pages from the tools
+man/verilog_lint.1: bazel-bin/verilog/tools/lint/verilog_lint
+	mkdir -p man
+	gflags2man --help_flag="--helpfull" --dest_dir man bazel-bin/verilog/tools/lint/verilog_lint
```
mithro: @hzeller - I think the `gfags2man` stuff should probably be a bazel rule?

docs.sh
mithro: I think this should probably be a bazel rule in some way?

hzeller:
There is now a [`bazel run :install -c opt -- -s /usr/local/bin`](https://github.com/google/verible#installation) which allow for native installation.

BUT, unlike a usual 'make ; sudo make install', bazel shall not be run as root, but we choose with the `-s` option if `install` shall be invoked with `sudo`.

This is somewhat different how a `Makefile` workflow would work...

mithro:
@hzeller is going to figure out how to make the documentation generation and the man page generation into bazel rules.

This Makefile will then just become a simple tool for those too lazy to remember the bazel commands.

mithro:
Assigning to @hzeller to take over.

hzeller:
Cleaning out old pull requests. Ideas of this have been implemented in various ways.

PR #148, opened 2020-01-22
Put task arguments on a new line

ldalek-antmicro:
Fixes #29

Issue #97, opened 2019-12-16
[line breaks] DPI-C function formatting


msfschaffner:
is:
```systemverilog
  import "DPI-C" function chandle spidpi_create(
      input string name, input int mode, input int loglevel
  );
```
should:
```systemverilog
  import "DPI-C" function chandle spidpi_create(input string name, input int mode, input int loglevel);
```
if it does not fit onto a single line, we prefer the following formats:
```systemverilog
  import "DPI-C" function chandle spidpi_create(
      input string name, 
      input int mode,
      input int loglevel
  );
```
```systemverilog
  import "DPI-C" function chandle spidpi_create(input string name, 
                                                input int mode,
                                                input int loglevel);
```

msfschaffner:
b/146354843

fangism:
Confirming with complete test case:

```systemverilog
module m;
  import "DPI-C" function chandle spidpi_create( 
      input string name, input int mode, input int loglevel
  );
endmodule
```

Indeed, that import declaration exceeds the 100-column limit.

Current partitioning:

```
$ verilog_format --show_token_partition_tree ~/testing/verilog/formatting/b146354843.sv
Full token partition tree:
{ ([<auto>]) @{}
  { ([<auto>]) @{0}
    { ([module m ;]) }
  }
  { (>>[<auto>]) @{1}
    { (>>[import "DPI-C" function chandle spidpi_create (]) }
    { (>>>>>>[<auto>]) @{1,1}
      { (>>>>>>[input string name ,]) }
      { (>>>>>>[input int mode ,]) }
      { (>>>>>>[input int loglevel]) }
    }
    { (>>[) ;]) }
  }
  { ([endmodule]) }
}
```

Closely related to #29.

Agree that either one of your "preferred formats" looks better.

Issue #29, opened 2019-11-25
Task formal arguments put on a new line


msfschaffner:
The issue is currently onliy occuring when a task/function has more than 1 arg.

should:
```systemverilog
  virtual task check_next_core_status(core_status_t core_status, 
                                      string error_msg = "",
                                      int timeout = 9999999);
```
is:
```systemverilog
  virtual task check_next_core_status(
      core_status_t core_status, string error_msg = "", int timeout = 9999999
  );
```

msfschaffner:
b/145157892

fangism:
Specifically, it aggressively wraps when it doesn't fit on a single line, at the moment.  It's not even using the wrap-optimizer at this time.  This is controlled by:
https://github.com/google/verible/blob/f74e900dbdd40c48fcaf1ff68db0c7f5bd342658/verilog/formatting/tree_unwrapper.cc#L660

ldalek-antmicro:
Hey @fangism,
Can you look at #90?

I've fixed this issue by skipping partitioning arguments in `tree_unwrapper.cc` and by
forcing line break after comma in kTaskHeader and kFunctionHeader. But unfortunately this approach
breaks some tests e.g. https://github.com/google/verible/pull/90/commits/6d7bfc63b979836dbbffde582ac085a377a919d6#diff-c44554418c4241ad93a4b74981f78b08L1350

Also I'm not sure about this test: :https://github.com/google/verible/pull/90/commits/6d7bfc63b979836dbbffde582ac085a377a919d6#diff-9886cf9de4b1ff6379851d2f4d707190L982

fangism:
Before we embark on any solution, we need to reach a consensus on what is acceptable.

@msfschaffner @snsokolov which of the following results would you consider acceptable?

1a) (from original comment) keep first port on same line (if it fits), align all other ports, close with last port
```systemverilog
  virtual task check_next_core_status(core_status_t core_status, 
                                      string error_msg = "",
                                      int timeout = 9999999);
```

1b) same as 1a, but closing on its own line
```systemverilog
  virtual task check_next_core_status(core_status_t core_status, 
                                      string error_msg = "",
                                      int timeout = 9999999
  );
```

1c) same as 1b, but closing on its own line, aligned to open-(
```systemverilog
  virtual task check_next_core_status(core_status_t core_status, 
                                      string error_msg = "",
                                      int timeout = 9999999
                                      );
```

2a) wrap first port, append the rest while they fit (allow more than one per line), close with last port
[...]

PR #197, opened 2020-02-12
Implement AppendFittingSubpartitions policy

### LINKED ISSUE #42: macro call arguments not sufficiently partitioned
should:
```systemverilog
    `ASSERT(MioWarl_A,
            padctrl.reg2hw.mio_pads[mio_sel].qe |=>
        !(|padctrl.mio_attr_q[mio_sel][padctrl_reg_pkg::AttrDw-1:2]),
            clk_i, !rst_ni)
```
is:
```systemverilog
    `ASSERT(MioWarl_A,
            padctrl.reg2hw.mio_pads[mio_sel].qe |=>
        !(|padctrl.mio_attr_q[mio_sel][padctrl_reg_pkg::AttrDw-1:2])
            , clk_i, !rst_ni)
```

ldalek-antmicro:
Implements algorithm from https://github.com/google/verible/pull/148#issuecomment-581972774
Fixes #42 #29 and (?) #97

### DIFF
--- a/common/formatting/token_partition_tree.cc
+++ b/common/formatting/token_partition_tree.cc
@@ -19,6 +19,7 @@
 #include <vector>
 
 #include "common/formatting/format_token.h"
+#include "common/formatting/line_wrap_searcher.h"
 #include "common/formatting/unwrapped_line.h"
 #include "common/util/container_iterator_range.h"
 #include "common/util/logging.h"
@@ -179,6 +180,251 @@ TokenPartitionTree* MoveLastLeafIntoPreviousSibling(TokenPartitionTree* tree) {
       *tree, tree->LeftmostDescendant()->Value().TokensRange().begin());
 
   return rightmost_leaf_parent;
+}
+
+//
+// TokenPartitionTree class wrapper used by AppendFittingSubpartitions and
+// ReshapeFittingSubpartitions for partition reshaping purposes.
+//
+class TokenPartitionTreeWrapper {
+ public:
+  TokenPartitionTreeWrapper(const TokenPartitionTree& node) : node_(&node) {}
+
+  // Grouping node with no corresponding TokenPartitionTree node
+  TokenPartitionTreeWrapper(const UnwrappedLine& unwrapped_line) : node_(nullptr) {
+    unwrapped_line_ = std::unique_ptr<UnwrappedLine>(new UnwrappedLine(unwrapped_line));
+  }
+
+  // At least one of node_ or unwrapped_line_ should not be nullptr
+  TokenPartitionTreeWrapper() = delete;
+
+  TokenPartitionTreeWrapper(const TokenPartitionTreeWrapper& other) {
+    CHECK((other.node_ != nullptr) || (other.unwrapped_line_ != nullptr));
+
+    if (other.node_) {
+      node_ = other.node_;
+    } else {
+      node_ = nullptr;
+      unwrapped_line_ = std::unique_ptr<UnwrappedLine>(
+          new UnwrappedLine(*other.unwrapped_line_));
+    }
+  }
+
+  // Return wrapped node value or concatenation of subnodes values
+  const UnwrappedLine& Value() const {
+    if (node_) {
+      return node_->Value();
+    } else {
+      return *unwrapped_line_;
+    }
+  }
+
+  // Concatenate subnodes value with other node value
+  UnwrappedLine Value(const TokenPartitionTree& other) const {
+    CHECK((node_ == nullptr) && (unwrapped_line_ != nullptr));
+    UnwrappedLine uw = *unwrapped_line_;
+    uw.SpanUpToToken(other.Value().TokensRange().end());
+    return uw;
+  }
+
+  // Update concatenated value of subnodes
+  void Update(const VectorTree<TokenPartitionTreeWrapper>* child) {
+    const auto& token_partition_tree_wrapper = child->Value();
+    const auto& uwline = token_partition_tree_wrapper.Value();
+
+    unwrapped_line_->SpanUpToToken(uwline.TokensRange().end());
+  }
+
+  // Fix concatenated value indentation
+  void SetIndentationSpaces(int indent) {
+    CHECK((node_ == nullptr) && (unwrapped_line_ != nullptr));
+    unwrapped_line_->SetIndentationSpaces(indent);
+  }
+
+  // Return address to wrapped node
+  const TokenPartitionTree* Node() const {
+    return node_;
+  }
+
+ private:
+  // Wrapped node
+  const TokenPartitionTree* node_;
+
+  // Concatenated value of subnodes
+  std::unique_ptr<UnwrappedLine> unwrapped_line_;
+};
+
+// Reshapes unfitted_partitions and returns result via fitted_partitions.
+// Function creates VectorTree<> with additional level of grouping.
+// Function expects two partitions: first used for computing indentation
+// second with subpartitions, e.g.
+// { (>>[<auto>]) @{1,0} // tree root
+//   { (>>[function fffffffffff (]) } // first subpartition
+//   { (>>>>>>[<auto>]) @{1,0,1}, // root node with subpartitions
+//     { (>>>>>>[type_a aaaa ,]) }
+//     { (>>>>>>[type_b bbbbb ,]) }
+//     { (>>>>>>[type_c cccccc ,]) }
+//     { (>>>>>>[type_d dddddddd ,]) }
+//     { (>>>>>>[type_e eeeeeeee ,]) }
+//     { (>>>>>>[type_f ffff ) ;]) }
+//   }
+// }
+// Function is iterating over subpartitions and tries to append
+// them if fits in line. Parameter wrap_first_subpartition is used to determine
+// whether wrap first subpartition or not.
+// Return value signalise whether first subpartition was wrapped.
+static bool AppendFittingSubpartitions(
+    VectorTree<TokenPartitionTreeWrapper>* fitted_partitions,
+    const TokenPartitionTree& unfitted_partitions,
+    const verible::BasicFormatStyle& style,
+    bool wrap_first_subpartition) {
+  bool wrapped_first_subpartition;
+
+  // expects two partitions
+  CHECK_EQ(unfitted_partitions.Children().size(), 2);
+
+  // first with header
+  const auto& header = unfitted_partitions.Children()[0];
+
+  // arguments
+  const auto& args = unfitted_partitions.Children()[1].Children();
+
+  // at least one argument
+  CHECK_GE(args.size(), 1);
+
+  // Create first partition group
+  // and populate it with function name (e.g. { [function foo (] })
+  auto* group = fitted_partitions->NewChild(header.Value());
+  auto* child = group->NewChild(header);
+
+  int indent;
+
+  // Try appending first argument
+  auto group_value_arg = group->Value().Value(args[0]);
+  if (wrap_first_subpartition || (verible::FitsOnLine(std::move(group_value_arg), style).fits == false)) {
+    // Use wrap indentation
+    indent = style.wrap_spaces + group_value_arg.IndentationSpaces();
+
+    // wrap line
+    group = group->NewSibling(args[0].Value()); // start new group
+    child = group->NewChild(args[0]); // append not fitting 1st argument
+    group->Value().SetIndentationSpaces(indent);
+
+    // Wrapped first argument
+    wrapped_first_subpartition = true;
+  } else {
+    // Compute new indentation level based on first partition
+    const auto& group_value = group->Value().Value();
+    const UnwrappedLine& uwline = group_value;
+    indent = verible::FitsOnLine(uwline, style).final_column;
+
+    // Append first argument to current group
+    child = group->NewChild(args[0]);
+    group->Value().Update(child);
+    // keep group indentation
+
+    // Appended first argument
+    wrapped_first_subpartition = false;
+  }
+
+  for (size_t idx = 1 ; idx < args.size() ; ++idx) {
+    const auto& arg = args[idx];
+
+    // Every group should have at least one child
+    CHECK_GT(group->Children().size(), 0);
+
+    // Try appending current argument to current line
+    UnwrappedLine uwline = group->Value().Value(arg);
+    if (verible::FitsOnLine(uwline, style).fits) {
+      // Fits, appending child
+      child = group->NewChild(arg);
+      group->Value().Update(child);
+    } else {
+      // Does not fit, start new group with current child
+      group = group->NewSibling(arg.Value());
+      child = group->NewChild(arg);
+      // no need to update because group was created
+      // with current child value
+
+      // Fix group indentation
+      group->Value().SetIndentationSpaces(indent);
+    }
+  }
+
+  return wrapped_first_subpartition;
+}
+
+void ReshapeFittingSubpartitions(TokenPartitionTree* node,
+                                 const verible::BasicFormatStyle& style) {
+  VectorTree<TokenPartitionTreeWrapper>* fitted_tree = nullptr;
+
+  // Leaf or simple node, e.g. '[function foo ( ) ;]'
+  if ((node->Children().size() == 0) || (node->Children().size() == 1)) {
+    // Nothing to to do
+    return ;
+  }
+
+  // Partition with arguments should have at least one argument
+  const auto& args = node->Children()[1].Children();
+  CHECK_GT(args.size(), 0);
+
+  VectorTree<TokenPartitionTreeWrapper> unwrapped_tree(node->Value());
+  VectorTree<TokenPartitionTreeWrapper> wrapped_tree(node->Value());
+
+  // Format unwrapped_lines. At first without forced wrap after first line
+  bool wrapped_first_token = AppendFittingSubpartitions(&unwrapped_tree,
+                                                        *node, style, false);
+
+  if (wrapped_first_token) {
+    // First token was forced to wrap so there's no need to
+    // generate wrapped version (it has to be wrapped)
+    fitted_tree = &unwrapped_tree;
+  } else {
+    // Generate wrapped version to compare results.
+    // Below function passes-trough lines that doesn't fit
+    // e.g. very looooooooooong arguments with length over column limit
+    // and leaves optimization to line_wrap_searcher.
+    // In this approach generated result may not be
+    // exactly correct beacause of additional line break done later.
+    AppendFittingSubpartitions(&wrapped_tree, *node, style, true);
+
+    // Compare number of grouping nodes
+    // If number of grouped node is equal then prefer unwrapped result
+    if (unwrapped_tree.Children().size() <= wrapped_tree.Children().size()) {
+      fitted_tree = &unwrapped_tree;
+    } else {
+      fitted_tree = &wrapped_tree;
+    }
+  }
+
+  // Rebuild TokenPartitionTree
+  TokenPartitionTree temporary_tree(node->Value());
+
+  // Iterate over partition groups
+  for (const auto& itr : fitted_tree->Children()) {
+    auto uwline = itr.Value().Value();
+    // Partitions groups should fit in line but we're
+    // leaving final decision to ExpandableTreeView
+    uwline.SetPartitionPolicy(PartitionPolicyEnum::kFitOnLineElseExpand);
+
+    // Create new grouping node
+    auto* group = temporary_tree.NewChild(std::move(uwline));
+
+    // Iterate over partitions in group
+    for (const auto& partition : itr.Children()) {
+      // access partition_node_type
+      const auto* node = partition.Value().Node();
+
+      // Append child (warning contains original indentation)
+      group->NewChild(std::move(*node));
+    }
+  }
+
+  // Remove moved nodes
+  node->Children().clear();
+
+  // Move back from temporary tree
+  node->AdoptSubtreesFrom(&temporary_tree);
 }
 
 }  // namespace verible
--- a/common/formatting/token_partition_tree.h
+++ b/common/formatting/token_partition_tree.h
@@ -19,6 +19,7 @@
 #include <iosfwd>
 #include <vector>
 
+#include "common/formatting/basic_format_style.h"
 #include "common/formatting/format_token.h"
 #include "common/formatting/unwrapped_line.h"
 #include "common/util/vector_tree.h"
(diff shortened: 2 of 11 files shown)


ldalek-antmicro:
This PR implements `ProxyTreeView` and `AppendableView`.
Second viewer implements kAppendFittingSubpartitions algorithm.
First one is a kind of "proxy" that applies different viewers dependent on partition policies:
1. `kAlwaysExpand` - none. Just keeps walking.
2. `kFitOnLineElseExpand` - `ExpandableTreeView`
3. `kAppendFittingSubpartitions` - `AppendableView`

This PR is just a draft. I'm thinking on making base class for viewers (`BasicTreeViewer`?)
and making some cleanups. But generally it presents the idea.

@fangism any suggestions?

ldalek-antmicro:
I agree with https://github.com/google/verible/pull/197#pullrequestreview-357681310. I'll split this
PR later.
For now I'm curious about updated solution. I've updated tree so it contain vector of unwrapped lines.
My idea is to pass between partition policy evaluation stages set of unwrapped lines. And depending
on partition we can operate only on this set of lines (for example in `kAppendFittingSubpartitions`) or look deeper (there is still available original tree).

@fangism How about this concept?

verilog/formatting/formatter.cc
```diff
@@ -195,66 +195,262 @@ Status FormatVerilog(absl::string_view text, absl::string_view filename,
...
+          for (auto& itr : node.Children()) {
+            std::copy(itr.Value().unwrapped_lines.begin(),
+                      itr.Value().unwrapped_lines.end(),
+                      std::back_inserter(unwrapped_lines));
+          }
+
+          const auto& unwrapped = MakeUnwrappedLines(unwrapped_lines, style, false);
+          const auto& wrapped   = MakeUnwrappedLines(unwrapped_lines, style, true);
+
+          if (unwrapped.size() <= wrapped.size()) {
```
fangism: Just to confirm: Is this a comparison of the number of lines of the above results?

fangism:
> I agree with [#197 (review)](https://github.com/google/verible/pull/197#pullrequestreview-357681310). I'll split this
> PR later.
> For now I'm curious about updated solution. I've updated tree so it contain vector of unwrapped lines.
> My idea is to pass between partition policy evaluation stages set of unwrapped lines. And depending
> on partition we can operate only on this set of lines (for example in `kAppendFittingSubpartitions`) or look deeper (there is still available original tree).
> 
> @fangism How about this concept?

I think I see, a set/vector of unwrapped lines is just a further generalization of what a 'node' holds (doesn't have to be a single unwrapped line).  That sounds reasonable.

It looks like in your code, you are conditionally merging multiple unwrapped lines into longer ones, with the intent of letting the line_wrap_searcher (optimizer) just handle it the same way as others.  That seems fine for now, especially because these were done with line-length already in mind, so there is less risk of oversized unwrapped lines failing to converge (search is limited as a safety precaution).

ldalek-antmicro:
Hey @fangism,
I've prepared implementation based on your advices. Mostly removed
unnecessary code (less code is better) and focused on `ApplyPreOrder()` function.
You've said that it's okay to reshape partition tree so I mess with it a bit.
This PR is still missing tests cases but can you tell me what do you think about this
implementation?

Also do you still want to split this PR into two separate PRs?

verilog/formatting/formatter.cc
```diff
@@ -230,15 +230,265 @@ static void DeterminePartitionExpansion(partition_node_type* node,
...
+      if (current_line.Size()) {
+        // Doesn't fit, push back last fitting line
+        fitted_lines.push_back(current_line);
+
+        // And start current_line at the beginning of current not-fitting line
+        // At this moment current_line is empty
+        current_line = UnwrappedLine(indent, expanded_lines[idx].TokensRange().begin());
+      } else {
+        // Handle empty current_line case
+
```
fangism: Just looking for clarification:
Is the intent here to let the non-fitting line go through the usual line-wrap optimization later?  If so, indicate this in a brief comment.

verilog/formatting/formatter.cc
```diff
@@ -230,15 +230,265 @@ static void DeterminePartitionExpansion(partition_node_type* node,
...
+    default: {
+      break;
+    }
+  }
+}
+
+//
+// This function formats expanded_lines in a simple greedy manner
+// that is by appending consecutive lines and wrapping only when necessary to stay
+// under the column limit.
```
fangism: Explain the significance of `expanded_lines[0]` as the line that establishes an indentation/alignment position (when `expanded_lines[1]` is appended).

verilog/formatting/formatter.cc
```diff
@@ -230,15 +230,265 @@ static void DeterminePartitionExpansion(partition_node_type* node,
...
+
+  if (expanded_lines.size() == 0) {
+    fitted_lines.clear();
+    return wrapped_second_line;
+  } else if (expanded_lines.size() == 1) {
+    fitted_lines.push_back(expanded_lines[0]);
+    return wrapped_second_line;
+  } /* else .size() >= 2 */
+
+  // Start current_line at the begging
```
fangism: "beginning"

verilog/formatting/formatter.cc
```diff
@@ -230,15 +230,265 @@ static void DeterminePartitionExpansion(partition_node_type* node,
...
+  } else {
+    // Don't have to wrap second line
+
+    // Compute new indentation level based on first line length
+    indent = verible::FitsOnLine(expanded_lines[0], style).final_column;
+
+    // current_line already spans up to end of second line
+  }
+
+  for (size_t idx = 2 ; idx < expanded_lines.size() ; ++idx) {
```
fangism: // The remaining lines are appended as long as they fit, wrapping to a position determined by `expanded_lines[1]`.

verilog/formatting/formatter.cc
```diff
@@ -230,15 +230,265 @@ static void DeterminePartitionExpansion(partition_node_type* node,
...
+  const auto partition_policy = uwline.PartitionPolicy();
+
+  // Handle only kAppendFittingSubpartitions policy, ignore rest of them
+  switch (partition_policy) {
+  case PartitionPolicyEnum::kAppendFittingSubPartitions:
+    break;
+
+  default:
+    return;
+  }
```
fangism: Simplify to:

```
if (partition_policy != ...) return;
```

verilog/formatting/formatter.cc
```diff
@@ -230,15 +230,265 @@ static void DeterminePartitionExpansion(partition_node_type* node,
...
+  // -----------
+  // function fffffffffff(type_a aaaa, type_b bbbbb);
+  //
+  // partition tree:
+  // ---------------
+  // { ([<auto>]) @{2}, policy: append-fitting-sub-partitions
+  //   { ([function fffffffffff (]) }
+  //   { (>>>>[<auto>]) @{2,1}, policy: fit-else-expand
+  //     { (>>>>[type_a aaaa ,]) }
+  //     { (>>>>[type_b bbbbb ,]) }
```
fangism: I think this should be: `{ (>>>>[type_b bbbbb ; )]) }` (and include the following `);`.

verilog/formatting/formatter.cc
```diff
@@ -230,15 +230,265 @@ static void DeterminePartitionExpansion(partition_node_type* node,
...
+  //   }
+  //   { ([) ;]) }
+  // }
+  //
+  // output:
+  // -------
+  // [function fffffffffff (]
+  // >>>>[type_a aaaa ,])
+  // >>>>[type_b bbbbb])
+  // [) ;]
```
fangism: I think this should be merged.

verilog/formatting/formatter.cc
```diff
@@ -230,15 +230,265 @@ static void DeterminePartitionExpansion(partition_node_type* node,
...
+
+  if (wrapped_first_token) {
+    // First token was forced to wrap so there's no need to
+    // generate wrapped version (it's already wrapped)
+    fitted_lines = &unwrapped;
+  } else {
+    // Generate wrapped version to compare results
+    AppendFittingSubpartitions(wrapped, unwrapped_lines, style, true);
+
+    // Compare number of fitted lines.
```
fangism: What happens if among those lines, there was a single unwrapped line that did not fit and is expected to need to be wrapped by line_wrap_searching?  It's unknown at this time how many (vertical) lines it would have actually taken after wrapping.  This is something to think about for later.  In the long run, I am interested in exploring dynamic programming approaches to building up subproblems' solutions to do a more holistic optimization over the space of subpartition formattings.

verilog/formatting/formatter.cc
```diff
@@ -257,6 +507,12 @@ static std::vector<UnwrappedLine> MakeUnwrappedLinesWorklist(
         DeterminePartitionExpansion(&node, style);
       });
 
+  // Reshape partiton tree with kAppendFittingSubPartitions policy
```
fangism: "partition"

verilog/formatting/formatter.cc
ldalek-antmicro: Done

verilog/formatting/formatter.cc
ldalek-antmicro: Done

verilog/formatting/formatter.cc
ldalek-antmicro: Done

verilog/formatting/formatter.cc
ldalek-antmicro: Done

verilog/formatting/formatter.cc
ldalek-antmicro: Done

verilog/formatting/formatter.cc
ldalek-antmicro: Merged.

verilog/formatting/formatter.cc
ldalek-antmicro: Done

verilog/formatting/formatter.cc
ldalek-antmicro: Done

ldalek-antmicro:
I think that I should break this PR into two parts:
1. kAppendFittingSubpartitions code
2. Task & Functions reshaping

but I don't know how test appending without modifying `tree_unwrapper`.
I was thinking about adding to `formatter_test.cc` function with manually built `TreeVector<UnwrappedLine>` and run `formatter` but I'm not sure how to do it
properly. @fangism do you have any ideas how to do it?

fangism:
> I think that I should break this PR into two parts:
> 
> 1. kAppendFittingSubpartitions code
> 2. Task & Functions reshaping
> 
> but I don't know how test appending without modifying `tree_unwrapper`.
> I was thinking about adding to `formatter_test.cc` function with manually built `TreeVector<UnwrappedLine>` and run `formatter` but I'm not sure how to do it
> properly. @fangism do you have any ideas how to do it?

Very good question.  Take a look at some of the tests in:
https://github.com/google/verible/blob/master/common/formatting/token_partition_tree_test.cc

There are some examples that manually build up small trees of hand-crafted unwrapped lines.  
To make the function prototype language-agnostic, all you need is:
https://github.com/google/verible/blob/6729d882840b9e792883411d024aedf0bf6f0d1e/common/formatting/basic_format_style.h#L21

which is the base class of the Verilog-specific `FormatStyle`.

Does that help?

ldalek-antmicro:
Yes, examples from https://github.com/google/verible/blob/master/common/formatting/token_partition_tree_test.cc do help. And I should put those tests in https://github.com/google/verible/blob/master/verilog/formatting/formatter_test.cc. Am I correct?

fangism:
> Yes, examples from https://github.com/google/verible/blob/master/common/formatting/token_partition_tree_test.cc do help. And I should put those tests in https://github.com/google/verible/blob/master/verilog/formatting/formatter_test.cc. Am I correct?

Your append function looks suitable to move to common/token_partition_tree.{h,cc}, so I would move it there, and test in token_partition_tree_test.cc.

ldalek-antmicro:
Hey @fangism,
I've moved code as you suggested but I'm not sure about this:
https://github.com/antmicro/verible/blob/append-fitting-subpartitions/common/formatting/token_partition_tree.cc#L370. Is it ok to keep
it like this or should this be somehow encapsulated?

fangism:
> Hey @fangism,
> I've moved code as you suggested but I'm not sure about this:
> https://github.com/antmicro/verible/blob/append-fitting-subpartitions/common/formatting/token_partition_tree.cc#L370. Is it ok to keep
> it like this or should this be somehow encapsulated?

If I'm reading this right, this is trying to re-construct an updated tree *view* from a modified subtree.
The passing of `*new` looks very suspect (memory-leak wise).
Is there a way you can leverage the existing `ExpandableTreeView` constructor to do this?

https://github.com/google/verible/blob/6a27838dd40c816c0139458390a7feb3d064f838/common/util/expandable_tree_view.h#L101

If necessary, you can refactor `ExpandableTreeView` to make that recursive reconstruction operation more accessible, like an `Update()` or `Rebuild()` method.

fangism:
@ldalek-antmicro Just checking in to see if this is coming along (and if you saw my last reply).  This work will un-gate all of the related issues around task/function/macro calls, and so is very much anticipated.  Let me know if I can advise on anything else too.

ldalek-antmicro:
Sorry for long response. I spent quite some time to analyze `formatter` code and do so experiments.

First of all I've pushed improved fitting algorithm. It's now using `VectorTree<>` class for partitioning and not `UnwrappedLine`. I think it's simpler now. It's using modified `TreeViewNodeInfo` to insert additional grouping nodes to fix partition structure.
This implementation works but has missing test suite.

Before I proceed any further I want to ask you few questions.
First thing is `formatter` formatting flow:
1. Token annotation
2. Tree unwrapping
3. Lines unwrapping (using ExpandableTreeView)
4. Line_wrap_search

Tree_unwrapper is using annotated tokens (passed with `const` keyword). You suggested
that instead of using `SetIndenationSpaces()` I could use `before.spaces` annotation.
I could do this but I'd have to drop `const` by using `const_cast` operator. It's what you intended?

This implementation is modifying `ExpandableTreeView` but it could be implemented right after
`.Unwrap()` function (after tree unwrapping). What do you think about moving it right after unwrapping (and not as post-traversal process)?

fangism:
> Sorry for long response. I spent quite some time to analyze `formatter` code and do so experiments.
> 
> First of all I've pushed improved fitting algorithm. It's now using `VectorTree<>` class for partitioning and not `UnwrappedLine`. I think it's simpler now. It's using modified `TreeViewNodeInfo` to insert additional grouping nodes to fix partition structure.
> This implementation works but has missing test suite.
> 
> Before I proceed any further I want to ask you two questions.
> First thing is `formatter` formatting flow:
> 
> 1. Token annotation
> 2. Tree unwrapping
> 3. Lines unwrapping (using ExpandableTreeView)
> 4. Line_wrap_search
> 
> Tree_unwrapper is using annotated tokens (passed with `const` keyword). You suggested
> that instead of using `SetIndenationSpaces()` I could use `before.spaces` annotation.
> I could do this but I'd have to drop `const` by using `const_cast` operator. It's what you intended?
> 
> This implementation is modifying `ExpandableTreeView` but it could be implemented right after
> `.Unwrap()` function (after tree unwrapping). What do you think about moving it right after unwrapping (and not as post-traversal process)?

I would like to keep Token-annotation and Tree-unwrapping independent steps, and parallelizable, and have Tree-unwrapping not touch anything related to token annotation.  Tree-unwrapping is a markup phase that annotates cut-points (in a hierarchical token-partition-tree manner).

After 1 and 2, we have a lot more options.  Today ExpandableTreeView helps choose appropriate partition sizes to pass on to line_wrap_search, however in the long run, not every unit of formatting work needs to undergo line_wrap_search, not everything needs to use UnwrappedLines/IndentationSpaces in the same manner.  As we introduce more subtree formatting policies like AppendFitting, I expect these will branch out to different paths.  

If the model of rendering UnwrappedLines the same way as before work conveniently, you're welcome to use it as you did in the draft: you altered and reshaped the token partition tree in a new pass to be able to re-use line_wrap_search and the rendering code that follows.  If you find it getting to be too awkward and constrained, maybe it would be cleaner to diverge.  (Imagine a tree of format handling paths that all converge in the final rendering.)

ldalek-antmicro:
Okay. I did some cleaning. Also I simplified design.
I think that the best would be stick to current flow (for now).
Current implementation reshapes partitions right after tree partitioning: https://github.com/google/verible/pull/197/files#diff-60e25030f24b313541dc73a69d41fc16R422
This implementation reshapes `tree_unwrapper` but it could also modify copy of that tree, for example:
```
const auto token_partition_tree = *format_tokens_partitions;
token_partition_tree.ApplyPreOrder(...);
format_tokens_partitions = &token_partition_tree;
```
I prefer the first one. But it can be changed.
This code needs some more tests & comments but I want to keep this way.
What do you think?

common/formatting/token_partition_tree.h
```diff
@@ -82,6 +84,71 @@ std::ostream& operator<<(std::ostream& stream,
...
+  std::shared_ptr<UnwrappedLine> unwrapped_line;
+};
+
+// Core part of kAppendFittingSubPartitions algorithm
+// Append fitting subpartitions in simple greedy manner
+bool AppendFittingSubpartitions(
+    VectorTree<TokenPartitionTreeWrapper>& fitted_partitions,
+    const TokenPartitionTree& unfitted_partitions,
+    const verible::BasicFormatStyle& style,
+    bool wrap_second_line);
```
fangism: I would move this function and the `TokenPartitionTreeUnwrapper` class into only the .cc file as implementation details, exposing only `ReshapeFittingSubpartitions` for direct use and direct testing.

If you really want to test through the internal function, consider moving these declarations into a `-internal.h` header (and have the `cc_library` and `cc_test` both depend on that in `srcs`).  That said, I still want to encourage testing through the public interfaces where possible.

common/formatting/token_partition_tree.h
```diff
@@ -82,6 +84,71 @@ std::ostream& operator<<(std::ostream& stream,
...
+
+// Evaluates two partition trees cases (first argument wrapped or not)
+// And reshapes partition_node_type tree accordingly
+void ReshapeFittingSubpartitions(TokenPartitionTree& node,
```
fangism: Pass in most modified object references by pointer, so `TokenPartitionTree* node`.
(Exceptions for this include `std::ostream&`.)

common/formatting/token_partition_tree.h
```diff
@@ -82,6 +84,71 @@ std::ostream& operator<<(std::ostream& stream,
...
+// Core part of kAppendFittingSubPartitions algorithm
+// Append fitting subpartitions in simple greedy manner
+bool AppendFittingSubpartitions(
+    VectorTree<TokenPartitionTreeWrapper>& fitted_partitions,
+    const TokenPartitionTree& unfitted_partitions,
+    const verible::BasicFormatStyle& style,
+    bool wrap_second_line);
+
+// Evaluates two partition trees cases (first argument wrapped or not)
+// And reshapes partition_node_type tree accordingly
```
fangism: Since this will be the public API, describe more here what the input partition is expected to look like.  e.g. subpartition[0] is like a function-declaration opener or function-call opener.

verilog/formatting/formatter.cc
```diff
@@ -413,6 +419,19 @@ Status Formatter::Format(const ExecutionControl& control) {
     }
   }
 
+  {
```
fangism: This location is good, it is clear that this is a transformational pass separate from the phases around it.

common/formatting/token_partition_tree_test.cc
```diff
@@ -398,4 +398,486 @@ TEST_F(MoveLastLeafIntoPreviousSiblingTest, TwoGenerations) {
                                     << *diff.right << '\n';
 }
 
+class TokenPartitionTreeWrapperTest
```
fangism: Related to earlier suggestion in token_partition_tree.h.
Is it possible to formulate these `TokenPartitionTreeWrapperTest`s in terms of only `ReshapeFittingSubpartitions`, the public interface?

common/formatting/token_partition_tree.h
```diff
@@ -19,7 +19,9 @@
 #include <iosfwd>
 #include <vector>
 
+#include "common/formatting/basic_format_style.h"
 #include "common/formatting/format_token.h"
+#include "common/formatting/line_wrap_searcher.h"
```
fangism: Is this header needed here, or only in the .cc file?

common/formatting/token_partition_tree_test.cc
```diff
@@ -398,4 +398,486 @@ TEST_F(MoveLastLeafIntoPreviousSiblingTest, TwoGenerations) {
...
+  auto* child = group->NewChild(tree.Children()[0]);
+
+  // FIXME: check structure
+
+  EXPECT_EQ(tree_wrapper.Children()[0].Children()[0].Value().Node(), &tree.Children()[0]);
+
+  const auto& tree_range = tree.Value().TokensRange();
+  const auto& tree_wrapper_range = tree_wrapper.Value().Value().TokensRange();
+  EXPECT_TRUE(tree_range.begin() == tree_wrapper_range.begin());
+  EXPECT_TRUE(tree_range.end() == tree_wrapper_range.end());
```
fangism: Does `EXPECT_EQ` work here?

common/formatting/token_partition_tree_test.cc
```diff
@@ -398,4 +398,486 @@ TEST_F(MoveLastLeafIntoPreviousSiblingTest, TwoGenerations) {
...
+  }
+}
+
+// --------
+class ReshapeFittingSubpartitionsTestFixture : public ::testing::Test,
+                                               public UnwrappedLineMemoryHandler {
+ public:
+  ReshapeFittingSubpartitionsTestFixture()
+      : sample_("function_fffffffffff( "
+                "type_a_aaaa, type_b_bbbbb, "
```
fangism: // combining what would normally be a type and a variable name into a single string for testing convenience
(if that is the reason)

fangism:
> Okay. I did some cleaning. Also I simplified design.
> I think that the best would be stick to current flow (for now).
> Current implementation reshapes partitions right after tree partitioning: https://github.com/google/verible/pull/197/files#diff-60e25030f24b313541dc73a69d41fc16R422
> This implementation reshapes `tree_unwrapper` but it could also modify copy of that tree, for example:
> 
> ```
> const auto token_partition_tree = *format_tokens_partitions;
> token_partition_tree.ApplyPreOrder(...);
> format_tokens_partitions = &token_partition_tree;
> ```
> 
> I prefer the first one. But it can be changed.
> This code needs some more tests & comments but I want to keep this way.
> What do you think?

Left comments.
The transformation pass on the TokenPartitionTree in formatter.cc is well-placed and localized.

common/formatting/token_partition_tree.h
ldalek-antmicro: Direct testing is enough. But thanks for info about `-internal.h`

common/formatting/token_partition_tree.h
ldalek-antmicro: Fixed

common/formatting/token_partition_tree.cc
```diff
@@ -181,4 +182,236 @@ TokenPartitionTree* MoveLastLeafIntoPreviousSibling(TokenPartitionTree* tree) {
...
+  }
+
+  // Return address to wrapped node
+  const TokenPartitionTree* Node() const {
+    return node;
+  }
+
+ private:
+  // Wrapped node
+  const TokenPartitionTree* node;
```
fangism: Name private members ending with `_`, so `node_`.

verilog/formatting/formatter.cc
ldalek-antmicro: Okay. Thanks!

common/formatting/token_partition_tree_test.cc
ldalek-antmicro: Fixed.

common/formatting/token_partition_tree.cc
```diff
@@ -181,4 +182,236 @@ TokenPartitionTree* MoveLastLeafIntoPreviousSibling(TokenPartitionTree* tree) {
...
+
+  // Concatenated value of subnodes
+  std::shared_ptr<UnwrappedLine> unwrapped_line;
```
fangism: `unwrapped_line_`

Can this be a `std::unique_ptr`?  If not, explain in comment.

common/formatting/token_partition_tree.h
ldalek-antmicro: Nope. Needed only in .cc. Fixed.

common/formatting/token_partition_tree_test.cc
ldalek-antmicro: Outdated. But using `EXPECT_EQ` in other tests.

common/formatting/token_partition_tree_test.cc
ldalek-antmicro: Yep. That's the reason. Added comment

common/formatting/token_partition_tree_test.cc
```diff
@@ -500,4 +500,762 @@ TEST_F(MergeConsecutiveSiblingsTest, TwoGenerations) {
...
+
+  EXPECT_EQ(saved_tree.Children()[1].Children()[2].Value().IndentationSpaces(),
+            tree.Children()[1].Children()[0].Value().IndentationSpaces());
+  EXPECT_EQ(saved_tree.Children()[1].Children()[3].Value().IndentationSpaces(),
+            tree.Children()[1].Children()[1].Value().IndentationSpaces());
+  EXPECT_EQ(saved_tree.Children()[1].Children()[4].Value().IndentationSpaces(),
+            tree.Children()[1].Children()[2].Value().IndentationSpaces());
+}
+
+// Tests with real-word example
```
fangism: "real-world" :)

common/formatting/token_partition_tree_test.cc
```diff
@@ -500,4 +500,762 @@ TEST_F(MergeConsecutiveSiblingsTest, TwoGenerations) {
...
+ public:
+  ReshapeFittingSubpartitionsTestFixture()
+      // combining what would normally be a type and a variable name
+      // into a single string for testing convenience
+      : sample_("function_fffffffffff( "
+                "type_a_aaaa, type_b_bbbbb, "
+                "type_c_cccccc, type_d_dddddddd, "
+                "type_e_eeeeeeee, type_f_ffff);"),
+        tokens_(absl::StrSplit(sample_, ' ')) {
+    EXPECT_EQ(tokens_.size(), 7);
```
fangism: `ASSERT_EQ` to stop test early if false

common/formatting/token_partition_tree_test.cc
```diff
@@ -500,4 +500,762 @@ TEST_F(MergeConsecutiveSiblingsTest, TwoGenerations) {
...
+  UnwrappedLine group1(0, begin);
+  group1.SpanUpToToken(header.TokensRange().end());
+  UnwrappedLine group2(0, arg1.TokensRange().begin());
+  group2.SpanUpToToken(arg2.TokensRange().end());
+  UnwrappedLine group3(0, arg3.TokensRange().begin());
+  group3.SpanUpToToken(arg4.TokensRange().end());
+  UnwrappedLine group4(0, arg5.TokensRange().begin());
+  group4.SpanUpToToken(arg6.TokensRange().end());
+
+  tree_type tree_expected{
```
fangism: Mark all `tree_expected` data as `const`.   (elsewhere too)

common/formatting/token_partition_tree_test.cc
```diff
@@ -500,4 +500,762 @@ TEST_F(MergeConsecutiveSiblingsTest, TwoGenerations) {
...
+      tree_type{arg6},
+    }
+  };
+
+  verible::BasicFormatStyle style;
+  style.column_limit = 40;
+  ReshapeFittingSubpartitions(&tree, style);
+
+  const auto diff = DeepEqual(tree, tree_expected, TokenRangeEqual);
+  EXPECT_TRUE(diff.left == nullptr) << "Expected:\n" << tree_expected
```
fangism: `EXPECT_EQ` on `nullptr`s

common/formatting/token_partition_tree.cc
ldalek-antmicro: Done

common/formatting/token_partition_tree.cc
ldalek-antmicro: Changed to `std::unique_ptr`

common/formatting/token_partition_tree_test.cc
ldalek-antmicro: Fixed :-)

common/formatting/token_partition_tree_test.cc
ldalek-antmicro: Changing to `ASSERT_EQ` gives: `error: returning a value from a constructor`.
That `EXPECT_EQ` was really just for me and I think it's not really necessary.
I removed it.

common/formatting/token_partition_tree_test.cc
ldalek-antmicro: Done

common/formatting/token_partition_tree_test.cc
ldalek-antmicro: Done

ldalek-antmicro:
I changed both lines:
https://github.com/google/verible/pull/197/commits/0c6b41de5de77227cd2327d634a6924b819df3e5#diff-26d82cd2477bae858141ce2a22032162R195
https://github.com/google/verible/pull/197/commits/0c6b41de5de77227cd2327d634a6924b819df3e5#diff-26d82cd2477bae858141ce2a22032162R208
to build with `--cxxopt=-std=c++11`

PR #199, opened 2020-02-13
Build binaries for releases using optimization.

hzeller:


hzeller:
Nice, the new binaries dropped into the release are much smaller.

Issue #203, opened 2020-02-18
Linter rules for parameters do not match multiple inline parameters


tgorochowik:
For example the following code:
```
module params;
  parameter int foO = 1, baR = 1;
endmodule
```

Executed like this:
```
verilog_lint -rules parameter-name-style ../params.sv
```

Produces:
```
params.sv:2:17: Non-type parameter names must be styled with UpperCamelCase or ALL_CAPS. [Style: constants] [parameter-name-style]
```

The second parameter (`baR`) is not matched.

tgorochowik:
@fangism what would be the preferred way to approach this? I guess we could either modify the rule code to search for multiple parameters after `kParamDeclaration`, or make sure each entry has its own `kParamDeclaration`.

This issue (and the chosen solution) also affects #162

fangism:
Good find.  I would add support for being able to extract multiple assignment-initializations from a single parameter declaration, and then extract and analyze the name from each assignment-initialization.

fangism:
b/149754819

PR #204, opened 2020-02-18
Forbid parameter names starting with 'disable', suggest 'enable' instead.

### LINKED ISSUE #162: Forbid parameter names starting with 'disable', suggest 'enable' instead.
Our style guide now explicitly discourages naming parameters starting with 'disable'.
The guidance is to use positive sense naming, so starting with 'enable' is recommended.

Task:
* Check all parameter/localparameter declarations for this (even where those keywords are not required, e.g. module parameters inside `#()`).
* Make sure to cover case variations like `DiSaBlE`
* Cover class parameters, package parameters.
* Unit-tests for all of the above.

tgorochowik:
Fixes #162 

This is affected by #203 so I think it makes sense to wait for that to be fixed first.

### DIFF
--- a/verilog/analysis/checkers/positive_meaning_parameter_name_rule.cc
+++ b/verilog/analysis/checkers/positive_meaning_parameter_name_rule.cc
@@ -1,0 +1,83 @@
+// Copyright 2017-2020 The Verible Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "verilog/analysis/checkers/positive_meaning_parameter_name_rule.h"
+
+#include <set>
+#include <string>
+
+#include "absl/strings/str_cat.h"
+#include "absl/strings/string_view.h"
+#include "absl/strings/match.h"
+#include "common/analysis/citation.h"
+#include "common/analysis/lint_rule_status.h"
+#include "common/analysis/matcher/bound_symbol_manager.h"
+#include "common/strings/naming_utils.h"
+#include "common/text/symbol.h"
+#include "common/text/syntax_tree_context.h"
+#include "common/text/token_info.h"
+#include "verilog/CST/parameters.h"
+#include "verilog/analysis/descriptions.h"
+#include "verilog/analysis/lint_rule_registry.h"
+#include "verilog/parser/verilog_token_enum.h"
+
+namespace verilog {
+namespace analysis {
+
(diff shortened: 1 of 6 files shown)


verilog/analysis/checkers/positive_meaning_parameter_name_rule_test.cc
```diff
@@ -0,0 +1,129 @@
...
+      {"class foo #(parameter int ",
+       {kToken, "disableOpt"}, "); endclass"},
+      {"class foo #(parameter int ",
+       {kToken, "DiSABle_Opt"}, "); endclass"},
+
+      {"package foo; parameter int ",
+       {kToken, "disable_sth",}, " = 1; endpackage"},
+  };
+  RunLintTestCases<VerilogAnalyzer, PositiveMeaningParameterNameRule>(kTestCases);
+}  // namespace
```
fangism: remove comment (doesn't end namespace)

verilog/analysis/checkers/positive_meaning_parameter_name_rule_test.cc
```diff
@@ -0,0 +1,129 @@
...
+      {"class foo #(parameter int ",
+       {kToken, "disable_opt"}, "); endclass"},
+      {"class foo #(parameter int ",
+       {kToken, "disableOpt"}, "); endclass"},
+      {"class foo #(parameter int ",
+       {kToken, "DiSABle_Opt"}, "); endclass"},
+
+      {"package foo; parameter int ",
+       {kToken, "disable_sth",}, " = 1; endpackage"},
+  };
```
fangism: Suggest test case:
* one with two non-conforming parameter names

verilog/analysis/checkers/positive_meaning_parameter_name_rule_test.cc
tgorochowik: Right! Thanks!

verilog/analysis/checkers/positive_meaning_parameter_name_rule_test.cc
tgorochowik: Sure, added!

tgorochowik:
This now depends on #207 (uses some code added in that PR), therefore this PR mail fail until #207 is merged and this is rebased.

fangism:
> This now depends on #207 (uses some code added in that PR), therefore the CI here will fail until #207 is merged and this is rebased.

Acknowledged.

github is giving me strange errors when I try to comment, but here a recommended change:

`for (const auto& id: identifiers)` in your rule's .cc file

verilog/analysis/checkers/positive_meaning_parameter_name_rule.cc
```diff
@@ -0,0 +1,83 @@
...
+}
+
+void PositiveMeaningParameterNameRule::HandleSymbol(
+    const verible::Symbol& symbol, const SyntaxTreeContext& context) {
+  verible::matcher::BoundSymbolManager manager;
+  if (matcher_.Matches(symbol, &manager)) {
+    if (IsParamTypeDeclaration(symbol)) return;
+
+    auto identifiers = GetAllParameterNameTokens(symbol);
+    for (auto id: identifiers) {
```
fangism: `const auto& id`

tgorochowik:
Rebased and ready to merge!

PR #211, opened 2020-02-20
Parse `alias` construct 

### LINKED ISSUE #81: Parse `alias` construct
`alias` constructs are currently not parsed.

Task:
update grammar and tests:
https://github.com/google/verible/blob/master/verilog/parser/verilog.y
https://github.com/google/verible/blob/master/verilog/parser/verilog_parser_unittest.cc

Add a new node enum to:
https://github.com/google/verible/blob/master/verilog/CST/verilog_nonterminals.h

Expect this to be fairly easy.

tgorochowik:
Fixes #81

### DIFF
--- a/verilog/parser/verilog.y
+++ b/verilog/parser/verilog.y
@@ -5479,6 +5479,18 @@ continuous_assign
     { $$ = MakeTaggedNode(N::kContinuousAssign, $1, $2, $3, $4, nullptr); }
   ;
 
+net_alias_assign_lvalue_list
+  : net_alias_assign_lvalue_list '=' lpvalue
+    { $$ = ExtendNode($1, $2, $3); }
+  | lpvalue '=' lpvalue
+    { $$ = MakeTaggedNode(N::kNetAliasLvalueList, $1, $2, $3); }
+  ;
+
+net_alias
+  : TK_alias net_alias_assign_lvalue_list ';'
+    { $$ = MakeTaggedNode(N::kNetAlias, $1, $2, $3); }
+  ;
+
 loop_generate_construct
   : TK_for '(' genvar_opt GenericIdentifier '=' expression ';'
     expression_opt ';' for_step_opt ')'
@@ -5556,7 +5568,8 @@ module_common_item
     { $$ = move($1); }
   | continuous_assign
     { $$ = move($1); }
-  /* TODO(fangism): net_alias */
+  | net_alias
+    { $$ = move($1); }
   | loop_generate_construct
     { $$ = move($1); }
   | conditional_generate_construct
--- a/verilog/parser/verilog_parser_unittest.cc
+++ b/verilog/parser/verilog_parser_unittest.cc
@@ -5203,6 +5203,38 @@ static const char* kRandSequenceTests[] = {
     "endtask\n",
 };
 
+static const char* kNetAliasTests[] = {
+    "module byte_swap (inout wire [31:0] A, inout wire [31:0] B);\n"
+    "  alias {A[7:0],A[15:8],A[23:16],A[31:24]} = B;\n"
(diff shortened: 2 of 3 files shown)


verilog/parser/verilog.y
```diff
@@ -5479,6 +5479,18 @@ continuous_assign
     { $$ = MakeTaggedNode(N::kContinuousAssign, $1, $2, $3, $4, nullptr); }
   ;
 
+net_alias_assign
+  : lpvalue '=' net_alias_assign
```
fangism: To treat this like other lists, I would flip the associativity like:
`: net_alias_assign '=' lpvalue`

Consider naming this rule `net_alias_assign_list` or `net_alias_lvalue_list`.

verilog/parser/verilog.y
```diff
@@ -5479,6 +5479,18 @@ continuous_assign
...
+    { $$ = ExtendNode($1, $2, $3); }
+  | lpvalue
+    { $$ = move($1); }
```
fangism: The semantic action for the first element of a list should make a singleton list, like:
`{ $$ = MakeTaggedNode(kAliasNetLvalueList, $1); }`
(might need a new node enum)

however, with a doubleton as the base case:
`{ $$ = MakeTaggedNode(kAliasNetLvalueList, $1, $2, $3); }`
which is the same as creating a singleton list followed by `ExtendNode`.

verilog/parser/verilog.y
```diff
@@ -5479,6 +5479,18 @@ continuous_assign
     { $$ = MakeTaggedNode(N::kContinuousAssign, $1, $2, $3, $4, nullptr); }
   ;
 
+net_alias_assign
+  : lpvalue '=' net_alias_assign
+    { $$ = ExtendNode($1, $2, $3); }
+  | lpvalue
```
fangism: Since the grammar for this requires at least two components to be aliased, the base case rule could be a doubleton instead of a singleton:
`| lpvalue '=' lpvalue`

verilog/parser/verilog.y
```diff
@@ -5479,6 +5479,18 @@ continuous_assign
...
+    { $$ = move($1); }
+  ;
+
+net_alias
+  : TK_alias lpvalue '=' net_alias_assign ';'
```
fangism: With the above list being at minimum a doubleton, this could be simplified to:
`: TK_alias net_alias_assign ';'`

verilog/parser/verilog_parser_unittest.cc
```diff
@@ -5774,6 +5806,10 @@ TEST(VerilogParserTestMatchAll, RandSequence) {
   TestVerilogParserMatchAll(kRandSequenceTests);
 }
 
+TEST(VerilogParserTestMatchAll, NetAlias) {
```
fangism: Somewhere around line 5620, don't forget to add:
`TestVerilogParser(kNetAliasTests);`

verilog/parser/verilog.y
tgorochowik: Sure, renamed

verilog/parser/verilog.y
tgorochowik: That makes sense, added the new node enum and refactored the code

verilog/parser/verilog.y
tgorochowik: Sure, changed

verilog/parser/verilog.y
tgorochowik: right!

verilog/parser/verilog_parser_unittest.cc
tgorochowik: Absolutely right, added.

Thanks for the review!

PR #229, opened 2020-03-13
Allow lint configuration to be placed in files

tgorochowik:
I am opening this early to ask for comments.

Some general info:
* The configuration from file is read after setting up the default configs, but before parsing the invocation arguments. This way the arguments can easily overwrite the configuration from the file
* I think the following should be considered (all of them should be optional):
  * a global configuration file (e.g. `~/.verible_lint.rules`)
  * a local configuration file (e.g. the top directory of a project), this assumes the linter is executed from the top dir of the project
  * any arbitrary file, specified with an argument
  * If we decide to use all of the options listed above we could:
    * combine the global file, local file and file from a flag (in that order), or
    * ignore global file if a local one exists, ignore both if another one is specified with a flag
    * always overwrite the config from files when the same flags are specified with the `-rules` switch
* for now the implementation does some string operations on the file to make the format compatible with the arguments format. The only difference is the settings can be put in separate lines to improve readability. Thanks to this we can reuse the same parsing code

Other options to consider:
* use a completely different format for the configuration (e.g. ini files, with a general `[rules]` section, and more rule-specific configuration)

Please share your thoughts.

CC @hzeller

### DIFF
--- a/verilog/analysis/verilog_linter.cc
+++ b/verilog/analysis/verilog_linter.cc
@@ -55,6 +55,8 @@
 ABSL_FLAG(verilog::RuleBundle, rules, {},
           "List of lint rules to enable.  "
           "Prefix a rule name with '-' to disable it.");
+ABSL_FLAG(std::string, rules_config, ".rules.verible_lint",
+          "Path to lint rules configuration file.");
 ABSL_FLAG(verilog::RuleSet, ruleset, verilog::RuleSet::kDefault,
           "[default|all|none], the base set of rules used by linter");
 
@@ -209,6 +211,23 @@ LinterConfiguration LinterConfigurationFromFlags() {
   // Turn on default ruleset.
   const auto& ruleset = absl::GetFlag(FLAGS_ruleset);
   config.UseRuleSet(ruleset);
+
+  // Read local configuration file
+  std::string content;
+  if (verible::file::GetContents(absl::GetFlag(FLAGS_rules_config), &content)) {
+
+    RuleBundle local_rules_bundle;
+    std::string error;
+    if (local_rules_bundle.ParseConfiguration(content, &error)) {
+      config.UseRuleBundle(local_rules_bundle);
+    } else {
+      LOG(ERROR) << "Unable to fully parse configuration: " << error << std::endl;
+    }
+  } else if(FLAGS_rules_config.IsModified()) {
+    // If flag is modified and  we were unable to open the file, report that
+    LOG(WARNING) << "Unable to open rules configuration file: "
+                 << absl::GetFlag(FLAGS_rules_config) << std::endl;
+  }
 
   // Turn on rules found in config flags.
   const auto& rules = absl::GetFlag(FLAGS_rules);
--- a/verilog/analysis/verilog_linter_configuration.cc
+++ b/verilog/analysis/verilog_linter_configuration.cc
@@ -83,6 +83,63 @@ std::string ProjectPolicy::ListPathGlobs() const {
                        });
 }
 
+bool RuleBundle::ParseConfiguration(absl::string_view text,
+                                    std::string *error) {
+  // Clear the vector to overwrite any existing value.
+  rules.clear();
+
+  for (absl::string_view part : absl::StrSplit(text, absl::ByAnyChar(",\n"),
+                                               absl::SkipEmpty())) {
+    // If prefix is '-', the rule is disabled.
+    // Note that part is guaranteed to be at least one character because
+    // of absl::SkipEmpty()
+    const bool prefix_minus = (part[0] == '-');
+
+    RuleSetting setting = {!prefix_minus, ""};
+
+    const auto rule_name_with_config = part.substr(prefix_minus ? 1 : 0);
+
+    // Independent of the enabled-ness: extract a configuration string
+    // if there is any assignment.
+    const auto equals_pos = rule_name_with_config.find('=');
+    if (equals_pos != absl::string_view::npos) {
+      const auto config = rule_name_with_config.substr(equals_pos + 1);
+      setting.configuration.assign(config.data(), config.size());
+    }
+    const auto rule_name = rule_name_with_config.substr(0, equals_pos);
+    const auto rule_name_set = analysis::GetAllRegisteredLintRuleNames();
+    const auto rule_iter = rule_name_set.find(rule_name);
+
+    // Check if text is a valid lint rule.
+    if (rule_iter == rule_name_set.end()) {
+      *error = absl::StrCat("invalid flag \"", rule_name, "\"");
+      return false;
+    } else {
+      // Map keys must use canonical registered string_views for guaranteed
+      // lifetime, not just any string-equivalent copy.
+      rules[*rule_iter] = setting;
+    }
+  }
+
+  return true;
+}
+
+// Parse and unparse for RuleBundle (for commandlineflags)
+std::string RuleBundle::UnparseConfiguration(const char separator) const {
+  std::vector<std::string> switches;
+  for (const auto& rule : rules) {
+    switches.push_back(absl::StrCat(
+        // If rule is set off, prepend "-"
+        rule.second.enabled ? "" : "-", rule.first,
+        // If we have a configuration, append assignment.
+        rule.second.configuration.empty() ? "" : "=",
+        rule.second.configuration));
+  }
+  // Concatenates all of rules into text.
+  return absl::StrJoin(switches.rbegin(), switches.rend(),
+                       std::string(1, separator));
+}
+
 bool LinterConfiguration::RuleIsOn(const analysis::LintRuleId& rule) const {
   const auto* entry = FindOrNull(configuration_, rule);
   if (entry == nullptr) return false;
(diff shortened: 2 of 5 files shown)


verilog/analysis/verilog_linter.cc
```diff
@@ -210,6 +211,26 @@ LinterConfiguration LinterConfigurationFromFlags() {
   const auto& ruleset = absl::GetFlag(FLAGS_ruleset);
   config.UseRuleSet(ruleset);
 
+  // Read local configuration file
+  std::ifstream local_config(".rules.verible_lint");
```
hzeller: There is `verible::file::GetContents()`, which might be a utility function to use here for less code https://github.com/google/verible/blob/master/common/util/file_util.h

verilog/analysis/verilog_linter.cc
```diff
@@ -210,6 +211,26 @@ LinterConfiguration LinterConfigurationFromFlags() {
...
+  std::vector<std::string> local_rules;
+  std::string str;
+  while (std::getline(local_config, str)) {
+    local_rules.push_back(str);
+  }
+
+  if (local_rules.size() > 0) {
+    const auto& local_rules_joined = absl::StrJoin(local_rules, ",");
```
hzeller: I think we should extend our parsing to read with newline separation in addition to comma, so that we don't have to awkwardly re-assemble everything with comma.
Also, not using something called AbslParseFlag() would probably be good to avoid confusion.
 
So I suggest to decompose AbslParseFlag() and probably add a method `RuleBundle::ParseConfiguration(absl::string_view, std::string *error)`. The implementation would be lifted from `AbslParseFlag()` and slightly extended to `absl::StrSplit()` on not only comma but also newline.

Then `AbslParseFlag()` just delegates to bundle->ParseConfiguration(...)
And here in verilog_linter,  we can just local_rules_bundle.ParseConfiguration(file_content,...)

verilog/analysis/verilog_linter.cc
hzeller: We should probably define a flag `--rule_config` (paging @fangism if he has a particular preference) and use that instead of the hardcoded `.rules.verible_lint` (which we still can make the default).

fangism: Sounds ok.

fangism:
> I am opening this early to ask for comments.
> 
> Some general info:
> 
> * The configuration from file is read after setting up the default configs, but before parsing the invocation arguments. This way the arguments can easily overwrite the configuration from the file
> * I think the following should be considered (all of them should be optional):
>   
>   * a global configuration file (e.g. `~/.verible_lint.rules`)
>   * a local configuration file (e.g. the top directory of a project), this assumes the linter is executed from the top dir of the project
>   * any arbitrary file, specified with an argument
>   * If we decide to use all of the options listed above we could:
>     
>     * combine the global file, local file and file from a flag (in that order), or
>     * ignore global file if a local one exists, ignore both if another one is specified with a flag
>     * always overwrite the config from files when the same flags are specified with the `-rules` switch
> * for now the implementation does some string operations on the file to make the format compatible with the arguments format. The only difference is the settings can be put in separate lines to improve readability. Thanks to this we can reuse the same parsing code
> 
> Other options to consider:
> 
> * use a completely different format for the configuration (e.g. ini files, with a general `[rules]` section, and more rule-specific configuration)
> 
> Please share your thoughts.
> 
> CC @hzeller

If we decide to implement upward search behavior, would the upward search occur from the current working directory where the tool is invoked, or upwards from each file's directory?  I'm not sure which is more conventional.  Could we start simple with just a `--rule_config` file only, without search-and-discover behavior?  Internally, we use the linter tool in a service which may not have access to local file systems, so we would need a switch to disable some behavior.

verilog/analysis/verilog_linter.cc
fangism: Sounds good.

tgorochowik: Right, that makes perfect sense, thanks!

verilog/analysis/verilog_linter.cc
tgorochowik: Sure, moved this out. Added split on newline, thanks!

verilog/analysis/verilog_linter.cc
tgorochowik: Added a flag, thanks!

verilog/analysis/verilog_linter.cc
```diff
@@ -210,6 +212,23 @@ LinterConfiguration LinterConfigurationFromFlags() {
...
+
+    RuleBundle local_rules_bundle;
+    std::string error;
+    if (local_rules_bundle.ParseConfiguration(content, &error)) {
+      config.UseRuleBundle(local_rules_bundle);
+    } else {
+      LOG(ERROR) << "Unable to fully parse configuration: " << error << std::endl;
+    }
+  } else if(FLAGS_rules_config.IsModified()) {
+    // If flag is modified and  we were unable to open the file, report that
```
tgorochowik: Does this assumption make sense? It will be similar even when we implement the upward searching

verilog/analysis/verilog_linter_configuration.cc
```diff
@@ -83,6 +83,46 @@ std::string ProjectPolicy::ListPathGlobs() const {
...
+    } else {
+      // Map keys must use canonical registered string_views for guaranteed
+      // lifetime, not just any string-equivalent copy.
+      rules[*rule_iter] = setting;
+    }
+  }
+
+  return true;
+}
+
```
tgorochowik: Note that I didn't move the tests to apply to this yet. Should the tests be just copied?

verilog/analysis/verilog_linter_configuration.cc
```diff
@@ -83,6 +83,46 @@ std::string ProjectPolicy::ListPathGlobs() const {
                        });
 }
 
+bool RuleBundle::ParseConfiguration(absl::string_view text, std::string *error) {
+  // Clear the vector to overwrite any existing value.
+  rules.clear();
+
+  for (absl::string_view part : absl::StrSplit(text, absl::ByAnyChar(",\n"),
```
tgorochowik: This is the only line that is different when comparing to the previous implementation.

verilog/analysis/verilog_linter_configuration.cc
tgorochowik: @fangism ?

verilog/analysis/verilog_linter.cc
```diff
@@ -55,6 +55,8 @@
 ABSL_FLAG(verilog::RuleBundle, rules, {},
           "List of lint rules to enable.  "
           "Prefix a rule name with '-' to disable it.");
+ABSL_FLAG(std::string, rules_config, ".rules.verible_lint",
+          "Path to lint rules configuration file.");
```
fangism: Let's make it clear whether this is an absolute or relative path.

verilog/analysis/verilog_linter.cc
fangism: If we do upward searching, we need to make it very clear that `rules_config` is interpreted as a file name that will be used to search, and not a one-time-use path.

verilog/analysis/verilog_linter_configuration.cc
fangism: I would move-adapt the tests to avoid redundancy.

tgorochowik:
Thanks for the reviews!

As we agreed to implement upward-searching as the next step, I skipped all the comments related to this.

I moved the tests to the test the new function, and added a few more. The ParseAbslFlag function now doesn't have any tests but I guess it's ok as long as it is just a wrapper for ParseConfiguration (?).

verilog/analysis/verilog_linter.cc
tgorochowik: Sure, will do when the search is implemented.

verilog/analysis/verilog_linter_configuration.cc
tgorochowik: Sure, moved, thanks!

verilog/analysis/verilog_linter.cc
tgorochowik: It's actually both right now. This help will need elaboration when we implement the search feature.

verilog/analysis/verilog_linter_configuration.cc
```diff
@@ -262,42 +302,7 @@ std::string AbslUnparseFlag(const RuleBundle& bundle) {
...
-      return false;
-    } else {
-      // Map keys must use canonical registered string_views for guaranteed
-      // lifetime, not just any string-equivalent copy.
-      bundle->rules[*rule_iter] = setting;
-    }
-  }
-
-  return true;
+  return bundle->ParseConfiguration(text, error);
```
fangism: Would it make sense to move the Unparse code+test into RuleBundle as well?  Might be a good idea to take a separator as a parameter (e.g. `','` vs. `'\n'`).

verilog/analysis/verilog_linter_configuration_test.cc
```diff
@@ -688,7 +687,27 @@ TEST(RuleBundleTest, ParseRuleBundleReject) {
...
+  bool success = bundle.ParseConfiguration(text, &error);
+  EXPECT_FALSE(success);
+  EXPECT_EQ(error, "invalid flag \"bad-flag\"");
+}
+
+TEST(RuleBundleTest, ParseRuleBundleAcceptMultiline) {
+  std::string text = "test-rule-1\n-test-rule-2";
+  RuleBundle bundle;
+  std::string error;
+  bool success = bundle.ParseConfiguration(text, &error);
```
fangism: Add `EXPECT_TRUE(error.empty());` ?

verilog/analysis/verilog_linter.cc
fangism: Keep in mind when the config-search is implemented, we will want a way to disable all searching behavior.  (e.g. an empty `rules_config` string might suffice.)

verilog/analysis/verilog_linter_configuration.cc
tgorochowik: Sure, that makes perfect sense. Changed the implementation.

verilog/analysis/verilog_linter_configuration_test.cc
tgorochowik: Sure, added, thanks!

verilog/analysis/verilog_linter.cc
tgorochowik: Sure, I will keep that in mind in the following PRs, thanks!

PR #231, opened 2020-03-16
Add missing macro definitions to fix CI failures

tgorochowik:
Without this the CI fails. Both with current master and for all new PRs (e.g. #229)


fangism:
@hzeller this fix needs to go in soon to unblock the others

PR #232, opened 2020-03-17
Fix dynamic_cast failure on [create-object-name-match] rule

### LINKED ISSUE #172: verilog_lint crash due to "type_id::create($sformatf("instr_%0d", i));"
I found the style check crashed due to "type_id::create($sformatf("instr_%0d", i));"

Here is [link](https://github.com/google/riscv-dv/blob/f7e35d7939a27ae17b0481eb070e9a36ea335d1f/src/riscv_instr_stream.sv) to the file
```
  virtual function void create_instr_instance();
    riscv_instr instr;
    for(int i = 0; i < instr_cnt; i++) begin
      instr = riscv_instr::type_id::create($sformatf("instr_%0d", i));
      instr_list.push_back(instr);
    end
  endfunction
```

Error message:
> verilog_lint: ./common/util/casts.h:30: To verible::down_cast(From*) [with To = const verible::SyntaxTreeLeaf*; From = verible::Symbol]: Assertion `f == nullptr || dynamic_cast<To>(f) != nullptr' failed.
> xargs: /tools/verible/verilog_lint: terminated by signal 6

### LINKED ISSUE #222: dynamic_cast failure on [create-object-name-match] rule
Hi,
I see verible supports the project: https://github.com/google/riscv-dv
However, I got error when I run the script: `verilog_style/run.sh`.

The script:
```
find src/ -type f \( -name "*.sv" -o -name "*.svh" \) \
  | grep -vFf ./verilog_style/exclude_filelist.f \
  | xargs /tools/verible/verilog_lint --rules=-macro-name-style
```

The error message:
```
verilog_lint: ./common/util/casts.h:30: To verible::down_cast(From*) [with To = const verible::SyntaxTreeLeaf*; From = verible::Symbol]: Assertion `f == nullptr || dynamic_cast<To>(f) != nullptr' failed.
xargs: /tools/verible/verilog_lint: terminated by signal 6
```

Could you help me take the look? Thank you

tgorochowik:
Closes #222
Closes #172 

The comment in the file actually says:

```
// Returns token information for a single string literal expression, or nullptr                                                                                                                                                                                                                                                                                              
// if the expression is not a string literal.                                                                                                                                                                                                                                                                                                                                
// `expr_node` should be a SyntaxTreeNode tagged as an expression. 
```

So this extra check in fact verifies if the given token can even be a string literal.

The problem in #222 was that the argument was in fact an expression (called as `l("name")` instead of just `"name"`) and the present checks failed to fully verify that.

### DIFF
--- a/verilog/analysis/checkers/create_object_name_match_rule.cc
+++ b/verilog/analysis/checkers/create_object_name_match_rule.cc
@@ -121,6 +121,12 @@ static absl::string_view StripOuterQuotes(absl::string_view text) {
 static const TokenInfo* ExtractStringLiteralToken(
     const SyntaxTreeNode& expr_node) {
   if (!expr_node.MatchesTag(NodeEnum::kExpression)) return nullptr;
+
+  // this check is limited to only checking string literal leaf tokens
+  if (expr_node.children().front().get()->Kind() != verible::SymbolKind::kLeaf) {
+    return nullptr;
+  }
+
   const auto* leaf_ptr =
       down_cast<const SyntaxTreeLeaf*>(expr_node.children().front().get());
   if (leaf_ptr != nullptr) {
--- a/verilog/analysis/checkers/create_object_name_match_rule_test.cc
+++ b/verilog/analysis/checkers/create_object_name_match_rule_test.cc
@@ -110,6 +110,16 @@ TEST(CreateObjectNameMatchTest, FunctionFailures) {
        "for (int i = 0; i<N; ++i)\n"
        "  foo_h[i] = foo::type_id::create($sformat(\"asdf_%d\", i)); "
        "endfunction endclass"},  // array assignments are not checked
+      {"class crash2;\n"
+       "  virtual function void a();\n"
+       "    abc = foo::type_id::create(xyz);\n"
+       "  endfunction\n"
+       "endclass\n"}, // non string literal - not verified
+      {"class c;\n"
+       "  virtual function void g();\n"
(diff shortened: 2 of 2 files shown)


tgorochowik:
The CI will of course fail until #231 is merged.

BTW. should some new test be added?

tgorochowik:
Fixes #222 
Fixes #172

fangism:
> The CI will of course fail until #231 is merged.
> 
> BTW. should some new test be added?

Yes, add a test that would have failed/crashed before this fix, but now passes.
If you can think of other variants, it would be good to cover those as well.

verilog/analysis/checkers/create_object_name_match_rule.cc
```diff
@@ -121,6 +121,11 @@ static absl::string_view StripOuterQuotes(absl::string_view text) {
 static const TokenInfo* ExtractStringLiteralToken(
     const SyntaxTreeNode& expr_node) {
   if (!expr_node.MatchesTag(NodeEnum::kExpression)) return nullptr;
+
+  if (expr_node.children().front().get()->Kind() != verible::SymbolKind::kLeaf) {
```
fangism: // this check is limited to only checking string literal leaf tokens

tgorochowik: Sure, added

tgorochowik:
@fangism added the comment you suggested + new test cases (two test cases to cover the errors from both #222 and #172 )

PR #235, opened 2020-03-18
CI: do not build optimized binaries for testing

tgorochowik:
> --compilation_mode (fastbuild|opt|dbg) (-c)
> This option takes an argument of fastbuild, dbg or opt, and affects various C/C++ code-generation options, such as the level of optimization and the completeness of debug tables. Bazel uses a different output directory for each different compilation mode, so you can switch between modes without needing to do a full rebuild every time.
> 
> * fastbuild means build as fast as possible: generate minimal debugging information (-gmlt -Wl,-S), and don't optimize. This is the default. Note: -DNDEBUG will not be set.
> * dbg means build with debugging enabled (-g), so that you can use gdb (or another debugger).
> * opt means build with optimization enabled and **with assert() calls disabled** (-O2 -DNDEBUG). Debugging information will not be generated in opt mode unless you also pass --copt -g.

tl;dr building with `-c opt` disables all asserts. This is not what we want for testing PRs.

Ref: #222 

@mithro are the same scripts used for binary releases? The flag still makes sense for binary releases, it should only be avoided in continuous code checks.


tgorochowik:
After a discussion with @hzeller I ran some more tests locally and verified that as long as `basel test (...)` is run without `-c opt` (which it is) it recompiles everything and runs tests with asserts enabled. This PR is therefore not needed, closing.

PR #237, opened 2020-03-21
Implement upward file searching of linter configuration

### LINKED ISSUE #236: Linter configuration file - upward searching
Now that the initial linter configuration file support is merged (#229), we can work on adding support for searching for the file.

After some research I think it would be nice to mimic what [yapf](https://github.com/google/yapf) does.

Here is the code responsible for searching:
https://github.com/google/yapf/blob/b306e98387b61a6adaeda4a9ef17c02e4184f559/yapf/yapflib/file_resources.py#L67-L109

To sum up:
* define a default file name to be searched for (i.e. `.rules.verible_lint`)
* look for that file starting in the directory of the linted source file
* look up the directories until the file is found
* it's okay if nothing is found
* optional: if not found look for a default global in `/etc/` or `~/.config/verible`

Additional configuration flags:
* `--rule_config` - defaults to empty, if specified the configuration is read from this particular file, the searching mechanism is not used
* `--no_rule_config` - defaults to false, if enabled the searching mechanism is not used and no configuration file will be used
* The two flags above could in theory be combined into one, but I think it's cleaner with two flags

I'm going to start working on the implementation described above. I am opening this issue earlier that the PR to let you guys add some comments before it's done if you done if you done agree with something.

tgorochowik:
Initial implementation, this feature probably still needs some discussion.

For now it mostly follows what was described in #236, taking the initial comments into account.

Fixes #236.

### DIFF
--- a/common/util/file_util.cc
+++ b/common/util/file_util.cc
@@ -28,6 +28,8 @@
 
 #include "absl/status/status.h"
 #include "absl/strings/str_cat.h"
+#include "absl/strings/str_split.h"
+#include "absl/strings/str_join.h"
 #include "absl/strings/string_view.h"
 #include "common/util/logging.h"
 
@@ -67,6 +69,50 @@ static absl::Status CreateErrorStatusFromErrno(const char *fallback_msg) {
     default:
       return absl::Status(StatusCode::kUnknown, system_msg);
   }
+}
+
+// TODO (after bump to c++17) rewrite this function to use std::filesystem
+absl::Status UpwardFileSearch(absl::string_view start,
+                              absl::string_view filename,
+                              std::string* result) {
+  static constexpr char dir_separator[] = "/";
+
+  /* Convert to absolute path */
+  char absolute_path[PATH_MAX];
+  if (realpath(std::string(start).c_str(), absolute_path) < 0) {
+    return absl::InvalidArgumentError("Invalid config path specified.\n");
+  }
+
+  /* Add trailing slash */
+  const std::string full_path = absl::StrCat(absolute_path, dir_separator);
+
+  VLOG(1) << "Upward search for " << filename << ", starting in " << start;
+
+  size_t search_up_to = full_path.length();
+  do {
+    const size_t separator_pos = full_path.rfind(dir_separator, search_up_to);
+
+    if (separator_pos == search_up_to || separator_pos == std::string::npos) {
+      break;
+    }
+
+    const std::string candidate =
+      absl::StrCat(full_path.substr(0, separator_pos + 1), filename);
+
+    if (access(candidate.c_str(), R_OK) != -1) {
+      *result = candidate;
+      VLOG(1) << "Found: " << *result;
+      return absl::OkStatus();
+    }
+
+    search_up_to = separator_pos - 1;
+
+    if (separator_pos == 0) {
+      break;
+    }
+  } while (true);
+
+  return absl::NotFoundError("No matching file found.\n");
 }
 
 absl::Status GetContents(absl::string_view filename, std::string *content) {
--- a/common/util/file_util.h
+++ b/common/util/file_util.h
@@ -35,6 +35,13 @@ absl::string_view Basename(absl::string_view filename);
 // there is no "." in the basename, this is equivalent to file::Basename(path).
 absl::string_view Stem(absl::string_view filename);
 
+// Search for 'filename' file starting at 'start' (can be file or directory)
+// and going upwards the directory tree. Returns true if anything is found and
+// puts the found file in 'result'.
+absl::Status UpwardFileSearch(absl::string_view start,
+                              absl::string_view filename,
+                              std::string* result);
+
 // Read file "filename" and store its content in "content"
 absl::Status GetContents(absl::string_view filename, std::string *content);
 
--- a/verilog/analysis/verilog_linter.cc
+++ b/verilog/analysis/verilog_linter.cc
@@ -57,7 +57,11 @@ ABSL_FLAG(verilog::RuleBundle, rules, {},
           "No prefix or a '+' prefix enables it, '-' disable it. "
           "Configuration values for each rules placed after '=' character.");
 ABSL_FLAG(std::string, rules_config, ".rules.verible_lint",
-          "Path to lint rules configuration file.");
+          "Path to lint rules configuration file. "
+          "Disables rule_config_search.");
+ABSL_FLAG(bool, rules_config_search, true,
+          "Look for lint rules configuration file, "
+          "searching upward from the location of each analyzed file.");
 ABSL_FLAG(verilog::RuleSet, ruleset, verilog::RuleSet::kDefault,
           "[default|all|none], the base set of rules used by linter");
 ABSL_FLAG(std::string, waivers_file, "",
@@ -228,7 +232,30 @@ std::vector<LintRuleStatus> VerilogLinter::ReportStatus(
   return statuses;
 }
 
-LinterConfiguration LinterConfigurationFromFlags() {
+absl::Status AppendLinterConfigurationFromFile(
+    LinterConfiguration *config, absl::string_view config_filename) {
+  // Read local configuration file
+  std::string content;
+
+  const absl::Status config_read_status =
+      verible::file::GetContents(config_filename, &content);
+  if (config_read_status.ok()) {
+    RuleBundle local_rules_bundle;
+    std::string error;
+    if (local_rules_bundle.ParseConfiguration(content, '\n', &error)) {
+      config->UseRuleBundle(local_rules_bundle);
+    } else {
+      LOG(WARNING) << "Unable to fully parse configuration: " << error;
+    }
+    return absl::OkStatus();
+  }
+
+  return config_read_status;
+}
+
+LinterConfiguration LinterConfigurationFromFlags(
+    absl::string_view config_filename) {
+  static constexpr absl::string_view linter_config_file = ".rules.verible_lint";
   LinterConfiguration config;
 
   // TODO move all of these calls to GetFlag outside of this function
@@ -237,24 +264,35 @@ LinterConfiguration LinterConfigurationFromFlags() {
   const auto& ruleset = absl::GetFlag(FLAGS_ruleset);
   config.UseRuleSet(ruleset);
 
-  // Read local configuration file
-  std::string content;
-  const absl::Status config_read_status =
-      verible::file::GetContents(absl::GetFlag(FLAGS_rules_config), &content);
-  if (config_read_status.ok()) {
-    RuleBundle local_rules_bundle;
-    std::string error;
-    if (local_rules_bundle.ParseConfiguration(content, '\n', &error)) {
-      config.UseRuleBundle(local_rules_bundle);
-    } else {
-      LOG(ERROR) << "Unable to fully parse configuration: " << error
-                 << std::endl;
+  if (FLAGS_rules_config.IsModified()) {
+    // Use configuration file from flag if specified
+    std::string config_file = absl::GetFlag(FLAGS_rules_config);
+
+    const absl::Status config_read_status =
+        AppendLinterConfigurationFromFile(&config, config_file);
+
+    if (!config_read_status.ok()) {
+      LOG(WARNING) << config_file
+                   << ": Unable to read rules configuration file "
+                   << config_read_status << std::endl;
     }
-  } else if (FLAGS_rules_config.IsModified()) {
-    // If flag is modified and  we were unable to open the file, report that
-    LOG(WARNING) << absl::GetFlag(FLAGS_rules_config)
-                 << ": Unable to read rules configuration file "
-                 << config_read_status << std::endl;
+
+  } else if (absl::GetFlag(FLAGS_rules_config_search)) {
+    // Search upward if search is enabled and no configuration file is
+    // specified
+    std::string resolved_config_file;
+    if (verible::file::UpwardFileSearch(config_filename,
+                                        linter_config_file,
+                                        &resolved_config_file).ok()) {
+      const absl::Status config_read_status =
+          AppendLinterConfigurationFromFile(&config, resolved_config_file);
+
+      if (!config_read_status.ok()) {
+        LOG(WARNING) << resolved_config_file
+                     << ": Unable to read rules configuration file "
+                     << config_read_status << std::endl;
+      }
+    }
   }
 
   // Turn on rules found in config flags.
--- a/verilog/analysis/verilog_linter.h
+++ b/verilog/analysis/verilog_linter.h
@@ -83,7 +83,12 @@ class VerilogLinter {
 };
 
 // Creates a linter configuration from global flags.
-LinterConfiguration LinterConfigurationFromFlags();
+LinterConfiguration LinterConfigurationFromFlags(
+    absl::string_view config_filename);
+
+// Expands linter configuration from a text file
+absl::Status AppendLinterConfigurationFromFile(
+    LinterConfiguration *config, absl::string_view config_filename);
 
 // VerilogLintTextStructure analyzes Verilog syntax tree for style violations
 // and syntactically detectable pitfalls.
--- a/verilog/tools/lint/verilog_lint.cc
+++ b/verilog/tools/lint/verilog_lint.cc
@@ -70,15 +70,13 @@ int main(int argc, char** argv) {
     return 0;
   }
 
-  const LinterConfiguration baseline_config(
-      verilog::LinterConfigurationFromFlags());
-
   int exit_status = 0;
   // All positional arguments are file names.  Exclude program name.
   for (const auto filename :
        verible::make_range(args.begin() + 1, args.end())) {
     // Copy configuration, so that it can be locally modified per file.
-    LinterConfiguration config(baseline_config);
+    const LinterConfiguration config(
+        verilog::LinterConfigurationFromFlags(filename));
 
     const int lint_status = verilog::LintOneFile(
         &std::cout, filename, config, absl::GetFlag(FLAGS_parse_fatal),
--- a/verilog/tools/lint/verilog_style_lint.bzl
+++ b/verilog/tools/lint/verilog_style_lint.bzl
@@ -208,19 +208,23 @@ _verilog_style_lint_diagnostics_test = rule(
     implementation = _style_lint_test_diagnostics_impl,
 )
 
-def _verilog_style_lint_test(name, srcs, flags = None, expect_fail = None):
+def _verilog_style_lint_test(name, srcs, flags = [], expect_fail = None):
     """Macro for running Verilog style lint tests in the silo.
 
     Args:
       name: name of test.
       srcs: list of source files to scan (e.g. can use glob()).
       flags: list of flags for running the verilog_lint binary.
+             Note: --norules_config_search is used for each test.
       expect_fail: if True, expect to find errors (invert status).
     """
+    forced_flags = [
+        "--norules_config_search",
+    ]
     _verilog_style_lint_report(
         name = name + "-report",
         srcs = srcs,
-        flags = flags,
+        flags = flags + forced_flags,
     )
     _verilog_style_lint_diagnostics_test(
         name = name,


common/util/file_util.h
```diff
@@ -34,6 +34,12 @@ absl::string_view Basename(absl::string_view filename);
 // there is no "." in the basename, this is equivalent to file::Basename(path).
 absl::string_view Stem(absl::string_view filename);
 
+// Search for 'filename' file starting at 'start' (can be file or directory)
+// and going upwards the directory tree. Returns true if anything is found and
+// puts the found file in 'result'.
+bool UpwardFileSearch(std::string start, std::string filename,
```
tgorochowik: This is an util so it can be used by others too (might be useful e.g. for waiver files if the decision will be to implement them).

verilog/analysis/verilog_linter.cc
```diff
@@ -205,28 +209,53 @@ std::vector<LintRuleStatus> VerilogLinter::ReportStatus(
...
+                   << config_file;
+    }
+  } else if (absl::GetFlag(FLAGS_rules_config_search)) {
+    // Search upward if search is enabled and no configuration file is
+    // specified
+    std::string config_file;
+    if (verible::file::UpwardFileSearch(std::string(filename),
+                                        linter_config_file,
+                                        config_file)) {
+      if (!AppendLinterConfigurationFromFile(config, config_file)) {
```
tgorochowik: I really think we're missing `--verbose` and `LOG(VERBOSE)` - this would be useful to log things like the file that is found and used here.

verilog/analysis/verilog_linter.cc
```diff
@@ -55,8 +55,12 @@
 ABSL_FLAG(verilog::RuleBundle, rules, {},
           "List of lint rules to enable.  "
           "Prefix a rule name with '-' to disable it.");
-ABSL_FLAG(std::string, rules_config, ".rules.verible_lint",
-          "Path to lint rules configuration file.");
+ABSL_FLAG(std::string, rules_config, "",
+          "Path to lint rules configuration file. "
+          "Disables rule_config_search.");
+ABSL_FLAG(bool, rules_config_search, true,
```
tgorochowik: We will need to somehow disable this by default for all tests.

common/util/file_util.cc
```diff
@@ -46,6 +48,34 @@ absl::string_view Stem(absl::string_view filename) {
...
+
+  std::vector<absl::string_view> search = absl::StrSplit(start, '/');
+
+  /* Push back linter configuration file name */
+  search.push_back(filename);
+
+  while (search.size() > 1) {
+    std::string candidate = absl::StrJoin(search,  "/");
+
+    if (access(candidate.c_str(), R_OK) != -1) {
```
tgorochowik: It probably makes sense to consider using `std:: filesystem`, but that would change the std requirements and as it wasn't used in other helpers I didn't use it here.

verilog/analysis/verilog_linter.cc
fangism: We do have `VLOG` (frequently used in the codebase).

verilog/analysis/verilog_linter.cc
fangism: For the sake of unit-testing, I recommend passing all flags in via some function parameter rather than accessing a global variable.  The style linter could use some clean-up in this regard.

common/util/file_util.cc
```diff
@@ -46,6 +48,34 @@ absl::string_view Stem(absl::string_view filename) {
              : filename.substr(0, last_dot_pos);
 }
 
+bool UpwardFileSearch(std::string start, std::string filename,
```
fangism: Recommend making (non-modified) string function parameters `absl::string_view`.

common/util/file_util.cc
```diff
@@ -46,6 +48,34 @@ absl::string_view Stem(absl::string_view filename) {
...
+                      std::string& result) {
+  /* Convert to absolute path */
+  char absolute_path[PATH_MAX];
+  if (realpath(std::string(start).c_str(), absolute_path) < 0) {
+    return false;
+  }
+  start = std::string(absolute_path);
+
+  std::vector<absl::string_view> search = absl::StrSplit(start, '/');
```
fangism: Maybe rename `search` to `path_components`?

common/util/file_util.cc
```diff
@@ -46,6 +48,34 @@ absl::string_view Stem(absl::string_view filename) {
...
+
+  /* Push back linter configuration file name */
+  search.push_back(filename);
+
+  while (search.size() > 1) {
+    std::string candidate = absl::StrJoin(search,  "/");
```
fangism: Just an implementation idea that reduces string-copying and allocation:

* full path consists of a directory component and file (base) name component.
* maintain a `string_view` of the directory component
* find all (iterator) positions that contain `/`
* shrink the string_view from the back with `remove_suffix`
* `absl::StrCat` the shrinking directory component with file name -> candidate.

verilog/analysis/verilog_linter.cc
tgorochowik: @fangism but `VLOG` still does not have any `verbose` severity, moreover it appears to work exactly the same as regular `LOG`. Which one should be used then?

Should I add logging the config file used with `[V]LOG(INFO)` then? It seems redundant to log it each time, but I think it would be useful if there was a `--verbose` switch.

fangism: With both `LOG` and `VLOG` messages get print to stderr if you run with `--alsologtostderr`.
When you use `VLOG(N)`, messages are controlled with `-v=M` for `M >= N`.
I'd recommend `VLOG(1)` to surface these messages with minimum verbosity.  Does that help?

tgorochowik: Thanks!

But it's not really what I meant. `VLOG(1)`  is basically `INFO`, this would be printed all the time with the default settings. I think that the information about the used config file should not be visible by default, unless the user asks for it. The `-v=X` flag also doesn't seem to work for me. Should it be available for linter?

verilog/analysis/verilog_linter.cc
tgorochowik: I modified the base of the integration tests for linter to always include this flag. Is that ok?

I agree that the flags could be handled differently and the code could be cleaned-up, but that would not really solve the problem with forcing a flag for all integration tests.

common/util/file_util.cc
tgorochowik: Sure, changed, thanks

common/util/file_util.cc
tgorochowik: Changed the implementation to somewhat reflect your suggestions. Is the current implementation acceptable?

tgorochowik:
I see that this needs to be rebased before it can be accepted.

@fangism @hzeller is there anything besides the rebase that needs to be addressed before this can be accepted?

common/util/file_util.cc
```diff
@@ -46,6 +48,39 @@ absl::string_view Stem(absl::string_view filename) {
              : filename.substr(0, last_dot_pos);
 }
 
+bool UpwardFileSearch(absl::string_view start, absl::string_view filename,
+                      std::string& result) {
+  /* Convert to absolute path */
+  char absolute_path[PATH_MAX];
+  if (realpath(std::string(start).c_str(), absolute_path) < 0) {
```
fangism, line 55: `start.data()` is also a const char*, no need to copy to std::string.

common/util/file_util.cc
```diff
@@ -46,6 +48,39 @@ absl::string_view Stem(absl::string_view filename) {
...
+    return false;
+  }
+
+  /* Add trailing slash */
+  std::string init = absl::StrCat(absolute_path, "/");
+
+  size_t r = 0, n = init.length();
+  do {
+    r = init.rfind("/", n);
+
```
fangism, line 65: In this loop, it would be a good idea to add some VLOG(1) statements to be able to trace this file access behavior for debugging which file was actually found.

common/util/file_util.cc
```diff
@@ -46,6 +48,39 @@ absl::string_view Stem(absl::string_view filename) {
...
+bool UpwardFileSearch(absl::string_view start, absl::string_view filename,
+                      std::string& result) {
+  /* Convert to absolute path */
+  char absolute_path[PATH_MAX];
+  if (realpath(std::string(start).c_str(), absolute_path) < 0) {
+    return false;
+  }
+
+  /* Add trailing slash */
+  std::string init = absl::StrCat(absolute_path, "/");
```
fangism, line 60: Could we use `std::filesystem::path::preferred_separator` for portability?  (ok to make a short local alias for that)

verilog/analysis/verilog_linter.cc
```diff
@@ -55,8 +55,12 @@
...
+          "Look for lint rules configuration file, "
+          "starting next to the linted up the directory tree.");
```
fangism, line 63: "searching upward from the location of each analyzed file" (?)

verilog/analysis/verilog_linter.cc
```diff
@@ -205,28 +209,53 @@ std::vector<LintRuleStatus> VerilogLinter::ReportStatus(
...
-  const auto& ruleset = absl::GetFlag(FLAGS_ruleset);
-  config.UseRuleSet(ruleset);
+const std::string linter_config_file = ".rules.verible_lint";
 
+bool AppendLinterConfigurationFromFile(
+    LinterConfiguration &config, std::string config_filename) {
   // Read local configuration file
   std::string content;
-  if (verible::file::GetContents(absl::GetFlag(FLAGS_rules_config), &content)) {
+  if (verible::file::GetContents(config_filename, &content)) {
```
fangism, line 218: Note that this access is relative to the current working directory.  Need to make sure behavior is consistent with documentation.

verilog/analysis/verilog_linter.cc
```diff
@@ -205,28 +209,53 @@ std::vector<LintRuleStatus> VerilogLinter::ReportStatus(
...
+
+    if (!AppendLinterConfigurationFromFile(config, config_file)) {
+      LOG(WARNING) << "Unable to open linter configuration file: "
+                   << config_file;
+    }
+  } else if (absl::GetFlag(FLAGS_rules_config_search)) {
+    // Search upward if search is enabled and no configuration file is
+    // specified
+    std::string config_file;
+    if (verible::file::UpwardFileSearch(std::string(filename),
```
fangism, line 251: no longer need to copy to std::string.

verilog/analysis/verilog_linter.cc
fangism: @hzeller is there a behavioral difference between internal/external logging?

verilog/analysis/verilog_linter.cc
```diff
@@ -205,28 +209,53 @@ std::vector<LintRuleStatus> VerilogLinter::ReportStatus(
...
-LinterConfiguration LinterConfigurationFromFlags() {
-  LinterConfiguration config;
-
-  // Turn on default ruleset.
-  const auto& ruleset = absl::GetFlag(FLAGS_ruleset);
-  config.UseRuleSet(ruleset);
+const std::string linter_config_file = ".rules.verible_lint";
 
+bool AppendLinterConfigurationFromFile(
+    LinterConfiguration &config, std::string config_filename) {
```
fangism, line 215: can we make config_filename an absl::string_view?

common/util/file_util.h
```diff
@@ -34,6 +34,12 @@ absl::string_view Basename(absl::string_view filename);
 // there is no "." in the basename, this is equivalent to file::Basename(path).
 absl::string_view Stem(absl::string_view filename);
 
+// Search for 'filename' file starting at 'start' (can be file or directory)
+// and going upwards the directory tree. Returns true if anything is found and
+// puts the found file in 'result'.
+bool UpwardFileSearch(absl::string_view start, absl::string_view filename,
+                      std::string& result);
```
fangism, line 41: [style] pass modified objects by pointers instead of reference, so `std::string* result`.

common/util/file_util.cc
tgorochowik: Sure, however the README file says:
> To build, you need the bazel build system and a C++11 compatible compiler (...)

To use your suggestion we would need `c++17`.

verilog/analysis/verilog_linter.cc
tgorochowik: I think this is consistent with what all the tools do when providing a path to file. This can either be relative to current working directory or an absolute path.
Do you think the help message should be saying that explicitly?

common/util/file_util.cc
tgorochowik: absolutely right, thanks!

common/util/file_util.cc
fangism: The next best thing may be to define this locally as a constant.

verilog/analysis/verilog_linter.cc
fangism: Hmm, this seems ok here.  I might have been thinking elsewhere around the upward file search.

verilog/analysis/verilog_linter.cc
```diff
@@ -206,28 +210,53 @@ std::vector<LintRuleStatus> VerilogLinter::ReportStatus(
...
-LinterConfiguration LinterConfigurationFromFlags() {
-  LinterConfiguration config;
-
-  // Turn on default ruleset.
-  const auto& ruleset = absl::GetFlag(FLAGS_ruleset);
-  config.UseRuleSet(ruleset);
+const std::string linter_config_file = ".rules.verible_lint";
 
+bool AppendLinterConfigurationFromFile(
+    LinterConfiguration &config, absl::string_view config_filename) {
```
fangism, line 216: Pass modified objects by pointer instead of reference (few exceptions): `config`.

verilog/analysis/verilog_linter.cc
```diff
@@ -206,28 +210,53 @@ std::vector<LintRuleStatus> VerilogLinter::ReportStatus(
...
+  }
+
+  return false;
+}
+
+LinterConfiguration LinterConfigurationFromFlags(absl::string_view filename) {
+  LinterConfiguration config;
+
+  // Turn on default ruleset.
+  const auto& ruleset = absl::GetFlag(FLAGS_ruleset);
```
fangism, line 237: Since this is new code, can we avoid accessing global flags here, and pass their values in?  It makes testing that much easier.

verilog/analysis/verilog_linter.h
```diff
@@ -83,7 +83,11 @@ class VerilogLinter {
 };
 
 // Creates a linter configuration from global flags.
-LinterConfiguration LinterConfigurationFromFlags();
+LinterConfiguration LinterConfigurationFromFlags(absl::string_view filename);
```
fangism, line 86: `config_filename`?

verilog/analysis/verilog_linter.cc
```diff
@@ -206,28 +210,53 @@ std::vector<LintRuleStatus> VerilogLinter::ReportStatus(
...
+    std::string config_file = absl::GetFlag(FLAGS_rules_config);
+
+    if (!AppendLinterConfigurationFromFile(config, config_file)) {
+      LOG(WARNING) << "Unable to open linter configuration file: "
+                   << config_file;
+    }
+  } else if (absl::GetFlag(FLAGS_rules_config_search)) {
+    // Search upward if search is enabled and no configuration file is
+    // specified
+    std::string config_file;
```
fangism, line 251: `resolved_config_file` ?  (name these to make clear the relationship between filenames)

common/util/file_util.cc
tgorochowik: Added

common/util/file_util.cc
tgorochowik: I think that the whole function could be rewritten to use `std::filesystem` - I added a comment with TODO

verilog/analysis/verilog_linter.cc
tgorochowik: It's not completely new - the `LinterConfigurationFromFlags` already existed.

I've added a TODO to clean up this function in the PR with waiver files. I will be happy to clean this up once one of these PRs is merged (I'd rather not do this earlier as both my PRs modify this function, but neither introduces it, so to avoid conflicts I'd rather at least one of them was merged before cleaning this up).

common/util/file_util.cc
```diff
@@ -46,6 +48,46 @@ absl::string_view Stem(absl::string_view filename) {
              : filename.substr(0, last_dot_pos);
 }
 
+// TODO (after bump to c++17) rewrite this function to use std::filesystem
+bool UpwardFileSearch(absl::string_view start, absl::string_view filename,
+                      std::string* result) {
+  static const std::string dir_separator = "/";
```
fangism, line 54: or `constexpr absl::string_view` (no allocation)

verilog/analysis/verilog_linter.cc
fangism: OK.

common/util/file_util.cc
```diff
@@ -46,6 +48,46 @@ absl::string_view Stem(absl::string_view filename) {
...
+  if (realpath(start.data(), absolute_path) < 0) {
+    return false;
+  }
+
+  /* Add trailing slash */
+  std::string init = absl::StrCat(absolute_path, dir_separator);
+
+  VLOG(1) << "Upward search for " << filename << ", starting in " << start;
+
+  size_t r = 0, n = init.length();
```
hzeller, line 67: I'd name `n` a bit more so that the code below reads more fluently without having to wonder what `n` means.
Maybe something like `search_up_to` ?

common/util/file_util.cc
```diff
@@ -46,6 +48,46 @@ absl::string_view Stem(absl::string_view filename) {
...
+  do {
+    r = init.rfind(dir_separator, n);
+
+    if (r == n) {
+      return false;
+    }
+
+    std::string candidate;
+    candidate = absl::StrCat(init.substr(0, r + 1), filename);
```
hzeller, line 76: If you assign it in the declaration, you can make it const.
```
const std::string candidate 
   = absl::StrCat(full_path.substr(0, separator_pos + 1), filename);
```

common/util/file_util.cc
```diff
@@ -46,6 +48,46 @@ absl::string_view Stem(absl::string_view filename) {
...
+
+    if (access(candidate.c_str(), R_OK) != -1) {
+      *result = candidate;
+      VLOG(1) << "Found: " << *result;
+      return true;
+    }
+
+    n = r - 1;
```
hzeller, line 84: With the suggestions above, this now becomes a more readable
`search_up_to = separator_pos - 1;`

common/util/file_util.cc
```diff
@@ -46,6 +48,46 @@ absl::string_view Stem(absl::string_view filename) {
...
+  /* Add trailing slash */
+  std::string init = absl::StrCat(absolute_path, dir_separator);
+
+  VLOG(1) << "Upward search for " << filename << ", starting in " << start;
+
+  size_t r = 0, n = init.length();
+  do {
+    r = init.rfind(dir_separator, n);
+
+    if (r == n) {
```
hzeller, line 71: This essentially hopes that the `realpath()` generated a path with slashes all the way back, that we have the same path separator on this system than we expect with our `dir_separator` etc.
So while unlikely that things go wrong, I'd make this defensive and formulate it as `if (separator_pos == n || separator_pos == std::string::npos)`

common/util/file_util.cc
tgorochowik: Thanks for the suggestions, changed

common/util/file_util.cc
```diff
@@ -46,6 +48,46 @@ absl::string_view Stem(absl::string_view filename) {
...
 }
 
+// TODO (after bump to c++17) rewrite this function to use std::filesystem
+bool UpwardFileSearch(absl::string_view start, absl::string_view filename,
+                      std::string* result) {
+  static const std::string dir_separator = "/";
+
+  /* Convert to absolute path */
+  char absolute_path[PATH_MAX];
+  if (realpath(start.data(), absolute_path) < 0) {
```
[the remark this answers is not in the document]
tgorochowik: Sure, changed it back to a copy + c_str

common/util/file_util.cc
```diff
@@ -46,6 +48,46 @@ absl::string_view Stem(absl::string_view filename) {
...
+    return false;
+  }
+
+  /* Add trailing slash */
+  std::string init = absl::StrCat(absolute_path, dir_separator);
```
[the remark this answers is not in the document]
tgorochowik: Renamed all the variables according to suggestions, thanks!

common/util/file_util.cc
```diff
@@ -46,6 +48,46 @@ absl::string_view Stem(absl::string_view filename) {
...
+
+  VLOG(1) << "Upward search for " << filename << ", starting in " << start;
+
+  size_t r = 0, n = init.length();
+  do {
+    r = init.rfind(dir_separator, n);
```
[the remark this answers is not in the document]
tgorochowik: To do that I've also had to change the loop to an endless loop + add a check with a break inside.

tgorochowik:
@hzeller thanks for the review, I addressed all your comments, could you take a look at it?

tgorochowik:
Rebased and resolved the conflicts.

tgorochowik:
Rebased again, converted the new functions to use `absl::Status` as others too at the moment.

tgorochowik:
@hzeller would you mind taking a look? this code is already starting to bit rot :)

verilog/analysis/verilog_linter.cc
hzeller: Yes, right  now the equivalent of -v=X is setting the environment variable `TF_CPP_VMODULE` (why TF ? well, I think this is because i copied it from tensor-flow and missed that reference. Probably should rename that to `CPP_VMODULE`)

verilog/analysis/verilog_linter.cc
```diff
@@ -228,33 +232,68 @@ std::vector<LintRuleStatus> VerilogLinter::ReportStatus(
...
 
-LinterConfiguration LinterConfigurationFromFlags() {
-  LinterConfiguration config;
-
-  // TODO move all of these calls to GetFlag outside of this function
-
-  // Turn on default ruleset.
-  const auto& ruleset = absl::GetFlag(FLAGS_ruleset);
-  config.UseRuleSet(ruleset);
+const std::string linter_config_file = ".rules.verible_lint";
```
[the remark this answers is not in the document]
tgorochowik: Sure, thanks for the review! Changed that to a local constexpr and pushed the update.

common/util/file_util.cc
```diff
@@ -69,6 +71,50 @@ static absl::Status CreateErrorStatusFromErrno(const char *fallback_msg) {
   }
 }
 
+// TODO (after bump to c++17) rewrite this function to use std::filesystem
+absl::Status UpwardFileSearch(absl::string_view start,
```
hzeller, line 75: Whlie playing with the thought of introducing `std::filesystem`, I noticed there is no unit-test for this function.

Background: I am considering that we should move forward with C++17; given that it is 2021, this sounds doable. (in fact, we have been compiling it for a while now as it was a temporary requirement for Kythe - nobody complained, but of course we need to double check that all releases on older platforms work.)

To be ready for more platform independent compilation (I am thinking of next frontier #307, now that MacOS is working), we can do that with `std::filesystem` and avoid Unix-specific calls. But of course prerequisite is to have unit tests to document the desired behavior...
(We need coverage ... #654 )

tgorochowik: What would be the best way to add unit tests for those? Setting up some fake virtual filesystem? Creating a tree in temp? We also need to make sure it's platform independent too

hzeller: Luckily we don't need a virtual filesystem, we can use the testing directory provided in the unit test framework with `testing::TempDir()`; that is a path provided by the sandboxed environment. We can create anything we want underneath.

Then use our own [`CreateDir()`](https://github.com/google/verible/blob/6ef4938d0a800b93ebf9fa7fdf692c4ed06c0471/common/util/file_util.h#L75) and  [`SetContents()`](https://github.com/google/verible/blob/6ef4938d0a800b93ebf9fa7fdf692c4ed06c0471/common/util/file_util.h#L69) functions to create a tree structure in the `testing::TempDir()`, and then test it with that.

Using that, we already have at least the test somewhat platform independent as we don't use platform specific functionality, only our own wrappers.

PR #246, opened 2020-03-27
Reshape method and function calls

### LINKED ISSUE #240: method call args not indented/wrapped/aligned
The following input:

```systemverilog
task t;
	tf.check(a, //comment1
		b, //comment2
		c);  // comment3
endtask
```

formats to:

```systemverilog
task t;
  tf.check(a,  //comment1
  b,  //comment2
  c);  // comment3
endtask
```

but should be more like (when not wrapping `a`):

```systemverilog
task t;
  tf.check(a,  //comment1
           b,  //comment2
           c);  // comment3
endtask
```

Current partitioning (`verilog_format --show_token_partition_tree`):

```
Full token partition tree:
  { ([<auto>], policy: fit-else-expand) @{0}
    { ([task t ;], policy: append-fitting-sub-partitions) }
    { (>>[<auto>], policy: fit-else-expand) @{0,1}
      { (>>[tf . check ( a , //comment1], policy: fit-else-expand) }
      { (>>[b , //comment2], policy: fit-else-expand) }
      { (>>[c ) ; // comment3], policy: fit-else-expand) }
    }
    { ([endtask], policy: always-expand) }
  }
}
```

The partition that encapsulates the method call should probably be 'append-fitting'.
@ldalek-antmicro would you like to tackle this one, since you worked on 'append-fitting'?

ldalek-antmicro:
Fixes #240

### DIFF
--- a/verilog/formatting/formatter_test.cc
+++ b/verilog/formatting/formatter_test.cc
@@ -597,10 +597,7 @@ static const std::initializer_list<FormatterTestCase> kFormatterTestCases = {
     },
     {
         "  parameter  int   foo=$bar(-  z,- y ) ;",
-        "parameter int foo = $bar(\n"
-        "    -z,\n"
-        "    -y\n"
-        ");\n",
+        "parameter int foo = $bar(-z, -y);\n",
     },
     {"  parameter int a=b&~(c<<d);", "parameter int a = b & ~(c << d);\n"},
     {"  parameter int a=~~~~b;", "parameter int a = ~~~~b;\n"},
--- a/verilog/formatting/formatter_test.cc
+++ b/verilog/formatting/formatter_test.cc
@@ -3516,6 +3516,138 @@ static const std::initializer_list<FormatterTestCase> kFormatterTestCases = {
      "  );\n"
      "endfunction : warranty\n"},
 
+    // Group of tests testing partitioning of arguments inside function calls
+    {
+        // function with function call inside if statement header
+        "function foo;if(aa(bb,cc));endfunction\n",
+        "function foo;\n"
+        "  if (aa(bb, cc));\n"
+        "endfunction\n"
+    },
+    {
+        // function with function call inside if statement header and with begin-end block
+        "function foo;if (aa(bb,cc,dd,ee))begin end endfunction\n",
+        "function foo;\n"
+        "  if (aa(bb, cc, dd, ee)) begin\n"
+        "  end\n"
+        "endfunction\n"
+    },
+    {
+        // function with kMethodCallExtension inside if statement header and with begin-end block
+        "function foo;if (aa.bb(cc,dd,ee))begin end endfunction\n",
+        "function foo;\n"
+        "  if (aa.bb(cc, dd, ee)) begin\n"
+        "  end\n"
+        "endfunction\n"
+    },
+    {
+        // nested kMethodCallExtension calls - one level
+        "function foo;aa.bb(cc.dd(a1), ee.ff(a2));endfunction\n",
+        "function foo;\n"
+        "  aa.bb(cc.dd(a1), ee.ff(a2));\n"
+        "endfunction\n"
+    },
+    {
+        // nested kMethodCallExtension calls - two level
+        "function foo;aa.bb(cc.dd(a1.b1(a2), b1), ee.ff(c1, d1));endfunction\n",
+        "function foo;\n"
+        "  aa.bb(cc.dd(a1.b1(a2), b1), ee.ff(\n"
+        "        c1, d1));\n"
+        "endfunction\n"
+    },
+
+    {
+        // simple initial statement with function call
+        "module m;initial aa(bb,cc,dd,ee);endmodule\n",
+        "module m;\n"
+        "  initial aa(bb, cc, dd, ee);\n"
+        "endmodule\n"
+    },
+    {
+        // expressions and function calls inside if-statement headers
+        "module m;initial begin if(aa(bb)==cc(dd))a=b;if (xx()) b = a;end endmodule\n",
+        "module m;\n"
+        "  initial begin\n"
+        "    if (aa(bb) == cc(dd)) a = b;\n"
+        "    if (xx()) b = a;\n"
+        "  end\n"
+        "endmodule\n"
+    },
+    {
+        // fuction with two arguments inside if-statement headers
+        "module\nm;initial\nbegin\nif(aa(bb,cc))x=y;end\nendmodule\n",
+        "module m;\n"
+        "  initial begin\n"
+        "    if (aa(bb, cc)) x = y;\n"
+        "  end\n"
+        "endmodule\n"
+    },
+    {
+        // kMethodCallExtension inside if-statement headers
+        "module m;initial begin if (aa.bb(cc)) x = y;end endmodule",
+        "module m;\n"
+        "  initial begin\n"
+        "    if (aa.bb(cc)) x = y;\n"
+        "  end\n"
+        "endmodule\n"
+    },
+    {
+        // initial statement with object method call
+        "module m; initial a.b(a,b,c); endmodule\n",
+        "module m;\n"
+        "  initial a.b(a, b, c);\n"
+        "endmodule\n"
+    },
+    {
+        // initial statement with method call on indexed object
+        "module m; initial a[i].b(a,b,c); endmodule\n",
+        "module m;\n"
+        "  initial a[i].b(a, b, c);\n"
+        "endmodule\n"
+    },
+    {
+        // initial statement with method call on function returned object
+        "module m; initial a(d,e,f).b(a,b,c); endmodule\n",
+        "module m;\n"
+        "  initial a(d, e, f).b(a, b, c);\n"
+        "endmodule\n"
+    },
+    {
+        // initial statement with indexed access to function returned object
+        "module m; initial a(a,b,c)[i]; endmodule\n",
+        "module m;\n"
+        "  initial a(a, b, c) [i];\n"
+        "endmodule\n"
+    },
+    {
+        // method call with no arguments on an object
+        "module m; initial foo.bar();endmodule\n",
+        "module m;\n"
+        "  initial foo.bar();\n"
+        "endmodule\n"
+    },
+    {
+        // method call with one argument on an object
+        "module m; initial foo.bar(aa);endmodule\n",
+        "module m;\n"
+        "  initial foo.bar(aa);\n"
+        "endmodule\n"
+    },
+    {
+        // method call with two arguments on an object
+        "module m; initial foo.bar(aa,bb);endmodule\n",
+        "module m;\n"
+        "  initial foo.bar(aa, bb);\n"
+        "endmodule\n"
+    },
+    {
+        // method call with three arguments on an object
+        "module m; initial foo.bar(aa,bb,cc);endmodule\n",
+        "module m;\n"
+        "  initial foo.bar(aa, bb, cc);\n"
+        "endmodule\n"
+    },
+
     {
         // This tests for if-statements with null statements
         "function foo;"
--- a/verilog/formatting/formatter_test.cc
+++ b/verilog/formatting/formatter_test.cc
@@ -597,7 +597,10 @@ static const std::initializer_list<FormatterTestCase> kFormatterTestCases = {
     },
     {
         "  parameter  int   foo=$bar(-  z,- y ) ;",
-        "parameter int foo = $bar(-z, -y);\n",
+        "parameter int foo = $bar(\n"
+        "    -z,\n"
+        "    -y\n"
+        ");\n",
     },
     {"  parameter int a=b&~(c<<d);", "parameter int a = b & ~(c << d);\n"},
     {"  parameter int a=~~~~b;", "parameter int a = ~~~~b;\n"},
@@ -3512,12 +3515,8 @@ static const std::initializer_list<FormatterTestCase> kFormatterTestCases = {
      "function void warranty;\n"
      "  foo.bar = fancyfunction(\n"
      "      aaaaaaaa.bbbbbbb,\n"
-     // TODO(fangism): ccccccccc is indented too much because it thinks it is
-     // part of a run-on string of tokens starting with aaaaaaaa.
-     // We need to inform that it should be on the same level as aaaaaaaa as a
-     // sibling item within the same balance group.
-     // Right now, there is no notion of sibling items within a balance group.
-     "          ccccccccc.ddddddddd);\n"
+     "      ccccccccc.ddddddddd\n"
+     "  );\n"
      "endfunction : warranty\n"},
 
     {
--- a/verilog/formatting/tree_unwrapper.cc
+++ b/verilog/formatting/tree_unwrapper.cc
@@ -643,7 +643,7 @@ void TreeUnwrapper::SetIndentationsAndCreatePartitions(
         //     function/system/method/random calls inside trailing assignments,
         //     if headers, ternary operators and so on
         VisitIndentedSection(node, style_.wrap_spaces,
-                             PartitionPolicyEnum::kAlwaysExpand);
+                             PartitionPolicyEnum::kFitOnLineElseExpand);
       } else {
         TraverseChildren(node);
       }
--- a/verilog/formatting/tree_unwrapper.cc
+++ b/verilog/formatting/tree_unwrapper.cc
@@ -626,6 +626,30 @@ void TreeUnwrapper::SetIndentationsAndCreatePartitions(
       break;
     }
 
+    // Indent only when applying kAppendFittingSubPartitions in parents
+    case NodeEnum::kArgumentList:
+    case NodeEnum::kIdentifierList: {
+      if (Context().DirectParentsAre({NodeEnum::kParenGroup,
+                                      NodeEnum::kRandomizeFunctionCall}) ||
+          Context().DirectParentsAre({NodeEnum::kParenGroup,
+                                      NodeEnum::kFunctionCall}) ||
+          Context().DirectParentsAre({NodeEnum::kParenGroup,
+                                      NodeEnum::kRandomizeMethodCallExtension}) ||
+          Context().DirectParentsAre({NodeEnum::kParenGroup,
+                                      NodeEnum::kSystemTFCall}) ||
+          Context().DirectParentsAre({NodeEnum::kParenGroup,
+                                      NodeEnum::kMethodCallExtension})) {
+        // TODO(fangism): Using wrap_spaces because of poor support of
+        //     function/system/method/random calls inside trailing assignments,
+        //     if headers, ternary operators and so on
+        VisitIndentedSection(node, style_.wrap_spaces,
+                             PartitionPolicyEnum::kAlwaysExpand);
+      } else {
+        TraverseChildren(node);
+      }
+      break;
+    }
+
     // The following constructs are flushed-left, not indented:
     case NodeEnum::kPreprocessorIfdefClause:
     case NodeEnum::kPreprocessorIfndefClause:
@@ -711,8 +735,10 @@ void TreeUnwrapper::SetIndentationsAndCreatePartitions(
     case NodeEnum::kGateInstance:
     case NodeEnum::kGenerateIfClause:
     case NodeEnum::kGenerateElseClause:
+    case NodeEnum::kGenerateIfHeader:
     case NodeEnum::kIfClause:
     case NodeEnum::kElseClause:
+    case NodeEnum::kIfHeader:
       // TODO(fangism): k{Assert,Assume,Expect}PropertyClause
     case NodeEnum::kAssertionClause:
     case NodeEnum::kAssumeClause:
@@ -830,6 +856,57 @@ void TreeUnwrapper::SetIndentationsAndCreatePartitions(
                            PartitionPolicyEnum::kFitOnLineElseExpand);
       break;
     }
+
+    case NodeEnum::kReferenceCallBase: {
+      // TODO(fangism): Create own section only for standalone calls
+      if (Context().DirectParentIs(NodeEnum::kStatement)) {
+        const auto& subnode = verible::SymbolCastToNode(*ABSL_DIE_IF_NULL(node.children().back()));
+        if (subnode.MatchesTag(NodeEnum::kRandomizeMethodCallExtension) &&
+            subnode.children().back() != nullptr) {
+          // TODO(fangism): Handle constriants
+          VisitIndentedSection(node, 0,
+                               PartitionPolicyEnum::kAlwaysExpand);
+        } else if (subnode.MatchesTagAnyOf({NodeEnum::kMethodCallExtension,
+                                            NodeEnum::kRandomizeMethodCallExtension,
+                                            NodeEnum::kFunctionCall})) {
+          VisitIndentedSection(node, 0,
+                               PartitionPolicyEnum::kAppendFittingSubPartitions);
+        } else {
+          TraverseChildren(node);
+        }
+      } else {
+        TraverseChildren(node);
+      }
+      break;
+    }
+
+    case NodeEnum::kRandomizeFunctionCall: {
+      // TODO(fangism): Create own section only for standalone calls
+      if (Context().DirectParentIs(NodeEnum::kStatement)) {
+        if (node.children().back() != nullptr) {
+          // TODO(fangism): Handle constriants
+          VisitIndentedSection(node, 0,
+                               PartitionPolicyEnum::kAlwaysExpand);
+        } else {
+          VisitIndentedSection(node, 0,
+                               PartitionPolicyEnum::kAppendFittingSubPartitions);
+        }
+      } else {
+        TraverseChildren(node);
+      }
+      break;
+    }
+
+    case NodeEnum::kSystemTFCall: {
+      // TODO(fangism): Create own section only for standalone calls
+      if (Context().DirectParentIs(NodeEnum::kStatement)) {
+        VisitIndentedSection(node, 0,
+                             PartitionPolicyEnum::kAppendFittingSubPartitions);
+      } else {
+        TraverseChildren(node);
+      }
+      break;
+    };
 
     case NodeEnum::kMacroCall: {
       // Single statements directly inside a controlled construct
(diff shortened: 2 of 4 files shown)


verilog/formatting/tree_unwrapper.cc
```diff
@@ -1288,6 +1307,36 @@ void TreeUnwrapper::ReshapeTokenPartitions(
...
+      auto* object_leaf = recent_partition->LeftmostDescendant();
+      if (object_leaf) {
+        const auto ftokens = object_leaf->Value().TokensRange();
+        if (ftokens.size() == 1) { // Only one token, check if we should merge it
+          auto& method_leaf = *ABSL_DIE_IF_NULL(object_leaf->NextLeaf());
+          if (PartitionStartsWithDot(method_leaf)) {
+            verible::MergeLeafIntoNextLeaf(object_leaf);
+          }
+        }
+      }
```
fangism: This section might be a better fit for `kMacroCallExtension` and `kHierarchyExtension`, rather than `kStatement`.  Try to keep the transformation as local as possible.

Moving the semicolon at the `kStatement` level should be ok though.

verilog/formatting/tree_unwrapper.cc
```diff
@@ -785,6 +785,20 @@ void TreeUnwrapper::SetIndentationsAndCreatePartitions(
...
 
+    case NodeEnum::kFunctionCall:
+    case NodeEnum::kMethodCallExtension: {
+      VisitIndentedSection(node, style_.indentation_spaces,
+                           PartitionPolicyEnum::kAppendFittingSubPartitions);
+                           //PartitionPolicyEnum::kFitOnLineElseExpand);
+      break;
+    }
+
+    case NodeEnum::kArgumentList: {
```
fangism: suggest grouping this case with others before line 700

verilog/formatting/tree_unwrapper.cc
```diff
@@ -1047,6 +1088,36 @@ static void IndentBetweenUVMBeginEndMacros(TokenPartitionTree* partition_ptr,
...
+  const SyntaxTreeNode& method_call_node =
+      SymbolCastToNode(*ABSL_DIE_IF_NULL(node.children().back()));
+  if (!method_call_node.MatchesTagAnyOf({NodeEnum::kMethodCallExtension})) {
+    return ;
+  }
+
+  auto& partition = *ABSL_DIE_IF_NULL(partition_ptr);
+  CHECK_GE(partition.Children().size(), 2);
+
+  auto& object_name = *ABSL_DIE_IF_NULL(partition.LeftmostDescendant());
```
fangism, line 1112: The invoking "object" could be more complex, and not necessarily a leaf partition.  (perhaps call this `object_tail`?)
`foo` could be a `foo[...]` or `foo(...)` or `(foo)`.

I'm concerned that `partition.LeftmostDescendant()` might not refer to the partition node you intend to move, in the case of more complicated invoking objects.  What I did in `kDataDeclaration` for module instances is flatten once before manipulating.

As far as partitioning shape in the more complex cases, I'm thinking (`bar(...)`, `.foo(...)`), and (`bar[...]`, `.foo(...)`).

verilog/formatting/tree_unwrapper_test.cc
```diff
@@ -995,16 +995,19 @@ const TreeUnwrapperTestData kUnwrapModuleTestCases[] = {
...
-                          "Num_blocks", "{",      "$bits", "(", "dbg", "::",
-                          "bus_t",      ")",      "'",     "(", "0",   ")",
-                          "}",          "}",      ")",     ","}),
-                    L(3, {".",        "bus_mid", "(",      "{", "brn", "::",
-                          "Num_bits", "{",       "$clog2", "(", "dbg", "::",
-                          "bus_t",    ")",       "'",      "(", "1",   ")",
-                          "}",        "}",       ")",      ","}),
+                PortActualList(3,  //
+                    N(3,  //
+                      L(3, {".", "bus_in", "(", "{", "brn", "::", "Num_blocks", "{"}),
```
fangism, line 1000: Shouldn't a new partition level start after `.bus_in(`, with `{...` opening the first call argument?

If this (or any) test is too complicated, feel free to split it into smaller ones that focus on a one or few concepts at a time.

verilog/formatting/tree_unwrapper.cc
```diff
@@ -1194,6 +1271,22 @@ void TreeUnwrapper::ReshapeTokenPartitions(
       ReshapeElseClause(node, &partition);
       break;
     }
+    case NodeEnum::kIfHeader: {
+      // Append next leaf to if header ([if (]) and treat as
+      // function/method/macro call (kAppendFittingSubPartitions)
+      if (partition.Children().size() > 1) {
+        auto& if_header = *ABSL_DIE_IF_NULL(partition.LeftmostDescendant());
+        verible::MergeLeafIntoNextLeaf(&if_header);
```
fangism, line 1279: My gut feeling here is that this Merge may be too aggressive for the general `if (...` cases.  The predicate expression may not always be a simple function/task call, it could be `if (x < y)` or `if (x(...) < y(...))`.  Pushing the `if (` into the next leaf would create an oddly shaped partition.   
Could we perhaps add a few conditions to limit this transformation to when the predicate starts "like a function/method call"?

ldalek-antmicro:
Okay. I've simplified partitioning code and addred more tests.

verilog/formatting/tree_unwrapper_test.cc
```diff
@@ -2116,6 +2122,233 @@ const TreeUnwrapperTestData kUnwrapModuleTestCases[] = {
...
+                          L(0, {"module", "m", ";"}),
+                          N(1,
+                            L(1, {"initial", "begin"}),
+                            N(2,
+                              N(2,
+                                N(2,
+                                  L(2, {"if", "("}),
+                                  N(2,
+                                    L(2, {"aa", "("}),
+                                    L(4, {"bb", ")"})),
```
fangism, line 2161: Shouldn't `aa` be wrapped as `N(2+2`? and `bb` be wrapped in as `L(2+2+2`?  (same question in many similiar places)

When things 'fit' we won't see the difference of course, but this will be noticeable in other expanded cases.

It is permissible to adjust indentation of subtrees *after* the decision to expand or not has been made.  So just because we see level `K` in these tree representations doesn't mean it is guaranteed to be indented `K` in the final result.  What is important here is the *relative* indentation differences between parent child partitions (as if expansion were forced across all levels).

fangism:
Checking on progress on this PR, is there anything else I can help with?

hzeller:
What is the state here, soon ready to merge ?
(I am asking as I'd like to slightly modify the interface of TokenInfo which also affects files open here; don't want to create conflicts)

ldalek-antmicro:
Do not hesitate to apply your own modifications. I've already conflicts with master branch. Few more doesn't make me difference.

verilog/formatting/tree_unwrapper.cc
```diff
@@ -1294,6 +1336,19 @@ void TreeUnwrapper::ReshapeTokenPartitions(
     case NodeEnum::kStatement: {
       // This handles cases like macro-calls followed by a semicolon.
       AttachTrailingSemicolonToPreviousPartition(&partition);
+
+      const auto* subtree = GetSubtreeAsSymbol(node, NodeEnum::kStatement, 0);
```
fangism, line 1340: Function calls in general can appear in expressions and statements.  Does it make sense to handle this partition reshaping in a case outside of `kStatement`?

verilog/formatting/tree_unwrapper.cc
```diff
@@ -1294,6 +1336,19 @@ void TreeUnwrapper::ReshapeTokenPartitions(
...
+      if (subtree != nullptr) {
+        const auto& subnode = SymbolCastToNode(*subtree);
+        if (NodeEnum(subnode.Tag().tag) == NodeEnum::kReferenceCallBase) {
+          if (partition.Children().size() >= 2) {
+            const auto& uwline = partition.Children()[1].Value();
+            if (uwline.PartitionPolicy() == PartitionPolicyEnum::kAppendFittingSubPartitions) {
+              verible::MergeLeafIntoNextLeaf(&partition.Children()[0]);
+            }
+          }
```
fangism, line 1349: In general, let's define these CST node accessors as helper functions in the CST (which can then be unit-tested).  It is hard to infer the intent in this form.

verilog/formatting/tree_unwrapper.cc
```diff
@@ -1321,6 +1376,49 @@ void TreeUnwrapper::ReshapeTokenPartitions(
...
+    case NodeEnum::kRandomizeFunctionCall: {
+      auto& last = *ABSL_DIE_IF_NULL(partition.RightmostDescendant());
+      if (PartitionIsCloseParen(last) || PartitionStartsWithSemicolon(last)) {
+        verible::MergeLeafIntoPreviousLeaf(&last);
+      }
+
+      auto& uwline = partition.Value();
+      const auto partition_policy = uwline.PartitionPolicy();
+      if (partition_policy == PartitionPolicyEnum::kAppendFittingSubPartitions) {
+        if (partition.Children().size() > 2) {
```
fangism, line 1389: Can you explain this with a comment?

verilog/formatting/tree_unwrapper.cc
```diff
@@ -1321,6 +1376,49 @@ void TreeUnwrapper::ReshapeTokenPartitions(
...
+      auto& last = *ABSL_DIE_IF_NULL(partition.RightmostDescendant());
+      if (PartitionIsCloseParen(last) || PartitionStartsWithSemicolon(last)) {
+        verible::MergeLeafIntoPreviousLeaf(&last);
+      }
+      break;
+    }
+
+    case NodeEnum::kIfHeader: {
+      if (partition.Children().size() > 1) {
+        const auto& uwline = partition.Children()[1].Value();
```
fangism, line 1408: Use one of the functions in 
https://github.com/google/verible/blob/master/verilog/CST/statement.h
for accessing substructure of `kIfHeader` nodes (instead of `[1]`).

verilog/formatting/tree_unwrapper.cc
```diff
@@ -1361,6 +1459,43 @@ void TreeUnwrapper::ReshapeTokenPartitions(
         break;
       }
 
+    case NodeEnum::kParamDeclaration: {
+      {
+        const auto* trailing_assign = GetSubtreeAsSymbol(
+            node, NodeEnum::kParamDeclaration, 2);
```
fangism, line 1465: define CST function for accessing substructure

verilog/formatting/tree_unwrapper.cc
```diff
@@ -1361,6 +1459,43 @@ void TreeUnwrapper::ReshapeTokenPartitions(
...
+            const auto& expression = GetSubtreeAsNode(
+                node, NodeEnum::kTrailingAssign, 1);
+            if (expression.children().size() == 1) {
+              const auto& expr = GetSubtreeAsNode(
+                  expression, NodeEnum::kExpression, 0);
+              switch (NodeEnum(expr.Tag().tag)) {
+                case NodeEnum::kRandomizeMethodCallExtension:
+                case NodeEnum::kRandomizeFunctionCall:
+                case NodeEnum::kFunctionCall:
+                case NodeEnum::kSystemTFCall: {
```
fangism, line 1478: Putting this logic here doesn't seem to cover general cases of expressions adequately.  The way I would think of this is:

* How should a standalone call/expression be shaped?  (without regard to parent context)
* How should the result of that shaping be integrated at the next level up, and what _incremental_ changes are needed to do the right thing?  (e.g. merging partitions, adjusting wrapping/indentation of sub-partitions).  Example contexts:
  * In statements
  * In initialization/defaults of parameters
  * On right-sides of assignments
  * Deeper inside other expressions

This will eliminate duplication and minimize deep special casing.

verilog/formatting/tree_unwrapper.cc
```diff
@@ -1425,13 +1560,26 @@ void TreeUnwrapper::ReshapeTokenPartitions(
     case NodeEnum::kContinuousAssignmentStatement: {  // e.g. assign a=0, b=2;
       // TODO(fangism): group into above similar assignment statement cases?
       //   Cannot easily move now due to multiple-assignments.
+      bool anyAppendFittingSubpartitions = std::find_if(
```
fangism, line 1563: `const bool`

verilog/formatting/tree_unwrapper_test.cc
fangism: I would encourage splitting into smaller focused test cases.  (You can still keep the existing ones.)

verilog/formatting/tree_unwrapper.cc
ldalek-antmicro: I access here UnwrapperLine child, not a CST node. Does CST accessors still apply here?

verilog/formatting/tree_unwrapper.cc
```diff
@@ -830,6 +835,32 @@ void TreeUnwrapper::SetIndentationsAndCreatePartitions(
...
+    case NodeEnum::kRandomizeFunctionCall:
+    case NodeEnum::kSystemTFCall:
+    case NodeEnum::kFunctionCall: {
+      VisitIndentedSection(node, 0,
+                           PartitionPolicyEnum::kAppendFittingSubPartitions);
+      break;
+    };
+
+    case NodeEnum::kIdentifierList: {
+      if (Context().ParentAtIs(1, NodeEnum::kRandomizeFunctionCall)) {
```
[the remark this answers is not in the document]
ldalek-antmicro: Replaced with DirectParentsAre but got this awful if-statement:
https://github.com/google/verible/pull/246/files#diff-7872d97b05fbc6f2a64cf14b8ae74d75R631
Is that ok?

verilog/formatting/tree_unwrapper.cc
ldalek-antmicro: It was fix for randomize function with constraints. Moved it here https://github.com/google/verible/pull/246/files#diff-7872d97b05fbc6f2a64cf14b8ae74d75R863.

Btw. how the ```kWithConstraints``` should be handled? Now it looks like this:
```
  { ([<auto>], policy: fit-else-expand) @{19}, (origin: "function f;...endfunction")
    { ([function f ;], policy: fit-else-expand, (origin: "function f;")) }
    { (>>[<auto>], policy: always-expand) @{19,1}, (origin: "s = std::ra...comment2
};")
      { (>>[s = std::randomize (], policy: fit-else-expand) }
      { (>>[<auto>], policy: always-expand) @{19,1,1}, (origin: "aaa,bbb,ccc,ddd,eee")
        { (>>[aaa ,], policy: fit-else-expand, (origin: "aaa")) }
        { (>>[bbb ,], policy: fit-else-expand, (origin: "bbb")) }
        { (>>[ccc ,], policy: fit-else-expand, (origin: "ccc")) }
        { (>>[ddd ,], policy: fit-else-expand, (origin: "ddd")) }
        { (>>[eee], policy: fit-else-expand, (origin: "eee")) }
      }
      { (>>[) with {], policy: uninitialized) }
      { (>>>>[<auto>], policy: always-expand) @{19,1,3}, (origin: "a == e;")
        { (>>>>[// comment1], policy: fit-else-expand) }
        { (>>>>[a == e ;], policy: fit-else-expand, (origin: "a == e;")) }
        { (>>>>[// comment2], policy: fit-else-expand) }
      }
      { (>>[} ;], policy: fit-else-expand) }
    }
    { ([endfunction], policy: always-expand, (origin: "endfunction")) }
  }
```
Should it look like ```if () begin end``` (with the ```with``` keyword appended to last argument?)

verilog/formatting/tree_unwrapper.cc
ldalek-antmicro: Statements and right-side assignments reuse current code.
Parameters (and other parents) are handled here https://github.com/google/verible/pull/246/files#diff-7872d97b05fbc6f2a64cf14b8ae74d75R863

I don't know how nested calls and if-statements should be handled.
For example if-statement can have binary expression, e.g. ```a + b(c, d, e)``` which currently is partitioned like:
```
      { (>>[<auto>], policy: append-fitting-sub-partitions) @{25,1,0}, (origin: "if (a + b(c,d,e));")
        { (>>[if ( a + b (], policy: fit-else-expand) }
        { (>>[<auto>], policy: always-expand) @{25,1,0,1}, (origin: "c,d,e")
          { (>>[c ,], policy: fit-else-expand, (origin: "c")) }
          { (>>[d ,], policy: fit-else-expand, (origin: "d")) }
          { (>>[e ) ) ;], policy: fit-else-expand, (origin: "e")) }
        }
      }
```
but such statement ```b(c,d,e) + a``` is partitioned like:
```
      { (>>[<auto>], policy: fit-else-expand) @{25,1,1}, (origin: "if (b(c,d,e)+c);")
        { (>>[<auto>], policy: append-fitting-sub-partitions) @{25,1,1,0}, (origin: "b(c,d,e)")
          { (>>[if ( b (], policy: fit-else-expand) }
          { (>>[<auto>], policy: always-expand) @{25,1,1,0,1}, (origin: "c,d,e")
            { (>>[c ,], policy: fit-else-expand, (origin: "c")) }
            { (>>[d ,], policy: fit-else-expand, (origin: "d")) }
            { (>>[e )], policy: fit-else-expand, (origin: "e")) }
          }
        }
        { (>>[+ c ) ;], policy: uninitialized) }
      }
```
I guess the ```+ c );``` should appended to last partition: ```e )```. But what about such expression:
```
      { (>>[<auto>], policy: append-fitting-sub-partitions) @{25,1,0}, (origin: "a(b,c)")
        { (>>[if ( a (], policy: fit-else-expand) }
        { (>>[<auto>], policy: always-expand) @{25,1,0,1}, (origin: "b,c")
          { (>>[b ,], policy: fit-else-expand, (origin: "b")) }
          { (>>[c )], policy: fit-else-expand, (origin: "c")) }
        }
      }
      { (>>[+], policy: uninitialized) }
      { (>>[<auto>], policy: append-fitting-sub-partitions) @{25,1,2}, (origin: "d(e,f)")
        { (>>[d (], policy: fit-else-expand) }
        { (>>[<auto>], policy: always-expand) @{25,1,2,1}, (origin: "e,f")
          { (>>[e ,], policy: fit-else-expand, (origin: "e")) }
          { (>>[f ) ) ;], policy: fit-else-expand, (origin: "f")) }
        }
      }
```
Should the ```+``` partition be appended to previous partition or prepended to next partition?
Or should the whole expression be flatten? I mean treat ```+``` as comma and ```if ()``` as function call?

verilog/formatting/tree_unwrapper.cc
ldalek-antmicro: Fixed. Btw. Is that partition policy propagation okay?

verilog/formatting/tree_unwrapper.cc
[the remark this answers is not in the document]
ldalek-antmicro: Replaced with DirectParentsAre but got this awful if-statement:
https://github.com/google/verible/pull/246/files#diff-7872d97b05fbc6f2a64cf14b8ae74d75R631
Is that ok?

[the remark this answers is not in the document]
fangism: Good enough for now.  It can always be hidden behind a function.

verilog/formatting/tree_unwrapper.cc
fangism: This looks good with the `) with {` partition in the middle.  (No need to attach it to the `eee` that precedes it.)

verilog/formatting/tree_unwrapper.cc
fangism: Sorry, these are token partitions already, and there are no equivalent functions because the structure is fluid.
Worth adding a comment about what `[1]` is intended to reference.

verilog/formatting/tree_unwrapper.cc
fangism: Yes, `b(c,d,e) + a` should probably contain `+ c)` up one level like this:

```
      { (>>[<auto>], policy: fit-else-expand) @{25,1,1}, (origin: "if (b(c,d,e)+c);")
        { (>>[<auto>], policy: append-fitting-sub-partitions) @{25,1,1,0}, (origin: "b(c,d,e)")
          { (>>[if ( b (], policy: fit-else-expand) }
          { (>>[<auto>], policy: always-expand) @{25,1,1,0,1}, (origin: "c,d,e")
            { (>>[c ,], policy: fit-else-expand, (origin: "c")) }
            { (>>[d ,], policy: fit-else-expand, (origin: "d")) }
            { (>>[e )], policy: fit-else-expand, (origin: "e")) }
          }
          { (>>[+ c ) ;], policy: uninitialized) }
        }
      }
```

For `a(b,c)+d(e,f)`, I'd lean towards attaching `+` to its left.  If you feel this change set is getting too large, you may handle it separately as a follow-up, I'll let you make that call.  It's ok to avoid handling binary expressions for now, and just focus on the partitioning around ','.

verilog/formatting/tree_unwrapper.cc
```diff
@@ -1454,13 +1575,26 @@ void TreeUnwrapper::ReshapeTokenPartitions(
     case NodeEnum::kContinuousAssignmentStatement: {  // e.g. assign a=0, b=2;
       // TODO(fangism): group into above similar assignment statement cases?
       //   Cannot easily move now due to multiple-assignments.
+      const bool anyAppendFittingSubpartitions = std::find_if(
```
fangism, line 1578: Just realized there's `std::any_of` for this, instead of `find_if`.
https://en.cppreference.com/w/cpp/algorithm/all_any_none_of

As for the policy itself `kAppendFittingSubpartitions` might be ok, but the bigger question is: does it need to be handled in this PR, which is focused on method and function calls?

verilog/formatting/tree_unwrapper.cc
```diff
@@ -1454,13 +1575,26 @@ void TreeUnwrapper::ReshapeTokenPartitions(
...
       partition.FlattenOnlyChildrenWithChildren();
       VLOG(4) << "after flatten:\n" << partition;
       AttachTrailingSemicolonToPreviousPartition(&partition);
       // Merge the 'assign' keyword with the (first) x=y assignment.
       // TODO(fangism): reshape for multiple assignments.
       verible::MergeConsecutiveSiblings(&partition, 0);
       VLOG(4) << "after merging 'assign':\n" << partition;
+
+      // Propagate kAppendFittingSubPartitions partition policy
+      if (anyAppendFittingSubpartitions && partition.Children().size() == 2) {
```
fangism, line 1594: Why `== 2`?  (Not clear to me.)

verilog/formatting/tree_unwrapper.cc
fangism: It looks like "if any direct children nodes are append-fitting, then make this node also append-fitting (for these constructs)".

I _think_ that's ok.  It feels safe because these statements are isolated from one another.

verilog/formatting/tree_unwrapper_test.cc
```diff
@@ -1203,14 +1205,14 @@ const TreeUnwrapperTestData kUnwrapModuleTestCases[] = {
...
-                          "bus_t",    ")",       "'",      "(", "1",   ")",
-                          "}",        "}",       ")",      ","}),
+                    N(3,
+                      L(3, {".", "bus_in", "(", "{", "brn", "::", "Num_blocks", "{", "$bits", "("}),
+                      L(5, {"dbg", "::", "bus_t"}),
+                      L(3, {")", "'", "(", "0", ")", "}", "}", ")", ","})),
+                    N(3,
+                      L(3, {".", "bus_mid", "(", "{", "brn", "::", "Num_bits", "{", "$clog2", "("}),
+                      L(5, {"dbg", "::", "bus_t"}),
+                      L(3, {")", "'", "(", "1", ")", "}", "}", ")", ","})),
```
fangism, line 1215: Looks ok.

verilog/formatting/tree_unwrapper_test.cc
```diff
@@ -2895,13 +2899,20 @@ const TreeUnwrapperTestData kClassTestCases[] = {
...
-                          L(1, {"localparam", "M", "=", "f", "(", "glb", "::",
-                                "arr", "[", "N", "]", ")", "+", "1", ";"})),
+                          N(1,
+                            L(1, {"localparam", "M", "=", "f", "("}),
+                            L(3, {"glb", "::", "arr", "[", "N", "]"}),
+                            L(1, {")", "+", "1", ";"})),
+                          N(1,
+                            L(1, {"localparam", "M", "=", "$f", "("}),
+                            L(3, {"glb", "::", "arr", "[", "N", "]"}),
+                            L(1, {")", "+", "1", ";"}))),
```
fangism, line 2915: Looks ok.

verilog/formatting/tree_unwrapper_test.cc
```diff
@@ -5620,8 +5643,19 @@ const TreeUnwrapperTestData kUnwrapFunctionTestCases[] = {
...
+                  N(3,
+                    L(3, {"x", ","}),
+                    L(3, {"1"})),
+                  L(1, {")", ";"})),
+                N(1,
+                  L(1, {"z", "=", "twister", "("}),
+                  N(3,
+                    L(3, {"x", ","}),
+                    L(3, {"2"})),
+                  L(1, {")", ";"}))),
```
fangism, line 5658: Looks good.

verilog/formatting/tree_unwrapper_test.cc
```diff
@@ -6347,7 +6393,9 @@ const TreeUnwrapperTestData kUnwrapFunctionTestCases[] = {
         FunctionDeclaration(
             0, FunctionHeader(0, {"function", "foo", ";"}),
             FlowControl(1, L(1, {"while", "(", "e", ")"}),
-                        L(2, {"coyote", "(", "sooper_genius", ")", ";"})),
+                        N(2,
+                          L(2, {"coyote", "("}),
+                          L(4, {"sooper_genius", ")", ";"}))),
```
fangism, line 6398: Looks good.

verilog/formatting/formatter_test.cc
```diff
@@ -597,7 +597,10 @@ static const std::initializer_list<FormatterTestCase> kFormatterTestCases = {
     },
     {
         "  parameter  int   foo=$bar(-  z,- y ) ;",
-        "parameter int foo = $bar(-z, -y);\n",
+        "parameter int foo = $bar(\n"
```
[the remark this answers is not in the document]
ldalek-antmicro: Yes, It was because of `kAlwaysExapand` partition policy applied to function calls arguments section.
Generally I tried to focus on function calls arguments rather. The simplest way to accomplish that was to
skip creating own section for function calls (with an exception for standalone calls) and just partition function arguments.
Which gives for example such partition tree:
```
  { ([<auto>], policy: fit-else-expand) @{49}, (origin: "parameter  ...  z,- y ) ;")
    { ([parameter int foo = $bar (], policy: fit-else-expand) }
    { (>>>>[<auto>], policy: always-expand) @{49,1}, (origin: "-  z,- y")
      { (>>>>[- z ,], policy: fit-else-expand, (origin: "-  z")) }
      { (>>>>[- y], policy: fit-else-expand, (origin: "- y")) }
    }
    { ([) ;], policy: uninitialized) }
  }
```
I thought that it'd be better to always expand such arguments list rather than fit some of them and left dangling `);` at the end, like in formatter tuning test:
```
module m;
  assign wwwwww[77:66] = sss(
      qqqq[33:22], vv[44:1]
  );
endmodule
```
(I left dangling `);` because I thought that it'd be better to address function calls in trailing assignments in separate PR)

ldalek-antmicro:
Yes I can. I will add them.

fangism:
> Yes I can. I will add them.

Can you add them to formatter_test.cc (end-to-end)?  It's good to see them in tree_unwrapper_test too.

Issue #260, opened 2020-04-08
Issue building on MacOS


Alex-Mann:
Hi there,

I understand there that the C++ compiler that comes with Xcode is not supported for building verible... but are there any more details for how sub out the compiler and build on MacOS? Currently I'm getting the following error:
```
❯ bazel build -c opt //...
INFO: Analyzed 1195 targets (5 packages loaded, 1752 targets configured).
INFO: Found 1195 targets...
ERROR: /Users/alexandermann/Work/verible/common/text/BUILD:183:1: C++ compilation of rule '//common/text:tree_utils' failed (Exit 1) wrapped_clang failed: error executing command external/local_config_cc/wrapped_clang '-D_FORTIFY_SOURCE=1' -fstack-protector -fcolor-diagnostics -Wall -Wthread-safety -Wself-assign -fno-omit-frame-pointer -g0 -O2 '-D_FORTIFY_SOURCE=1' -DNDEBUG ... (remaining 28 argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox
In file included from common/text/tree_utils.cc:15:
./common/text/tree_utils.h:115:55: error: unknown type name 'nullptr_t'; did you mean 'std::nullptr_t'?
const SyntaxTreeNode* CheckOptionalSymbolAsNode(const nullptr_t& symbol, E) {
                                                      ^~~~~~~~~
                                                      std::nullptr_t
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/__nullptr:56:31: note: 'std::nullptr_t' declared here
    typedef decltype(nullptr) nullptr_t;
                              ^
In file included from common/text/tree_utils.cc:15:
./common/text/tree_utils.h:130:55: error: unknown type name 'nullptr_t'; did you mean 'std::nullptr_t'?
const SyntaxTreeLeaf* CheckOptionalSymbolAsLeaf(const nullptr_t& symbol, E) {
                                                      ^~~~~~~~~
                                                      std::nullptr_t
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/__nullptr:56:31: note: 'std::nullptr_t' declared here
    typedef decltype(nullptr) nullptr_t;
                              ^
2 errors generated.
INFO: Elapsed time: 10.413s, Critical Path: 2.66s
INFO: 10 processes: 10 darwin-sandbox.
[...]

PR #263, opened 2020-04-09
In create-version-header: use less system-specific features.

hzeller:
Using 'date' to format some date only really works with
GNU date, so just use the raw time_t timestamp and format
that in code.

While at it, remove dependency on 'grep'.

This will remove one of the problems compiling on
Mac, found in #260.

### DIFF
--- a/common/util/create_version_header.sh
+++ b/common/util/create_version_header.sh
@@ -16,13 +16,9 @@
 # The file where bazel stores name/value pairs
 BAZEL_VOLATILE=bazel-out/volatile-status.txt
 
-GIT_DATE="$(grep -s GIT_DATE $BAZEL_VOLATILE | cut -f2- -d' ')"
-GIT_DESC="$(grep -s GIT_DESCRIBE $BAZEL_VOLATILE | cut -f2- -d' ')"
-
-# Timestamp is always provided by bazel.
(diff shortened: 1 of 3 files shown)


common/util/create_version_header.sh
```diff
@@ -16,13 +16,9 @@
...
-TS_INT="$(grep -s BUILD_TIMESTAMP $BAZEL_VOLATILE | cut -f2 -d' ')"
-test -z "$TS_INT" || TS_STRING="$(date +"%Y-%m-%dT%H:%M:%SZ" -u -d @$TS_INT)"
-
-test -z "$GIT_DATE"  || echo "#define VERIBLE_GIT_DATE $GIT_DATE"
-test -z "$GIT_DESC"  || echo "#define VERIBLE_GIT_DESC $GIT_DESC"
-test -z "$TS_STRING" || echo "#define VERIBLE_BUILD_TIMESTAMP \"$TS_STRING\""
+(while read NAME VALUE ; do
+   case $NAME in
+     # Only select the ones we're interested in.
+     GIT_* | BUILD_TIMESTAMP) echo "#define VERIBLE_${NAME} ${VALUE}"
```
fangism, line 22: customary to end case with `;;`

PR #273, opened 2020-04-17
[git] Source diff awk file with absolute path

eunchan:
git-verilog_format.sh uses the script directory to source diff awk file.
If the script is called with relative path it cannot source the file
correctly as the tool change the directory to the repository top.

This commit is to change the path to absolute path so that wherever the
script is called inside repo top, it can source the awk file.


eunchan:
Revised. Thanks for the pointer.

PR #283, opened 2020-04-28
System calls such as  and  fail if encrypted

jrrk2:
This patch recognizes a SystemTFIdentifier and turns off encryption for that token. Otherwise the obfuscated Verilog cannot be compiled if it contains such calls.

I await further guidance on testcase integration.


verilog/transform/obfuscate.cc
```diff
@@ -47,9 +47,12 @@ static void ObfuscateVerilogCodeInternal(absl::string_view content,
       case verilog_tokentype::PP_Identifier:
         *output << (*subst)(token.text);
         break;
+        // System calls fail if encrypted
```
fangism, line 50: suggest: "Preserve all $ID calls, including system task/function calls, and VPI calls"

jrrk2:
@googlebot I signed it!

fangism:
You may mark in your commit message:

```
fixes #282 
```

jrrk2:
@googlebot I signed it!

jrrk2:
@googlebot I fixed it.

Issue #276, opened 2020-04-20
[unpacked-dimensions-range-ordering] rule should exclude module array instances


fangism:
The following test case:

```systemverilog
module test (input logic [1:0] blah);
endmodule

module tb;
test test_i[3:0] (
  .blah (2'b0)
);
endmodule
```

is caught by `verilog_lint` as "[unpacked-dimensions-range-ordering]".
We would like to exempt module instance arrays from this check.

fangism:
b/154451633

fangism:
Running `verilog_syntax --printtree` on that input reveals that we can use the syntax tree context to refine the finding.  The tree dump includes:

```
...
          Node (tag: kGateInstanceRegisterVariableList) {
            Node (tag: kGateInstance) {
              (#SymbolIdentifier @65-71: "test_i")
              Node (tag: kUnpackedDimensions) {
                Node (tag: kDeclarationDimensions) {
                  Node (tag: kDimensionRange) {
                    (#'[' @71-72: "[")
                    Node (tag: kExpression) {
                      Node (tag: kNumber) {
                        (#TK_DecNumber @72-73: "3")
                      }
                    }
                    (#':' @73-74: ":")
                    Node (tag: kExpression) {
                      Node (tag: kNumber) {
                        (#TK_DecNumber @74-75: "0")
                      }
                    }
                    (#']' @75-76: "]")
                  }
                }
              }
...
```

So the check for exemption should use `kGateInstance`.

Ciprian167:
If this is still free, I'd like to take a shot at it. I cannot self assign it yet.

fangism:
Thank you for offering, let me know if you have any questions.

Ciprian167:
I think this could be waived in UnpackedDimensionsRule::HandleSymbol by changing the return condition to something like:
``` C++
if (!ContextIsInsideUnpackedDimensions(context) || 
     context.IsInside(NodeEnum::kGateInstance)) return;
```
What do think?

fangism:
> I think this could be waived in UnpackedDimensionsRule::HandleSymbol by changing the return condition to something like:
> 
> ```c++
> if (!ContextIsInsideUnpackedDimensions(context) || 
>      context.IsInside(NodeEnum::kGateInstance)) return;
> ```
> 
> What do think?

I think that might do.  Suggest building up some test cases to verify.

PR #288, opened 2020-05-04
Fix for #276: unpacked dimensions rules doesn't apply to array instances

Ciprian167:


verilog/analysis/checkers/unpacked_dimensions_rule_test.cc
```diff
@@ -109,6 +109,13 @@ TEST(UnpackedDimensionsRuleTests, CheckRanges) {
       // inner range is packed dimensions, should not be detected as unpacked
       {"int some_array [bit [7:0]];"},
       {"int some_array [bit [0:7]];"},
+
+      // array of instances
+      {"module test (input logic [1:0] blah);",
+       "endmodule",
+       "module tb;",
+       "  test test_i[3:0] (.blah (2'b0));",
```
fangism, line 117: Let's also add a case for `test_i[0:3]`

verilog/analysis/checkers/unpacked_dimensions_rule_test.cc
```diff
@@ -109,6 +109,13 @@ TEST(UnpackedDimensionsRuleTests, CheckRanges) {
       // inner range is packed dimensions, should not be detected as unpacked
       {"int some_array [bit [7:0]];"},
       {"int some_array [bit [0:7]];"},
+
+      // array of instances
+      {"module test (input logic [1:0] blah);",
+       "endmodule",
```
fangism, line 115: This looks like a packed-dimensions usage in the ports, which is not checked by this rule.  It's ok to keep this as a negative test case, but I'd re-arrange the comment to reflect what covers `// array of instances`, below.

Ciprian167: You're right. I'll remove the packed-dimensions in ports just to keep thing clear.  Thanks for feedback.

verilog/analysis/checkers/unpacked_dimensions_rule_test.cc
```diff
@@ -109,6 +109,19 @@ TEST(UnpackedDimensionsRuleTests, CheckRanges) {
...
+       "module tb;",
+       "  test test_i[3:0] (.blah (1'b0));",
+       "endmodule"},
+       
+       {"module test (input logic blah);",
+       "endmodule",
+       "module tb;",
+       "  test test_i[0:3] (.blah (1'b0));",
+       "endmodule"}
```
fangism, line 124: Just noticed, you can omit the commas between joined string literals.  But you might want to place a space or \n between "endmodule" and "module" or they'll be fused together.

Just make sure `bazel test //...` works.  :)

Ciprian167: Thanks for feedback. I've seen tests written in both ways, is there a preferred one?

fangism: I would only separate with a `,` when marking an expected finding, like `{"good text", {kFoo, "bad-text"}, "more good text"}`

Ciprian167: OK. I will take this into account next time. Thanks.

PR #290, opened 2020-05-06
verilog_format: accept multiple files to all process.

hzeller:
* Don't CHECK()-fail when no argument is given, but print usage instead.
  * Allow for arbitrary number of filename arguments.
  * Don't --inplace overwrite a file if the formatting did not
    generate any change (this is crucial to use in build-systems where
    typically the timestamp indicates if something needs to be
    re-built).

### DIFF
--- a/verilog/tools/formatter/verilog_format.cc
+++ b/verilog/tools/formatter/verilog_format.cc
@@ -121,6 +121,11 @@ ABSL_FLAG(bool, format_module_instantiations, false,
           "If true, format module instantiations (data declarations), "
           "else leave them unformatted.  This is a short-term workaround.");
 
+std::ostream &FileMsg(absl::string_view filename) {
+  std::cerr << filename << ": ";
+  return std::cerr;
+}
+
 bool formatOneFile(absl::string_view filename,
                    const verilog::formatter::LineNumberSet &lines_to_format) {
   const bool inplace = absl::GetFlag(FLAGS_inplace);
@@ -128,8 +133,9 @@ bool formatOneFile(absl::string_view filename,
   const auto& stdin_name = absl::GetFlag(FLAGS_stdin_name);
 
   if (inplace && is_stdin) {
-    std::cerr << "--inplace is incompatible with stdin.  Ignoring --inplace "
-        "and writing to stdout." << std::endl;
+    FileMsg(filename)
+        << "--inplace is incompatible with stdin.  Ignoring --inplace "
+        << "and writing to stdout." << std::endl;
   }
 
   const auto diagnostic_filename = is_stdin ? stdin_name : filename;
@@ -137,7 +143,7 @@ bool formatOneFile(absl::string_view filename,
   // Read contents into memory first.
   std::string content;
   if (!verible::file::GetContents(filename, &content)) {
-    std::cerr << filename << ": Couldn't read" << std::endl;
+    FileMsg(filename) << "Couldn't read" << std::endl;
     return false;
   }
 
@@ -181,21 +187,21 @@ bool formatOneFile(absl::string_view filename,
       // Fall back to printing original content regardless of error condition.
       std::cout << content;
     }
-    // Print the error message last so it shows up in user's console.
-    std::cerr << filename << ": " << format_status.message() << " ";
     switch (format_status.code()) {
       case StatusCode::kCancelled:
       case StatusCode::kInvalidArgument:
+        FileMsg(filename) << format_status.message() << std::endl;
         break;
       case StatusCode::kDataLoss:
-        std::cerr << "\nProblematic formatter output is:\n"
-                  << formatted_output << "<<EOF>>";
+        FileMsg(filename) << format_status.message()
+                          << "; problematic formatter output is\n"
+                          << formatted_output << "<<EOF>>" << std::endl;
         break;
       default:
-        std::cerr << "[other error status]";
+        FileMsg(filename) << format_status.message() << "[other error status]"
+                          << std::endl;
         break;
     }
-    std::cerr << std::endl;
 
     return absl::GetFlag(FLAGS_failsafe_success);
   }
@@ -206,11 +212,11 @@ bool formatOneFile(absl::string_view filename,
     // with tools that look for timestamp changes (such as make).
     if (content != formatted_output) {
       if (!verible::file::SetContents(filename, formatted_output)) {
-        std::cerr << filename <<  ": error writing to file" << std::endl;
+        FileMsg(filename) << "error writing to file" << std::endl;
         return false;
       }
     } else {
-      std::cerr << filename << ": Already formatted, no change." << std::endl;
+      FileMsg(filename) << "Already formatted, no change." << std::endl;
     }
   } else {
     std::cout << formatted_output;
--- a/verilog/tools/formatter/verilog_format.cc
+++ b/verilog/tools/formatter/verilog_format.cc
@@ -30,6 +30,7 @@
 #include <vector>
 
 #include "absl/flags/flag.h"
+#include "absl/flags/usage.h"
 #include "absl/status/status.h"
 #include "absl/strings/str_cat.h"
 #include "absl/strings/str_join.h"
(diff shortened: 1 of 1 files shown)


hzeller:
So you mean the test that I can't write yet because I can't use sh-tests ? :)

fangism:
> So you mean the test that I can't write yet because I can't use sh-tests ? :)

You can have sh_tests, they just have to be liberated from internals.

I'll keep reviewing...

verilog/tools/formatter/verilog_format.cc
```diff
@@ -216,22 +194,62 @@ int main(int argc, char** argv) {
...
 
   // Safe to write out result, having passed above verification.
   if (inplace && !is_stdin) {
-    if (!verible::file::SetContents(filename, formatted_output)) {
-      std::cerr << "Error writing to file: " << filename << std::endl;
-      std::cerr << "Printing to stdout instead." << std::endl;
-      std::cout << formatted_output;
-      return 1;
+    // Don't write if the output is exactly as the input, so that we don't mess
+    // with tools that look for timestamp changes (such as make).
```
fangism, line 205: In this case can we print a message to stderr or stdout saying that file was already properly formatted?
I think in multi-file inplace mode it would be good to have file-by-file confirmation as a mark of progress, in case the formatter fails/terminates/crashes on one of the inputs.  Then we at least know the last successful file for triage.

verilog/tools/formatter/verilog_format.cc
```diff
@@ -216,22 +194,62 @@ int main(int argc, char** argv) {
...
+  if (!verible::ParseInclusiveRanges(
+          &lines_to_format, LineRanges::values.begin(),
+          LineRanges::values.end(), &std::cerr, '-')) {
+    std::cerr << "Error parsing --lines." << std::endl;
+    std::cerr << "Got: --lines=" << AbslUnparseFlag(LineRanges()) << std::endl;
+    return 1;
+  }
+
+  if (!lines_to_format.empty() && file_args.size() > 2) {
+    std::cerr << "Note, applying --lines rules to all files" << std::endl;
```
fangism, line 244: This feels weird/surprising to do, can we error out instead?  Maybe see what clang_format does in this case.

verilog/tools/formatter/verilog_format.cc
```diff
@@ -216,22 +194,62 @@ int main(int argc, char** argv) {
...
     std::cout << formatted_output;
   }
 
-  return 0;
+  return true;
+}
+
+int main(int argc, char** argv) {
+  const auto usage = absl::StrCat("usage: ", argv[0],
+                                  " [options] <file>\n"
```
fangism, line 223: `<files...>`

verilog/tools/formatter/verilog_format.cc
```diff
@@ -216,22 +194,62 @@ int main(int argc, char** argv) {
...
+                                  "To pipe from stdin, use '-' as <file>.");
```
fangism, line 224: Should we now report a user-error if `-` is one of many files?  Few utilities have a meaningful behavior when stdin is one of multiple inputs, `diff` being one of them.

verilog/tools/formatter/verilog_format.cc
hzeller: Done.

verilog/tools/formatter/verilog_format.cc
hzeller: Done.

verilog/tools/formatter/verilog_format.cc
hzeller: done.

verilog/tools/formatter/verilog_format.cc
hzeller: Having stdin as one of many files is a benign situation (this is the only file we then output to stdout, which is exactly as expected).

But having multiple files without `--inplace` doesn't really make sense, so error on that now.

verilog/tools/formatter/verilog_format.cc
```diff
@@ -205,11 +206,11 @@ bool formatOneFile(absl::string_view filename,
...
     if (content != formatted_output) {
       if (!verible::file::SetContents(filename, formatted_output)) {
-        std::cerr << "Error writing to file: " << filename << std::endl;
-        std::cerr << "Printing to stdout instead." << std::endl;
-        std::cout << formatted_output;
+        std::cerr << filename <<  ": error writing to file" << std::endl;
         return false;
       }
+    } else {
+      std::cerr << filename << ": no change." << std::endl;
```
fangism, line 213: suggest: "Already formatted, no change."

verilog/tools/formatter/verilog_format.cc
```diff
@@ -240,8 +241,20 @@ int main(int argc, char** argv) {
     return 1;
   }
 
-  if (!lines_to_format.empty() && file_args.size() > 2) {
-    std::cerr << "Note, applying --lines rules to all files" << std::endl;
+  // Some sanity checks if multiple files are given.
+  if (file_args.size() > 2) {
+    if (!lines_to_format.empty()) {
+      std::cerr << "Providing --lines but gave multiple different files."
```
fangism, line 247: suggest: "--lines=... only works for a single file"

We should also update the help string for this flag similarly.

verilog/tools/formatter/verilog_format.cc
```diff
@@ -240,8 +241,20 @@ int main(int argc, char** argv) {
...
+                << std::endl;
+      return 1;
+    }
+
+    if (!absl::GetFlag(FLAGS_inplace)) {
+      // Dumping all to stdout doesn't really make sense.
+      std::cerr << "Multiple files given. Use --inplace "
+          "for a meaningful handling of these." << std::endl;
```
fangism, line 255: suggest: "--inplace is required when processing multiple files."

verilog/tools/formatter/verilog_format.cc
```diff
@@ -179,24 +182,22 @@ bool formatOneFile(absl::string_view filename,
       std::cout << content;
     }
     // Print the error message last so it shows up in user's console.
-    std::cerr << format_status.message() << std::endl;
+    std::cerr << filename << format_status.message() << " ";
```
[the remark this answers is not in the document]
hzeller: Added ": ".

verilog/tools/formatter/verilog_format.cc
```diff
@@ -179,24 +182,22 @@ bool formatOneFile(absl::string_view filename,
...
-        std::cerr << "[other error status]" << std::endl;
+        std::cerr << "[other error status]";
         break;
     }
-    if (absl::GetFlag(FLAGS_failsafe_success)) {
-      // original text was preserved, and --inplace modification is skipped.
-      return true;
-    }
-    return false;
+    std::cerr << std::endl;
```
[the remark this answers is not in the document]
hzeller: but then the accumulation message `[other error status]` would be on its own line.

If we don't make it a continuation, we have tp put the format_status.message() print on every switch/case instead of above.

verilog/tools/formatter/verilog_format.cc
hzeller: done.

verilog/tools/formatter/verilog_format.cc
hzeller: Sounds better.done.

verilog/tools/formatter/verilog_format.cc
hzeller: Sounds better. Done.

verilog/tools/formatter/verilog_format.cc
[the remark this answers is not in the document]
hzeller: Ok, repeating the message now which allows for full newline printing in each of them. Introduced `ostream& FileMsg(filename)` to reduce and unify boilerplate for each message that prints filenames with correct `<filename>:<space>` spacing.

verilog/tools/formatter/verilog_format.cc
```diff
@@ -121,23 +121,29 @@ ABSL_FLAG(bool, format_module_instantiations, false,
           "If true, format module instantiations (data declarations), "
           "else leave them unformatted.  This is a short-term workaround.");
 
+std::ostream &FileMsg(absl::string_view filename) {
```
fangism, line 124: nit: `&` hugs left in most of this code base
/me ducks

PR #294, opened 2020-05-07
File operations return status code

hzeller:
Now that we use absl::Status in a lot of place, make the file operations return more detailed information.


common/util/file_util.cc
```diff
@@ -50,20 +50,28 @@ absl::string_view Stem(absl::string_view filename) {
...
+      break;
+    case ENOENT:
+      translated_code = StatusCode::kNotFound;
+      break;
+    case EEXIST:
+      translated_code = StatusCode::kAlreadyExists;
+      break;
+    case EINVAL:
+      translated_code = StatusCode::kInvalidArgument;
+      break;
```
fangism, line 74: worth declaring a function-local `static const map<int, StatusCode>*` for these?

hzeller: I think it makes it more ugly and more slow, so wouldn't do it.

hzeller: I've described it now more compactly. PTAL.

hzeller:
(re-done in #295)

Issue #293, opened 2020-05-07
Support multiple waiver files 


imphil:
With the PR https://github.com/google/verible/pull/245 (issue https://github.com/google/verible/issues/105) one waiver file per Verible run is now supported by passing the name of a waiver file to the `--waivers_file` argument. For our use cases it would be beneficial to be able to specify multiple waiver files. 

For example on OpenTitan, we have one waiver file per IP block to keep them separated from each other. IPs which depend on each other will then use the combined set of waivers (and RTL files, etc.).

This feature is supported by Ascent Lint and Verilator Lint, and it would be great to have the same functionality in Verible.

Would it be possible to allow multiple `--waivers_file` options to enable that? In existing tools, the waivers are simply concatenated together, nothing fancy beyond that happens.

PR #305, opened 2020-05-21
Add support for multiple waiver files

tgorochowik:
Fixes: #293

cc @imphil

### DIFF
--- a/README.md
+++ b/README.md
@@ -372,7 +372,9 @@ All lines in between will be waived for rule-X
 // verilog_lint: waive-stop rule-X
 ```
 
-Another option is to use a configuration file with the `--waivers_file` flag.
+Another option is to use a configuration file with the `--waiver_files` flag.
+The flag accepts a single file or a list of files (comma-separated). Specifying
+multiple files is equivalent to concatenating the files in order of appearance.
 The format of this file is as follows: `waive --rule=rule-name-1 --line=10 waive
 --rule=rule-name-2 --line=5:10 waive --rule=rule-name-3 --regex="^\s*abc$"`
 
--- a/verilog/analysis/verilog_linter.cc
+++ b/verilog/analysis/verilog_linter.cc
@@ -60,9 +60,9 @@ ABSL_FLAG(std::string, rules_config, ".rules.verible_lint",
           "Path to lint rules configuration file.");
 ABSL_FLAG(verilog::RuleSet, ruleset, verilog::RuleSet::kDefault,
           "[default|all|none], the base set of rules used by linter");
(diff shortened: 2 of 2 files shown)


tgorochowik:
@fangism would you mind taking a look at this too? As far as I understand the lack of this feature is what's blocking verible from running in _enforcing_ mode in OpenTitan checks

README.md
```diff
@@ -372,7 +372,8 @@ All lines in between will be waived for rule-X
 // verilog_lint: waive-stop rule-X
 ```
 
-Another option is to use a configuration file with the `--waivers_file` flag.
+Another option is to use a configuration file with the `--waiver_files` flag.
+The flag accepts a single file or a list of files (comma-separated).
```
fangism, line 376: Can you explain how the effects of multiple files are accumulated?
That is, is it equivalent to concatenating files in the order of appearance? or something else?

verilog/analysis/verilog_linter.cc
```diff
@@ -147,23 +147,27 @@ absl::Status VerilogLinter::Configure(
     syntax_tree_linter_.AddRule(std::move(rule));
   }
 
+  absl::Status rc = absl::OkStatus();
   if (!configuration.external_waivers.empty()) {
```
fangism, line 151: If you pass `absl::SkipEmpty()` to `StrSplit` you can remove this `if` check and rely only on the `for` loop.

verilog/analysis/verilog_linter.cc
```diff
@@ -262,7 +266,7 @@ LinterConfiguration LinterConfigurationFromFlags() {
   config.UseRuleBundle(rules);
 
   // Apply external waivers
-  config.external_waivers = absl::GetFlag(FLAGS_waivers_file);
+  config.external_waivers = absl::GetFlag(FLAGS_waiver_files);
```
fangism, line 269: This will move with your other PR, right?

README.md
tgorochowik: Yes, it is as if they were concatenated. The order doesn't matter for now as waivers can only disable something so there should be no conflicts, but I guess it makes sense to add such a comment in case we add more commands in the future.

verilog/analysis/verilog_linter.cc
tgorochowik: Sure, will do!

verilog/analysis/verilog_linter.cc
tgorochowik: Yes, either this or the other PR will have to be rebased after one of them is merged.

PR #306, opened 2020-05-22
linter: Move Flags handling closer to main

tgorochowik:
Cleaning up the flag accesses as promised in #237

### DIFF
--- a/verilog/analysis/verilog_linter.cc
+++ b/verilog/analysis/verilog_linter.cc
@@ -233,75 +233,24 @@ std::vector<LintRuleStatus> VerilogLinter::ReportStatus(
   return statuses;
 }
 
-absl::Status AppendLinterConfigurationFromFile(
-    LinterConfiguration* config, absl::string_view config_filename) {
-  // Read local configuration file
-  std::string content;
-
-  const absl::Status config_read_status =
-      verible::file::GetContents(config_filename, &content);
-  if (config_read_status.ok()) {
-    RuleBundle local_rules_bundle;
-    std::string error;
-    if (local_rules_bundle.ParseConfiguration(content, '\n', &error)) {
-      config->UseRuleBundle(local_rules_bundle);
-    } else {
-      LOG(WARNING) << "Unable to fully parse configuration: " << error;
-    }
-    return absl::OkStatus();
-  }
-
-  return config_read_status;
-}
-
 LinterConfiguration LinterConfigurationFromFlags(
     absl::string_view linting_start_file) {
   LinterConfiguration config;
 
-  // TODO move all of these calls to GetFlag outside of this function
+  const verilog::LinterOptions options = {
+    .ruleset = absl::GetFlag(FLAGS_ruleset),
+    .rules = absl::GetFlag(FLAGS_rules),
+    .config_file = absl::GetFlag(FLAGS_rules_config),
+    .config_file_is_custom = FLAGS_rules_config.IsModified(),
+    .rules_config_search = absl::GetFlag(FLAGS_rules_config_search),
+    .linting_start_file = std::string(linting_start_file),
+    .waiver_files = absl::GetFlag(FLAGS_waiver_files)
+  };
 
-  // Turn on default ruleset.
-  const auto& ruleset = absl::GetFlag(FLAGS_ruleset);
-  config.UseRuleSet(ruleset);
-
-  if (FLAGS_rules_config.IsModified()) {
-    // Use configuration file from flag if specified
-    std::string config_file = absl::GetFlag(FLAGS_rules_config);
-
-    const absl::Status config_read_status =
-        AppendLinterConfigurationFromFile(&config, config_file);
-
-    if (!config_read_status.ok()) {
-      LOG(WARNING) << config_file
-                   << ": Unable to read rules configuration file "
-                   << config_read_status << std::endl;
-    }
-
-  } else if (absl::GetFlag(FLAGS_rules_config_search)) {
-    // Search upward if search is enabled and no configuration file is
-    // specified
-    static constexpr absl::string_view linter_config = ".rules.verible_lint";
-    std::string resolved_config_file;
-    if (verible::file::UpwardFileSearch(linting_start_file, linter_config,
-                                        &resolved_config_file)
-            .ok()) {
-      const absl::Status config_read_status =
-          AppendLinterConfigurationFromFile(&config, resolved_config_file);
-
-      if (!config_read_status.ok()) {
-        LOG(WARNING) << resolved_config_file
-                     << ": Unable to read rules configuration file "
-                     << config_read_status << std::endl;
-      }
-    }
-  }
-
-  // Turn on rules found in config flags.
-  const auto& rules = absl::GetFlag(FLAGS_rules);
-  config.UseRuleBundle(rules);
-
-  // Apply external waivers
-  config.external_waivers = absl::GetFlag(FLAGS_waiver_files);
+   absl::Status config_status = config.ConfigureFromOptions(options);
+   if (!config_status.ok()) {
+     LOG(WARNING) << "Unable to configure linter for: " << linting_start_file;
+   }
 
   return config;
 }
--- a/verilog/analysis/verilog_linter_configuration.cc
+++ b/verilog/analysis/verilog_linter_configuration.cc
@@ -34,6 +34,7 @@
 #include "common/analysis/token_stream_lint_rule.h"
 #include "common/util/container_util.h"
 #include "common/util/enum_flags.h"
+#include "common/util/file_util.h"
 #include "common/util/logging.h"
 #include "verilog/analysis/default_rules.h"
 #include "verilog/analysis/lint_rule_registry.h"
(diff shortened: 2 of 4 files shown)


tgorochowik:
I added the changes you asked for and added a generic unit test.

verilog/analysis/verilog_linter_configuration.h
```diff
@@ -130,6 +130,19 @@ struct ProjectPolicy {
...
+  // strings, ints, bools, and unprocessed values from flags, no other derived
+  // information. Reasonable default values may be specified here for each
+  // member
+  const RuleSet& ruleset;
+  const RuleBundle& rules;
+  absl::string_view config_file;
+  bool config_file_is_custom;
+  bool rules_config_search;
+  absl::string_view waivers_file;
+  absl::string_view linting_start_file;
```
fangism, line 143: Add a brief description for each of these members.

verilog/analysis/verilog_linter_configuration_test.cc
```diff
@@ -775,5 +775,22 @@ TEST(RuleBundleTest, ParseRuleBundleSkipComments) {
   }
 }
 
+// ConfigureFromFlags Tests
+TEST(ConifigureFromFlagsTest, Basic) {
```
fangism, line 779: `ConfigureFromFlagsTest` (spelling)

verilog/analysis/verilog_linter_configuration_test.cc
```diff
@@ -775,5 +775,22 @@ TEST(RuleBundleTest, ParseRuleBundleSkipComments) {
...
+      false,
+      ".",
+      "filename"};
+
+  LinterConfiguration config;
+  auto status = config.ConfigureFromFlags(flags);
+
+  EXPECT_TRUE(status.ok());
+}
+
```
fangism, line 794: Good so far.  Can you test to cover some of the other behavior/code paths in `ConfigureFromFlags`?

tgorochowik: Thanks for the review! I've added some more tests. I don't think we can add unit tests for all of what is being done there without extra refactoring of the code

fangism: Can you leave a `//TODO:` note around here with what could/should be covered in the next refactoring?

tgorochowik: Sure added a comment explaining what I meant - basically that if we also read the configuration files in `main()` we will be able to test the configuration form the content of the configuration files too.

hzeller: Isn't the result of reading a configuration file just a rule-set ? Passing around a LinterConfiguration should contain all that. 
If we want to extend the notion of the _flags_ values being set, we should not call this not LinterConfigurationFlags, but LinterConfigurationOptions, so that it is less tied to the notion of flags (also that is a common way to name such things).

hzeller: (for testing reading from files, there is a [ScopedTestFile](https://github.com/google/verible/blob/master/common/util/file_util.h#L59) when this becomes important)

verilog/analysis/verilog_linter_configuration.h
```diff
@@ -130,6 +130,30 @@ struct ProjectPolicy {
...
+  // information. Reasonable default values may be specified here for each
+  // member
+
+  // The base set of rules used by linter
+  const RuleSet& ruleset;
+  // Bundle of rules to enable/disable in addition to the base set
+  const RuleBundle& rules;
+  // Path to a file with extra linter configuration, applied on top of the
+  // base 'ruleset' and extra 'rules'
+  absl::string_view config_file;
```
hzeller, line 144: I would not store string_views for the config file, linting_start_file and waivers_file.
The struct is meant to be passed around with the configuration, but we don't know anything about the lifetime of the strings we set it to. So all these should be std::string.

verilog/analysis/verilog_linter_configuration.h
```diff
@@ -130,6 +130,30 @@ struct ProjectPolicy {
   std::string ListPathGlobs() const;
 };
 
+struct LinterConfigurationFlags {
+  // strings, ints, bools, and unprocessed values from flags, no other derived
+  // information. Reasonable default values may be specified here for each
+  // member
+
+  // The base set of rules used by linter
+  const RuleSet& ruleset;
```
hzeller, line 139: This essentially is a pointer to a rule-set and given that the struct is meant to be passed around self-contained, I think this should be a  copy of a rule-set. 

If this is not meant to be an instance but in fact a reference (e.g. because you want to build trees of rules-sets), I'd make it a const pointer, so that the lifetime-implications are very clear (and they shoudl also be mentioned in the field documentation).

verilog/analysis/verilog_linter_configuration.h
```diff
@@ -130,6 +130,30 @@ struct ProjectPolicy {
   std::string ListPathGlobs() const;
 };
 
+struct LinterConfigurationFlags {
```
hzeller, line 133: This name sounds very specific to be tied to flags. But in general this is more like a configuration struct, which we commonly would call somethinsomethingOptions; `LinterOptions` maybe ? ('Configuration' could be added but would not add more to the meaning).

verilog/analysis/verilog_linter_configuration.cc
```diff
@@ -282,6 +283,71 @@ bool LinterConfiguration::operator==(const LinterConfiguration& config) const {
...
+    absl::string_view config_filename) {
+  // Read local configuration file
+  std::string content;
+
+  const absl::Status config_read_status =
+      verible::file::GetContents(config_filename, &content);
+  if (config_read_status.ok()) {
+    RuleBundle local_rules_bundle;
+    std::string error;
+    if (local_rules_bundle.ParseConfiguration(content, '\n', &error)) {
```
hzeller, line 296: It feels like we have done the 'read config into string, then send it to ParseConfiguration' somewhere else already ? Maybe we should have RuleBundle implement ParseConfigurationFile()...
(might be something for a later refactoring; if so, add TODO)

verilog/tools/lint/verilog_lint.cc
```diff
@@ -74,9 +91,24 @@ int main(int argc, char** argv) {
   // All positional arguments are file names.  Exclude program name.
   for (const auto filename :
        verible::make_range(args.begin() + 1, args.end())) {
-    // Copy configuration, so that it can be locally modified per file.
-    const LinterConfiguration config(
-        verilog::LinterConfigurationFromFlags(filename));
+    // local configuration, so that it can be locally modified per file.
+    LinterConfiguration config;
+
+    const verilog::LinterConfigurationFlags flags = {
```
hzeller, line 97: So, personally, I'd probably solve this differently: keep the flags wherever they were before, but define the LinterConfigurationFlags struct (or, as suggested LinterConfigurationOptions).
The default constructore of LinterConfigurationOptions would just set the values all from the flags, so that you just would then assign values in main() that you'd like to be overriden.

That way, we keep the encapsulation of flags and keep the use of flags consistent in case we use juse the linter library in another tool. But it also allows for full unit-testing, as all the Options can be set manually.

verilog/analysis/verilog_linter.cc
```diff
@@ -51,23 +50,6 @@
 #include "verilog/parser/verilog_token_classifications.h"
 #include "verilog/parser/verilog_token_enum.h"
 
-// TODO(hzeller): make --rules repeatable and cumulative
-ABSL_FLAG(verilog::RuleBundle, rules, {},
```
hzeller, line 55: I'd leave the flags here or in whatever compilation unit we define the constructor of LlinterConfigurationFlags (or, as I'd like to call it LinterConfigurationOptions). See comments below.

verilog/analysis/verilog_linter_configuration_test.cc
```diff
@@ -775,5 +775,76 @@ TEST(RuleBundleTest, ParseRuleBundleSkipComments) {
...
+  LinterConfiguration config;
+
+  LinterConfigurationFlags flags = {
+      RuleSet::kAll,
+      RuleBundle(),
+      "-",
+      false,
+      false,
+      "filename"
+      "filename"};
```
hzeller, line 805: So a lot of unnamed parameters are a little bit hard to follow as we don't see the definition of LinterConfigurationFlags side-by-side. If the structs would only contain 2-3 parameters, {}-initiaalization would be fine.

Until C++20 catches up with C-89 and allows designated initializers (C++, why oh why arent they there?), we probably have to do manual init:
```
  flags.config_file = "-"
  flags.is_custom = false;
  // ...
```

I suspect that will also clash with the use of const references in the struct, but discussed above.

verilog/tools/lint/verilog_lint.cc
tgorochowik: Sure, I can move them back. I moved them here as the official guidelines ([Flags Best Practices](https://abseil.io/docs/cpp/guides/flags)) say:

> * Prefer to define flags only in the file containing the binary’s main() function. Although Abseil flags may be defined anywhere in any namespace, avoid any usage outside of main() as it will otherwise be difficult to resolve conflicts.
> * Prefer to reference flags only from within the file containing the binary’s main() function, for the same reason.

And I thought that's what @fangism meant by cleaning the flags up - these are just guidelines however so I guess we can use a mix of both approaches.

verilog/analysis/verilog_linter_configuration.cc
tgorochowik: I think this is the exact code you remember, I just moved it to a different file

verilog/analysis/verilog_linter_configuration.h
tgorochowik: Sure, makes sense, renamed

verilog/analysis/verilog_linter_configuration.h
tgorochowik: Sure, thanks, changed to a copy

verilog/analysis/verilog_linter_configuration.h
tgorochowik: Right, changed to strings

verilog/analysis/verilog_linter.cc
tgorochowik: Moved the flags back

verilog/analysis/verilog_linter_configuration_test.cc
tgorochowik: I think in our case we can use the designated initializes as we are initializing everything anyway. Thanks for the tip!

verilog/tools/lint/verilog_lint.cc
fangism: Sorry, it was my idea to move them to main(), but they may still remain in the library, but be referenced from main().

verilog/analysis/verilog_linter_configuration_test.cc
fangism: Keep in mind, the current minimum standard for building is `-std=c++11`.

verilog/tools/lint/verilog_lint.cc
hzeller: ah ok. I think it is better for them to be where the struct is defined to set some defaults.

The ABSL guidelines are good, but it is also good to ignore them when it makes more sense :D

Issue #281, opened 2020-04-28
';' after `uvm_* statements is unwanted


fangism:
We've resolved to be consistent on this subject.

All `` `uvm_* `` _statements_, shall not be terminated with `;`.

Task:

* Examine the syntax tree difference between macro-call statements with vs. without `;` (use `verilog_syntax --printtree`)
* Identify all macro-call statements that start with `` `uvm_ `` in a new `SyntaxTreeLintRule` subclass.
* Check if they terminate with a `;`, and report them.
* Write unit-test test cases.

fangism:
b/155128436

svenka3:
As a regular user and reveiwer of UVM code, I would like to point out that, though this rule is a good one, many user environments violate this one consistently. It is so hard to educate them and trainers across the globe have given up on this one (Cliff Cumming's words at DVCon US 2020). Also I have heard one EDA vendor arguing it is still a legal SV syntax (A null or something). Also most of the simulators accept this extra ;

Hence, my suggestion would be not to add this rule - most violated one it shall become.

Regards
Srini

fangism:
Hi there, this request came from a consensus reached by internal users, and doesn't necessarily mean it has to be imposed on everyone else -- just those who want it moving forward for consistency, and with less interest about fixing legacy code.  With ever-improving configurability in the style linter, you can have it your own way and opt-out.

fangism:
Also tools never get exhausted trying to educate users.  :)

Ciprian167:
@fangism If it's still free, I would like to take a look at it. Could you please assing it to me?

PR #308, opened 2020-05-27
Initial implementation of what could be a solution to #281.

Ciprian167:
UvmMacroSemicolonRule checks if there are any unwanted ';' after a `uvm_* macro call.
Also, a new method has been added in CST/macro.h that makes rule implementation a bit cleaner.


verilog/analysis/checkers/uvm_macro_semicolon_rule_test.cc
```diff
@@ -0,0 +1,68 @@
...
+namespace verilog {
+namespace analysis {
+namespace {
+
+using verible::LintTestCase;
+using verible::RunLintTestCases;
+
+TEST(UvmMacroSemicolonRule, BaseTests) {
+  constexpr int kToken = MacroCallId;
+  const std::initializer_list<LintTestCase> kTestCases = {
```
fangism, line 35: Let's add some trivial negative test cases, like:

* empty file
* empty module
* empty function
* empty class
* empty task
* empty package

verilog/analysis/checkers/uvm_macro_semicolon_rule_test.cc
```diff
@@ -0,0 +1,68 @@
...
+       "initial begin\n"
+         "`uvm_config_db#(virtual top_if)::set"
+         "(\"uvm_test_top\", \" \",  \"if\", if);"
+        "end\n"
+     "endmodule"},
+    {"`uvm_analysis_imp_decl(_RX)\n"}
+  };
+
+  RunLintTestCases<VerilogAnalyzer, UvmMacroSemicolonRule>(kTestCases);
+}
```
fangism, line 64: More test case ideas:

* uvm macro calls inside common flow-control (`if`, `for`) constructs, some inside begin/end blocks, some bare.

verilog/analysis/checkers/uvm_macro_semicolon_rule.cc
```diff
@@ -0,0 +1,94 @@
...
+  
+  if(IsMacroCall(symbol)) {
+    const auto& macro_call_id = GetMacroCallId(symbol);
+
+    if(absl::StartsWithIgnoreCase(std::string(macro_call_id.text),"`uvm_")) {
+      auto &paren_group = GetMacroCallParenGroup(symbol);      
+      auto parameters = std::count_if(paren_group.children().begin(), paren_group.children().end(), 
+                        IsMacroCallClose);
+      if(!parameters) {
+        violations_.insert( LintViolation(symbol, kMessage, context) );
```
fangism, line 80: Suggesting passing the token that represents the offending `;`.

Also construct a message using `absl::StrCat` that looks something like:
_"UVM macro call `uvm_foobar should not be followed by a semicolon ';'."_

(You don't have to use a constant `kMessage` in its entirety.)

verilog/analysis/checkers/uvm_macro_semicolon_rule.cc
```diff
@@ -0,0 +1,94 @@
...
+}
+
+
+void UvmMacroSemicolonRule::HandleSymbol(
+    const verible::Symbol& symbol, const verible::SyntaxTreeContext& context) {
+  
+  if(IsMacroCall(symbol)) {
+    const auto& macro_call_id = GetMacroCallId(symbol);
+
+    if(absl::StartsWithIgnoreCase(std::string(macro_call_id.text),"`uvm_")) {
```
fangism, line 75: No need to copy to `std::string`, that function accepts an `absl::string_view`.

verilog/analysis/checkers/uvm_macro_semicolon_rule.cc
```diff
@@ -0,0 +1,94 @@
...
+      auto &paren_group = GetMacroCallParenGroup(symbol);      
+      auto parameters = std::count_if(paren_group.children().begin(), paren_group.children().end(), 
+                        IsMacroCallClose);
```
fangism, line 78: Not clear to me why it is necessary to inspect the parameters, can you explain?

verilog/analysis/checkers/uvm_macro_semicolon_rule.cc
```diff
@@ -0,0 +1,94 @@
...
+  if (symbol->Kind() == verible::SymbolKind::kLeaf) {
+    const auto& leaf = SymbolCastToLeaf(*symbol.get());
+    return leaf.get().token_enum == verilog_tokentype::MacroCallCloseToEndLine;
+  }
+  return false;
+}
+
+
+void UvmMacroSemicolonRule::HandleSymbol(
+    const verible::Symbol& symbol, const verible::SyntaxTreeContext& context) {
```
fangism, line 70: FYI: an alternative to `HandleSymbol` is overriding `HandleLeaf` and `HandleNode`.  That way you will have already narrowed down the case a little.

fangism:
One more thing I forgot:

Around here:

https://github.com/google/verible/blob/629d8908d05d895ca1586e98bd73284ad41b77f4/verilog/analysis/default_rules.h#L64

Add a comment to enable this new rule in production. `// TODO(b/155128436): "your-new-rule"`

Ciprian167:
Thanks for reviewing this. The strategy is to start this on every `kMacroCall`, then go downwards. Get its corresponding `kMacroCallId` and see if it matches ``` `uvm_*```. If it matches, it will try to find `MacroCallCloseToEndLine` in the initial node's `kParenGroup`. If it doesn't find anything then it will register a violation. 

My first idea was to rely on `kStatement` and implement something similar to what you have suggested, but there are some cases where relying on this wouldn't be enough. For example the following:
```systemverilog
class c;
    `uvm_analysis_imp_decl(_RX);
endclass
```
Would generate the following syntax tree:
```
  ...
  Node (tag: kClassDeclaration) {
    Node (tag: kClassHeader) {
      (#"class" @0-5: "class")
      (#SymbolIdentifier @6-7: "c")
      (#';' @7-8: ";")
    }
    Node (tag: kClassItems) {
      Node (tag: kMacroCall) {
        (#MacroCallId @13-35: "`uvm_analysis_imp_decl")
        Node (tag: kParenGroup) {
          (#'(' @35-36: "(")
          Node (tag: kMacroArgList) {
            Node (tag: kExpression) {
              Node (tag: kReferenceCallBase) {
                Node (tag: kReference) {
                  Node (tag: kLocalRoot) {
                    Node (tag: kUnqualifiedId) {
                      (#SymbolIdentifier @36-39: "_RX")
                    }
                  }
                }
              }
            }
          }
          (#')' @39-40: ")")
        }
        (#';' @40-41: ";")
      }
    }
    (#"endclass" @42-50: "endclass")
  }
  ...
```
So we will have to search for ';' or `MacroCallCloseToEndLine` anyway. 

I guess my implementation makes some assumptions on CST's structure, which may be problematic if this changes. If you have any suggestion on what might work, I am more than happy to explore it.

verilog/CST/macro.cc
```diff
@@ -44,6 +44,10 @@ const TokenInfo& GetMacroGenericItemId(const Symbol& s) {
   return GetSubtreeAsLeaf(s, NodeEnum::kMacroGenericItem, 0).get();
 }
 
+const SyntaxTreeNode& GetMacroCallParenGroup(const Symbol& s) {
+  return GetSubtreeAsNode(s, NodeEnum::kMacroCall, 1, NodeEnum::kParenGroup);
+}
+
```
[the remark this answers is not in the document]
Ciprian167: Done.

verilog/analysis/checkers/uvm_macro_semicolon_rule.cc
```diff
@@ -0,0 +1,94 @@
...
+namespace analysis {
+
+using verible::GetVerificationCitation;
+using verible::LintViolation;
+
+// Register UvmMacroSemicolonRule
+VERILOG_REGISTER_LINT_RULE(UvmMacroSemicolonRule);
+
+absl::string_view UvmMacroSemicolonRule::Name() { return "uvm-macro-semicolon"; }
+const char UvmMacroSemicolonRule::kTopic[] = "uvm-macro-semicolon-convention";
```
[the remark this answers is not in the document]
Ciprian167: Done.

verilog/analysis/checkers/uvm_macro_semicolon_rule.cc
Ciprian167: Good point. I will try to change this and search for ';' instead of `MacroCallCloseToEndLine`, because right now, there is no offending token I can pass.

verilog/analysis/checkers/uvm_macro_semicolon_rule.cc
Ciprian167: Done.

verilog/analysis/checkers/uvm_macro_semicolon_rule.cc
Ciprian167: Well, I was trying to look for `MacroCallCloseToEndLine`. But on a second look, I realise it's was a bad idea, and this is not really necesarry. I will try to search for ';' instead which should be outside parameter list.

verilog/analysis/checkers/uvm_macro_semicolon_rule.cc
Ciprian167: Yeah, I was thinking to use HandleNode. But I remember reading in some comments that HandleSymbol is somehow favored. Maybe I missread it, in that case I will use HandleNode.

fangism:
> Thanks for reviewing this. The strategy is to start this on every `kMacroCall`, then go downwards. Get its corresponding `kMacroCallId` and see if it matches `` `uvm_* ``. If it matches, it will try to find `MacroCallCloseToEndLine` in the initial node's `kParenGroup`. If it doesn't find anything then it will register a violation.
> 
> My first idea was to rely on `kStatement` and implement something similar to what you have suggested, but there are some cases where relying on this wouldn't be enough. For example the following:
> 
> ```systemverilog
> class c;
>     `uvm_analysis_imp_decl(_RX);
> endclass
> ```
> 
> Would generate the following syntax tree:
> 
> ```
>   ...
>   Node (tag: kClassDeclaration) {
>     Node (tag: kClassHeader) {
>       (#"class" @0-5: "class")
>       (#SymbolIdentifier @6-7: "c")
>       (#';' @7-8: ";")
>     }
>     Node (tag: kClassItems) {
>       Node (tag: kMacroCall) {
>         (#MacroCallId @13-35: "`uvm_analysis_imp_decl")
>         Node (tag: kParenGroup) {
>           (#'(' @35-36: "(")
>           Node (tag: kMacroArgList) {
>             Node (tag: kExpression) {
>               Node (tag: kReferenceCallBase) {
>                 Node (tag: kReference) {
>                   Node (tag: kLocalRoot) {
>                     Node (tag: kUnqualifiedId) {
>                       (#SymbolIdentifier @36-39: "_RX")
>                     }
>                   }
>                 }
>               }
>             }
>           }
>           (#')' @39-40: ")")
>         }
>         (#';' @40-41: ";")
>       }
>     }
>     (#"endclass" @42-50: "endclass")
>   }
>   ...
> ```
> 
> So we will have to search for ';' or `MacroCallCloseToEndLine` anyway.
> 
> I guess my implementation makes some assumptions on CST's structure, which may be problematic if this changes. If you have any suggestion on what might work, I am more than happy to explore it.

Right, the macro calls never contain the `;` themselves, so it doesn't help to look downward the tree node.
And you're also right that some strategies are more sensitive to CST structures than others.  Macros in general can appear in many places, so for a rule like this I would try to look for a way that is less reliant on structure.  If you have to rely on CST structure it's ok, that's why we have tests to prevent breakages.  

Some thoughts: off the top of my head:

* Detect when we've just closed a macro with `)` or `MacroCallCloseToEndLine`, or perhaps the macro itself is a plain `MacroIdItem` (see `MacroGenericItem` in verilog.y).  Should be able to look shallowly up-stack (context) to determine this.
* Look shallowly up the stack a little more to determine if this is a direct child of a `kClassItemList` or is acting as a `kStatement`.  (Maybe even define this function for re-use in CST/)
* Remember this state as "having-just-completed-a-macro-call-of-interest", maybe save the macro call node for diagnostics.
* On encountering a `;` check for "having-just-completed-a-macro-call-of-interest", and report a violation there.  Reset that state variable in all other cases.

Any time I suggest looking up the context stack, it is also suitable to create some auxiliary state variables in the lint rule class.  e.g. it's more efficient to remember that you're already inside a module with a bool than to query up the stack all the time.  A fresh lint rule object is created for every file, so you don't have to worry about state persisting across files.

Looking at my notes on the discussion behind this request, it was mostly focused on the UVM info/warning/error macro calls, and less so on the class item macros, so it is ok we start with only statements, and support class items in a follow-up (hopefully a small code change then).

Ciprian167:
Thanks for your advice @fangism .  With this latest commit I tried to follow all your recommandations. I changed the rule strategy (please check the commit comment for implementation details) and added some new tests. Let me know what you think.

Ciprian167:
Thanks for feedback. My initial feeling was to go with ``DirectParentIs`` or ``DirectParentIsOneOf``,  because it didn't felt right to search (possibly) the whole stack, but it didn't really worked for all cases. 

As for nested macros, it's kind of covered. For example:
```systemverilog
`uvm_macro(`uvm_other_macro());
```
throws:
```
uvm_macros_test.svh:1:33: UVM macro call, `uvm_macro should not be followed by a semicolon ';'. [Verification-Style: uvm-macro-semicolon-convention] [uvm-macro-semicolon]
```
which is correct. 

But the following:
```systemverilog
`uvm_macro(`uvm_other_macro(););
`uvm_macro(`uvm_other_macro();)
```
throws:
```text
uvm_macros_test.svh:1:30: UVM macro call, `uvm_macro should not be followed by a semicolon ';'. [Verification-Style: uvm-macro-semicolon-convention] [uvm-macro-semicolon]
uvm_macros_test.svh:2:30: UVM macro call, `uvm_macro should not be followed by a semicolon ';'. [Verification-Style: uvm-macro-semicolon-convention] [uvm-macro-semicolon]
```

So, I guess the message is a bit misleading. To be honest, it feels strange to use something like this. But it definetly can be improved.

fangism:
> Thanks for feedback. My initial feeling was to go with `DirectParentIs` or `DirectParentIsOneOf`, because it didn't felt right to search (possibly) the whole stack, but it didn't really worked for all cases.
> 
> As for nested macros, it's kind of covered. For example:
> 
> ```systemverilog
> `uvm_macro(`uvm_other_macro());
> ```
> 
> throws:
> 
> ```
> uvm_macros_test.svh:1:33: UVM macro call, `uvm_macro should not be followed by a semicolon ';'. [Verification-Style: uvm-macro-semicolon-convention] [uvm-macro-semicolon]
> ```
> 
> which is correct.
> 
> But the following:
> 
> ```systemverilog
> `uvm_macro(`uvm_other_macro(););
> `uvm_macro(`uvm_other_macro();)
> ```
> 
> throws:
> 
> ```
> uvm_macros_test.svh:1:30: UVM macro call, `uvm_macro should not be followed by a semicolon ';'. [Verification-Style: uvm-macro-semicolon-convention] [uvm-macro-semicolon]
> uvm_macros_test.svh:2:30: UVM macro call, `uvm_macro should not be followed by a semicolon ';'. [Verification-Style: uvm-macro-semicolon-convention] [uvm-macro-semicolon]
> ```
> 
> So, I guess the message is a bit misleading. To be honest, it feels strange to use something like this. But it definetly can be improved.

Ok, thanks for checking.  Good enough for now.

PR #323, opened 2020-06-08
Use bazel install for installing binaries instead of copy them.

hzeller:


hzeller:
(currently waiting for CI to finish to see if this would work. Will ask for review later)

hzeller:
Using bazel install, to implement part of #186

PR #324, opened 2020-06-09
Create manpages from the binary in the install location.

hzeller:
...instad of extracting it manually from the bazel-bin/
location.

This is in preparation for the rename of the files. We can't expect the old names to exist anymore in bazel-bin/ soon, but they still will be in the installed location.


hzeller:
(will wait for CI to finish before requesting review).

PR #325, opened 2020-06-09
Create manpage names from new binary names.

hzeller:
Also, add symbolic links so that manpages to the old names still
work.


mithro:
LGTM!

PR #339, opened 2020-06-23
Remove space before `#(`

### LINKED ISSUE #287: Unwanted space before `#(` in actual type parameters
Example formatter output of reduced test case:

```systemverilog
task t;
  a_pkg::foo #(x)::func();
endtask
```

Should be:

```systemverilog
task t;
  a_pkg::foo#(x)::func();
endtask
```

wtatarski:
Fixes #287

### DIFF
--- a/verilog/formatting/token_annotator_test.cc
+++ b/verilog/formatting/token_annotator_test.cc
@@ -177,6 +177,7 @@ struct AnnotateWithContextTestCase {
   FormatStyle style;
   verible::TokenInfo left_token;
   verible::TokenInfo right_token;
+  InitializedSyntaxTreeContext left_context;
   InitializedSyntaxTreeContext right_context;
   ExpectedInterTokenInfo expected_annotation;
 };
@@ -1843,12 +1844,14 @@ TEST(TokenAnnotatorTest, AnnotateFormattingWithContextTest) {
           {'=', "="},
           {verilog_tokentype::TK_StringLiteral, "\"hello\""},
           {},  // any context
+          {},  // any context
           {1, SpacingOptions::Undecided},
       },
       {
           DefaultStyle,
           {'=', "="},
           {verilog_tokentype::TK_EvalStringLiteral, "`\"hello`\""},
+          {},  // any context
           {},  // any context
           {1, SpacingOptions::Undecided},
       },
@@ -1858,12 +1861,14 @@ TEST(TokenAnnotatorTest, AnnotateFormattingWithContextTest) {
           {verilog_tokentype::TK_EOL_COMMENT, "//comment1"},
           {verilog_tokentype::PP_ifdef, "`ifdef"},
           {},  // any context
+          {},  // any context
           {1, SpacingOptions::MustWrap},
       },
       {
           DefaultStyle,
           {verilog_tokentype::SymbolIdentifier, "id"},
           {verilog_tokentype::PP_ifdef, "`ifdef"},
+          {},  // any context
           {},  // any context
           {1, SpacingOptions::MustWrap},
       },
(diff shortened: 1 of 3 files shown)


wtatarski:
@googlebot I signed it!

wtatarski:
Probably we have to edit this case: 
https://github.com/google/verible/blob/e6b80381092049520dae421089fc73fd584f106e/verilog/formatting/token_annotator_test.cc#L3658-L3666

Do you have an idea how to detect kUnqualifiedId from left context?

fangism:
> Probably we have to edit this case:
> https://github.com/google/verible/blob/e6b80381092049520dae421089fc73fd584f106e/verilog/formatting/token_annotator_test.cc#L3658-L3666
> 
> Do you have an idea how to detect kUnqualifiedId from left context?

This struct:

https://github.com/google/verible/blob/e6b80381092049520dae421089fc73fd584f106e/verilog/formatting/token_annotator_test.cc#L176

was written at a time when only `right_context` was provided/needed for inter-token decisions and testing.
Now would be a good time to add a `left_context` member.  :-)  (Then you'd need to add a blank `{}` left-context to all existing test cases.)

wtatarski:
Added left_context. I think it should be OK now.

verilog/formatting/token_annotator_test.cc
```diff
@@ -3712,38 +3954,51 @@ TEST(TokenAnnotatorTest, AnnotateFormattingWithContextTest) {
...
           {1, SpacingOptions::Undecided},
       },
+      {
+          DefaultStyle,
+          {SymbolIdentifier, "id_before_pound"},
+          {'#', "#"},
+          {NodeEnum::kUnqualifiedId},
+          {},  // default context
+          {0, SpacingOptions::Undecided},
+      },
```
fangism, line 3969: One more small request, can you add one more test case to cover left_context = kQualifiedId?

wtatarski: Done

hzeller:
@wtatarski the CLA check did not finish, did you register it with your antmicro address ? (it compares the email addresses)

wtatarski:
@googlebot I signed it!

wtatarski:
@googlebot I signed it!

fangism:
I've reviewed the test changes internally, and I think we missed a detail:

The following spacings are considered "good":

```systemverilog
my_fifo #(.Width($bits(nibble_t)), .Depth(2)) my_fifo_nibble(.in, .out);  // module instance

class foo extends bar #(32, 8);  // unqualified base class
  ...
endclass

foo_h = my_class#(.X(1), .Y(0))::type_id::create("foo_h");  // static method call

my_pkg::x_class#(8, 1) bar;  // package-qualified name
```

The following spacings are "bad":

```systemverilog
my_fifo#(.Width(4), .Depth(2)) my_fifo_2by4(.in, .out);  // instantiation

class foo extends bar#(32, 8);  // unqualified base class
  ...
endclass

foo_h = my_class #(.X(1), .Y(0))::type_id::create("foo_h");  // static method call

my_pkg::x_class #(8, 1) bar;  // package-qualified name
```

Looking over the test cases, it seems we accidentally removed the space before `#(` in module instances (whose base types are unqualified).  Can we go back and fix/undo those?  It may require a little more context examination to distinguish between those cases.

wtatarski:
I've reverted changes in tests related to space before `#(` in module instances. Obviously those tests failed. I will try to fix this case.

wtatarski:
I pushed new fix. Please let me know if this is correct.

wtatarski:
I squashed commits and rebased.

PR #354, opened 2020-07-05
Partial implementation for Issue 313

Ciprian167:
This implementation will create two left alignment columns, one for the `SymbolIdentifier` directly under `kActualNamePort`, and one for `kParenGroup` to align `'(signal_name)'`. This will ignore comments, preproc directives and `.*`. There is no point in ignoring positional port connection as it can't be used together with named connections.

The initial issue had the following desired output:
 ```systemverilog
 foo bar (
 .aa  (aa),
 .aaa (aaa)
);
```
Current output
 ```systemverilog
 foo bar (
 .aa (aa),
 .aaa(aaa)
);
```

It is also missing unit testing.


Ciprian167:
@fangism I couldn't really figure out how to insert a space before the `(`. If you could give me some pointers it would be appreciated. Also, not sure what to do about testing for this class.

Ciprian167:
It seems there is an issue when there are multiple modules in the same file. I get segmentation fault. Ignore for now.

PR #362, opened 2020-07-14
Added support for lint waivers without regex and line

### LINKED ISSUE #343: Support lint waivers without regex and line
I have the following Verible lint error that I want to waive:

```
../../../rtl/ibex_register_file_ff.sv:13:8: Declared module does not match the first dot-delimited component of file name: "ibex_register_file_ff" [Style: file-names] [module-filename]
```

This waiver doesn't need a regex or line since it applies to the whole file. However, a waiver like this fails:

```
waive --rule=module-filename --location="ibex_register_file_ff\.sv"
```

```
E common/analysis/lint_waiver.cc:450] ../../../lint/verible_waiver.vbw:1:68: command error: Insufficient waiver configuration
```

Would it be possible to make both `--line` and `--regex` optional?

--- 

As a workaround, I'm currently using

```
waive --rule=module-filename --regex=".*" --location="ibex_register_file_ff\.sv"
```

The equivalent Verilator waiver looks like this: https://github.com/lowRISC/ibex/blob/master/lint/verilator_waiver.vlt#L28

wtatarski:
Fixes #343

### DIFF
--- a/common/analysis/lint_waiver.cc
+++ b/common/analysis/lint_waiver.cc
@@ -350,7 +350,7 @@ static absl::Status WaiveCommandHandler(
         if (!location_match) return absl::OkStatus();
 
         // Check if everything required has been set
-        if (rule == nullptr || (!can_use_regex && !can_use_lineno)) {
+        if (rule == nullptr) {
           return WaiveCommandError(token_pos, waive_file,
                                    "Insufficient waiver configuration");
         }
@@ -375,6 +375,19 @@ static absl::Status WaiveCommandHandler(
         if (can_use_lineno) {
           waiver->WaiveLineRange(rule, line_start - 1, line_end);
         }
+
+	if (!can_use_regex && !can_use_lineno) {
+	  unsigned int number_of_lines = 0;
+	  std::string content;
+	  absl::Status status = verible::file::GetContents(lintee_filename,
+							   &content);
+	  if (!status.ok()) {
+	    return WaiveCommandError(token_pos, waive_file, status.ToString());
+	  }
+
+	  number_of_lines = std::count(content.begin(), content.end(), '\n');
+	  waiver->WaiveLineRange(rule, 1, number_of_lines);
+	}
 
         return absl::OkStatus();
 
--- a/verilog/tools/lint/lint_tool_test.sh
+++ b/verilog/tools/lint/lint_tool_test.sh
@@ -1,0 +1,110 @@
+#!/bin/bash
+# Copyright 2017-2020 The Verible Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+MY_OUTPUT_FILE="${TEST_TMPDIR}/myoutput.txt"
+readonly MY_OUTPUT_FILE
+
+# Process script flags and arguments.
+[[ "$#" == 1 ]] || {
+  echo "Expecting 1 positional argument, verible-verilog-lint path."
+  exit 1
+}
+lint_tool="$1"
+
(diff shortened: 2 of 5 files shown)


common/analysis/lint_waiver.cc
```diff
@@ -376,6 +377,24 @@ static absl::Status WaiveCommandHandler(
...
+	  if (!user_file.is_open()) {
+	    return WaiveCommandError(token_pos, waive_file,
+				     "lintee_filename was not opened");
+	  }
+
+	  while (std::getline(user_file, line)) {
+	    number_of_lines++;
+	  }
+
+	  user_file.close();
```
fangism, line 394: Recommend using:

https://github.com/google/verible/blob/1886f72085356d3169d35a8b496f97fe1c662959/common/util/file_util.h#L45

to get file contents, and then `std::count(contents.begin(), contents.end(), '\n')` to count lines.

wtatarski:
Will it to correct to place tests in https://github.com/google/verible/blob/master/verilog/tools/lint/BUILD ?
Create waiver files and use `verilog_style_lint.test`? Something like this:
```
verilog_style_lint.test(                                                        
    name = "verilog_lint-test",                                                 
    srcs = ["testdata/instance_parameters.sv"],                                 
    flags = [                                                                   
        "--waiver_files waive.vlt",                                             
    ],                                                                          
    expect_fail = False,                                                        
)  
```

fangism:
> Will it to correct to place tests in https://github.com/google/verible/blob/master/verilog/tools/lint/BUILD ?
> Create waiver files and use `verilog_style_lint.test`? Something like this:
> 
> ```
> verilog_style_lint.test(                                                        
>     name = "verilog_lint-test",                                                 
>     srcs = ["testdata/instance_parameters.sv"],                                 
>     flags = [                                                                   
>         "--waiver_files waive.vlt",                                             
>     ],                                                                          
>     expect_fail = False,                                                        
> )  
> ```

You could, for example, place test data under somewhere like 'verilog/tools/lint/testdata/waivers/...' and access that data from `sh_test`s.  I hadn't considered using `verilog_style_lint.test` because that test doesn't give control over where the command is invoked, and I know some of the lint-waiver behavior is working-dir-sensitive (doing things like upward searching).  An `sh_test` would give you total freedom to set up the environment.

wtatarski:
I added some tests according to your conception. Please let me know if this is correct. Would you like me to add more tests related to waiver?

verilog/tools/lint/lint_tool_test.sh
```diff
@@ -0,0 +1,110 @@
...
+
+grep -q "ERROR: Missing the value for the flag 'waiver_files'" "$MY_OUTPUT_FILE" || {
+  echo "Expected \"ERROR: Missing the value for the flag 'waiver_files'\" in $MY_OUTPUT_FILE but didn't find it.  Got:"
+  cat "$MY_OUTPUT_FILE"
+  exit 1
+}
+
+WAIVER_FILE="verilog/tools/lint/testdata_waivers/waive_module-parameter.vlt"
+
+"$lint_tool" "$TEST_FILE" --waiver_files $WAIVER_FILE > "$MY_OUTPUT_FILE" 2>&1
```
fangism, line 85: quote `"$WAIVER_FILE"` (elsewhere too)

verilog/tools/lint/lint_tool_test.sh
```diff
@@ -0,0 +1,110 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+MY_OUTPUT_FILE="${TEST_TMPDIR}/myoutput.txt"
+readonly MY_OUTPUT_FILE
+
+# Process script flags and arguments.
+[[ "$#" == 1 ]] || {
+  echo "Expecting 1 positional argument, verible-patch-tool path."
```
fangism, line 21: I think you mean `verible-verilog-lint`, not `verible-patch-tool`.

common/analysis/lint_waiver.cc
```diff
@@ -376,6 +376,19 @@ static absl::Status WaiveCommandHandler(
           waiver->WaiveLineRange(rule, line_start - 1, line_end);
         }
 
+	if (!can_use_regex && !can_use_lineno) {
+	  unsigned int number_of_lines = 0;
```
fangism, line 380: This initialization is dead, might as well declare and set value where it is first used a few lines below.
We can take care of that on importing this PR.

Issue #313, opened 2020-05-30
align module instance named port connections


fangism:
Like this:

```systemverilog
module m;
  foo #(
    ...
  ) bar (
    .a   (a),
    .aa  (aa),
    .aaa (aaa)
  );
endmodule
```

If a port is abbreviated with just `.aa`, then there is no need for that line to align with anything else.

Recommend following the example:

https://github.com/google/verible/blob/85f8ff8134957d220400e29b05036318523ea22e/verilog/formatting/align.cc#L80

I expect the handling around new column markers in the case of named port connections to be simpler.

fangism:
b/152805837

Ciprian167:
I would like to give this a try. @fangism, could you please assign it to me?

fangism:
Assigned, thanks for volunteering.

fangism:
Quickly checking in, have you been able to make some progress or have any questions?

Ciprian167:
I didn't got the chance to start it yet. I will come back with questions soon.

fangism:
No problem.  Make sure you sync up through at least https://github.com/google/verible/commit/bdbc19b016898410e8eb3c0b327d822f1acbd994 because I've made some interface improvements that should help with controlling alignment.

fangism:
Just checking in again.  Have any questions?

Ciprian167:
> Just checking in again.  Have any questions?
Sorry I didn't had much time this past month. I'm giving it try today.

Ciprian167:
@fangism I've got a couple of questions:
1. Right now I implemented the named port scanner inside /verible/verilog/formatting/align.cc. Should I move it out in a separate file?

2. I'm trying to workout the following corner cases when interlacing wildcard connections, implicit connections and named connections. 
First, the following:
 ```systemverilog
module m;
    foo bar(.a(a), .aa, .bbb(b));
endmodule
```
Gets formatted:
```systemverilog
module m;
  foo bar (
      .a  (a),
      .aa,
      .bbb(b)
  );
endmodule
```
I guess `.a(a)` should not be aligned with `.bbb(b)`, since they are divided by an implicit connection, which is not aligned with any of them. I assume, I have to do something similar to `GetSubpartitionsBetweenBlankLines` and break partitions separated by blank lines, `.*`, and `.a`, right?

Second, the following:
 ```systemverilog
module m;
    foo bar(.a(a), .*, .bbb(b));
endmodule
```
Ouputs:
```systemverilog
module m;
  foo bar (
[...]

PR #367, opened 2020-07-15
Align module instance named port connections (for #313)

Ciprian167:


verilog/formatting/tree_unwrapper.cc
```diff
@@ -516,6 +516,7 @@ void TreeUnwrapper::InterChildNodeHook(const SyntaxTreeNode& node) {
   switch (tag) {
     // TODO(fangism): cover all other major lists
     case NodeEnum::kPortDeclarationList:
+    case NodeEnum::kPortActualList:
```
fangism, line 519: Thanks for this fix.  Can you also include some tests inside tree_unwrapper_test.cc that cover this directly, using some of the above test cases with comments?  Those tests verify that partitioning structure directly.

verilog/formatting/formatter_test.cc
```diff
@@ -5190,6 +5190,121 @@ static const std::initializer_list<FormatterTestCase> kFormatterTestCases = {
...
+        ".aaa(aaa));\n"
+        "endmodule\n",
+        "module m;\n"
+        "  foo bar (\n"
+        "      .a  (a),\n"
+        "      //.aa(aa),\n"
+        "      .aaa(aaa)\n"
+        "  );\n"
+        "endmodule\n"
+    },
```
fangism, line 5270: more suggested test cases:

* An instance with only implicit `.xyz` connections.
* An instance with empty ports, but containing only comments.  The formatted result should at least indent the comments correctly.
* An instance with one or more port with a trailing //comment, e.g. `.aa(aa),  // this is aa`
* An instance with a blank line separating two groups of named ports.
* A port connection that includes a concatenation, like `.p({q,r,s})`
* A port connection that includes a (non-macro) subrange slice, like `.p(q[r:s])`

verilog/formatting/align.cc
```diff
@@ -66,6 +66,15 @@ static bool TokensAreAllComments(const T& tokens) {
              }) == tokens.end();
 }
 
+template <class T>
+static bool TokensHaveParenthesis(const T& tokens) {
+  return std::find_if(
+              tokens.begin(), tokens.end(),
+              [](const typename T::value_type& token) {
+                return token.TokenEnum() == '(';
+              }) == tokens.end();
```
fangism, line 75: The name of this function is HaveParen, but the logic here (iterator == end()) is the inverse "not-have".

verilog/formatting/align.cc
```diff
@@ -86,6 +95,74 @@ static bool IgnorePortDeclarationPartition(
...
+    return true;
+  
+  // ignore wildcard connections .*
+  if (verilog_tokentype(token_range.front().TokenEnum()) == verilog_tokentype::TK_DOTSTAR) {
+    return true;
+  }
+  
+  // ignore implicit connections .aaa
+  if (verible::SymbolCastToNode(*uwline.Origin()).MatchesTag(NodeEnum::kActualNamedPort) && 
+      TokensHaveParenthesis(token_range)) {
```
fangism, line 117: This works for now, but in the long run, I'd like logic like this to go into functions in verilog/CST.
This one would be something like:
`bool ActualNamedPortIsImplicit(const Symbol&);`
which would examine the CST node substructure directly instead of scanning/searching for `(`.

verilog/formatting/align.cc
fangism: also c++11 introduces `std::any_of` so you don't have to compare against `end()`

verilog/formatting/formatter_test.cc
Ciprian167: > - An instance with empty ports, but containing only comments. The formatted result should at least indent the comments correctly.

This doesn't really test the kPortActualList tabular alignment. But I tried two versions:
```systemverilog
module m;
foo bar(
   //comment1
//comment2
);
endmodule
```
Gets formatted to:
```systemverilog
module m;
  foo bar (
  //comment1
  //comment2
  );
endmodule
```
I feel that it should be?
```systemverilog
module m;
  foo bar (
      //comment1
      //comment2
  );
endmodule
```

And the following doesn't get formatted at all. But this feels like a different issue.
```systemverilog
module m;
  foo bar( //comment1
  //comment2
  );
endmodule
```

fangism: Ok, we'll address this later, I thought/hoped it might already be working with your changes.

Ciprian167:
It's my pleasure. I'm quite new to all this and I like how you and the others are managing this project. It's so clean, organized and very welcoming for a newbie like me. Keep it up!

PR #376, opened 2020-07-28
Added support for comments in waiver files

### LINKED ISSUE #341: Support for comments in waiver files
Waiver files for Verible lint (passed with `--waiver_files`) do not (seem to) support comments. Can that support be added, in line with the comment support in rules files, which was added in https://github.com/google/verible/issues/274?

Comments in waiver are essential to give a bit of motivation why a waiver is necessary (and to add a license header).

wtatarski:
Fixes #341

### DIFF
--- a/common/analysis/command_file.lex
+++ b/common/analysis/command_file.lex
@@ -51,10 +51,13 @@ FlagWithArg {ParamPrefix}{Name}=
 Arg {Value}
 Param {Value}
 
+StartComment #
+EOLComment {StartComment}({InputCharacter}*)
 
 %x COMMAND
 %x ARG
 %x QUOTED_ARG
+%x COMMENT
 
 %%
 
@@ -78,6 +81,12 @@ Param {Value}
   UpdateLocation();
   yy_push_state(ARG);
   return CFG_TK_FLAG_WITH_ARG;
+}
+
+<COMMAND>{EOLComment} {
+  UpdateLocation();
+  yy_push_state(COMMENT);
+  return CFG_TK_COMMENT;
 }
 
 <ARG>{Quote} {
@@ -115,6 +124,24 @@ Param {Value}
 
 <INITIAL>{LineTerminator} {
   UpdateLocation();
+}
+
+{EOLComment} {
+  UpdateLocation();
+  yy_push_state(COMMENT);
+  return CFG_TK_COMMENT;
+}
+
+<COMMENT>{LineTerminator} {
+  UpdateLocation();
+  yy_pop_state();
+  return CFG_TK_NEWLINE;
+}
+
+<COMMENT><<EOF>> {
+  UpdateLocation();
+  yy_pop_state();
+  return CFG_TK_NEWLINE;
 }
 
 <*>{LineTerminator} {
--- a/common/analysis/command_file_lexer.cc
+++ b/common/analysis/command_file_lexer.cc
@@ -42,6 +42,7 @@ CommandFileLexer::CommandFileLexer(absl::string_view config)
         break;
     }
   }
+  Restart(config);
 }
 
 bool CommandFileLexer::TokenIsError(const verible::TokenInfo& token) const {
--- a/common/analysis/command_file_lexer.h
+++ b/common/analysis/command_file_lexer.h
@@ -45,6 +45,7 @@ enum ConfigTokenEnum {
   CFG_TK_ARG,
   CFG_TK_PARAM,
   CFG_TK_NEWLINE,
+  CFG_TK_COMMENT,
   CFG_TK_ERROR,
 };
 
--- a/common/analysis/command_file_lexer_test.cc
+++ b/common/analysis/command_file_lexer_test.cc
@@ -1,0 +1,83 @@
+// Copyright 2017-2020 The Verible Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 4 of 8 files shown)


common/analysis/command_file.lex
```diff
@@ -51,10 +51,13 @@ FlagWithArg {ParamPrefix}{Name}=
 Arg {Value}
 Param {Value}
 
+CommentPrefix #|#{Space}
+Comment {CommentPrefix}{Value}
```
fangism, line 55: Recommend formulating this as a a 'start' and 'rest-of-line', e.g.

https://github.com/google/verible/blob/d811cea2c593985cabbd043d338daa4a8ed08493/verilog/parser/verilog.lex#L175

'Value' in this case should also allow for spaces and tabs (which isn't in 'AnyChar', above).

And suggest naming this `EOLComment`.

I see that the newline is excluded from the comment token and handled separately, that's ok.

wtatarski:
@fangism  Would you like me to add tests related only to waiver comments or the whole command file lexer?

fangism:
> @fangism Would you like me to add tests related only to waiver comments or the whole command file lexer?

Preferably the whole lexer, if possible.  Fortunately it's not a big lexer.  We'd greatly appreciate it.

common/analysis/command_file_lexer.h
```diff
@@ -60,6 +61,7 @@ class CommandFileLexer : public FlexLexerAdapter<veribleCommandFileFlexLexer> {
 
   // Runs the Lexer and attached command handlers
   std::vector<TokenRange> GetCommandsTokenRanges();
+  bool KeepSyntaxTreeTokens(const verible::TokenInfo& t);
```
fangism, line 64: Implementation doesn't use `this`, so this should be a `static` method.

common/analysis/command_file_lexer.cc
```diff
@@ -74,4 +75,13 @@ std::vector<TokenRange> CommandFileLexer::GetCommandsTokenRanges() {
   return commands;
 }
 
+bool CommandFileLexer::KeepSyntaxTreeTokens(const verible::TokenInfo& t) {
+  switch (t.token_enum()) {
+    case CFG_TK_NEWLINE:
```
fangism, line 80: don't we intend to remove comment tokens before parsing?

or if this function is only used for the sake of testing (not needed publicly in non-tests), you could define it as a free function only in the test.  (It looks like you handle comments separately in the parsing state machine)

common/analysis/command_file_lexer_test.cc
```diff
@@ -0,0 +1,64 @@
...
+static void TestFilteredLexer(Args&&... args) {
+  verible::TestLexer<FilteredCommandFileLexer>(std::forward<Args>(args)...);
+}
+
+static std::initializer_list<LexerTestData> kCommentTests = {
+    {{CFG_TK_COMMENT, "#comment"}, {TK_EOF, "\0"}},
+    {{CFG_TK_COMMAND, "waive"}, {TK_EOF, "\0"}},
+};
+
+
```
fangism, line 60: Thanks for starting these tests.

Can you add a test case to cover each token type at least once?

common/text/token_info_test_util.h
```diff
@@ -34,7 +34,10 @@ namespace verible {
 struct ExpectedTokenInfo : public TokenInfo {
   // This pseudo token_enum signals to the test harness to not bother
   // checking the token enum, and check only the string contents.
-  enum { kDontCare = -1 };
+  enum {
+    kDontCare = -2,
+    kSkip
```
fangism, line 39: Explain the difference between these two.

common/analysis/command_file_lexer_test.cc
fangism: Thanks, tests look good.

common/analysis/command_file_lexer_test.cc
```diff
@@ -0,0 +1,83 @@
...
+template <typename... Args>
+static void TestFilteredLexer(Args&&... args) {
+  verible::TestLexer<FilteredCommandFileLexer>(std::forward<Args>(args)...);
+}
+
+static std::initializer_list<LexerTestData> kTokenTests = {
+    {{CFG_TK_COMMENT, "#comment"}, {TK_EOF, "\0"}},
+    {{CFG_TK_COMMAND, "waive"}, {TK_EOF, "\0"}},
+    {{CFG_TK_COMMAND, "waive"},
+     {ExpectedTokenInfo::kSkip, " "},
```
fangism, line 68: For the Verilog lexer, I return a token even for whitespaces, and then filter them out.  This way, every byte in the original text is accounted for somehow: with tokens, whose string_view ranges are contiguous and completely cover the entire text.  

If you went that route, could you then do without `kSkip`?  and just use `" "` (`kDontCare`) for all spaces?

common/text/token_info_test_util.h
wtatarski: `kDontCare` compares length of the token. `kSkip` doesn't compare anything, just skip token.

common/analysis/command_file_lexer_test.cc
wtatarski: Actually test doesn't pass with `" "`. There must be `kSkip`

fangism: In that case, can we rename this `kNoToken`?

common/text/token_info_test_util.h
fangism: Let's rename this `kNoToken` and describe in comment here.

PR #380, opened 2020-08-05
Simplify, speed-up and error-proof op<< for PartitionPolicyEnum

hzeller:
Using switch/case to output enum values is the best for readability and performance-wise
(about 8x speed-up vs the map implementation).

Also, this makes sure it is immune to future issues if the enum
is extended, as the compiler will warn about a new value not handled
in the switch.

### DIFF
--- a/common/formatting/unwrapped_line.cc
+++ b/common/formatting/unwrapped_line.cc
@@ -28,25 +28,26 @@
 #include "common/strings/display_utils.h"
 #include "common/text/tree_utils.h"
 #include "common/util/container_iterator_range.h"
-#include "common/util/container_util.h"
 #include "common/util/spacer.h"
 
 namespace verible {
 
-using verible::container::FindOrDie;
-
 std::ostream& operator<<(std::ostream& stream, PartitionPolicyEnum p) {
-  static const auto* enum_names =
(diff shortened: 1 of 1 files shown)


hzeller:
(while digging through tree related things, fixing stuff on the go)

common/formatting/unwrapped_line.cc
```diff
@@ -28,25 +28,26 @@
...
+  case PartitionPolicyEnum::kFitOnLineElseExpand:
+    return stream << "fit-else-expand";
+  case PartitionPolicyEnum::kTabularAlignment:
+    return stream << "tabular-alignment";
+  case PartitionPolicyEnum::kSuccessfullyAligned:
+    return stream << "aligned-success";
+  case PartitionPolicyEnum::kAppendFittingSubPartitions:
+    return stream << "append-fitting-sub-partitions";
+  }
+  LOG(FATAL) << "Unknown partition";
```
fangism, line 50: "Unknown partition policy" (and maybe the int value of `p`?)

hzeller: Done.
Performance is just a nice side effect of improving the readability of unnecessary complicated code.

Issue #185, opened 2020-02-05
source code indexing, symbol/fact extraction, cross-reference linking


fangism:
Gist:
* extract symbol definitions and declarations and references
* establish use/definition links
* symbols would be properly scoped

Purpose:
* code navigation (IDEs)
* auto-completion
* all-of-the-above, but incremental, and operable on local workspace changes

Representation? any structure that is easily exportable/convertible to interfaces expected by other tools such as IDEs.
For example https://github.com/kythe/kythe is one platform for serving source code information and metadata.

fangism:
@MinaToma work is ongoing at https://github.com/MinaToma/verible

fangism:
@ivan444 FYI

fangism:
b/163065644

ivan444:
This is somewhat done. The indexing is working. The remaining tasks are covered with the finer grained bugs. This issue is not useful anymore.

PR #386, opened 2020-08-06
Add Kythe verification tests to Verible

ivan444:
Issues #185

Releasing Kythe verification tests (without BUILD files due to custom skylark rules).


ivan444:
@minatoma FYI.

PR #400, opened 2020-08-10
package_test_WIP

muhdmud:
This is WIP for 370.

Please have a look if i am the right track, i created another test instead of changing the current one, in case i made a mistake

### DIFF
--- a/verilog/CST/package.cc
+++ b/verilog/CST/package.cc
@@ -39,4 +39,9 @@ const verible::TokenInfo& GetPackageNameToken(const verible::Symbol& s) {
   return name_node.get();
 }
 
+const verible::SyntaxTreeLeaf* GetPackageNameToken2(const verible::Symbol& s)
+{
+  return &verible::GetSubtreeAsLeaf(s, NodeEnum::kPackageDeclaration, 2);
+}
+
 }  // namespace verilog
--- a/verilog/CST/package.cc
+++ b/verilog/CST/package.cc
@@ -34,14 +34,14 @@ std::vector<verible::TreeSearchMatch> FindAllPackageDeclarations(
 }
 
 const verible::TokenInfo& GetPackageNameToken(const verible::Symbol& s) {
-  const auto& name_node =
-      verible::GetSubtreeAsLeaf(s, NodeEnum::kPackageDeclaration, 2);
-  return name_node.get();
+    const auto& name_node =
+        GetPackageNameLeaf(s);
+    return name_node.get();
 }
 
-const verible::SyntaxTreeLeaf* GetPackageNameToken2(const verible::Symbol& s)
+const verible::SyntaxTreeLeaf& GetPackageNameLeaf(const verible::Symbol& s)
 {
-  return &verible::GetSubtreeAsLeaf(s, NodeEnum::kPackageDeclaration, 2);
+  return verible::GetSubtreeAsLeaf(s, NodeEnum::kPackageDeclaration, 2);
 }
 
 }  // namespace verilog
--- a/verilog/CST/package_test.cc
+++ b/verilog/CST/package_test.cc
@@ -36,14 +36,24 @@
 #include "common/util/logging.h"
 #include "verilog/analysis/verilog_analyzer.h"
 
+#include "common/analysis/syntax_tree_search.h"
+#include "common/analysis/syntax_tree_search_test_utils.h"
+
+
 #undef EXPECT_OK
 #define EXPECT_OK(value) EXPECT_TRUE((value).ok())
+
+#undef ASSERT_OK
+#define ASSERT_OK(value) ASSERT_TRUE((value).ok())
 
 namespace verilog {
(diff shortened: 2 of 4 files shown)


verilog/CST/package_test.cc
```diff
@@ -91,6 +101,35 @@ endclass
...
+
+    const auto decls = FindAllPackageDeclarations(*root);
+
+    //EXPECT_EQ(decls.size(), 1);
+    //std::vector<TreeSearchMatch> types;
+    const auto& package_node =
+    down_cast<const SyntaxTreeNode&>(*decls.front().match);
+    const auto& token = GetPackageNameToken(package_node);
+    EXPECT_EQ(token.text(), "mod");
+    //std::ostringstream diffs;
```
fangism, line 128: Take a look at the tests in https://github.com/google/verible/blob/master/verilog/CST/declaration_test.cc

You can write test cases that embed expected tokens that way.  `ExactMatchFindings` will do all of the element comparison work for you.  It also work when you expect no matches.  (Trying breaking some tests in that file to see how it works.)

muhdmud: ok have been playing with it, have somemore testcases.
and package seems to be always tagged in pairs with endpackage, is there a way to split them up into below test :

package p;
//some code

endpackage

or is this where getpackagenametoken comes in?

fangism: You can find more package declaration test cases and examples here (search for endpackage):

https://github.com/google/verible/blob/master/verilog/parser/verilog_parser_unittest.cc

And yes, many language elements are paired with `end*` keywords.

muhdmud: thanks so much for the help! Have added more testcases and removed the commented code. Are these good to move to next task or anymore change needed?

fangism: Last suggestion for this exercise, as it pertains to the original task of finding declared identifiers:
Create or modify tests that pick out the declared package name identifiers, `{"package ", {kTag, "my_package"}, ...}`
(You may also replace/delete older tests that are more than covered by your new one.)

muhdmud: ok noted, i have sample test like this below : 

{"package ", {kTag, "foo"}, "; \n endpackage",
      "package ", {kTag, "bar"}, "; \n endpackage"}


However, i am struggling to fix below, do i need to perform some casting? since GetPackageNameToken returns TokenInfo, instead of symbol that is expected in TreeSearchMatch object

const auto declarations = FindAllPackageDeclarations(*ABSL_DIE_IF_NULL(root));

    std::vector<TreeSearchMatch> declIdentifiers;
    for (const auto& decl : declarations) {
      const auto& packageToken = GetPackageNameToken(*decl.match);
      declIdentifiers.push_back(TreeSearchMatch{&packageToken, {}});
    }

fangism: Not casting but, you may need to have `GetPackageNameToken` return `SyntaxTreeLeaf` instead of the `TokenInfo` inside it.

muhdmud: ok i made another function getpackagenametoken2 to test it first,  since getpackagenametoken affects some code in verilog/analysis/checkers/package_filename_rule* 

also just placed a couple testcases for now, if the code looks good, il add more testcases during the weekend

verilog/CST/package.cc
```diff
@@ -39,4 +39,9 @@ const verible::TokenInfo& GetPackageNameToken(const verible::Symbol& s) {
   return name_node.get();
 }
 
+const verible::SyntaxTreeLeaf* GetPackageNameToken2(const verible::Symbol& s)
```
fangism, line 42: Suggest naming this function `GetPackageNameLeaf` and have it return `const verible::SyntaxTreeLeaf&`, and have `GetPackageNameToken` call this one (to avoid replicating logic).

verilog/CST/package_test.cc
```diff
@@ -36,59 +36,182 @@
...
-      FindAllPackageDeclarations(*ABSL_DIE_IF_NULL(root));
-  EXPECT_EQ(package_declarations.size(), 2);
+TEST(GetPackageNameTokenTest, Testing) {
+
+  constexpr int kTag = 1;
+  const SyntaxTreeSearchTestCase testcases[] = {
+
+      {"package ", {kTag, "foo"}, "; \n endpackage"},
+      {"package ", {kTag, "bar"}, "; \n endpackage"}
+  };
```
fangism, line 190: Exactly, this is the right idea.

verilog/CST/package.cc
muhdmud: changed the function as requested, also added more testcases

muhdmud:
thanks for the help! what usually happens now? can close and comment? or need to wait for merge?

fangism:
> thanks for the help! what usually happens now? can close and comment? or need to wait for merge?

We will import this to the top of tree.  @hzeller, can you take this from here?

PR #401, opened 2020-08-13
Extracting Facts Tree from CST

MinaToma:


verilog/tools/kythe/indexing_facts_tree.h
```diff
@@ -0,0 +1,111 @@
...
+        end_location_(token.right(base)),
+        value_(token.text()){};
+
+  // This function is for debugging only and isn't intended to be a textual
+  // representation of this class.
+  std::string DebugString() const;
+
+  int StartLocation() const { return start_location_; }
+  int EndLocation() const { return end_location_; }
+  std::string Value() const { return value_; }
```
fangism, line 51: return `const&`

hzeller: (will fix it on import)

PR #414, opened 2020-08-31
Class extraction

MinaToma:
Kythe facts extraction for class name and class instances.


hzeller:
Can you base this relative to a current head as suggested in 
https://github.com/google/verible/pull/407#issuecomment-677844490

When I attempt to merge this now with the current state, copybara considers 155 files to be changed (since it is looking since the beginning of the branch changes), and then syncing to upstream create a merge conflict in a bunch of files. So it is essentially makes it very difficult to merge with copybara (copybara doesn't see rebases, so all kinds of things are off).

MinaToma:
Moved to PR #415 
@fangism  @hzeller  @ivan444

MinaToma:
> Key suggestion/comment is to treat classes like any other variable if possible.
> I don't know how this looks with C++ extraction, but consider:
> 
> ```c++
> type_t variable_name;
> ```
> 
> It may not be known what `type_t` is (built-in type or user-defined type).

Kythe has:
- node/kind record (which is used for classes).
- node/kind varaible (which is used for variables).
I don't the the difference between the two but I believe a class is called a record for a reason.

@fangism  if I may suggest we can treat it as a class for now and in verification tests we can figure out more about this.

PR #424, opened 2020-09-03
Macro Extractor

### LINKED ISSUE #391: Kythe: map SystemVerilog macros to Kythe schema
Extend Kythe translation library to support macro definitions and calls/references.

Depends on: #387
Umbrella issue: #185

MinaToma:
Kythe Facts Extraction for macro definition/calls.

Fixes #391


verilog/CST/macro_test.cc
```diff
@@ -174,5 +178,47 @@ TEST(GetFunctionFormalPortsGroupTest, WithFormalPorts) {
   }
 }
 
+TEST(FindAllMacroDefinitions, MacroName) {
+  constexpr int kTag = 1;  // value doesn't matter
+  const SyntaxTreeSearchTestCase kTestCases[] = {
+      {""},
+      {"`define ", {kTag, "PRINT_STRING"}, "(str1) $display(\"%s\\n\", str1)"},
```
fangism, line 185: Suggested test cases:

macro definition appearing inside the following bodies:

* `module`
* `package`
* `class`
* `function`
* `task`

(Technically all macro definitions live inside the same global preprocessor namespace, no matter where they appear.)

fangism (changes requested):
Also TODO/new-issue: track macro definitions and calls appearing inside macro definition bodies

Yes, SystemVerilog preprocessing is a sick language that allows defines inside defines inside conditionals...

ivan444 (approved):

fangism (changes requested):
Can you add at least one unit test for a macro definition that shows that even if a definition appears inside a scope (like module or class), that it is still available in the (effective) "root namespace"?

fangism (commented):
Looks good so far.  Do you plan to cover:

* macro formal arguments, like `a` and `b` from `` `define FOO(a, b) ``?
* macro references like ``parameter int x = `bar; `` and `` `FOO(`BAR)``

or in follow-up PRs?

MinaToma:
> Looks good so far. Do you plan to cover:
> 
> * macro formal arguments, like `a` and `b` from `` `define FOO(a, b) ``?
> * macro references like `` parameter int x = `bar;  `` and ``  `FOO(`BAR) ``
> 
> or in follow-up PRs?

The first case is not covered and I didn't consider as the body can't be extracted and to do that CST needs to change.
The second one maybe covered I can't verify that because `parameter int` is not covered yet but the following works:
- $display("%d\n", `TEN); --> `TEN` is captured 
- `PRINT_3_STRINGS("Grand", "Tour", "S4"); --> `PRINT_3_STRINGS` is captured
- `PRINT_3_INTS(x, y, z); --> `PRINT_3_INTS` is captured and `x` ,`y` and `z` are ready to be captured after extracting `int`.

fangism:
> > Looks good so far. Do you plan to cover:
> > 
> > * macro formal arguments, like `a` and `b` from `` `define FOO(a, b) ``?
> > * macro references like `` parameter int x = `bar;  `` and ``  `FOO(`BAR) ``
> > 
> > or in follow-up PRs?
> 
> The first case is not covered and I didn't consider it as the body can't be extracted and to do that CST needs to change.
> The second one maybe covered I can't verify that because `parameter int` is not covered yet but the following works:
> 
> * $display("%d\n", `TEN); --> `TEN` is captured
> * `PRINT_3_STRINGS("Grand", "Tour", "S4"); --> `PRINT_3_STRINGS` is captured
> * `PRINT_3_INTS(x, y, z); --> `PRINT_3_INTS`is captured and`x` ,`y`and`z`are ready to be captured after extracting`int`.

The first case does not need the macro definition body, `a` and `b` are un-typed parameters (that can then be referenced for substitution inside the body).  Linking macro parameter references within the body can be a task for later, it may require more lexer work.

The second case was specifically targeted at macro references (`` `BAR``) as macro call arguments.

Both can be done later (but we should file issues to track).

verilog/tools/kythe/indexing_facts_tree_extractor.cc
```diff
@@ -123,6 +124,28 @@ void IndexingFactsTreeExtractor::Visit(const verible::SyntaxTreeNode& node) {
...
+    // e.g leaves tagged with MacroIdentifier.
+    case NodeEnum::kUnqualifiedId: {
+      const verible::SyntaxTreeLeaf* leaf =
+          ABSL_DIE_IF_NULL(AutoUnwrapIdentifier(node));
+
+      // Leaves tagged with MacroIdentifier.
+      if (leaf->get().token_enum() == verilog_tokentype::MacroIdentifier) {
+        ExtractMacroReference(*leaf);
+        break;
+      }
```
MinaToma, line 145: does this cover cases like:  parameter int x = TEN?

MinaToma: I check it in CST and this should be covered by this part of code.

fangism: You have to examine the CST structure `--printtree` to confirm.

I would however recommend that for macros, you use a implement a `::Visit(const SyntaxTreeLeaf&)` to simplify this.  Macro identifiers are neither qualified nor unqualified as they may appear anywhere.  This might actually cover `` `BAR`` in  `` `FOO(`BAR) `` as well.

fangism (changes requested):

MinaToma:
> Looks good so far. Do you plan to cover:
> 
> * macro formal arguments, like `a` and `b` from `` `define FOO(a, b) ``?
> * macro references like `` parameter int x = `bar;  `` and ``  `FOO(`BAR) ``
> 
> or in follow-up PRs?

The arguments are extracted now and for cases like `` parameter int x = `bar;  `` and ``  `FOO(`BAR) `` both are covered now.

fangism (approved):
Thanks for covering the macro references and formal arguments.  Nice work with testing.

fangism (approved):

fangism (approved):

MinaToma:
moved to PR #430 because of conflicts and problems on merge.

Issue #392, opened 2020-08-06
Kythe: map SystemVerilog packages to Kythe schema


ivan444:
Extend Kythe translation library to support packages.

Depends on: #387
Umbrella issue: #185

PR #427, opened 2020-09-08
Package Extractor

MinaToma:
Kythe facts extraction for Packages.

Convers:
- package name (also package name after `endpackage`).
- package body.
- package imports: `import pkg::*` and `import pkg::my_var`

Fixes: #392


verilog/CST/package.cc
```diff
@@ -42,4 +42,22 @@ const verible::SyntaxTreeLeaf& GetPackageNameLeaf(const verible::Symbol& s) {
...
+  const auto& package_name = verible::GetSubtreeAsLeaf(
+      verible::SymbolCastToNode(*label_node), NodeEnum::kLabel, 1);
+  return &package_name;
+}
+
+const verible::Symbol* GetPackageItemList(
+    const verible::Symbol& package_declaration) {
+  return verible::GetSubtreeAsSymbol(package_declaration,
+                                     NodeEnum::kPackageDeclaration, 4);
+}
```
MinaToma, line 61: @fangism how can this function be tested? can you provide some similar tests?

fangism: Example:

```systemverilog
package p;
  parameter int foo = 1;      // start matching tagged region here
  function void f();          // end matching here
endpackage
```

Follow the same outline as declaration_test.cc: analyze to get a syntax tree, match/find all package declarations, `GetPackageItemList` from each package (if not null), store the `TreeSearchMatch` into pas into `ExactMatchFindings`.

Does this help?

fangism (approved):

verilog/tools/kythe/kythe_facts_extractor.cc
```diff
@@ -24,6 +24,29 @@
...
+
+    Visit(child);
+  }
+
+  for (auto x : package_scope_context) {
+    LOG(INFO) << x.first;
+    for (auto y : x.second) {
+      LOG(INFO) << y.signature;
+    }
+    LOG(INFO) << "";
```
ivan444, line 46: Do you plan to keep the logging / the whole loop? Seems noisy.

verilog/tools/kythe/kythe_facts_extractor.h
```diff
@@ -189,6 +201,27 @@ class KytheFactsExtractor {
...
+  //   function my_fun(); endfunction
+  //   class my_class; endclass
+  // endpackage
+  //
+  // Creates the following:
+  // {
+  //   "pkg1": ["my_fun", "my_class"],
+  //   "pkg2": ["my_fun", "my_class"]
+  // }
+  std::map<std::string, std::vector<VName>> package_scope_context;
```
ivan444, line 223: This is a private variable. Append "_" to the end of the name.

verilog/tools/kythe/kythe_facts_extractor.cc
```diff
@@ -323,7 +381,9 @@ VName KytheFactsExtractor::PrintAnchorVName(const Anchor& anchor) {
 
 std::string KytheFactsExtractor::CreateScopeRelativeSignature(
     absl::string_view signature) const {
-  return absl::StrCat(signature, "#", vnames_context_.top().signature);
+  return vnames_context_.empty()
+             ? std::string(signature)
```
ivan444, line 385: Can you explain this change? Can't figure it out from the context.

verilog/tools/kythe/kythe_facts_extractor.cc
```diff
@@ -24,6 +24,29 @@
...
+  Visit(root);
+}
+
+void KytheFactsExtractor::CreatePackageScopes(const IndexingFactNode& root) {
+  for (const verible::VectorTree<IndexingNodeData>& child : root.Children()) {
+    if (child.Value().GetIndexingFactType() != IndexingFactType::kPackage) {
+      continue;
+    }
+
+    Visit(child);
```
ivan444, line 38: How do you prevent outputting duplicate VNames? (Visit is called for every package child, and for the code root)

verilog/tools/kythe/kythe_facts_extractor.cc
MinaToma: This is just to generalize the case for creating the scope-relative-signature, consider the below:

```
class bar;
endclass: bar

module foo();
   class zar;
   endclass;
endmodule
```

For the first class the vnames_context is empty but it's not for the second.

verilog/tools/kythe/kythe_facts_extractor.cc
MinaToma: Thanks for this note, I didn't notice that.

verilog/tools/kythe/kythe_facts_extractor.cc
ivan444: That's very implicit and makes the code difficult to understand. Can you make this behavior more explicit? Even if it needs forking some code.

hzeller:
mmh, trying to merge this also looks difficult. Lots of little conflicts in `indexing_facts_tree_extractor.cc`

MinaToma:
> mmh, trying to merge this also looks difficult. Lots of little conflicts in `indexing_facts_tree_extractor.cc`

Will this be solved if it is merged with MinaToma/verible/master?

verilog/tools/kythe/kythe_facts_extractor.cc
```diff
@@ -307,6 +379,62 @@ VName KytheFactsExtractor::ExtractClassInstances(
   return class_instance_vname;
 }
 
+VName KytheFactsExtractor::ExtractPackageImport(
+    const IndexingFactNode& import_fact_node) {
+  const auto& anchors = import_fact_node.Value().Anchors();
+  const Anchor& package_name = anchors[0];
```
ivan444, line 385: Is it guaranteed that there will be at least one anchor? Is a size check needed?

verilog/tools/kythe/kythe_facts_extractor.cc
```diff
@@ -307,6 +379,62 @@ VName KytheFactsExtractor::ExtractClassInstances(
...
+
+    const VName imported_item_anchor = PrintAnchorVName(imported_item_name);
+    GenerateEdgeString(imported_item_anchor, kEdgeRef, defintion_vname);
+
+    // Add the found definition to the current scope as if it was declared in
+    // our scope so that it can be caputred without "::".
+    scope_context_.top().push_back(defintion_vname);
+  }
+  // case of import pkg::*.
+  else {
```
ivan444, line 408: This is uncommon formatting. Can you move the comment inside the else-block?

hzeller:
You could try getting MinaToma/verible/master synced with google/verible/master and then re-base this pull request relative to your new MinaToma/verible/master.
(though our recent experiments always were a bit mixed experience...)

I suspect that we had problems as git attempts to rebase every commit in-between in a sequence on the new base and that that caused issues. Maybe there is a way to squash-rebase everything, i.e. just have the resulting end of that branched rebased.

That then should essentially be the same as [taking the diff](https://patch-diff.githubusercontent.com/raw/google/verible/pull/427.diff) and apply it to a freshly synced master.

verilog/tools/kythe/kythe_facts_extractor.cc
```diff
@@ -24,7 +24,30 @@
 namespace verilog {
 namespace kythe {
 
-void KytheFactsExtractor::Visit(const IndexingFactNode& node) {
+void KytheFactsExtractor::ExtractKytheFacts(const IndexingFactNode& root) {
+  CreatePackageScopes(root);
```
ivan444, line 28: This won't handle package imported from other files. Only those defined in the same file, right?

This is intentional, IIUC. (for now, until we have support for processing multiple files)

Or am I missing something?

ivan444 (approved):

verilog/tools/kythe/kythe_facts_extractor.cc
hzeller: Specifically, the [Google styleguide](https://google.github.io/styleguide/cppguide.html#Conditionals) always wants the closing brace directly followed by the else
```
} else {
```

... which then in turn requires to put the comment inside the block.
```
} else {
   // case of import pkg::*
  ...
```

Though
```
} else {  // case of import pkg::*
```

would also be acceptable.

verilog/tools/kythe/kythe_facts_extractor.cc
MinaToma: For package imports the facts tree node would contain two anchors in case of import pkg::myVar one for pkg and the other for myVar and in case of import pkg::* we will have only one anchor for pkg.

verilog/tools/kythe/kythe_facts_extractor.cc
MinaToma: Yes this is intended for single files but in case the same KyhteFactsExtractor object is used for processing other files the packages would still be there.
But for now this works fine for single files.

verilog/tools/kythe/kythe_facts_extractor.cc
MinaToma: Thanks for the info I will fix it.

verilog/CST/package_test.cc
```diff
@@ -345,5 +347,148 @@ TEST(GetPackageNameTokenTest, ValidPackage) {
...
+    EXPECT_TRUE(test.ExactMatchFindings(lists, code_copy, &diffs))
+        << "failed on:\n"
+        << code << "\ndiffs:\n"
+        << diffs.str();
+  }
+}
+
+TEST(PackageImportTest, GetImportedPackageName) {
+  constexpr int kTag = 1;  // value doesn't matter
+  const SyntaxTreeSearchTestCase kTestCases[] = {
```
fangism, line 428: More suggested test cases:

```systemverilog
module m
  import a_pkg::foo;
  import b_pkg::bar;
  ();
endmodule
```

(including variants with `::*`)

These can also be used below for `GetImportedItemName`.

verilog/tools/kythe/kythe_facts_extractor.cc
```diff
@@ -24,7 +24,30 @@
...
+  IndexingFactNodeTagResolver(root);
+}
+
+void KytheFactsExtractor::CreatePackageScopes(const IndexingFactNode& root) {
+  // Searches for packages as a first pass and saves their scope to be used for
+  // imports.
+  for (const verible::VectorTree<IndexingNodeData>& child : root.Children()) {
```
fangism, line 35: `const IndexingFactNode&`?  (you may use the `typedef`, it makes it obvious that the children are the same type as the object itself)

If you elect to change this, check elsewhere too for consistency.

verilog/tools/kythe/kythe_facts_extractor.cc
```diff
@@ -222,6 +272,28 @@ VName KytheFactsExtractor::ExtractVariableReferenceFact(
...
+
+  const VName package_vname(file_path_,
+                            CreateScopeRelativeSignature(
+                                CreatePackageSignature(package_name.Value())));
+  const VName package_name_anchor = PrintAnchorVName(package_name);
+
+  GenerateFactString(package_vname, kFactNodeKind, kNodePackage);
+  GenerateEdgeString(package_name_anchor, kEdgeDefinesBinding, package_vname);
+
+  if (anchors.size() > 1) {
```
fangism, line 289: You've already referenced `anchors[1]` above.  Should that reference declaration `package_end_label` be moved inside here to protect from out-of-bounds access?

verilog/tools/kythe/kythe_facts_extractor.cc
```diff
@@ -307,6 +379,62 @@ VName KytheFactsExtractor::ExtractClassInstances(
...
+    const Anchor& imported_item_name = anchors[1];
+    const VName& defintion_vname =
+        *ABSL_DIE_IF_NULL(SearchForDefinitionVNameInPackage(
+            package_name.Value(), imported_item_name.Value()));
+
+    const VName imported_item_anchor = PrintAnchorVName(imported_item_name);
+    GenerateEdgeString(imported_item_anchor, kEdgeRef, defintion_vname);
+
+    // Add the found definition to the current scope as if it was declared in
+    // our scope so that it can be caputred without "::".
```
fangism, line 404: captured

(check elsewhere too)

verilog/tools/kythe/kythe_facts_extractor.cc
```diff
@@ -307,6 +379,62 @@ VName KytheFactsExtractor::ExtractClassInstances(
...
+    for (const VName& vname : current_package_scope->second) {
+      scope_context_.top().push_back(vname);
+    }
+  }
+
+  return package_vname;
+}
+
+const VName* KytheFactsExtractor::SearchForDefinitionVNameInPackage(
+    std::string package_name, std::string reference_name) {
```
fangism, line 422: I think these parameters can be `absl::string_view`.

And it looks (at-a-glance) like this method could be `const`.

verilog/tools/kythe/kythe_facts_extractor.h
```diff
@@ -197,13 +248,16 @@ class KytheFactsExtractor {
...
+
 // Creates the signature for Class names.
 std::string CreateClassSignature(absl::string_view);
 
 // Creates the signature for module instantiations.
 std::string CreateVariableSignature(absl::string_view);
 
-// Creates the signature for funcitons or tasks.
+// Creates the signature for functions or tasks.
 std::string CreateFunctionOrTaskSignature(absl::string_view);
```
fangism, line 261: Do these need to be in the public header (are they called outside of this library?), or could they be private-ized to the .cc implementation file?

fangism (changes requested):
The imports are looking good!

fangism:
Just FYI @muhdmud This PR will also touch CST/package.

verilog/tools/kythe/kythe_facts_extractor.h
```diff
@@ -129,6 +130,28 @@ class KytheFactsExtractor {
     }
   };
 
+  // Searches for a defintion suitable for the given reference within package
```
fangism, line 133: "definition" (same misspelling in at least one other place)

fangism (approved):
Just one misspelling.  :)

fangism (approved):

MinaToma:
moved to PR #443 because of conflicts.

Issue #390, opened 2020-08-06
Kythe: map SystemVerilog functions to Kythe schema


ivan444:
Extend Kythe translation library to support task/function definitions and calls.

Depends on: #387
Umbrella issue: #185

MinaToma:
This is done. The followups are in #439 #437 #438

PR #417, opened 2020-09-01
Function and task extractor

MinaToma:
Kythe facts extractor for function and tasks.

Coverd:
- function and task name.
- function and task call (without ports).
- function and task ports in declaration.

WIP #390 
Fixes #418

PR #429, opened 2020-09-09
Function and task extractor

### LINKED ISSUE #418: Kythe: extract Non-ANSI style ports in module declaration
Extend Kythe facts extractor to extract Non-ANSI ports in module declaration.

```
module foo(a, b);
    input a;
    input b;
endmodule
```

Depends on: #388

MinaToma:
Kythe facts extractor for function and tasks.

Coverd:
- function and task name.
- function and task call (without ports).
- function and task ports in declaration.

WIP #390 
Fixes #418


fangism (approved):
Approved, based on #417 .

PR #447, opened 2020-09-14
Module port data type forwarding

### LINKED ISSUE #419: Kythe: extract data type forwarding in module declaration
Extend Kythe facts extractor to extract  ports with no data type in module declaration.

```
module foo(input a, b);
endmodule
```

Depends on: #388

MinaToma:
Fixes #419

### DIFF
--- a/verilog/tools/kythe/indexing_facts_tree_extractor.cc
+++ b/verilog/tools/kythe/indexing_facts_tree_extractor.cc
@@ -119,11 +119,6 @@ void IndexingFactsTreeExtractor::Visit(const SyntaxTreeNode& node) {
 
       break;
     }
-    case NodeEnum::kPortDeclaration:
-    case NodeEnum::kPortReference: {
-      ExtractModulePort(node);
-      break;
-    }
     case NodeEnum::kIdentifierUnpackedDimensions: {
       ExtractInputOutputDeclaration(node);
       break;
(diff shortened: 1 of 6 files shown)


verilog/tools/kythe/indexing_facts_tree_extractor.cc
```diff
@@ -217,17 +212,27 @@ void IndexingFactsTreeExtractor::ExtractModuleHeader(
     return;
   }
 
-  Visit(*port_list);
+  bool has_propagated_type = false;
```
fangism, line 215: Can you add some //comments to explain this section, with examples of input code?
Are you making any assumptions with this particular implementation?

verilog/tools/kythe/indexing_facts_tree_extractor.cc
```diff
@@ -217,17 +212,27 @@ void IndexingFactsTreeExtractor::ExtractModuleHeader(
...
+    } else if (tag == NodeEnum::kPort) {
+      ExtractModulePort(GetPortReferenceFromPort(port_node),
+                        has_propagated_type);
+    }
+  }
 }
 
 void IndexingFactsTreeExtractor::ExtractModulePort(
-    const SyntaxTreeNode& module_port_node) {
+    const SyntaxTreeNode& module_port_node, bool has_propagated_type = 0) {
```
fangism, line 233: If you always invoke this method with the parameter, there's no need to provide a default value.

verilog/tools/kythe/indexing_facts_tree_extractor.cc
```diff
@@ -217,17 +212,27 @@ void IndexingFactsTreeExtractor::ExtractModuleHeader(
...
   const auto tag = static_cast<verilog::NodeEnum>(module_port_node.Tag().tag);
 
-  // TODO(minatoma): Fix case like:
-  // module m(input a, b); --> b is treated as a reference but should be a
-  // definition.
```
fangism, line 229: Retain a comment that explains that this is handling both cases where `b` is a non-ANSI port declaration and `b` follows a typed ANI-style port declaration with type/direction.

verilog/tools/kythe/indexing_facts_tree_extractor_test.cc
```diff
@@ -422,6 +422,10 @@ TEST(FactsTreeExtractor, ModuleWithPortsNonANSIStyleTest) {
                                                         {kTag, "a"},
                                                         ", ",
                                                         {kTag, "b"},
+                                                        ", input ",
```
fangism, line 425: I suggest we add a type to this example, like `input wire z`.  But it's good to test that both cases work.  I might even recommend that the new capability be split out into a separate test case because mixing non-ANSI and ANSI-style is technically invalid Verilog.  Verilog has this notion of implict net-type (like `wire`), but if a type is missing like in `input x`, then we just won't assign a type to it.  We should, however, have some fact about a ports direction (but that may follow-up separately).

fangism (changes requested):

fangism (approved):

MinaToma:
@hzeller are we facing the same merge issue here?

hzeller:
This looks good to merge.

PR #448, opened 2020-09-14
Package member referencing 

### LINKED ISSUE #440: Kythe: extend package extractor to extract package member direct access
Extend Kythe facts extractor to cover cases like `display(pkg::some_var)`

MinaToma:
Fixes #440


MinaToma:
@hzeller do we have merge issues here?

fangism (approved):

verilog/tools/kythe/kythe_facts_extractor.h
```diff
@@ -229,6 +229,9 @@ class KytheFactsExtractor {
   void GenerateEdgeString(const VName& source, absl::string_view name,
                           const VName& target) const;
 
+  // Create the signature for the given name (variable, module ..etc).
+  const std::string CreateSignature(absl::string_view name) const;
```
fangism, line 233: The return type doesn't need `const` (unless it is a reference).

fangism: Since this function doesn't use any class members, it should be a `static` method.

hzeller:
this looks like it wants to be rebased on a recent master. There are all kinds od conflicts in places that used the old signature functions and now use the new one.

verilog/tools/kythe/indexing_facts_tree_extractor.cc
```diff
@@ -586,5 +590,20 @@ void IndexingFactsTreeExtractor::ExtractPackageImport(
...
 }
 
+void IndexingFactsTreeExtractor::ExtractQualifiedId(
+    const verible::SyntaxTreeNode& qualified_id) {
+  IndexingNodeData member_reference_data(IndexingFactType::kMemberReference);
+
+  const std::vector<TreeSearchMatch>& unqualified_ids =
+      FindAllUnqualifiedIds(qualified_id);
+
+  for (const TreeSearchMatch unqualified_id : unqualified_ids) {
```
hzeller, line 600: Make this a `const TreeSearchMatch&` to avoid a copy.

hzeller:
I think I found a way to get this merged. No need to create a new pull request.
If you just fix the conflict, I should be able to get it in.

fangism (approved):

verilog/tools/kythe/indexing_facts_tree_extractor.cc
```diff
@@ -604,5 +608,20 @@ void IndexingFactsTreeExtractor::ExtractPackageImport(
   facts_tree_context_.top().NewChild(package_import_data);
 }
 
+void IndexingFactsTreeExtractor::ExtractQualifiedId(
+    const verible::SyntaxTreeNode& qualified_id) {
+  IndexingNodeData member_reference_data(IndexingFactType::kMemberReference);
+
+  const std::vector<TreeSearchMatch>& unqualified_ids =
```
ivan444, line 615: this shouldn't be a reference (it's a dangling reference as FindAllUnqualifiedIds returns a value)

ivan444 (approved):

PR #462, opened 2020-09-17
Primitive type extractor

### LINKED ISSUE #442: Kythe: add support to integers, strings, bit and logic
Extend Kythe facts extractor to support extraction for integers, strings, bits, and logic for both definitions and references.

MinaToma:
Fixes #442

### DIFF
--- a/verilog/CST/declaration.h
+++ b/verilog/CST/declaration.h
@@ -90,6 +90,8 @@ std::vector<verible::TreeSearchMatch> FindAllRegisterVariables(
     const verible::Symbol&);
 std::vector<verible::TreeSearchMatch> FindAllGateInstances(
     const verible::Symbol&);
+std::vector<verible::TreeSearchMatch> FindAllVariableDeclarationAssign(
+    const verible::Symbol&);
 
 // For a given data declaration (includes module instantiation), returns the
 // subtree containing qualifiers.  e.g. from "const foo bar, baz;",
@@ -131,6 +133,11 @@ const verible::TokenInfo& GetModuleInstanceNameTokenInfoFromGateInstance(
 // name. e.g. int b1; returns TokenInfo for "b1".
 const verible::TokenInfo& GetInstanceNameTokenInfoFromRegisterVariable(
     const verible::Symbol&);
+
+// For a given node tagged with kVariableDeclarationAssign return the
+// unqualified id inside that node.
+const verible::SyntaxTreeLeaf& GetUnqualifiedIdFromVariableDeclaratioAssign(
+    const verible::Symbol& variable_declaration_assign);
 
 }  // namespace verilog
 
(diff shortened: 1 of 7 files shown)


ivan444 (approved):

verilog/CST/declaration.h
```diff
@@ -132,6 +134,12 @@ const verible::TokenInfo& GetModuleInstanceNameTokenInfoFromGateInstance(
 const verible::TokenInfo& GetInstanceNameTokenInfoFromRegisterVariable(
     const verible::Symbol&);
 
+// For a given node tagged with kVariableDeclarationAssign return the
+// unqualified id inside that node.
```
fangism, line 138: there can be many unqualified ids.  can you provide an example?  from "x" extracts "y"?

verilog/tools/kythe/indexing_facts_tree_extractor.cc
```diff
@@ -116,6 +116,18 @@ void IndexingFactsTreeExtractor::Visit(const SyntaxTreeNode& node) {
...
+        ExtractPrimitiveVariables(node, register_variables);
+        break;
+      }
+
+      const std::vector<TreeSearchMatch> variable_declaration_assign =
+          FindAllVariableDeclarationAssignment(node);
+
+      if (!variable_declaration_assign.empty()) {
+        ExtractPrimitiveVariables(node, variable_declaration_assign);
+        break;
```
fangism, line 130: Does this nave to be done using FindAll... or could you leverage the visitation that is already happening?

verilog/tools/kythe/testdata/more_testdata/primitive_types.sv
```diff
@@ -0,0 +1,78 @@
...
+    function int my_function(int my_arg1, int my_arg2);
+        //- @x defines/binding XDef
+        //- XDef.node/kind variable
+        //- XDef.complete definition
+        //- XDef childof MyFunction
+        //- @y defines/binding YDef
+        //- YDef.node/kind variable
+        //- YDef.complete definition
+        //- YDef childof MyFunction
+        int x, y;
```
fangism, line 67: do you have one test case with `int x = value;` (declare and assign)?
or even `int x = 1, y = 0;` (multiple)?

fangism (changes requested):

verilog/tools/kythe/indexing_facts_tree_extractor.cc
MinaToma: I think it can be leveraged.

MinaToma: `kVariableDeclarationAssignment` is contained inside a `kDataDeclaration` Node so it will enter the switch case for `kDataDeclaration` and the traversal would stop. so I think it's okay to leave it this way.

fangism: ok for now, but consider defaulting to traversing more subnodes might simplify some logic down the road.  It might reduce the number of cases you have to handle explicitly, but add a little more bookkeeping work to the extractor class if the information you need isn't already available in the Context().

fangism (approved):

verilog/tools/kythe/indexing_facts_tree_extractor.cc
MinaToma: Okay got that.

PR #468, opened 2020-09-18
Generate label prefix rule

### LINKED ISSUE #311: New lint rule for checking labels of generate blocks
Some style guides recommend prefixing all generate block labels in a design with an identifier like `g_*` or `gen_*`. Hence, a Verible lint rule that can check for such patterns in generate labels would be useful.

```systemverilog
if (ParamXY) begin : gen_xy
    ...
end
```

tgorochowik:
Fixes #311

### DIFF
--- a/verilog/analysis/checkers/generate_label_prefix_rule.cc
+++ b/verilog/analysis/checkers/generate_label_prefix_rule.cc
@@ -1,0 +1,92 @@
+// Copyright 2017-2020 The Verible Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "verilog/analysis/checkers/generate_label_prefix_rule.h"
+
+#include <vector>
+#include <string>
+
+#include "absl/strings/str_cat.h"
+#include "absl/strings/match.h"
+#include "common/analysis/citation.h"
+#include "common/analysis/lint_rule_status.h"
+#include "common/analysis/matcher/bound_symbol_manager.h"
+#include "common/analysis/matcher/matcher.h"
+#include "common/text/symbol.h"
+#include "common/text/syntax_tree_context.h"
+#include "verilog/analysis/descriptions.h"
+#include "verilog/analysis/lint_rule_registry.h"
+#include "verilog/CST/identifier.h"
+#include "verilog/CST/seq_block.h"
+
+namespace verilog {
+namespace analysis {
+
+using verible::GetStyleGuideCitation;
+
+// Register the lint rule
+VERILOG_REGISTER_LINT_RULE(GenerateLabelPrefixRule);
+
+absl::string_view GenerateLabelPrefixRule::Name() { return "generate-label-prefix"; }
+const char GenerateLabelPrefixRule::kTopic[] = "generate-statements";
+const char GenerateLabelPrefixRule::kMessage[] =
+    "All generate block labels must start with g_ or gen_";
+
+std::string GenerateLabelPrefixRule::GetDescription(
+    DescriptionType description_type) {
+  return absl::StrCat(
+      "Checks that every generate block label starts with g_ or gen_. See ",
+      GetStyleGuideCitation(kTopic), ".");
+}
+
+void GenerateLabelPrefixRule::HandleSymbol(
+    const verible::Symbol& symbol, const verible::SyntaxTreeContext& context) {
+  verible::matcher::BoundSymbolManager manager;
+  if (matcher_.Matches(symbol, &manager)) {
+    // Exclude case generate statements, as kGenerateBlock is generated for
+    // each 'case' item too.
+    if (context.IsInside(NodeEnum::kGenerateCaseItemList)) {
+      return;
+    }
+
+    for (const auto& child : SymbolCastToNode(symbol).children()) {
+      const verible::TokenInfo* label = nullptr;
+      switch (NodeEnum(SymbolCastToNode(*child).Tag().tag)) {
+        case NodeEnum::kBegin:
+          label = GetBeginLabelTokenInfo(*child);
+          break;
+        case NodeEnum::kEnd:
+          label = GetEndLabelTokenInfo(*child);
+          break;
+        default:
+          continue;
+      }
+
+      if (label != nullptr) {
+        if (!(absl::StartsWith(label->text(), "g_") ||
+              absl::StartsWith(label->text(), "gen_"))) {
+          violations_.insert(verible::LintViolation(*label, kMessage, context));
+        }
+      }
+    }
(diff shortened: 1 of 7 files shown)


verilog/analysis/checkers/generate_label_prefix_rule.h
```diff
@@ -0,0 +1,67 @@
...
+#include "common/text/syntax_tree_context.h"
+#include "verilog/CST/verilog_matchers.h"  // IWYU pragma: keep
+#include "verilog/analysis/descriptions.h"
+
+namespace verilog {
+namespace analysis {
+
+// GenerateLabelPrefixRule checks that all generate block labels start
+// with g_ or gen_
+
```
fangism, line 36: delete this blank line to attach this comment to the class

verilog/analysis/checkers/generate_label_prefix_rule.cc
```diff
@@ -0,0 +1,82 @@
...
+      "Checks that every generate block label starts with g_ or gen_. See ",
+      GetStyleGuideCitation(kTopic), ".");
+}
+
+void GenerateLabelPrefixRule::HandleSymbol(
+    const verible::Symbol& symbol, const verible::SyntaxTreeContext& context) {
+  verible::matcher::BoundSymbolManager manager;
+  if (matcher_.Matches(symbol, &manager)) {
+    if (context.IsInside(NodeEnum::kGenerateRegion)) {
+      const auto& matchingEnd = GetMatchingEnd(symbol, context);
```
fangism, line 58: const auto*

verilog/analysis/checkers/generate_label_prefix_rule.cc
```diff
@@ -0,0 +1,82 @@
...
+
+      std::vector<const verible::TokenInfo*> labels;
+      labels.push_back(GetBeginLabelTokenInfo(symbol));
+      labels.push_back(GetEndLabelTokenInfo(*matchingEnd));
```
fangism, line 62: if (matchingEnd != nullptr)

(otherwise not safe)

verilog/analysis/checkers/generate_label_prefix_rule_test.cc
```diff
@@ -0,0 +1,154 @@
...
+       "    end : gen_endif\n"
+       "  end : g_endini\n"
+       "end\n"
+       "endgenerate\nendmodule\n"},
+
+      {"module m;\n"
+       "generate\n"
+       "begin\n",
+       "  initial begin : ",
+       {SymbolIdentifier, "wrong"},
```
fangism, line 126: This is technically a statement, not a generate construct.  I don't think this rule should cover here, so I would expect this to be a negative test.  (Same with closely related tests)

verilog/analysis/checkers/generate_label_prefix_rule_test.cc
```diff
@@ -0,0 +1,154 @@
...
+       "\n"
+       "    if (1) begin : gen_if\n"
+       "    end : ",
+       {SymbolIdentifier, "also_wrong"},
+       "\n"
+       "  end : g_endini\n"
+       "end\n"
+       "endgenerate\nendmodule\n"},
+  };
```
fangism, line 147: Suggested test cases:

* loop generate statements (not inside `generate` and `endgenerate`), so `module m; if (x) ...`
  * with and without `begin` `end` labels
* conditional generate statements (not inside `generate` and `endgenerate`), so `module ; for(...`
  * with and without `begin` `end` labels
* case generate statement (doesn't have `begin` `end`, so a negative test)
* empty generate region (doesn't have `begin` `end`, so a negative test)
* always/initial/final constructs with begin/end labels at the module-item level (not inside generate/endgenerate), should be negative test

verilog/tools/lint/testdata/generate_begin_module.sv
```diff
@@ -1,6 +1,6 @@
 module generate_begin_module;
   generate
-    begin : block1
+    begin : gen_block1
```
fangism, line 3: :-)  it's getting to be more work to make test cases fail exactly one rule.  :-)

verilog/tools/lint/testdata/generate_label_prefix.sv
```diff
@@ -0,0 +1,7 @@
+module generate_label_prefix;
+  generate
```
fangism, line 2: See suggested test cases.  This should also fail without `generate` `endgenerate`.

fangism (changes requested):
Thank you for starting this!

verilog/analysis/checkers/generate_label_prefix_rule.cc
tgorochowik: Added this check, but is it even possible though? Can we have a `begin` statement without matching `end`?

tgorochowik:
@fangism thanks for the review. Could you look at my updates?

verilog/analysis/checkers/generate_label_prefix_rule_test.cc
fangism: This case is still missing (and is considered a generate label):

```systemverilog
module m;
if (x) begin : bad_label
end : bad_label
endmodule
```


Negative test:

```systemverilog
module m;
initial begin : OkNotAGenerateLabel
end : OkNotAGenerateLabel
endmodule
```

verilog/analysis/checkers/generate_label_prefix_rule.cc
```diff
@@ -0,0 +1,90 @@
...
+    DescriptionType description_type) {
+  return absl::StrCat(
+      "Checks that every generate block label starts with g_ or gen_. See ",
+      GetStyleGuideCitation(kTopic), ".");
+}
+
+void GenerateLabelPrefixRule::HandleSymbol(
+    const verible::Symbol& symbol, const verible::SyntaxTreeContext& context) {
+  verible::matcher::BoundSymbolManager manager;
+  if (matcher_.Matches(symbol, &manager)) {
```
fangism, line 56: Alternative to consider:

Instead of using a matcher on all begin/end pairs, there are only a few situations that need to be handled, and those can be a switch-case on the node's tag.  For each case, directly access the begin/end labels by descending into the subtree.  There are some functions in CST/statement.h that may help with conditional/loop generates.

That said, as long as the tests pass, I'll accept your choice of implementation.

fangism (changes requested):

tgorochowik:
@fangism thanks for your comments. I added the missing test cases and reworked the rule to match `kGenerateBlock` instead of `kBegin` and interate over its children and handle them in a switch statement, I am not sure if that's exactly what you meant, but I guess it does check fewer statements now.

verilog/analysis/checkers/generate_label_prefix_rule.cc
```diff
@@ -0,0 +1,92 @@
...
+    // Exclude case generate statements, as kGenerateBlock is generated for
+    // each 'case' item too.
+    if (context.IsInside(NodeEnum::kGenerateCaseItemList)) {
+      return;
+    }
+
+    for (const auto& child : SymbolCastToNode(symbol).children()) {
```
fangism, line 63: I was hoping to avoid a search loop, and use direct subtree access, however, looking at https://cs.opensource.google/verible/verible/+/master:verilog/parser/verilog.y, in particular the construction of `kGenerateBlock`, it seems the structure of this node is not quite consistent (the `kLabel` node has two forms).  Consider this a TODO for another time.

fangism (approved):
The tests look good, and that's the most important part.  Thanks again for working on this!

PR #505, opened 2020-09-28
Iterative definition resolving

### LINKED ISSUE #435: Kythe: Fix bug in resolving references inside classes and packages
Consider this case

```
class my_class;
    function int my_funtion()
        return x;
    endfunction
    int x;
endclass
```

```
package pkg;
    function int my_funtion()
        return pkg::x;
    endfunction
    int x;
endpackage
```

both cases will fail to find the definition of `x` in the function but it should be resolved correctly.

MinaToma:
Fixes #435


verilog/tools/kythe/kythe_facts.cc
```diff
@@ -53,6 +53,14 @@ std::string Signature::ToBase64() const {
   return absl::Base64Escape(ToString());
 }
 
+bool VName::operator==(const VName& other) const {
+  return this->ToString() == other.ToString();
+}
+
+bool VName::operator<(const VName& other) const {
+  return this->ToString() < other.ToString();
```
ivan444, line 61: (No need to fix for now)

While this comparison works, it's far from optimal. ToString() creates a new string just for a sake of comparison. While comparing two short strings is fast, allocating them isn't.

Doing a comparison by hand might make sense here.
Also, keep in mind that manually writing comparators is error prone (seems simple, but it's easy to introduce a mistake). Using tricks like with std::tie[1] might be useful.

[1] http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n3950.html

verilog/tools/kythe/kythe_facts.h
```diff
@@ -101,6 +104,50 @@ struct VName {
 
 std::ostream& operator<<(std::ostream&, const VName&);
 
+// Facts for kythe.
```
ivan444, line 107: Please expand the comment.

verilog/tools/kythe/testdata/bad_goal.sv
```diff
@@ -1,6 +0,0 @@
-// Copyright 2020 Google LLC.
-// SPDX-License-Identifier: Apache-2.0
-
-//- @bad_goal ref _
-module bad_goal;
```
ivan444, line 5: This was probably a test to check for a bad expectation (i.e., it expects the verifier to fail on it).

Might be useful, but it's also fine to delete it.

verilog/tools/kythe/testdata/more_testdata/function_call_hierarchy.sv
```diff
@@ -1,5 +1,4 @@
-// Copyright 2020 Google LLC.
-// SPDX-License-Identifier: Apache-2.0
```
ivan444, line 2: Please keep the copyright header. (in all files)

verilog/tools/kythe/kythe_facts_extractor.cc
```diff
@@ -26,20 +26,17 @@ namespace verilog {
...
-  // imports.
-  for (const IndexingFactNode& child : root.Children()) {
-    if (child.Value().GetIndexingFactType() != IndexingFactType::kPackage) {
-      continue;
-    }
-
-    VName package_vname = ExtractPackageDeclaration(child);
-    Visit(child, package_vname);
+  std::size_t number_of_extracted_facts = 0;
+  do {
```
ivan444, line 30: Add a comment above this loop. Describe the algorithm. It's not the most intuitive one.

ivan444 (changes requested):

verilog/tools/kythe/kythe_facts.cc
fangism: Agreed on tuple-comparison recommendation.  (At least leave a TODO.)

verilog/tools/kythe/kythe_facts.cc
```diff
@@ -64,5 +72,43 @@ std::ostream& operator<<(std::ostream& stream, const VName& vname) {
...
+  stream << fact.ToString();
+  return stream;
+}
+
+bool Edge::operator==(const Edge& other) const {
+  return this->ToString() == other.ToString();
+}
+
+bool Edge::operator<(const Edge& other) const {
+  return this->ToString() < other.ToString();
```
fangism, line 99: Same here, TODO ok.

verilog/tools/kythe/kythe_facts.h
```diff
@@ -101,6 +104,50 @@ struct VName {
...
+struct Fact {
+  Fact(const VName& vname, absl::string_view name, absl::string_view value)
+      : node_vname(vname), fact_name(name), fact_value(value) {}
+
+  bool operator==(const Fact& o) const;
```
fangism, line 112: The argument name in the header and definition should either match, or if it is obvious, can be omitted in the header.

verilog/tools/kythe/kythe_facts_extractor.cc
```diff
@@ -26,20 +26,17 @@ namespace verilog {
...
+    number_of_extracted_facts = facts_.size();
+    IndexingFactNodeTagResolver(root);
+  } while (number_of_extracted_facts != facts_.size());
+
+  for (const Fact& fact : facts_) {
```
fangism, line 35: Related to a suggestion from a while ago.  Consider isolating the collection of facts from printing, for example, accepting a listener function that may choose to collect or call some other function(s) on facts/edges.  For example a caller may wish to do some alternative sorting on facts/edges (than the ordering you've established).

No change needed for this PR.

verilog/tools/kythe/kythe_facts.h
fangism: In particular, document where this structure/definition comes from in relation to Kythe (links to source would be good too).

verilog/tools/kythe/kythe_facts_extractor.cc
```diff
@@ -289,37 +296,22 @@ void KytheFactsExtractor::ExtractDataTypeReference(
...
-        vertical_scope_resolver_.SearchForDefinition(anchor.Value());
-
-    if (port_vname_definition == nullptr) {
-      continue;
-    }
 
-    const VName port_vname_anchor = PrintAnchorVName(anchor);
-    GenerateEdgeString(port_vname_anchor, kEdgeRef, *port_vname_definition);
-  }
+  const VName module_instance_anchor = PrintAnchorVName(instance_name);
```
fangism, line 310: Recommend a few comments around these blocks, maybe a short explanation of each fact and edge w.r.t. an example.  "Given 'foo bar()', this creates facts about..."

(and generally other places too)

verilog/tools/kythe/kythe_facts_extractor.h
```diff
@@ -207,9 +208,13 @@ class KytheFactsExtractor {
   // Output stream for capturing, redirecting, testing and verifying the
   // output.
   std::ostream* stream_;
-};
 
-std::ostream& operator<<(std::ostream&, const KytheFactsPrinter&);
+  // Used to save all the generated facts uniquiely.
```
fangism, line 212: Uniquely

verilog/tools/kythe/kythe_facts_extractor.h
```diff
@@ -42,8 +43,12 @@ class KytheFactsPrinter {
   const IndexingFactNode& root_;
 };
 
+std::ostream& operator<<(std::ostream&, const KytheFactsPrinter&);
+
 // Responsible for traversing IndexingFactsTree and processing its different
 // nodes to produce kythe indexing facts.
+// Iteratively extracts facts and keeps running till no new facts are found in
```
fangism, line 50: s/till/until/

fangism, line 50: Iteratively: can you bound the number of iterations?  Is this truly a fixed-point calculation?

verilog/tools/kythe/testdata/more_testdata/reversed_definition.sv
```diff
@@ -0,0 +1,85 @@
...
+//- @class1 defines/binding _
+class class1;
+
+    //- @my_function defines/binding _
+    function int my_function();
+        //- @instance1 ref Instance1
+        //- @instance2 ref Instance2
+        //- @instance3 ref Instance3
+        //- @var1 ref Var1
+        return instance1::instance2::instance3::var1;
```
fangism, line 12: Like C++, `::` should be used for package/class-statics/nested-classes, and `.` for object members.  To access a global `var1`, here one would only write `class4::var1`.

Also if this was meant to be an invalid/negative test case, do indicate so in comments.

verilog/tools/kythe/testdata/more_testdata/reversed_definition.sv
```diff
@@ -0,0 +1,85 @@
...
+    //- @instance3 defines/binding Instance3
+    static class4 instance3 = new();
+endclass
+
+//- @class4 defines/binding MyClass4
+class class4;
+    //- @var1 defines/binding Var1
+    static int var1;
+endclass
+
```
fangism, line 39: Another general recommendation on these verification tests.  I'm not opposed to covering multiple features at a time in these tests, but if these have generally independent sections (where no references edges cross these boundaries), it would help to see this with some separator comments like `//===== description ======`.

fangism (changes requested):
What happened with the deleted test files?  Have their contents been re-distributed?  (Just need confirmation.)

verilog/tools/kythe/testdata/more_testdata/reversed_definition.sv
MinaToma: This is only a handling for the `instance.var` because member isn't extracted yet. 
Will leave a TODO.

verilog/tools/kythe/kythe_facts_extractor.cc
MinaToma: I totally agree with you. But let's forget about it in this PR and comments will be added while I work on this every PR.

verilog/tools/kythe/testdata/more_testdata/reversed_definition.sv
MinaToma: The test files are somehow self contained tests the extra facts are extracted only on the way but every file is intended to be a single test.

fangism (approved):

ivan444 (approved):

PR #520, opened 2020-10-02
Fix module-level package import indentation

### LINKED ISSUE #498: [indentation] un-indented package import statement in module declaration
**Test case**

```systemverilog
module alert_handler_esc_timer_fpv import alert_pkg::*; (
  // i/o ports
);
```

**Actual output**

```systemverilog
module alert_handler_esc_timer_fpv
import alert_pkg::*;
(
  // i/o ports
);
```

Ideally the `import` statement should be indented, e.g.

```systemverilog
module alert_handler_esc_timer_fpv
  import alert_pkg::*;
( 
  // i/o ports
);
```

ahmedkrmn:
Fixes #498.

**Input:**
```verilog
module foo import bar::*; (baz); endmodule
```
**Formatter output before:**
```verilog
module foo
import bar::*;
(
    baz
);
endmodule
```

**Formatter output after:**
```verilog
module foo
  import bar::*;
(
    baz
);
endmodule
```

### DIFF
--- a/verilog/formatting/formatter_test.cc
+++ b/verilog/formatting/formatter_test.cc
@@ -6963,6 +6963,30 @@ static constexpr FormatterTestCase kFormatterTestCases[] = {
      "    type TT = my_pkg::my_type2_t\n"
      ");\n"
      "endinterface\n"},
+     {// wildcard import package at module header
+     "module foo import bar::*; (baz); endmodule\n",
+     "module foo\n"
+     "  import bar::*;\n"
+     "(\n"
+     "    baz\n"
+     ");\n"
+     "endmodule\n"},
+     {// import package at module header
+     "module foo import bar::baz; (qux); endmodule\n",
+     "module foo\n"
+     "  import bar::baz;\n"
+     "(\n"
+     "    qux\n"
+     ");\n"
+     "endmodule\n"},
+     {// wildcard import mutiple packages at module header
+     "module foo import bar::*,baz::*; (qux); endmodule\n",
+     "module foo\n"
+     "  import bar::*, baz::*;\n"
+     "(\n"
+     "    qux\n"
+     ");\n"
+     "endmodule\n"},
     //{   // parameterized class with 'parameter_declaration' and MACRO
     //    "class foo #(parameter int a = 2,\n"
     //    "parameter int aaa = `MACRO);\n"
--- a/verilog/formatting/tree_unwrapper.cc
+++ b/verilog/formatting/tree_unwrapper.cc
@@ -946,6 +946,7 @@ void TreeUnwrapper::SetIndentationsAndCreatePartitions(
 
     // For the following constructs, always expand the view to subpartitions.
     // Add a level of indentation.
+    case NodeEnum::kPackageImportList:
     case NodeEnum::kPackageItemList:
     case NodeEnum::kInterfaceClassDeclaration:
     case NodeEnum::kCasePatternItemList:


verilog/formatting/formatter_test.cc
```diff
@@ -6963,6 +6963,14 @@ static constexpr FormatterTestCase kFormatterTestCases[] = {
      "    type TT = my_pkg::my_type2_t\n"
      ");\n"
      "endinterface\n"},
+     {// import package at module header
+     "module foo import bar::*; (baz); endmodule\n",
+     "module foo\n"
+     "    import bar::*;\n"
```
fangism, line 6969: Other suggested test cases:

* multiple import statements
* non-wildcard import: `import bar::baz;`

verilog/formatting/tree_unwrapper.cc
```diff
@@ -972,6 +972,7 @@ void TreeUnwrapper::SetIndentationsAndCreatePartitions(
     }
 
       // Add a level of grouping that is treated as wrapping.
+    case NodeEnum::kPackageImportList:
```
fangism, line 975: Placed here, this "wraps" which means indenting 4 spaces `style_.wrap_spaces`.
Let's indent this only 2 spaces, look for `style_.indent_spaces`.

(Test case will need a minor adjustment.)

fangism (changes requested):
Nicely done!  Some small suggestions.

verilog/formatting/tree_unwrapper.cc
ahmedkrmn: Thanks for the clarification. 
I thought it needed to match the indentation level of the module i/o ports. That's why I used an indentation of 4 spaces.

fangism: To be fair, there is disagreement over 2 vs. 4 spaces in some contexts.
In such cases, we've surfaced a few knobs to the users who need them.

See uses of `IndentationStyle` in here:
https://cs.opensource.google/verible/verible/+/master:verilog/formatting/format_style.h

and some of their corresponding uses here:
https://cs.opensource.google/verible/verible/+/master:verilog/tools/formatter/verilog_format.cc

No need to address this now, we can add such a style option and flag as a follow-up later.

verilog/formatting/formatter_test.cc
```diff
@@ -6963,6 +6963,30 @@ static constexpr FormatterTestCase kFormatterTestCases[] = {
...
+     ");\n"
+     "endmodule\n"},
+     {// wildcard import mutiple packages at module header
+     "module foo import bar::*,baz::*; (qux); endmodule\n",
+     "module foo\n"
+     "  import bar::*, baz::*;\n"
+     "(\n"
+     "    qux\n"
+     ");\n"
+     "endmodule\n"},
```
fangism, line 6989: request for one more test case with two separate imports, like:

```systemverilog
module foo
  import bar1::*;
  import bar2::quux;
...
```

fangism (changes requested):

verilog/formatting/formatter_test.cc
ahmedkrmn: This will cause a syntax error. Should I add this to the parser?

fangism: Oh, my mistake.  In that case, this PR is done.  I'll approve we'll take it from here.

fangism (approved):

verilog/formatting/formatter_test.cc
ahmedkrmn: @fangism Should I open a new issue for the separate imports?

fangism: I tested your change, and found that it already works for me, so I'll add tests to cover this myself.  Thanks though!

ahmedkrmn: If there is anything else that I should do, please let me know!

fangism: This one is done, just waiting for colleague to merge.

PR #522, opened 2020-10-05
Multi file extractor

### LINKED ISSUE #441: Kythe: extend current solution for kythe facts extraction to support multi file extraction and cross file referencing
Current solution for Kythe works only with single files but it should be extended to support cross file referencing like importing package from other files.

MinaToma:
Fixes #441


verilog/tools/kythe/verilog_kythe_extractor.cc
```diff
@@ -60,8 +61,7 @@ static int ExtractOneFile(absl::string_view content,
     std::cout << std::endl
               << (!parse_ok ? " (incomplete due to syntax errors): " : "")
               << std::endl;
-
-    std::cout << verilog::kythe::KytheFactsPrinter(facts_tree) << std::endl;
+    multi_file_kythe_facts_extractor.ExtractKytheFacts(facts_tree);
```
ivan444, line 64: Here you no longer print the tree. Is that intended?

verilog/tools/kythe/scope_resolver.h
```diff
@@ -168,7 +175,28 @@ class FlattenedScopeResolver {
   void MapSignatureToScopeOfSignature(const Signature& signature,
                                       const Signature& other_signature);
 
+  ScopeContext& GetScopeContext() { return scope_context_; }
+
  private:
+  // Return the global scope of the current scope_resolver.
+  const Scope* GetGlobalScope() const;
```
ivan444, line 182: Can this ever be null? If not, consider returning a const reference.

verilog/tools/kythe/multi_file_kythe_facts_extractor.h
```diff
@@ -0,0 +1,47 @@
...
+#include "verilog/tools/kythe/scope_resolver.h"
+
+namespace verilog {
+namespace kythe {
+
+// Responsible for extracting Kythe facts from multi-files in which they may
+// have cross-files references.
+// The given trees (files) should be ordered by dependency between files in
+// which the first one doesn't depend on other files.
+class MultiFileKytheFactsExtractor {
```
ivan444, line 30: MultiFileKytheFactsExtractor makes the whole system difficult to use (the sole existence of two extractors is confusing).

Can you extend KytheFactsExtractor to support handling multiple files and remove MultiFileKytheFactsExtractor (and this file)?

ivan444 (changes requested):

verilog/tools/kythe/scope_resolver.h
MinaToma: It can be null in the first pass for a single file as the global will be under construction and not saved to the scopes yet.

verilog/tools/kythe/verilog_kythe_extractor.cc
MinaToma: Yes we now have two different entries `KytheFactsPrinter` for processing single files which won't consider any dependencies and `MultiFileFactsExtractor` which is for multi-files and cross-referencing in files.

verilog/tools/kythe/kythe_facts_extractor.cc
```diff
@@ -558,108 +547,86 @@ VName KytheFactsExtractor::ExtractClassInstances(
...
-      return;
-    }
 
-    const VName class_anchor = CreateAnchor(containing_block_name);
-    CreateEdge(class_anchor, kEdgeRef, *containing_block_vname);
-
-    definition_signature = containing_block_vname->signature;
+  // Extract the list of reference_names.
+  std::vector<std::string> names;
+  for (const Anchor anchor : anchors) {
```
ivan444, line 613: Make it a reference to avoid unnecessary copy

verilog/tools/kythe/kythe_facts_extractor.h
```diff
@@ -223,6 +222,10 @@ class KytheFactsExtractor {
   std::set<Edge> edges_;
 };
 
+// Returns the file path from the given tree which this tree contains facts
```
ivan444, line 225: Can you rephrase the comment? I find it difficult to read.

verilog/tools/kythe/kythe_facts_extractor.cc
```diff
@@ -725,16 +692,18 @@ std::string GetFilePathFromRoot(const IndexingFactNode& root) {
...
-  return stream;
-}
+std::ostream& KytheFactsPrinter::Print(std::ostream& stream,
+                                       const IndexingFactNode& root) {
+  if (!files_scope_resolvers_.empty()) {
+    files_scope_resolvers_.push_back(
+        ScopeResolver(files_scope_resolvers_.back()));
+  } else {
+    files_scope_resolvers_.push_back(ScopeResolver(nullptr));
+  }
```
ivan444, line 702: I wouldn't expect the Print() function to alter the state of the object (unless it's related to printing).

Can you move this code to another function? Simply call the print for the whole tree and not file-by-file.

ivan444 (changes requested):

verilog/tools/kythe/verilog_kythe_extractor.cc
```diff
@@ -92,6 +93,8 @@ verible-verilog-kythe-extractor files...)");
   const auto args = verible::InitCommandLine(usage, &argc, &argv);
 
   int exit_status = 0;
+  std::vector<verilog::kythe::IndexingFactNode> indexing_facts_trees;
```
fangism, line 96: Ok for now, but in follow-ups consider managing multiple files' facts trees inside a class/structure intended for multi-file management.  In the long run, single-file operation should be a degenerate case of multi-file operation.

fangism (changes requested):
@ivan444 @hzeller Until external kythe verifier testing works, we will need to accommodate new test additions when we import PRs like this.

verilog/tools/kythe/kythe_facts_extractor.cc
```diff
@@ -558,108 +547,86 @@ VName KytheFactsExtractor::ExtractClassInstances(
...
+    // Search for member hierarchy in the scopes.
+    const std::vector<const VName*> defintion_vname =
+        scope_resolver_->SearchForDefinitions(
+            {package_name_anchor.Value(), imported_item_name.Value()});
+
+    // Loop over the found definitions and create kythe facts.
+    for (size_t i = 0; i < defintion_vname.size(); i++) {
+      const VName current_anchor = CreateAnchor(anchors[i]);
+      if (i == 0) {
+        CreateEdge(current_anchor, kEdgeRefImports, *defintion_vname[i]);
```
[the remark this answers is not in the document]
MinaToma: It's not guaranteed to find a definition so this loop is safe as is.

kythe-browse-multi-file.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash -e
```
ivan444, line 1: Do we need this script? Can't all be handled by the existing one?

verilog/tools/kythe/kythe_facts_extractor.cc
ivan444: This is not yet done. David made a similar comment.

verilog/tools/kythe/kythe_facts_extractor.cc
```diff
@@ -726,9 +693,20 @@ std::string GetFilePathFromRoot(const IndexingFactNode& root) {
 }
 
 std::ostream& KytheFactsPrinter::Print(std::ostream& stream) const {
-  KytheFactsExtractor kythe_extractor(GetFilePathFromRoot(root_), &stream);
+  std::vector<ScopeResolver> scope_resolvers;
+
+  for (const IndexingFactNode& root : trees_) {
```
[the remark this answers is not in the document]
ivan444: +1
Please move this out of the `Print` function.

verilog/tools/kythe/scope_resolver.h
```diff
@@ -168,7 +175,28 @@ class FlattenedScopeResolver {
   void MapSignatureToScopeOfSignature(const Signature& signature,
                                       const Signature& other_signature);
 
- private:
+  ScopeContext& GetScopeContext() { return scope_context_; }
```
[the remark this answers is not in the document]
MinaToma: This is changed during traversal so it has to bee mutable.

verilog/tools/kythe/verilog_kythe_extractor.cc
ivan444: David, can you expand on your comment? I want to make sure you and I don't send different messages. This looks OK to me.

CST --> Middle layer works file-by-file. Middle layer --> Kythe takes all files at once.

Maybe it would help to encapsulate all in a single class?

ivan444 (changes requested):

MinaToma:
@ivan @fangism concerning how to deal with multi-files, my intention was to have `KytheFactsExtractor` as the one responsible for extracting kythe facts for single files and having another class `MultiFileKytheFactsExtractor` which is responsible for all the logic about extracting multi-files like creating the scopes and whatever can be modified later can be addressed in that class.

ivan444:
> @ivan444 @fangism concerning how to deal with multi-files, my intention was to have `KytheFactsExtractor` as the one responsible for extracting kythe facts for single files and having another class `MultiFileKytheFactsExtractor` which is responsible for all the logic about extracting multi-files like creating the scopes and whatever can be modified later can be addressed in that class.

As David also mentioned, why have two? Extracting a single file is just a simplification of extracting multiple files. From user perspective it's much better to have a single class (easier to use).

verilog/tools/kythe/verilog_kythe_extractor.cc
fangism: For example, encapsulation within a single structure/class would be a good place to establish a mapping between root-level symbol definitions and their files of origin.  Why search through each file's root scope when a top-level map could give you the same information more efficiently?  That is to say that the root-level symbol table (or equivalent) should handle multiple files.

fangism:
> @ivan444 @fangism concerning how to deal with multi-files, my intention was to have `KytheFactsExtractor` as the one responsible for extracting kythe facts for single files and having another class `MultiFileKytheFactsExtractor` which is responsible for all the logic about extracting multi-files like creating the scopes and whatever can be modified later can be addressed in that class.

Ok, I'll look for that implementation in a follow-up PR.

kythe-multi-file-verifier.sh
```diff
@@ -0,0 +1,37 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+usage() {
+  cat <<EOF
+
+Extracts Kythe facts from the given verilog file and runs Kythe verifier on the produced facts.
```
fangism, line 19: ... on the given Verilog files ...

kythe-multi-file-verifier.sh
```diff
@@ -0,0 +1,37 @@
...
+EOF
+}
+
+set -o pipefail
+KYTHE_BINDIR="/opt/kythe/tools"
+KYTHE_OUT="./kythe-out"
+
+# The files are expected to be self-contained single-file test cases.
+VERILOG_MULTI_FILE_TEST_FILES="./verilog/tools/kythe/testdata/more_testdata/multi_file_test/*.sv"
```
fangism, line 28: This is fine for now, but in the future, we'll want a way to test multiple independent groups of files.  @ivan444 has some ideas about this.

verilog/tools/kythe/scope_resolver.h
```diff
@@ -151,12 +155,19 @@ class VerticalScopeResolver : public verible::AutoPopStack<Scope*> {
...
 // }
 // this way definitions can be found using the signatures.
-class FlattenedScopeResolver {
+class ScopeResolver {
  public:
-  // Searches for a VName with the given name in the scope with the given
-  // signature.
-  const VName* SearchForVNameInScope(const Signature& signature,
-                                     absl::string_view name) const;
+  explicit ScopeResolver(const ScopeResolver* previous_file_scope_resolver)
```
fangism, line 160: // Is this allowed to be nullptr? if yes, what does that mean?  if not `ABSL_DIE_IF_NULL`, or consider taking the address of a `const&` argument as a signal to the reader.

verilog/tools/kythe/scope_resolver.cc
```diff
@@ -15,12 +15,20 @@
...
+bool ScopeMemberItem::operator==(const ScopeMemberItem& other) const {
+  return this->vname == other.vname;
+}
+bool ScopeMemberItem::operator<(const ScopeMemberItem& other) const {
+  return this->vname < other.vname;
+}
+
 void Scope::AddMemberItem(const ScopeMemberItem& member_item) {
-  members_.push_back(member_item);
+  members_.insert(member_item);
```
fangism, line 31: Note: `insert` returns a `pair<iterator, bool>` where the bool is true if insertion succeeded, false if rejected because the element with key already exists.  Should this be signaled to the caller in a return value?  What happens if this code encounters a duplicate definition/declaration?  It should work gracefully, taking a best guess if needed, but might be worth `LOG`-ing.  (And this design choice should be documented.)

verilog/tools/kythe/scope_resolver.h
```diff
@@ -58,7 +62,7 @@ class Scope {
   Signature signature_;
 
   // list of the members inside this scope.
-  std::vector<ScopeMemberItem> members_;
+  std::set<ScopeMemberItem> members_;
```
fangism, line 65: Yes!

verilog/tools/kythe/scope_resolver.h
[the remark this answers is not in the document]
fangism: From a class-design perspective, try to avoid this if possible because this may invite the caller to violate class-invariant properties (like relationships between members).  It takes more analysis to conclude whether or not this is possible.  If the number of mutating operations on `ScopeContext` is sufficiently small, consider expose only those mutable operations.

In this case, I would name this `GetMutableScopeContext()` to convey intent more clearly where it is called.

verilog/tools/kythe/scope_resolver.cc
```diff
@@ -50,16 +57,97 @@ const VName* VerticalScopeResolver::SearchForDefinition(
...
+
+const Scope* ScopeResolver::SearchForScope(const Signature& signature) const {
   const auto scope = scopes_.find(signature);
-  if (scope == scopes_.end()) {
-    return nullptr;
+  if (scope != scopes_.end()) {
+    return &scope->second;
+  }
+
+  // Try to find the definition in the previous files' scopes.
```
fangism, line 142: // Comment here that this is a linear-time search over files.

verilog/tools/kythe/scope_resolver.h
```diff
@@ -189,6 +223,10 @@ class FlattenedScopeResolver {
   //   "pkg2": ["my_fun", "my_class"]
   // }
   std::map<Signature, Scope> scopes_;
+
+  // Pointer to the previous file's discovered scopes (if a previous file
+  // exists). This is used for definition finding in cross-file referencing.
```
fangism, line 228: // This forms a null-terminated singly-linked list across files.

verilog/tools/kythe/scope_resolver_test.cc
```diff
@@ -91,27 +100,39 @@ TEST(FlattenedScopeTests, SearchForDefinition) {
...
-  const VName* vname4 =
-      flattened_scope_resolver.SearchForVNameInScope(signatures[2], names[1]);
-  EXPECT_EQ(vname4, nullptr);
+  {
+    Scope global_scope(signatures[7]);
+    ScopeContext::AutoPop p1(&scope_resolver.GetScopeContext(), &global_scope);
+    scope_resolver.AddDefinitionToScopeContext(vnames[0]);
+    scope_resolver.MapSignatureToScope(signatures[0], scope);
+    {
+      const std::vector<const VName*> vname =
```
fangism, line 109: `vnames =` (plural)

verilog/tools/kythe/scope_resolver_test.cc
```diff
@@ -91,27 +100,39 @@ TEST(FlattenedScopeTests, SearchForDefinition) {
...
+    }
+    {
+      const std::vector<const VName*> vname =
+          scope_resolver.SearchForDefinitions({names[0], names[3]});
+      EXPECT_EQ(vname.size(), 1);
+    }
+    {
+      const std::vector<const VName*> vname =
+          scope_resolver.SearchForDefinitions({names[2], names[1]});
+      EXPECT_EQ(vname.size(), 0);
```
fangism, line 129: EXPECT_TRUE(vnames.empty());

verilog/tools/kythe/scope_resolver_test.cc
```diff
@@ -91,27 +100,39 @@ TEST(FlattenedScopeTests, SearchForDefinition) {
...
+    {
+      const std::vector<const VName*> vname =
+          scope_resolver.SearchForDefinitions({names[0], names[2]});
+      EXPECT_EQ(vname.size(), 2);
+      EXPECT_EQ(vname[1]->signature, signatures[2]);
+    }
+    {
+      const std::vector<const VName*> vname =
+          scope_resolver.SearchForDefinitions({names[0], names[3]});
+      EXPECT_EQ(vname.size(), 1);
```
fangism, line 124: Can you assert anything specific about `vnames[0]`?

verilog/tools/kythe/scope_resolver_test.cc
```diff
@@ -134,123 +155,152 @@ TEST(FlattenedScopeTests, SearchForNestedDefinition) {
...
+    {
+      const Scope* found_scope = scope_resolver.SearchForScope(signatures[5]);
+      EXPECT_NE(found_scope, nullptr);
+    }
+    {
+      const Scope* found_scope = scope_resolver.SearchForScope(signatures[6]);
+      EXPECT_EQ(found_scope, nullptr);
+    }
+    {
+      const std::vector<const VName*> vname =
```
fangism, line 188: `vnames` (plural, vector)

fangism (changes requested):

verilog/tools/kythe/kythe_facts_extractor.h
```diff
@@ -34,13 +34,14 @@ namespace kythe {
 // Usage: stream << KytheFactsPrinter(*tree_root);
 class KytheFactsPrinter {
  public:
-  explicit KytheFactsPrinter(const IndexingFactNode& root) : root_(root) {}
+  explicit KytheFactsPrinter(const std::vector<IndexingFactNode>& root)
```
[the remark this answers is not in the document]
MinaToma: I will consider this because more modification will change when we extracts includes.

verilog/tools/kythe/scope_resolver.cc
MinaToma: The members here can be duplicated only because the multi-iterations but they all map to the same member item.

verilog/tools/kythe/scope_resolver.h
MinaToma: This is intended to be nulltr in case it is the first files which has no dependencies on other files.

MinaToma:
Let's ignore the implementation details of multi-files now because they will change in the coming PR for includes.

kythe-multi-file-verifier.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash -e
```
ivan444, line 1: Do we need this script? Can the other one handle both multi and single files?

verilog/tools/kythe/kythe_facts_extractor.cc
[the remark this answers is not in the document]
ivan444: I'm approving this PR even though this is not done yet. The code for handling includes will address this problem.

ivan444 (approved):

kythe-multi-file-verifier.sh
MinaToma: for now, yes but in the coming PR this should be the general case.

fangism (approved):
Understood that you have some upcoming follow-up changes.

PR #524, opened 2020-10-06
Add numeric format string style checker

### LINKED ISSUE #329: Check for style-compliant format-string numeric specifiers
According to our style guide, we should always display numbers in the following manner:

```systemverilog
// good examples
$display("0x%0x", some_hex_value);
$display("0b%0b", some_binary_value);
$display("'h%0h", some_hex_value);
$display("'b%0b", some_binary_value);  // also acceptable
$display("%0d", some_decimal_value);  // no prefix wanted
```

```systemverilog
// bad examples
$display("%0x", some_hex_value);
$display("%0b", some_binary_value);
$display("0d%0d", some_decimal_value);
$display("'d%0d", some_decimal_value);
```

See:
https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md#number-formatting

This includes any task function that is `$display`-like, including `$sformat`.

ldalek-antmicro:
Fixes #329

### DIFF
--- a/verilog/analysis/checkers/numeric_format_string_style_rule.cc
+++ b/verilog/analysis/checkers/numeric_format_string_style_rule.cc
@@ -1,0 +1,167 @@
+// Copyright 2017-2020 The Verible Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "verilog/analysis/checkers/numeric_format_string_style_rule.h"
+
+#include <set>
+#include <string>
+
+#include "absl/strings/match.h"
+#include "absl/strings/str_cat.h"
+#include "absl/strings/str_split.h"
+#include "common/analysis/citation.h"
+#include "common/analysis/lint_rule_status.h"
+#include "common/analysis/token_stream_lint_rule.h"
+#include "common/strings/naming_utils.h"
+#include "common/text/token_info.h"
+#include "verilog/analysis/descriptions.h"
+#include "verilog/analysis/lint_rule_registry.h"
+#include "verilog/parser/verilog_lexer.h"
+#include "verilog/parser/verilog_token_classifications.h"
+#include "verilog/parser/verilog_token_enum.h"
+
+namespace verilog {
+namespace analysis {
+
+using verible::GetStyleGuideCitation;
+using verible::LintRuleStatus;
+using verible::LintViolation;
+using verible::TokenInfo;
+using verible::TokenStreamLintRule;
+
+// Register the lint rule
+VERILOG_REGISTER_LINT_RULE(NumericFormatStringStyleRule);
+
+absl::string_view NumericFormatStringStyleRule::Name() { return "numeric-format-string-style"; }
+const char NumericFormatStringStyleRule::kTopic[] = "number-formatting";
+const char NumericFormatStringStyleRule::kMessage[] =
+  "Formatting string must contain proper style-compilant numeric specifiers.";
+
+std::string NumericFormatStringStyleRule::GetDescription(
+    DescriptionType description_type) {
+  return absl::StrCat("Checks that string literals with numeric format specifiers "
+    "have proper prefixes for hex and bin values and no prefixes for decimal values.");
+}
+
+template<typename T> class TD;
+
+void NumericFormatStringStyleRule::CheckAndReportViolation(
+    const TokenInfo& token, size_t pos, size_t len,
+    std::initializer_list<unsigned char> prefixes) {
+  const absl::string_view text(token.text());
+
+  // Check for prefix
+  if (pos >= 2 && (text[pos-2] == '0' || text[pos-2] == '\'')) {
+    const auto ch = text[pos-1];
+    if (std::none_of(prefixes.begin(), prefixes.end(),
+        [&ch](const char& _ch) { return _ch == ch; })) {
+      // Report whole prefix with a "0" or "'"
+      violations_.insert(LintViolation(TokenInfo(token.token_enum(),
+                                                 text.substr(pos - 2, len + 2)),
+                                       kMessage));
+    }
+  } else {
+    // Report just radix
+    violations_.insert(LintViolation(TokenInfo(token.token_enum(),
+                                               text.substr(pos, len)),
+                                     kMessage));
+  }
+}
+
+void NumericFormatStringStyleRule::HandleToken(const TokenInfo& token) {
+  const auto token_enum = static_cast<verilog_tokentype>(token.token_enum());
+  const absl::string_view text(token.text());
+
+  // TODO(fangism): Refactor this. Right now it's a copy-paste from macro_name_style_rule.cc
+  if (IsUnlexed(verilog_tokentype(token.token_enum()))) {
+    // recursively lex to examine inside macro definition bodies, etc.
+    VerilogLexer lexer(text);
+    while (true) {
+      const TokenInfo& subtoken(lexer.DoNextToken());
+      if (subtoken.isEOF()) break;
+      HandleToken(subtoken);
+    }
+    return;
+  }
+
+  if (token_enum != TK_StringLiteral)
+    return ;
+
+  for (size_t pos = 0 ; pos < text.size() ;) {
+    const auto& ch = text[pos];
+
+    // Skip ordinary characters
+    if (ch != '%') {
+      ++pos;
+      continue ;
+    }
+
+    // Examine formatting directive
+    for (size_t itr = pos + 1 ; itr < text.size() ; ++itr) {
+      // Skip field with display data size
+      if (absl::ascii_isdigit(text[itr]))
+        continue ;
+
+      const auto radix = text[itr];
+      const auto len = itr - pos + 1;
+
+      // Format radix
+      switch (radix) {
+        // binary value
+        case 'b':
+        case 'B': {
+          CheckAndReportViolation(token, pos, len, { 'b' });
+          break;
+        }
+
+        // hexdecimal value
+        case 'h':
+        case 'H':
+        case 'x':
+        case 'X': {
+          CheckAndReportViolation(token, pos, len,  { 'h', 'x' });
+          break;
+        }
+
+        // decimal value
+        case 'd':
+        case 'D': {
+          // Detect prefixes starting with "0" and "'"
+          if (pos >= 2 && (text[pos-2] == '0' || text[pos-2] == '\'')) {
+            // Report whole string with a "0" or "'"
+            violations_.insert(LintViolation(TokenInfo(token.token_enum(),
+                                                       text.substr(pos - 2, len + 2)),
+                                             kMessage));
+          }
+          break;
+        }
+
+        default:
+          break;
+      }
+
+      // continue with the following character
+      pos = itr + 1;
+      break;
+    }
+  }
+
+}
+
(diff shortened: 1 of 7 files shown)


verilog/analysis/checkers/numeric_format_string_style_rule.cc
```diff
@@ -0,0 +1,143 @@
...
+using verible::LintRuleStatus;
+using verible::LintViolation;
+using verible::TokenInfo;
+using verible::TokenStreamLintRule;
+
+// Register the lint rule
+VERILOG_REGISTER_LINT_RULE(NumericFormatStringStyleRule);
+
+absl::string_view NumericFormatStringStyleRule::Name() { return "numeric-format-string-style"; }
+const char NumericFormatStringStyleRule::kTopic[] = "format-string";
```
fangism, line 47: topic string is "number-formatting"

verilog/analysis/checkers/numeric_format_string_style_rule.cc
```diff
@@ -0,0 +1,143 @@
...
+          break;
+        }
+
+        default:
+          break;
+      }
+
+      // TODO(fangism): Different message for different formatting string violations?
+      if (violation)
+        violations_.insert(LintViolation(token, kMessage));
```
fangism, line 128: Instead of the whole token, can we synthesize a token whose .text() spans the offending format specifier (and its prefix where applicable).  `string_view::substr()` should do it.  This would give a very precise location/byte-offset.

verilog/analysis/checkers/numeric_format_string_style_rule_test.cc
```diff
@@ -0,0 +1,75 @@
...
+      {"module test; initial $display(\"%0d\", some_decimal_value); endmodule"}, // no prefix wanted
+      {"module test; initial $display(", {kToken, "\"0d%0d\""}, ", some_dec_value); endmodule"},
+      {"module test; initial $display(", {kToken, "\"'d%0d\""}, ", some_dec_value); endmodule"},
+      {"module test; initial $display(\"Value: %0d\", some_decimal_value); endmodule"}, // no prefix wanted
+      {"module test; initial $display(", {kToken, "\"Value: 0d%0d\""}, ", some_dec_value); endmodule"},
+      {"module test; initial $display(", {kToken, "\"Value: 'd%0d\""}, ", some_dec_value); endmodule"},
+
+      // mixed
+      {"module test; initial $display(", {kToken, "\"0x%0x, %x\""}, ", some_hex_value); endmodule"},
+  };
```
fangism, line 69: Suggested test cases:

* single string literal with multiple violations inside
* string literal outside of a $display call (one case enough), like `parameter string foo = "...";`
* string literal inside $sformat call (one case enough)

Bonus:

* string literals inside macro definition bodies, and macro call arguments

To do this you'll want to to recursively lex `token.text()` like done here:
https://cs.opensource.google/verible/verible/+/master:verilog/analysis/verilog_equivalence.cc;l=92

fangism (changes requested):
Thanks for working on this!

verilog/analysis/checkers/numeric_format_string_style_rule.cc
fangism: ... and support finding multiple violations in unique locations within one string literal.  :)

verilog/analysis/checkers/numeric_format_string_style_rule_test.cc
```diff
@@ -0,0 +1,112 @@
...
+      {"module test; initial $display(", {kToken, "\"Value: 0d%0d\""}, ", some_dec_value); endmodule"},
+      {"module test; initial $display(", {kToken, "\"Value: 'd%0d\""}, ", some_dec_value); endmodule"},
+
+      // invalid prefix
+      {"module test; initial $display(", {kToken, "\"Value: 0b%0x\""}, ", some_hex_value); endmodule"},
+      {"module test; initial $display(", {kToken, "\"Value: 0x%0b\""}, ", some_binary_value); endmodule"},
+
+      // TODO(fangism): Are those ok? Or just "0d" and "'d" are invalid prefixes for decimal values?
+      //{"module test; initial $display(", {kToken, "\"Value: 0x%0d\""}, ", some_dec_value); endmodule"},
+      //{"module test; initial $display(", {kToken, "\"Value: 0b%0d\""}, ", some_dec_value); endmodule"},
```
ldalek-antmicro, line 73: @fangism How about those tests?

verilog/analysis/checkers/numeric_format_string_style_rule.cc
```diff
@@ -0,0 +1,143 @@
...
+        // decimal value
+        case 'd':
+        case 'D': {
+          // Innocent until proven guilty
+          violation = false;
+
+          // TODO(fangism): Detecting only 4 prefixes.
+          //   Should it be more generic?, e.g.
+          //   if (pos >= 2 && absl::ascii_isalpha(text[pos-1]))
+          //     violation = true
```
ldalek-antmicro, line 113: @fangism And this comment?

verilog/analysis/checkers/numeric_format_string_style_rule_test.cc
fangism: Good suggestions, let's catch these scenarios as report them as mismatches.

verilog/analysis/checkers/numeric_format_string_style_rule.cc
fangism: I suggest limiting it to a handful of common radix prefix cases.

verilog/analysis/checkers/numeric_format_string_style_rule_test.cc
```diff
@@ -0,0 +1,112 @@
...
+      {"module test; initial $display(", {kToken, "\"Value: 0x%0x %0x %0x\""}, ", hex1, hex2, hex3); endmodule"},
+      {"module test; initial $display(", {kToken, "\"Value: %0x 'h%0x %0x\""}, ", hex1, hex2, hex3); endmodule"},
+      {"module test; initial $display(", {kToken, "\"Value: %0x %0x %0x\""}, ", hex1, hex2, hex3); endmodule"},
+
+      {"module test; initial $display(", {kToken, "\"Value: 0b%0b %0b %0b\""}, ", bin1, bin2, bin3); endmodule"},
+      {"module test; initial $display(", {kToken, "\"Value: %0b 'b%0b %0b\""}, ", bin1, bin2, bin3); endmodule"},
+      {"module test; initial $display(", {kToken, "\"Value: %0b %0b %0b\""}, ",  bin1, bin2, bin3); endmodule"},
+
+      {"module test; initial $display(", {kToken, "\"Value: 0d%0d 'D%0d %0d\""}, ", dec1, dec2, dec3); endmodule"},
+      {"module test; initial $display(", {kToken, "\"Value: %0d 0D%0d 'd%0d\""}, ", dec1, dec2, dec3); endmodule"},
```
fangism, line 85: These expected violations should be as narrow as possible, also allowing for multiple violations reported per string literal.

verilog/analysis/checkers/numeric_format_string_style_rule_test.cc
```diff
@@ -0,0 +1,117 @@
...
+      {"module test; initial $display(\"0x%0x\", some_hex_value); endmodule"},
+      {"module test; initial $display(\"'h%0h\", some_hex_value); endmodule"},
+      {"module test; initial $display(\"Value: 0x%0x\", some_hex_value); endmodule"},
+      {"module test; initial $display(\"Value: 'x%0x\", some_hex_value); endmodule"},
+      {"module test; initial $display(\"Value: ", {kToken, "0X%0x"}, "\", some_hex_value); endmodule"},
+      {"module test; initial $display(\"Value: ", {kToken, "'X%0x"}, "\", some_hex_value); endmodule"},
+      {"module test; initial $display(\"Value: ", {kToken, "%0x"}, "\", some_hex_value); endmodule"},
+
+      // TODO(fangism): Guessing that "'h0%0x" is okay?
+      //{"module test; initial $display(\"'h%0x\", some_hex_value); endmodule"},
```
ldalek, line 50: @fangism Guessing that `'h%0x` is okay?

fangism: Yes

verilog/analysis/checkers/numeric_format_string_style_rule.cc
```diff
@@ -0,0 +1,141 @@
...
+          if (pos >= 2 && (text[pos-2] == '0' || text[pos-2] == '\'')) {
+            // Report whole string with a "0" or "'"
+            violations_.insert(LintViolation(TokenInfo(token.token_enum(),
+                                                       text.substr(pos - 2, len + 2)),
+                                             kMessage));
+          }
+          break;
+        }
+
+        // TODO(fangism): Unknown radix, report it somehow?
```
fangism, line 123: Ok to ignore, we can't be 100% certain it is a radix prefix because it is inside a format string as plain text.

verilog/analysis/checkers/numeric_format_string_style_rule_test.cc
fangism: New test cases look good.  Do you plan work the macro definition bodies and call arguments, or shall we take this as-is and save that for later?

fangism (changes requested):
Last task:

Can you add a test case to verilog/tools/lint/testdata and test rule for it to the verilog/tools/lint/BUILD?

ldalek-antmicro:
I'll tackle this 'bonus' test case.

I guess it should be handled like here: https://github.com/google/verible/blob/master/verilog/analysis/checkers/macro_name_style_rule.cc#L63 + I should make use of `ShouldRecursivelyAnalyzeToken()`?

fangism:
> I'll tackle this 'bonus' test case.
> 
> I guess it should be handled like here: https://github.com/google/verible/blob/master/verilog/analysis/checkers/macro_name_style_rule.cc#L63 + I should make use of `ShouldRecursivelyAnalyzeToken()`?

Yes, exactly like that in macro_name_style_rule, which is simple enough.  Thanks!

verilog/analysis/checkers/numeric_format_string_style_rule.cc
```diff
@@ -0,0 +1,170 @@
...
+  return absl::StrCat("Checks that string literals with numeric format specifiers "
+    "have proper prefixes for hex and bin values and no prefixes for decimal values.");
+}
+
+void NumericFormatStringStyleRule::HandleToken(const TokenInfo& token) {
+  const auto token_enum = static_cast<verilog_tokentype>(token.token_enum());
+  const absl::string_view text(token.text());
+
+  // TODO(fangism): Move it to some utils? It's copy-paste from macro_name_style_rule.cc
+  //     And maybe we could use ShouldRecursivelyAnalyzeToken().
```
fangism, line 62: `ShouldRecursivelyAnalyzeToken` does little more than wrap around `IsLexed`, and just exists to conform to a `std::function` interface where it is used, it may as well just be replaced with an anonymous lambda.  And it's kind of self-documenting.

If I were to refactor this pattern out, it would be something like:

```c++
  if (IsUnlexed(verilog_tokentype(token.token_enum()))) {
    RecursivelyLexTokens([this](const TokenInfo& subtoken){ HandleToken(subtoken); });
    return;
  }
``` 

where

```c++
  // Lexes a token into subtokens while applying a generic function 'func'.
  void RecursivelyLexTokens(const std::function<void(const TokenInfo&)>& func) {
    VerilogLexer lexer(text);
    while (true) {
      const TokenInfo& subtoken(lexer.DoNextToken());
      if (subtoken.isEOF()) break;
      func(subtoken);
    }
  }
```

could be defined in verilog/parser/lexer.cc as a public free-standing function.
How does this look to you?

fangism (changes requested):
Suggested a refactoring to answer your comment.  
If you want to skip it and leave it for me, you can keep the TODO.  Let me know.

verilog/analysis/checkers/numeric_format_string_style_rule.cc
ldalek-antmicro: I've updated a comment. Leaving refactoring for you.

fangism (approved):

PR #529, opened 2020-10-11
Preserving the module name and port names during obfuscation

hchsiao:
I've done some modifications to meet our application requirement.
We obfuscate everything except module name and port names before delivering our IP to customers since the customers should at least be able to recognize our interface when integrating our product into their design.

I think this could be useful for other use cases and it may save someone's time, so I issued this PR.
This addresses #530.

(Note: this is my first time submitting to a public repo, any sorts of suggestions for improvement are welcomed. Thanks, guys!)


hchsiao:
@googlebot I signed it!

fangism (commented):
Thank you for your contribution, this looks like a useful option to have.  I'll review this soon with some suggestions.

fangism:
The high level approach of populating the obfuscation map with preserved identifiers is good.  It is minimally invasive.

I can suggest some simplifications, that may accomplish what you want with less code.

Preserving interfaces feels like preserving all identifiers encountered under `kModuleHeader` CST nodes, which covers the module name itself, imports, parameters, and ports.  Do you agree?

If that is the case, we could break this down into a few operations:

1) `FindAllModuleHeaders` using functions in https://cs.opensource.google/verible/verible/+/master:verilog/CST/module.h (or make a new one there)
2) A generic `CollectAllIdentifiers` function that works on any CST subtree.  You can probably just use [`SearchSyntaxTree`](https://cs.opensource.google/verible/verible/+/master:common/analysis/syntax_tree_search.cc;drc=e64869b7ae41c4c2ee041ab9c5d1c9d58da57b94;bpv=1;bpt=1;l=87?gsn=SearchSyntaxTree&gs=kythe%3A%2F%2Fgithub.com%2Fgoogle%2Fverible%3Flang%3Dc%252B%252B%3Fpath%3Dcommon%2Fanalysis%2Fsyntax_tree_search.cc%23_Ww4UlOgtMeBSgSv0VJmAHgbTW3H1C5zzL-hxxSmoB0&gs=kythe%3A%2F%2Fgithub.com%2Fgoogle%2Fverible%3Flang%3Dc%252B%252B%3Fpath%3Dcommon%2Fanalysis%2Fsyntax_tree_search.h%23TWorgFiq5yrr9u5uHLwDT3RTv5OsQ7yaDeRqySBZzy4) with a predicate that is true for CST leaves that are identifiers, and not have to write a new visitor.  
3) Uniquify strings with a `std::set<absl::string_view, verible::StringViewCompare>` (using copy-less heterogenous lookup).  Producing a separate intermediate result avoids having to modify the obfuscator altogether, and becomes more reusable, and is easier to test.
3) Populate the obfuscation map with that set of strings to preserve them.

How does this sound to you?  The less new code you need to write, the easier testing will be.

Also please file a feature request, which this PR will address, describing your application.  That will make it easier to search for and track this issue, and provide an opportunity for comments.

hchsiao:
@fangism Sounds perfect! I should have explored the code base more before digging into implementation.

I've updated the implementation according to your suggestion. Does this conform to what you mean?

Also, the string set `std::set<absl::string_view, verible::StringViewCompare>` is currently implemented as `std::set<std::string, verible::StringViewCompare>`. If I directly replace `std::string` to `absl::string_view`, the de-referenced string will be corrupted outside of the function producing the set (when I print set elements). I am not sure if I did something wrong. Since the performance of this part may not be critical, do you think it is okay to just use `std::string`?

Speaking of testing, I will start writing tests if this implementation is on the right track.

fangism:
> @fangism Sounds perfect! I should have explored the code base more before digging into implementation.
> 
> I've updated the implementation according to your suggestion. Does this conform to what you mean?
> 
> Also, the string set `std::set<absl::string_view, verible::StringViewCompare>` is currently implemented as `std::set<std::string, verible::StringViewCompare>`. If I directly replace `std::string` to `absl::string_view`, the de-referenced string will be corrupted outside of the function producing the set (when I print set elements). I am not sure if I did something wrong. Since the performance of this part may not be critical, do you think it is okay to just use `std::string`?
> 
> Speaking of testing, I will start writing tests if this implementation is on the right track.

Yes, caveat, when you use `string_views`, you must guarantee that the backing memory for those strings (in this case, their are owned in `TextStructureView`) out-live all views.  If you're still having issues, you may use `std::set<std::string>` and add a TODO comment to investigate this later.

verilog/transform/obfuscate.cc
```diff
@@ -141,6 +144,38 @@ static absl::Status VerifyEquivalence(absl::string_view original,
   return absl::OkStatus();
 }
 
+// Find all modules and collect interface names
+absl::Status CollectInterfaceNames(absl::string_view content,
```
fangism, line 148: Since this function doesn't do any "transform" operations, it would be a better fit inside `verilog/analysis`.  Maybe `verilog/analysis/extractors.cc`?

verilog/transform/obfuscate.cc
```diff
@@ -141,6 +144,38 @@ static absl::Status VerifyEquivalence(absl::string_view original,
...
+      verilog::VerilogAnalyzer::AnalyzeAutomaticMode(content, NULL);
+  const auto lex_status = ABSL_DIE_IF_NULL(analyzer)->LexStatus();
+  const auto parse_status = analyzer->ParseStatus();
+
+  if (lex_status.ok() && parse_status.ok()) {
+    const auto& syntax_tree = analyzer->SyntaxTree();
+    const auto mod_headers =
+      FindAllModuleHeaders(*ABSL_DIE_IF_NULL(syntax_tree).get());
+
+    // For each module, collect all indentifiers under the module
```
fangism, line 162: identifiers

verilog/transform/obfuscate.cc
```diff
@@ -141,6 +144,38 @@ static absl::Status VerifyEquivalence(absl::string_view original,
...
+    // header tree into the StringSet.
+    for(auto const& mod_header: mod_headers) {
```
fangism, line 164: Most other places, we've been using `const auto&` instead of `auto const&`, though I know they're equivalent.

verilog/transform/obfuscate.cc
```diff
@@ -141,6 +144,38 @@ static absl::Status VerifyEquivalence(absl::string_view original,
...
+      const auto if_leafs =
+        FindAllSymbolIdentifierLeafs(*mod_header.match);
```
fangism, line 166: If `` `MACRO`` are referenced in headers, should those be preserved too (`MacroIdentifier`, `MacroCallId`, `MacroIdItem`)?

verilog/transform/obfuscate.cc
```diff
@@ -141,6 +144,38 @@ static absl::Status VerifyEquivalence(absl::string_view original,
...
+      for(auto const& if_leaf_match: if_leafs) {
+        const auto if_leaf = SymbolCastToLeaf(*if_leaf_match.match);
```
fangism, line 168: `const auto& if_leaf`

verilog/transform/obfuscate.cc
```diff
@@ -141,6 +144,38 @@ static absl::Status VerifyEquivalence(absl::string_view original,
...
+        absl::string_view if_name = if_leaf.get().text();
+        if_names->insert(std::string(if_name)); // TODO: use absl::string_view
```
fangism, line 170: We can only use string_view if the `analyzer` object (which contains `TextStructureView`) outlives the set.  Let's just stick to `std::string` in this case.  Later on, if we find multiple uses of the same parse/analysis, then we could refactor `analyzer` to live longer.

verilog/transform/obfuscate.cc
```diff
@@ -141,6 +144,38 @@ static absl::Status VerifyEquivalence(absl::string_view original,
...
+      }
+    }
+    return absl::OkStatus();
+  }
+  return absl::InvalidArgumentError(
```
fangism, line 175: Move the error handling earlier like:

```c++
if (!lex_status.ok()) return lex_status;
if (!parse_status.ok()) return parse_status;
```

This will already bear precise error messages.

verilog/transform/obfuscate.h
```diff
@@ -23,6 +23,12 @@
 
 namespace verilog {
 
+// Collect all identifiers under module header subtree in the CTS.
+// This could be useful when interface names are required to be
+// preserved.
+absl::Status CollectInterfaceNames(absl::string_view content,
+                                   verible::StringSet* if_names);
```
fangism, line 30: I'd just spell out `std::set<std::string>` to make it clear without having to lookup a typedef.

fangism:
In this PR's description, you may say `fixes #530`.

fangism:
In addition to the unit tests, update the documentation on the obfuscator here:
https://cs.opensource.google/verible/verible/+/master:verilog/tools/obfuscator/README.md

Running `--help` can regenerate the usage/options for you.

fangism:
Checking in to see if you need any more help finishing.

fangism (commented):
The implementation looks clean.  Let me know if you need any guidance on testing.

The CST/identifier and module tests can follow surrounding examples.

extractor_test.cc can just pass in code text and expect a set of strings to match.

I'll provide a starter obfuscate_test.sh soon, stay tuned.

hchsiao:
Thanks for helping! I've got another urgent job so this is delayed.
I will work on unit-tests for those codes recently.
And I definitely need a starter for writing tests :).

verilog/transform/obfuscate.cc
hchsiao: In our flow, we expand macros before doing obfuscation so this does not affect our usage. However, I can imagine someone would define macros as their parameter (e.g. bit-width) and include an all-in-one `define.h` in the design. In this case, we may need another option (like `--preserve_macro`) to preserve macro related tokens as `define.h` may not be transformed together.

Should I make the modification with this PR, or we should open another issue for this?

fangism: I'd say separate issue, separate PR.

fangism:
> Thanks for helping! I've got another urgent job so this is delayed.
> I will work on unit-tests for those codes recently.
> And I definitely need a starter for writing tests :).

For testing the binary, feel free to add test commands to:
https://cs.opensource.google/verible/verible/+/master:verilog/tools/obfuscator/obfuscate_test.sh

verilog/CST/identifier_test.cc
```diff
@@ -178,5 +178,31 @@ TEST(GetIdentifierTest, IdentifierUnpackedDimensions) {
...
+      {"function foo(); endfunction", {"foo"}},
+      {"function myclass::foo(); endfunction", {"myclass", "foo"}},
+      {"task goo(); endtask", {"goo"}},
+      {"task fff::goo(); endtask", {"fff", "goo"}},
+  };
+  for (const auto test : kTestCases) {
+    VerilogAnalyzer analyzer(test.first, "");
+    ASSERT_OK(analyzer.Analyze());
+    const auto& root = analyzer.Data().SyntaxTree();
+    const auto symb_ids = FindAllSymbolIdentifierLeafs(*root);
```
fangism, line 196: Recommend using the `TestVerilogSyntaxRangeMatches` form in the examples above, which will not only check string contents, but also exact byte-offset ranges of text.  Your test case data becomes arrays of `SyntaxTreeSearchTestCase`, which lets you tag the substrings of interest in-line.

A `std::set<string_view>` won't be able to discern duplicate identifiers in the actual set, but the `SyntaxTreeSearchTestCase` form will be able to.

verilog/CST/verilog_matchers_test.cc
```diff
@@ -56,6 +56,24 @@ TEST(VerilogMatchers, MacroCallIdLeafTests) {
...
+      {SymbolIdentifierLeaf(), EmbedInClassMethod("`uvm_foo"), 3},
+      {SymbolIdentifierLeaf(), "`uvm_foo(\"foo\");", 2},
+      {SymbolIdentifierLeaf(), "`uvm_foo(\"foo\")\n", 2},
+      {SymbolIdentifierLeaf(), EmbedInClassMethod("`uvm_foo(\"foo\");"), 4},
+      {SymbolIdentifierLeaf(), EmbedInClassMethod("`uvm_foo(\"foo\")\n"), 4},
+      {SymbolIdentifierLeaf(), EmbedInClassMethod("uvm_foo(\"foo\");"), 4},
+      {SymbolIdentifierLeaf(), EmbedInClassMethod("$uvm_foo(\"foo\");"), 4},
+  };
+  for (const auto& test : tests)
+    verible::matcher::RunRawMatcherTestCase<VerilogAnalyzer>(test);
```
fangism, line 74: *sigh* I never got around to cleaning up this entire test file to use a substring-tagged annotation like `SyntaxTreeSearchTestCase`.  Huge TODO for us.

verilog/analysis/extractors.cc
```diff
@@ -0,0 +1,58 @@
...
+
+  if (!lex_status.ok()) return lex_status;
+  if (!parse_status.ok()) return parse_status;
+
+  const auto& syntax_tree = analyzer->SyntaxTree();
+  const auto mod_headers =
+    FindAllModuleHeaders(*ABSL_DIE_IF_NULL(syntax_tree).get());
+
+  // For each module, collect all identifiers under the module
+  // header tree into the StringSet.
```
fangism, line 44: nit: StringSet was an old name.  Can just say "set<string>" for short.

verilog/analysis/extractors.cc
```diff
@@ -0,0 +1,58 @@
...
+                                   std::set<std::string>* if_names) {
+  VLOG(1) << __FUNCTION__;
+
+  const auto analyzer =
+      verilog::VerilogAnalyzer::AnalyzeAutomaticMode(content, NULL);
+  const auto lex_status = ABSL_DIE_IF_NULL(analyzer)->LexStatus();
+  const auto parse_status = analyzer->ParseStatus();
+
+  if (!lex_status.ok()) return lex_status;
+  if (!parse_status.ok()) return parse_status;
```
fangism, line 37: FYI, it is possible to have a syntax error inside the module (could be false positive in the case of parser bugs), and still have a partial syntax tree to work with.  Up to you whether or not you want to allow this, but rejecting on syntax error is ok too.

verilog/tools/obfuscator/verilog_obfuscate.cc
```diff
@@ -95,6 +103,20 @@ Output is written to stdout.
...
+  if (preserve_interface) {
+    const auto status =
+        verilog::analysis::CollectInterfaceNames(content, &preserved);
+    if (!status.ok()) {
+      std::cerr << status.message();
+      return 1;
+    }
+    for (auto const& preserved_name : preserved)
+      subst.encode(preserved_name, preserved_name);
+  }
```
fangism, line 118: minor suggestion: Enclose this block inside `{}` to help the reader quickly see that variables inside (like `preserved`) have short scope/lifetime.  Or just move `preserved` inside the `if` clause.

verilog/tools/obfuscator/verilog_obfuscate.cc
```diff
@@ -50,6 +52,12 @@ ABSL_FLAG(                //
     "reverse to de-obfuscate the source code, and do not obfuscate any unseen "
     "identifiers.  There is no need to --save_map with this option, because "
     "no new substitutions are established.");
+ABSL_FLAG(                //
+    bool, preserve_interface, false,  //
```
fangism, line 56: Can you also document this new feature in obfuscator/README.md?  (Run `--help` and copy-paste.)

verilog/analysis/extractors.h
```diff
@@ -0,0 +1,40 @@
...
+#include "absl/status/status.h"
+#include "absl/strings/string_view.h"
+
+namespace verilog {
+namespace analysis {
+
+// Collect all identifiers under module header subtree in the CTS.
+// This could be useful when interface names are required to be
+// preserved.
+absl::Status CollectInterfaceNames(absl::string_view content,
```
fangism, line 34: Suggested test cases for analysis_extractors_test.cc:

* non-modules (function, class, ...) -- should yield empty set
* minimalist module
* module with import declarations in header
* module with parameters
* module with ports
* module with some body statements, to show that none of those identifiers get included in the resulting set
* one source file with two+ modules

Optional test cases:
* nested modules (should internal modules be obfuscated?) -- or punt with TODO

fangism (changes requested):
Welcome back!  Thanks for resuming work on this, this is coming along nicely.

verilog/CST/verilog_matchers_test.cc
hchsiao: There’s a lot of work to do 😂. For now I’ll leave this as it is to make the file uniform.

verilog/analysis/extractors.cc
hchsiao: Added TODO to not always rejecting syntax error

verilog/tools/obfuscator/verilog_obfuscate.cc
hchsiao: You’re right! Thanks for pointing it out.

verilog/tools/obfuscator/verilog_obfuscate.cc
hchsiao: Sure! 😉

verilog/analysis/extractors.h
hchsiao: Since the `CollectInterfaceNames` extractor should collect names from all modules (including nested ones), the inner module exclusion/inclusion should be an obfuscator issue. I've added a TODO note about this to the obfuscator source file. Is this OK?

hchsiao:
The check item 'verible-ci / ubuntu eoan' seems to have some problems. Would it be OK to ignore it for now?

fangism:
> The check item 'verible-ci / ubuntu eoan' seems to have some problems. Would it be OK to ignore it for now?

Ignore that one, looks unrelated to your PR.  If this is ready, can you update the status from Draft to Open, and request review, to signal to us?

hchsiao:
@fangism Seems like I forgot to click the re-request review button after fixing the code according to your comments.

Anyway, I just wanted to say thank you for providing those invaluable comments that are not only kind and clever but also informative. It makes me start thinking to participate in more open-source activities, from which I can help someone and, in the meantime, strengthen my skills. :)

verilog/analysis/extractors.h
fangism: Yes, this is ok.

fangism (approved):
Everything looks good, we'll take it from here!

fangism:
> @fangism Oops! seems like I forgot to click the re-request review button after fixing the code according to your comments.
> 
> Anyway, I just wanted to say thank you for providing those invaluable comments that are not only kind and clever but also informative. It makes me start thinking to participate in more open-source activities, from which I can help someone and, in the meantime, strengthen my skills. :)

That is helpful feedback to us, thank you for that, and for working through the life of (one of your first?) an open-source contribution.  We love it when contributors help with the features they request/need, and everyone gains something.  We'd be happy to hear from you more.

PR #542, opened 2020-10-16
Add space between return keyword and return value

### LINKED ISSUE #493: want space between return keyword and following expression
**Test case**

```systemverilog
  function int foo(logic [31:0] data);
    return {<<8{data}};
  endfunction
```

**Actual output**

```systemverilog
  function int foo(logic [31:0] data);
    return{<<8{data}};
  endfunction
```

[should be the original spacing after `return`]

ahmedkrmn:
Fixes #493.

**Input:**
```verilog
  function int foo(logic [31:0] data);
    return {<<8{data}};
  endfunction
```
**Formatter output before:**
```verilog
  function int foo(logic [31:0] data);
    return{<<8{data}};
  endfunction
```

**Formatter output after:**
```verilog
  function int foo(logic [31:0] data);
    return {<<8{data}};
  endfunction
```

### DIFF
--- a/verilog/formatting/token_annotator.cc
+++ b/verilog/formatting/token_annotator.cc
@@ -201,6 +201,10 @@ static WithReason<int> SpacesRequiredBetween(
     return {1, "Require space after semicolon"};
   }
 
+  if (left.TokenEnum() == TK_return){
+    return {1, "Space between return keyword and return value"};
+  }
+
   if (right_context.IsInsideFirst({NodeEnum::kStreamingConcatenation}, {})) {
     if (left.TokenEnum() == TK_LS || left.TokenEnum() == TK_RS) {
       return {0, "No space around streaming operators"};
--- a/verilog/formatting/token_annotator_test.cc
+++ b/verilog/formatting/token_annotator_test.cc
@@ -4672,6 +4672,55 @@ TEST(TokenAnnotatorTest, AnnotateFormattingWithContextTest) {
           {/* any context */},
           {0, SpacingOptions::MustWrap},
       },
+      // Space between return keyword and return value
+      {
+          DefaultStyle,
+          {verilog_tokentype::TK_return, "return"},
+          {'{', "{"},
+          {/* unspecified context */},
+          {/* unspecified context */},
+          {1, SpacingOptions::Undecided},
+      },
+      {
+          DefaultStyle,
+          {verilog_tokentype::TK_return, "return"},
+          {'(', "("},
+          {/* unspecified context */},
+          {/* unspecified context */},
+          {1, SpacingOptions::Undecided},
+      },
+      {
+          DefaultStyle,
+          {verilog_tokentype::TK_return, "return"},
+          {'-', "-"},
+          {/* unspecified context */},
+          {/* unspecified context */},
+          {1, SpacingOptions::Undecided},
+      },
+      {
+          DefaultStyle,
+          {verilog_tokentype::TK_return, "return"},
+          {'!', "!"},
+          {/* unspecified context */},
+          {/* unspecified context */},
+          {1, SpacingOptions::Undecided},
+      },
+      {
+          DefaultStyle,
+          {verilog_tokentype::TK_return, "return"},
+          {'~', "~"},
+          {/* unspecified context */},
+          {/* unspecified context */},
+          {1, SpacingOptions::Undecided},
+      },
+      {
+          DefaultStyle,
+          {verilog_tokentype::TK_return, "return"},
+          {verilog_tokentype::SystemTFIdentifier, "$foo"},
+          {/* unspecified context */},
+          {/* unspecified context */},
+          {1, SpacingOptions::Undecided},
+      },
   };
   int test_index = 0;
   for (const auto& test_case : kTestCases) {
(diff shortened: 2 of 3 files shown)


ahmedkrmn:
Hi @fangism,
Can you please label this PR as "hacktoberfest-accepted" so that it would count as a contribution for [Hacktoberfest](https://hacktoberfest.digitalocean.com/hacktoberfest-update) this year.

verilog/formatting/token_annotator_test.cc
```diff
@@ -4672,6 +4672,15 @@ TEST(TokenAnnotatorTest, AnnotateFormattingWithContextTest) {
           {/* any context */},
           {0, SpacingOptions::MustWrap},
       },
+      {
+          // return {<<8{data}};
+          DefaultStyle,
+          {verilog_tokentype::TK_return, "return"},
+          {'{', "{"},
+          {/* unspecified context */},
+          {NodeEnum::kStreamingConcatenation},
```
fangism, line 4681: Should be context-independent so: `{/* unspecified context */},` instead of `{NodeEnum::kStreamingConcatenation},`

verilog/formatting/token_annotator_test.cc
```diff
@@ -4672,6 +4672,15 @@ TEST(TokenAnnotatorTest, AnnotateFormattingWithContextTest) {
...
+          {1, SpacingOptions::Undecided},
+      },
```
fangism, line 4683: More suggested return test cases:

* `return (`
* `return $foo` (is a `SystemTFIdentifier`)
* `return -` (like `return -1`)
* `return !` (like `return !condition`)
* `return ~` (like `return ~condition`)

verilog/formatting/token_annotator.cc
```diff
@@ -206,8 +206,12 @@ static WithReason<int> SpacesRequiredBetween(
       return {0, "No space around streaming operators"};
     } else if (left.format_token_enum == FormatTokenType::numeric_literal ||
                left.format_token_enum == FormatTokenType::identifier ||
-               left.format_token_enum == FormatTokenType::keyword) {
+               (left.format_token_enum == FormatTokenType::keyword &&
+                strcmp(verilog_symbol_name(left.TokenEnum()), "\"return\""))) {
       return {0, "No space around streaming operator slice size"};
+    } else if (left.format_token_enum == FormatTokenType::keyword &&
+               !strcmp(verilog_symbol_name(left.TokenEnum()), "\"return\"")) {
+      return {1, "Space between return keyword and return value"};
```
fangism, line 214: I would suggest instead of editing inside the condition for `kStreamingConcatenation` that we think of this as a matter of rule precedence, that is `return ...` should always have a space no matter what is on the right.  See the block around line 159 (if keyword).  It was intended to cover `return` but wasn't reached first.

* If it is possible/easy to reorder rules that don't break existing tests and fix the new ones great!  (What happens if this whole block were moved further down?)
* If that proves difficult, the next easiest option is to add one case like: `if (left.token_enum() == TK_return) ...`

fangism (changes requested):
Thank you again for working on this!

fangism:
> Hi @fangism,
> Can you please label this PR as "hacktoberfest-accepted" so that it would count as a contribution for [Hacktoberfest](https://hacktoberfest.digitalocean.com/hacktoberfest-update) this year.

Done.

verilog/formatting/token_annotator.cc
ahmedkrmn: I tried your suggestion of re-arranging the blocks. I moved this block that I updated to the bottom - along with some other blocks to pass all the cases - but still, no spaces were added after the return keyword.
I believe adding this condition as you've suggested before the block is the easiest option:
```cpp
  if (left.TokenEnum() == TK_return){
    return {1, "Space between return keyword and return value"};
  }
```

verilog/formatting/token_annotator_test.cc
ahmedkrmn: I tried setting it to `{/* unspecified context */}`, but the test case always passes even if no space is added after the return keyword.

verilog/formatting/token_annotator.cc
fangism: Ok, thanks for checking.  Let's go with this solution.

verilog/formatting/token_annotator_test.cc
fangism: This test doesn't have/need context about what _original_ spacing was, because it is purely based on two adjacent token _types_.  Only the end-to-end test in formatter_test.cc has _original_ code to consider.  (Did that address your concern?)

verilog/formatting/token_annotator_test.cc
fangism: Also mirror some of these tests over to formatter_test.cc, e.g.

```
{"function int f;return-1;endfunction",
 "function int f;\n"
 "  return -1;\n"
 "endfunction\n"},
{"function int f ;return    ! x\n;endfunction",
 "function int f;\n"
 "  return !x;\n"
 "endfunction\n"},

```

fangism (approved):

PR #553, opened 2020-10-25
Scopes for if conditions and for loops

### LINKED ISSUE #503: [kythe] scoping of local nets/variables in various generate blocks
## Input test case

```systemverilog
module m;
  if (x) begin : gen_x
    logic temp;
  end
  if (y) begin : gen_y
    logic temp;
  end
  // no temp in this scope
endmodule
```

Make sure that the two `temp`s are defined in distinct scopes.
These begin/end blocks may not always come with labels like `gen_x` and `gen_y`, in which case you may assign any unique identifier internally.
Same with `for(genvar i = 0; ...) begin ... end`.
Verify with test cases.

### LINKED ISSUE #504: [kythe] genvar variables declared inside for-loops should be distinct
## Input test case

Try to reduce the test down to a single demonstrative file.

```systemverilog
module m;
  for (genvar i=0; i<32; i++) begin
    assign y[i] = x[i];
  end
  for (genvar i=0; i<32; i++) begin
    assign y[i] = x[i];
  end
endmodule
```

The loop-variable `i`s should be distinct def/ref sets, but right now are combined.

MinaToma:
Fixes #504
Fixes #503


verilog/tools/kythe/indexing_facts_tree_extractor_test.cc
```diff
@@ -3154,140 +3154,6 @@ TEST(PackageImportTest, PackageDirectMemberReference) {
   EXPECT_EQ(result_pair.right, nullptr) << *result_pair.right;
 }
 
-TEST(FactsTreeExtractor, ForLoopInitializations) {
-  constexpr int kTag = 1;  // value doesn't matter
```
MinaToma, line 3158: I removed this test because now we create the scope with random id and it can't be tested but this is covered in verification test with 2 or more tests.
We can keep a counter for the scopes we create for loops (keep our own counter instead of using random id).

MinaToma:
@fangism does this create a scope?
```
int x;
initial begin
    int x;
end
```

fangism:
> @fangism should this create a scope?
> 
> ```
> int x;
> initial begin
>     int x;
> end
> ```

I don't think so because declarations shouldn't be allowed inside these bare statement blocks.

verilog/tools/kythe/indexing_facts_tree_extractor.cc
```diff
@@ -14,6 +14,9 @@
 
 #include "verilog/tools/kythe/indexing_facts_tree_extractor.h"
 
+#include <stdlib.h>
```
fangism, line 17: `<cstdlib>`

verilog/tools/kythe/indexing_facts_tree_extractor.cc
```diff
@@ -255,6 +258,21 @@ void IndexingFactsTreeExtractor::Visit(const SyntaxTreeNode& node) {
       ExtractTypeDeclaration(node);
       break;
     }
+    case NodeEnum::kLoopGenerateConstruct:
```
fangism, line 261: Do we need `kConditionalGenerateConstruct`, `kGenerateRegion`?

verilog/tools/kythe/indexing_facts_tree_extractor.cc
```diff
@@ -1458,6 +1476,23 @@ void IndexingFactsTreeExtractor::ExtractTypeDeclaration(
...
 }
 
+void IndexingFactsTreeExtractor::ExtractTemporaryScope(
+    const verible::SyntaxTreeNode& node) {
+  IndexingFactNode temp_scope_node(
+      IndexingNodeData{IndexingFactType::kTemporaryScope});
+
+  // Generate unique id for this scope.
+  temp_scope_node.Value().AppendAnchor(
+      Anchor(absl::StrCat("temp-scope-", getpid(), "-", random()), 0, 0));
```
fangism, line 1486: Choosing `-` is good because it cannot appear in any user space identifier.

I would also use one counter **per enclosing scope** instead of `random()` for determinism, and maximum stability (to incremental changes).  
e.g. `int next_anonymous_id = 0;`

verilog/tools/kythe/indexing_facts_tree_extractor.h
```diff
@@ -224,6 +224,9 @@ class IndexingFactsTreeExtractor : public verible::TreeContextVisitor {
   // e.g counter #(.N(r)) extracts "N".
   void ExtractParamByName(const verible::SyntaxTreeNode& param_by_name);
 
+  // Creates new scope with unique id for if conditions and loops.
```
fangism, line 227: specifically, intended for conditional/loop generate constructs.

verilog/tools/kythe/indexing_facts_tree_extractor_test.cc
fangism: Suggested above, keep one counter per enclosing scope.

verilog/tools/kythe/indexing_facts_tree_extractor.h
```diff
@@ -224,6 +224,9 @@ class IndexingFactsTreeExtractor : public verible::TreeContextVisitor {
...
+  void ExtractTemporaryScope(const verible::SyntaxTreeNode& node);
```
fangism, line 228: Design: this is not really a temporary scope, it needs to persist semantically, even if it is not named.  It should not be discarded, which is what "Temporary" is suggesting.

verilog/tools/kythe/kythe_facts_extractor.h
```diff
@@ -171,6 +171,9 @@ class KytheFactsExtractor {
   // Extracts Kythe facts from class instance node and return its VName.
   VName ExtractClassInstances(const IndexingFactNode& class_instance_fact_node);
 
+  // Creates a new temporary scope for if conditions and loops.
```
fangism, line 174: An extractor should not be creating anything, even if temporary.  The unnamed scopes created by the indexer should persist.

fangism (commented):
> > @fangism should this create a scope?
> > ```
> > int x;
> > initial begin
> >     int x;
> > end
> > ```
> 
> I don't think so because declarations shouldn't be allowed inside these bare statement blocks.

Hmm, I take that back, I think local variable declarations are allowed, and can only be referenced in their respective blocks.

verilog/tools/kythe/testdata/more_testdata/do_while_loop.sv
```diff
@@ -0,0 +1,20 @@
+//- @m1 defines/binding _
+module m1;
+    //- @temp defines/binding Temp0
+    //- @y defines/binding Y
+    int temp, y;
+
+    initial begin
+        do begin
+            //- @temp defines/binding Temp8
+            int temp;
```
fangism, line 10: FYI: Normally compilers can diagnose these as _shadowing violations_, where introducing a name in one scope precludes the use of the same name referring to an enclosing scope.  The language permits this to happen without regard to any ill-effects or readability.

verilog/tools/kythe/verilog_extractor_indexing_fact_type.h
```diff
@@ -36,6 +36,7 @@ enum class IndexingFactType {
   kClassInstance,
   kModuleInstance,
   kDataTypeReference,
+  kTemporaryScope,
```
fangism, line 39: `kAnonymousScope`?

fangism (changes requested):

verilog/tools/kythe/indexing_facts_tree_extractor.h
ivan444: +1

Unnamed or Loop scope sounds better. And it should persist.

verilog/tools/kythe/verilog_extractor_indexing_fact_type.h
ivan444: or Loop?

ivan444 (approved):

verilog/tools/kythe/indexing_facts_tree_extractor.cc
MinaToma: I don know where these tags appear? can you point me.

fangism: https://cs.opensource.google/verible/verible/+/master:verilog/CST/verilog_nonterminals.h

e.g. `module m; if (condition) ...` and `module m; for(genvar i=0; ...) ...`

MinaToma: I think we don't we have already test cases for these two examples and they are working fine.

fangism: The following works already?

```systemverilog
module m #(int N) ();
  for (genvar i=0; i<N; ++i) begin
    wire foo, bar;
    assign foo = bar;
  end
  for (genvar i=0; i<N; ++i) begin
    wire foo, bar;
    assign foo = bar;
  end
endmodule
```

The loops are duplicate, but aside from N, all other symbols should be distinct from the other loop.

BTW, the relevant passage on scoping from the SV-LRM is chapter 27.  For loops 27.4, and for conditional 27.5.

Can you help me with a pointer?  I haven't mentally cached our corpus of test cases.

If only I had a tool that could search files by language-feature (CST-based)...

verilog/tools/kythe/testdata/more_testdata/for_loop.sv
```diff
@@ -0,0 +1,55 @@
+//- @m1 defines/binding _
+module m1;
```
MinaToma, line 2: @fangism check this file for `for (genvar ...)`

verilog/tools/kythe/testdata/more_testdata/if_condition.sv
```diff
@@ -0,0 +1,85 @@
+//- @m2 defines/binding _
+module m2;
```
MinaToma, line 2: @fangism check `if` conditions here.

verilog/tools/kythe/indexing_facts_tree_extractor.cc
MinaToma: I added some comments below for the files with these examples.

verilog/tools/kythe/testdata/more_testdata/do_while_loop.sv
```diff
@@ -0,0 +1,20 @@
+//- @m1 defines/binding _
+module m1;
+    //- @temp defines/binding Temp0
+    //- @y defines/binding Y
+    int temp, y;
+
+    initial begin
```
fangism, line 7: FYI, you only need begin/end where there is more than one statement, so here: `initial do begin ...` should work too.

verilog/tools/kythe/indexing_facts_tree_extractor.cc
fangism: Thanks for the comments.

fangism (changes requested):

verilog/tools/kythe/indexing_facts_tree_extractor.cc
```diff
@@ -1458,6 +1474,23 @@ void IndexingFactsTreeExtractor::ExtractTypeDeclaration(
...
 }
 
+void IndexingFactsTreeExtractor::ExtractAnonymousScope(
+    const verible::SyntaxTreeNode& node) {
+  IndexingFactNode temp_scope_node(
+      IndexingNodeData{IndexingFactType::kAnonymousScope});
+
+  // Generate unique id for this scope.
+  temp_scope_node.Value().AppendAnchor(
+      Anchor(absl::StrCat("anonymous-scope-", next_anonymous_id++), 0, 0));
```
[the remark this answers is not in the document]
MinaToma: The 0,0 aren't important at all. I think we need to refactor a ctor for Anchor with default values for these two.

verilog/tools/kythe/testdata/more_testdata/conditional_generate.sv
```diff
@@ -0,0 +1,29 @@
+//- @m1 defines/binding _
```
fangism, line 1: This test looks good, suggest a few more variants to cover/mix-up:

* single generate items without `begin`/`end`
* module instantiations, like `foo bar(.baz(baz));`
* net declarations using `wire`

```systemverilog
module m;
  if (x) foo1 bar();
  else foo2 bar();   // this bar should be distinct
endmodule
```

I also almost forgot that case-generates are another kind of conditional generate. `module m; case ... endcase endmodule`

verilog/tools/kythe/testdata/more_testdata/for_loop.sv
fangism: Can you create a version of this test where loop-statements are inside a task or function?  (This is very typical.)

verilog/tools/kythe/testdata/more_testdata/for_loop.sv
```diff
@@ -0,0 +1,55 @@
...
+
+    initial begin
+        //- @#1i defines/binding I1
+        //- @#2i ref I1
+        //- @#3i ref I1
+        for (int i = 0; i < 50; i++) begin
+            //- @temp defines/binding Temp1
+            int temp;
+            //- @temp ref Temp1
+            assign temp = 1;
```
fangism, line 48: suggest a new small test case that uses function calls inside loop statements like `y[i] = f(x[i]);`
to show that `f` resolves to the same function no matter which loop (can be duplicate) it appears in.

fangism (changes requested):
Thought of some more tests to cover, then I think we're good.

fangism (approved):

ivan444 (approved):

PR #555, opened 2020-10-29
Fix bug in extracting facts from file_list_node and print function

MinaToma:
Fixing bug for extracting facts from `file_list_node` in the iterative loop which is not intended.
Simplifying the logic in `KytheFactsPrinter`'s `Print` function.

### DIFF
--- a/verilog/tools/kythe/verilog_kythe_extractor.cc
+++ b/verilog/tools/kythe/verilog_kythe_extractor.cc
@@ -17,16 +17,16 @@
 #include <iostream>
 #include <string>
 
-#include "google/protobuf/io/zero_copy_stream_impl.h"
 #include "absl/flags/flag.h"
 #include "absl/status/status.h"
 #include "absl/strings/str_cat.h"
 #include "absl/strings/string_view.h"
 #include "absl/strings/substitute.h"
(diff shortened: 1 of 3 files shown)


verilog/tools/kythe/verilog_kythe_extractor.cc
```diff
@@ -64,10 +64,8 @@ VNameRef ConvertToVnameRef(const VName& vname,
 
 // Prints Kythe facts in proto format to stdout.
 void PrintKytheFactsProtoEntries(const IndexingFactNode& file_list_facts_tree) {
-  KytheFactsExtractor kythe_extractor(
-      GetFileListDirFromRoot(file_list_facts_tree),
-      /*previous_files_scopes=*/nullptr);
-  auto indexing_data = kythe_extractor.ExtractKytheFacts(file_list_facts_tree);
+  auto indexing_data =
+      KytheFactsExtractor::ExtractFileList(file_list_facts_tree);
```
ivan444, line 68: Can you keep the name `ExtractKytheFacts`? `ExtractFileList` being the entry point is a bit confusing. `ExtractKytheFacts` is what the users want.

ivan444 (approved):

fangism (approved):

PR #557, opened 2020-10-30
Type parameter qualified reference

### LINKED ISSUE #543: [kythe] safely handle type-parameter qualified references
## Input test case

```systemverilog
class S #(parameter type T);
  T::member_type foo;
  typedef int some_type;
endclass

S#(logic)::some_type bar;
```

Make sure that

1) in `T::member_type`, do not emit any reference for `member_type` because `T` is not resolvable.
2) in `S#(logic)::some_type`, `some_type` does resolve to the typedef definition in `class S`.  This is possible because SystemVerilog` does not have a notion of template specialization like C++ does.

If this is already working, make sure it is covered in a test somewhere.

MinaToma:
Fixes #543

### DIFF
--- a/verilog/tools/kythe/indexing_facts_tree_extractor.cc
+++ b/verilog/tools/kythe/indexing_facts_tree_extractor.cc
@@ -1499,7 +1499,7 @@ void IndexingFactsTreeExtractor::ExtractTypeDeclaration(
         GetIdentifierFromTypeDeclaration(type_declaration);
     facts_tree_context_.top().NewChild(
         IndexingNodeData({Anchor(type_name->get(), context_.base)},
-                         IndexingFactType::kVariableDefinition));
+                         IndexingFactType::kTypeDeclaration));
     return;
   }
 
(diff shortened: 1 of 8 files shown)


ivan444 (approved):

verilog/tools/kythe/indexing_facts_tree_extractor.cc
```diff
@@ -1493,7 +1493,15 @@ void IndexingFactsTreeExtractor::ExtractTypeDeclaration(
   if (tag != NodeEnum::kDataType) return;
   const SyntaxTreeNode* primitive =
       GetStructOrUnionOrEnumTypeFromDataType(*type);
-  if (primitive == nullptr) return;
+  if (primitive == nullptr) {
```
fangism, line 1496: // Then this is a user-defined type.

verilog/tools/kythe/indexing_facts_tree_extractor.cc
```diff
@@ -1493,7 +1493,15 @@ void IndexingFactsTreeExtractor::ExtractTypeDeclaration(
...
+    // Extract type name.
+    const SyntaxTreeLeaf* type_name =
+        GetIdentifierFromTypeDeclaration(type_declaration);
+    facts_tree_context_.top().NewChild(
+        IndexingNodeData({Anchor(type_name->get(), context_.base)},
```
fangism, line 1501: check for type_name nullptr first

fangism (changes requested):

fangism (approved):

fangism (approved):

PR #558, opened 2020-10-30
Compact bit selection/indexing expressions

### LINKED ISSUE #361: Formatter: Avoid whitespace in bit selection
The formatter introduces whitespace around operators inside bit-selections. I'd argue these are better off as compact expressions, I note whitespace isn't added in declarations.

For example in https://github.com/lowRISC/ibex/blob/master/rtl/ibex_icache.sv

Before formatting:
```systemverilog
logic [INDEX_W-1:0]                  inval_index_d, inval_index_q;
```

After formatting:
```systemverilog
logic [INDEX_W-1:0] inval_index_d, inval_index_q;
```

The extra whitespace between the dimension and the names is removed but the expression inside the `[]` remains compact.

Before formatting:
```systemverilog
assign lookup_addr_aligned = {lookup_addr_ic0[ADDR_W-1:LINE_W],{LINE_W{1'b0}}};
```

After formatting:
```systemverilog
assign lookup_addr_aligned = {lookup_addr_ic0[ADDR_W - 1:LINE_W], {LINE_W{1'b0}}};
```

Similar syntax (though used to select bits rather than declare dimensions) but this time whitespace is added around the `-`

ldalek-antmicro:
Fixes #361

### DIFF
--- a/verilog/formatting/format_style.h
+++ b/verilog/formatting/format_style.h
@@ -96,6 +96,9 @@ struct FormatStyle : public verible::BasicFormatStyle {
   // reduce the risk of harming already decent code.
   bool try_wrap_long_lines = true;
 
+  // Compact binary expressions inside indexing / bit selection operators
+  bool compact_indexing_and_selections = true;
+
   // TODO(fangism): introduce the following knobs:
   //
   // Unless forced by previous line, starting a line with a comma is
--- a/verilog/formatting/formatter_test.cc
+++ b/verilog/formatting/formatter_test.cc
@@ -593,14 +593,12 @@ static constexpr FormatterTestCase kFormatterTestCases[] = {
                                                   // (symmetrized)
     },
     {
-        "  parameter  int   foo=bar [ a +b ] ;",  // binary inside index expr
-        "parameter int foo = bar[a + b];\n",      // allowed to be 1 space
-                                                  // (symmetrized)
+        "  parameter  int   foo=bar [ a +b ] ;",  // compact binary inside
+        "parameter int foo = bar[a+b];\n",        // index expression
     },
     {
-        "  parameter  int   foo=bar [ a  +b ] ;",  // binary inside index expr
-        "parameter int foo = bar[a + b];\n",       // limited to 1 space (and
-                                                   // symmetrized)
+        "  parameter  int   foo=bar [ a  +b ] ;",  // compact binary inside
+        "parameter int foo = bar[a+b];\n",         // index expression
     },
     {
         // with line continuations
--- a/verilog/formatting/token_annotator.cc
+++ b/verilog/formatting/token_annotator.cc
@@ -132,7 +132,8 @@ static bool IsAnySemicolon(const PreFormatToken& ftoken) {
 static WithReason<int> SpacesRequiredBetween(
     const PreFormatToken& left, const PreFormatToken& right,
     const SyntaxTreeContext& left_context,
-    const SyntaxTreeContext& right_context) {
+    const SyntaxTreeContext& right_context,
+    const FormatStyle& style) {
   VLOG(3) << "Spacing between " << verilog_symbol_name(left.TokenEnum())
           << " and " << verilog_symbol_name(right.TokenEnum());
   // Higher precedence rules should be handled earlier in this function.
@@ -241,6 +242,10 @@ static WithReason<int> SpacesRequiredBetween(
     // TODO(fangism): make this behavior configurable
     if (right.format_token_enum == FormatTokenType::binary_operator &&
         InRangeLikeContext(right_context)) {
+      if (style.compact_indexing_and_selections && !InDeclaredDimensions(right_context)) {
+        return {0, "Compact binary expressions inside indexing / bit selection operator []"};
+      }
+
       int spaces = right.OriginalLeadingSpaces().length();
       if (spaces > 1) {
         spaces = 1;
@@ -560,7 +565,7 @@ static SpacePolicy SpacesRequiredBetween(
   // Default for unhandled cases, 1 space to be conservative.
   constexpr int kUnhandledSpacesDefault = 1;
   const auto spaces =
-      SpacesRequiredBetween(left, right, left_context, right_context);
+      SpacesRequiredBetween(left, right, left_context, right_context, style);
   VLOG(1) << "spaces: " << spaces.value << ", reason: " << spaces.reason;
 
   if (spaces.value == kUnhandledSpacesRequired) {
--- a/verilog/formatting/token_annotator_test.cc
+++ b/verilog/formatting/token_annotator_test.cc
@@ -4753,6 +4753,12 @@ struct OriginalSpacingSensitiveTestCase {
   ExpectedInterTokenInfo expected_annotation;
 };
 
+static const auto CompactIndexSelectionStyle = [](){
+  auto style = DefaultStyle;
+  style.compact_indexing_and_selections = false;
+  return style;
+}();
+
 // These tests are allowed to be sensitive to original inter-token spacing.
 TEST(TokenAnnotatorTest, OriginalSpacingSensitiveTests) {
   const OriginalSpacingSensitiveTestCase kTestCases[] = {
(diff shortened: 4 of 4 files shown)


verilog/formatting/token_annotator.cc
```diff
@@ -426,6 +430,11 @@ static WithReason<int> SpacesRequiredBetween(
 
     // Spacing in ranges
     if (InRangeLikeContext(right_context)) {
+      // Bit selection
```
fangism, line 433: ... and indexing inside []

verilog/formatting/token_annotator.cc
```diff
@@ -241,6 +241,10 @@ static WithReason<int> SpacesRequiredBetween(
     // TODO(fangism): make this behavior configurable
     if (right.format_token_enum == FormatTokenType::binary_operator &&
         InRangeLikeContext(right_context)) {
+      if (!InDeclaredDimensions(right_context)) {
+        return {0, "Remove space before binary operator inside []."};
```
fangism, line 245: ... inside [] for indexing.

verilog/formatting/formatter_test.cc
```diff
@@ -594,12 +594,12 @@ static constexpr FormatterTestCase kFormatterTestCases[] = {
     },
     {
         "  parameter  int   foo=bar [ a +b ] ;",  // binary inside index expr
-        "parameter int foo = bar[a + b];\n",      // allowed to be 1 space
+        "parameter int foo = bar[a+b];\n",      // allowed to be 1 space
```
fangism, line 597: update the // comment too

verilog/formatting/formatter_test.cc
```diff
@@ -594,12 +594,12 @@ static constexpr FormatterTestCase kFormatterTestCases[] = {
...
                                                   // (symmetrized)
     },
     {
         "  parameter  int   foo=bar [ a  +b ] ;",  // binary inside index expr
-        "parameter int foo = bar[a + b];\n",       // limited to 1 space (and
+        "parameter int foo = bar[a+b];\n",       // limited to 1 space (and
```
fangism, line 602: update the // comment too

verilog/formatting/formatter_test.cc
```diff
@@ -5633,7 +5633,7 @@ static constexpr FormatterTestCase kFormatterTestCases[] = {
     {"module mp ;property p1 ; a|->## [0  : 1]  b;endproperty endmodule",
      "module mp;\n"
      "  property p1;\n"
-     "    a |-> ##[0 : 1] b;\n"  // limit to one space, symmetrize
+     "    a |-> ##[0:1] b;\n"  // limit to one space, symmetrize
```
fangism, line 5636: I think this is ok, even though semantically, it is a delay range context, not a declared range or indexing.

Also update // comment

fangism (changes requested):
Something to think about, but not required for this PR:

https://cs.opensource.google/verible/verible/+/master:verilog/formatting/format_style.h

is where I've been slowly adding configurable knobs.  That style structure is available to most formatter passes like the token-annotator.  When introducing a change that could likely have different user preferences, I consider adding an option in this style struct.  This lets us test out different configurations and combinations internally.  (formatter_test.cc has test sections that override the defaults.)  However, very few of these struct fields are exposed to users (intentionally) because it is too early to have users rely on stable configurations schemas at this time.

Keep this in mind as we continue to work on more opinionated formatter features.

And welcome back!

verilog/formatting/token_annotator.cc
ldalek-antmicro: Done

verilog/formatting/token_annotator.cc
ldalek-antmicro: Done

verilog/formatting/formatter_test.cc
ldalek-antmicro: Done

verilog/formatting/formatter_test.cc
ldalek-antmicro: Done

verilog/formatting/formatter_test.cc
ldalek-antmicro: Done

ldalek-antmicro:
Hey @fangism 
I've added an option for controlling a compaction of dimension range.
What do you think? It's okay or too soon for that?

verilog/formatting/format_style.h
```diff
@@ -96,6 +96,9 @@ struct FormatStyle : public verible::BasicFormatStyle {
   // reduce the risk of harming already decent code.
   bool try_wrap_long_lines = true;
 
+  // Compact bit selection / indexing ranges
+  bool compact_non_declaration_ranges = true;
```
fangism, line 100: Suggest naming this with what it includes, rather than what it excludes, something like: `compact_indexing_and_selections`, so we know what it is intended for (in case we make a mistake down the road and find that this implementation is too inclusive).  Also make it clear in the comment that this is concerning a subset of syntax inside []s.

verilog/formatting/formatter_test.cc
```diff
@@ -5633,7 +5631,7 @@ static constexpr FormatterTestCase kFormatterTestCases[] = {
     {"module mp ;property p1 ; a|->## [0  : 1]  b;endproperty endmodule",
      "module mp;\n"
      "  property p1;\n"
-     "    a |-> ##[0 : 1] b;\n"  // limit to one space, symmetrize
+     "    a |-> ##[0:1] b;\n"  // compact bit slice range
```
fangism, line 5634: Actually, this is a (cycle?) delay range, not a bit slice, nor declared dimensions.
If you look at its CST, is it distinguishable from indexing and declared dimensions?
If it is, can we update the logic to exclude this case?
If it is not, we should consider re-tagging these CST nodes to distinguish them semantically.
[Example of a re-tagging a CST subtree.](https://cs.opensource.google/verible/verible/+/master:verilog/parser/verilog.y;drc=259321c5bb95882d2ac872ca03de6c8d35fca733;l=1897)

verilog/formatting/token_annotator.cc
```diff
@@ -241,6 +242,10 @@ static WithReason<int> SpacesRequiredBetween(
     // TODO(fangism): make this behavior configurable
     if (right.format_token_enum == FormatTokenType::binary_operator &&
         InRangeLikeContext(right_context)) {
+      if (style.compact_non_declaration_ranges && !InDeclaredDimensions(right_context)) {
+        return {0, "Remove space before binary operator inside [] for indexing."};
```
fangism, line 246: Specify that this is targeting indexing/bit-selection uses of [].

verilog/formatting/token_annotator.cc
```diff
@@ -426,6 +431,11 @@ static WithReason<int> SpacesRequiredBetween(
 
     // Spacing in ranges
     if (InRangeLikeContext(right_context)) {
+      // Bit selection and indexing inside []
+      if (style.compact_non_declaration_ranges && !InDeclaredDimensions(right_context)) {
+        return {0, "Limit spaces before ':' in bit slice/indexing to 0"};
+      }
```
fangism, line 437: I suggest we leave this out this time (changing spacing around `:`), and only target the binary operators as you've done above.

verilog/formatting/token_annotator_test.cc
```diff
@@ -5187,19 +5187,123 @@ TEST(TokenAnnotatorTest, OriginalSpacingSensitiveTests) {
...
-          {1, SpacingOptions::Undecided},  // limit to 1
+          {1, SpacingOptions::Undecided}, // do not compact
+      },
+      {
+          // [a  +b]
+          [](){
+            auto style = DefaultStyle;
+            style.compact_non_declaration_ranges = false;
+            return style;
+          }(),
```
fangism, line 5266: To reduce repetition, let's pull this into its own static initializer:

```C++
static const CompactIndexSelectionStyle = [](){...}();
```

fangism (changes requested):

verilog/formatting/format_style.h
ldalek-antmicro: Done

verilog/formatting/formatter_test.cc
ldalek-antmicro: N/A anymore in this PR (removed).

verilog/formatting/token_annotator.cc
ldalek-antmicro: Done

verilog/formatting/token_annotator.cc
ldalek-antmicro: Removed.

verilog/formatting/token_annotator_test.cc
ldalek-antmicro: Done.

fangism (approved):

Issue #556, opened 2020-10-30
travis-ci: builds time out


fangism:
Since kythe was added as a dependency, most travis-builds have been timing out.

First bad:
https://travis-ci.org/github/google/verible/jobs/739104819

Last good:
https://travis-ci.org/github/google/verible/builds/738921915
https://travis-ci.org/github/google/verible/builds/738413728

Things one could look into, to name a few:

* .github/travis/script.sh -- adding some time commands or timestamps
* bazel -j parallelization, already max'd?
* Maintaining a persistent bazel build cache from build-to-build (but also have a way to reset the cache, just-in-case)

fangism:
b/171743791

fangism:
@mithro anyone on AntMicro team available to help?

fangism:
Most recent timeout: https://travis-ci.org/github/google/verible/builds/740056016

mithro:
@kgugala - Can you get someone to look at this?

kgugala:
@fangism can we migrate the CI to e.g. github actions? there is 6h time limit in actions

fangism:
No objection from me.  (I am still interested if there are some easy speed-ups though.)

mglb:
I've added timestamps to script.sh. Results: 
* Environment initialization takes < 1 min
* `bazel test` takes about 29 min
* `bazel build` runs until timeout.

See build log for more details: https://travis-ci.com/github/antmicro/verible/jobs/423717516

Running test and build in separate jobs solves the problem. Each job runs in parallel for 29 minutes: https://travis-ci.com/github/antmicro/verible/builds/197663893

fangism:
Thanks, I commented in the PR: can we try matching the `-c opt` for both build and test, for the sake of caching?

tgorochowik:
@fangism should we still migrate the CI to github actions? It probably makes sense to do that to avoid similar issues in the future.

mithro:
@tgorochowik - Yes please move to GitHub Actions ASAP.

fangism:
All former Travis-CI actions have been migrated to github actions.

PR #561, opened 2020-11-03
CI: Run test and build in separate jobs

mglb:
Solves problem from issue #556


.github/travis/script.sh
```diff
@@ -22,14 +22,18 @@ export TRAVIS_TAG=${TRAVIS_TAG:-$(git describe --match=v*)}
 BAZEL_CXXOPTS=(--cxxopt=-std=c++17)
 
 case "$MODE" in
-compile-n-test)
-    bazel test --noshow_progress "${BAZEL_CXXOPTS[@]}" //...
-    bazel build -c opt --noshow_progress "${BAZEL_CXXOPTS[@]}" //...
+test)
+    bazel test --noshow_progress --curses=no --show_timestamps "${BAZEL_CXXOPTS[@]}" //...
```
fangism, line 26: Just noticed: `build` and `test` should use the same options, both should use `-c opt`.  Different options will double the work due to bazel cache missing.  If the old `compile-n-test` added the missing `-c opt`, the second command should be much faster, I wager nearly instantaneous because of cache hits.  Can we try this experiment first?  

`test` should be a superset of `build`.  One could argue for removing `build` altogether and letting `test` do all of the work.

tgorochowik: @fangism note that for `-c opt` means build with optimization enabled and with `assert()` calls disabled (`-O2 -DNDEBUG`)

So the decision here is whether we want to disable asserts for testing. I think it would be beneficial to leave this as is at least for PRs (although it would be better if it stayed on for master builds too) - the `-c opt` flag is really only needed for creating releases to make the binaries smaller.

mglb: It doesn't help: https://travis-ci.com/github/antmicro/verible/jobs/426145668

mglb:
@googlebot I signed it!

.github/travis/script.sh
fangism: So Verible's codebase doesn't use any traditional `assert`s from `<cassert>`.  It uses `CHECK*` for assertions, which are always-on.

tgorochowik: ack, thanks!

fangism: At line 1147 of https://travis-ci.com/github/antmicro/verible/jobs/426145668

```
(09:30:53) INFO: Build options --cxxopt, --define, and --host_cxxopt have changed, discarding analysis cache.
```

If you see that, that may be an early warning sign of cache-missing a lot.  Can you verify that the options match?
Or could we eliminate `build` if `test` is a superset anyway?

fangism: Could we reduce this to `test -c opt` which is a superset of `build`?  We need to fix CI soon.

.github/travis/script.sh
```diff
@@ -22,14 +22,18 @@ export TRAVIS_TAG=${TRAVIS_TAG:-$(git describe --match=v*)}
...
 case "$MODE" in
-compile-n-test)
-    bazel test --noshow_progress "${BAZEL_CXXOPTS[@]}" //...
-    bazel build -c opt --noshow_progress "${BAZEL_CXXOPTS[@]}" //...
+test)
+    bazel test -c opt //...
+    ;;
+
+compile)
+    bazel build -c opt --noshow_loading_progress --curses=no --progress_report_interval 60 --show_timestamps "${BAZEL_CXXOPTS[@]}" //...
```
fangism, line 30: Keep the same bazel options in `test` for consistency.

.github/travis/success.sh
```diff
@@ -21,7 +21,11 @@ git config --local user.name "Deployment Bot"
 git config --local user.email "verible-dev@googlegroups.com"
 
 case $MODE in
-compile-n-test)
+test)
```
fangism, line 24: Just wondering: is there a concern with `test` latency that is keeping this separate/distinct from `compile`?

fangism (changes requested):

.github/travis/script.sh
```diff
@@ -22,14 +22,18 @@ export TRAVIS_TAG=${TRAVIS_TAG:-$(git describe --match=v*)}
...
 case "$MODE" in
-compile-n-test)
-    bazel test --noshow_progress "${BAZEL_CXXOPTS[@]}" //...
-    bazel build -c opt --noshow_progress "${BAZEL_CXXOPTS[@]}" //...
+test)
+    bazel test -c opt //...
+    ;;
+
+compile)
+    bazel build -c opt //...
```
fangism, line 30: You'll want `"${BAZEL_CXXOPTS[@]}"` as c++17 it is required by one of Verible's prerequisites (until we fix it).

fangism (changes requested):
One last change.

.github/travis/script.sh
wtatarski: done

fangism (approved):

PR #565, opened 2020-11-03
Fix to include with non-literal string

### LINKED ISSUE #562: [kythe] crash on `include directive with non-string-literal
## Input test case

```systemverilog
`include`d
```

Crashes:

```
Crash Type: CHECK failure
Crash Address:
Crash State:
  symbol.Kind() == SymbolKind::kLeaf in tree_utils.cc
  verible::SymbolCastToLeaf
  verilog::kythe::IndexingFactsTreeExtractor::ExtractInclude
```

Syntactically, the argument to include could be more complicated than a string literal; it could be other string concatenations and "advanced" preprocessing uses.

Suggest making it fail-safe, and ignore any time the right side is not a plain string-literal.

MinaToma:
Fixes #562

### DIFF
--- a/verilog/CST/macro.cc
+++ b/verilog/CST/macro.cc
@@ -32,6 +32,11 @@ std::vector<verible::TreeSearchMatch> FindAllMacroDefinitions(
   return SearchSyntaxTree(root, NodekPreprocessorDefine());
 }
 
+std::vector<verible::TreeSearchMatch> FindAllPreprocessorInclude(
+    const verible::Symbol& root) {
+  return SearchSyntaxTree(root, NodekPreprocessorInclude());
+}
+
 std::vector<verible::TreeSearchMatch> FindAllMacroCalls(const Symbol& root) {
   return SearchSyntaxTree(root, NodekMacroCall());
 }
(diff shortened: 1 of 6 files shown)


verilog/CST/macro.cc
```diff
@@ -83,10 +83,10 @@ const verible::SyntaxTreeLeaf& GetMacroArgName(
   return GetSubtreeAsLeaf(macro_formal_arg, NodeEnum::kMacroFormalArg, 0);
 }
 
-const verible::SyntaxTreeLeaf& GetFileFromPreprocessorInclude(
+const verible::Symbol& GetFileFromPreprocessorInclude(
     const verible::Symbol& preprocessor_include) {
-  return GetSubtreeAsLeaf(preprocessor_include, NodeEnum::kPreprocessorInclude,
-                          1);
+  return *ABSL_DIE_IF_NULL(verible::GetSubtreeAsSymbol(
```
ivan444, line 88: Is it ok to add a check here?

ivan444 (approved):

verilog/CST/macro.cc
MinaToma: yes. after all the include must contain some information after `include` keyword.

fangism: Yes, and ideally, this should be `CHECK`ed on CST node construction for includes.  Checking here will do for now.

verilog/tools/kythe/indexing_facts_tree_extractor.cc
```diff
@@ -1312,9 +1312,19 @@ absl::string_view StripOuterQuotes(absl::string_view text) {
 
 void IndexingFactsTreeExtractor::ExtractInclude(
     const verible::SyntaxTreeNode& preprocessor_include) {
-  const SyntaxTreeLeaf& included_filename =
+  const verible::Symbol& included_filename_symbol =
       GetFileFromPreprocessorInclude(preprocessor_include);
 
+  // Terminate if this isn't a string literal.
+  if (included_filename_symbol.Kind() != verible::SymbolKind::kLeaf ||
```
fangism, line 1319: if you call `Tag()` once, that will also give you `.kind` (and avoid double virtual call).

verilog/tools/kythe/indexing_facts_tree_extractor.cc
```diff
@@ -1312,9 +1312,19 @@ absl::string_view StripOuterQuotes(absl::string_view text) {
 
 void IndexingFactsTreeExtractor::ExtractInclude(
     const verible::SyntaxTreeNode& preprocessor_include) {
-  const SyntaxTreeLeaf& included_filename =
+  const verible::Symbol& included_filename_symbol =
       GetFileFromPreprocessorInclude(preprocessor_include);
 
+  // Terminate if this isn't a string literal.
```
fangism, line 1318: Can we move this logic to a CST function?  This one will be very useful.

```c++
// Returns ... else nullptr if the argument is not a simple string-literal.
const SyntaxTreeLeaf* GetStringLiteralFromPreprocessorInclude(const Symbol& pp_include_arg);
```

fangism (changes requested):

fangism (approved):

PR #583, opened 2020-11-09
Add suggest_parentheses rule for Verible::Linter

### LINKED ISSUE #383: Recommend parentheses around certain expression patterns
Though parsers and compilers know fully well the precedence of operators, humans can still benefit from extra parenthesis.

example:
https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md#parentheses

Task: detect such situations (starting with the above ternary expression example) and report them

Suggest: subclassing a SyntaxTreeLintRule and for various expression types, looking downwards at its operands and comparing operators/precedences.  

Feel free to suggest more specific operator classes and cases.

omarahmed1111:
This PR introduces:
* Implementation of suggest_parentheses rule for the linter.
* Change kTernaryExpression to kConditionExpression globally.

suggest-parentheses rule is a new rule for the linter to suggest parentheses around nested expressions. The rule is not enabled by-default. Now, the rule only suggest parentheses around nested kConditionExpression in kConditionExpression.

Fixes #383


verilog/analysis/checkers/suggest_parentheses_rule_test.cc
```diff
@@ -0,0 +1,44 @@
...
+using verible::RunLintTestCases;
+
+TEST(SuggestParenthesesTest, AcceptTests) {
+  const std::initializer_list<LintTestCase> kSuggestParenthesesTestCases = {
+      {""},
+      {"module m;\nendmodule\n"},
+      {"module m;\ninitial begin end\nendmodule"},
+      {"module m;\n", "assign foo = condition_a? a : b;\n", "endmodule"},
+      {"module m;\n", "assign foo = condition_a? (condition_b? a : b) : c;\n", "endmodule"},
+      {"module m;\n", "assign foo = condition_a? (condition_b? a : b) : (condition_c? c : d);\n", "endmodule"},
```
fangism, line 35: As per the earlier comment, this should also be accepted with un-parenthesized `condition_c? c : d`.  (Can you add this as a separate test case?)

verilog/analysis/checkers/suggest_parentheses_rule_test.cc
```diff
@@ -0,0 +1,44 @@
...
+      {"module m;\n", "assign foo = condition_a? (condition_b? (condition_c? a : b) : c) : d;\n", "endmodule"},
+  };
```
fangism, line 37: Add some test cases that expect to find violations like this:

```c++
constexpr int kTag = 1;  // don't care
const std::initializer_list<LintTestCase> kSuggestParenthesesTestCases = {
  ...
  {"module m;\nassign foo = condition_a? ", {kTag, "condition_b? a : b"}, " : c;\nendmodule"},
};
```

This notation `{kTag, "condition_b? a : b"}` says: "expect to find a violation on this exact substring range."

verilog/analysis/checkers/suggest_parentheses_rule.cc
```diff
@@ -0,0 +1,85 @@
...
+      const auto falseCasetag = static_cast<verilog::NodeEnum>(falseCase->Tag().tag);
+
+      const verible::Symbol* trueCaseChild = UnwrapExpression(*trueCase);
+      const verible::Symbol* falseCaseChild = UnwrapExpression(*falseCase);
+
+      const auto trueCaseChildtag = static_cast<verilog::NodeEnum>(trueCaseChild->Tag().tag);
+      const auto falseCaseChildtag = static_cast<verilog::NodeEnum>(falseCaseChild->Tag().tag);
+      
+      if((trueCasetag == NodeEnum::kExpression && trueCaseChildtag == NodeEnum::kConditionExpression) || trueCasetag == NodeEnum::kConditionExpression){
+        violations_.insert(LintViolation(node, kMessage, context));
```
fangism, line 68: Instead of highlighting the whole `node`, the violation's text should narrow down to `*trueCase` because that is where parentheses can be added to remedy the problem.

verilog/analysis/checkers/suggest_parentheses_rule.cc
```diff
@@ -0,0 +1,85 @@
...
+using verible::LintViolation;
+using verible::SyntaxTreeContext;
+using verible::SymbolTag;
+
+absl::string_view SuggestParenthesesRule::Name() {
+  return "suggest-parentheses";
+}
+const char SuggestParenthesesRule::kTopic[] = "parentheses";
+const char SuggestParenthesesRule::kMessage[] =
+    "Use parentheses around nested expressions";
```
fangism, line 38: I would make this message more specific to the implemented check:

"Parenthesize condition expressions that appear in the true-clause of another condition expression."

As more checks are added, they will get their own specific messages.

verilog/analysis/checkers/suggest_parentheses_rule.cc
```diff
@@ -0,0 +1,85 @@
...
+
+std::string SuggestParenthesesRule::GetDescription(
+    DescriptionType description_type) {
+  return absl::StrCat(
+      "Checks that parentheses are wrapped in nested expressions. See ",
```
fangism, line 43: This message can be generic: "Recommend extra parentheses around subexpressions where it helps readability."

verilog/analysis/checkers/suggest_parentheses_rule.h
```diff
@@ -0,0 +1,52 @@
...
+#ifndef VERIBLE_VERILOG_ANALYSIS_CHECKERS__SUGGEST_PARENTHESES_RULE_H_
+#define VERIBLE_VERILOG_ANALYSIS_CHECKERS__SUGGEST_PARENTHESES_RULE_H_
+
+#include "common/analysis/syntax_tree_lint_rule.h"
+#include "verilog/analysis/descriptions.h"
+
+namespace verilog {
+namespace analysis {
+
+// checks that nested expressions are wrapped properly in parentheses.
```
fangism, line 24: // Suggests where extra parentheses can help readability.
// For example:
//   a1 ? a2 ? b : c : d
// would be improved as the equivalent:
//   a1 ? (a2 ? b : c) : d

verilog/analysis/checkers/suggest_parentheses_rule.h
```diff
@@ -0,0 +1,52 @@
...
+  void HandleNode(const verible::SyntaxTreeNode& node,
+                  const verible::SyntaxTreeContext& context) override;
+
+  verible::LintRuleStatus Report() const override;
+
+ private:
+  // Link to style guide rule.
+  static const char kTopic[];
+
+  // Diagnostic message for constraint name violations.
```
fangism, line 43: // Diagnostic message for rule violations.

(nothing to do with constraints)

verilog/analysis/checkers/suggest_parentheses_rule_test.cc
```diff
@@ -0,0 +1,44 @@
...
+
```
fangism, line 38: Suggest testing a few more contexts like:

* `parameter foo = expresssion;`
*  
```systemverilog
always @(posedge clk) begin
  left <= expression;
end
```
* 
```systemverilog
function f;
  g = h(expression);
endfunction
```

verilog/parser/verilog.y
```diff
@@ -380,6 +380,7 @@ is not locally defined, so the grammar here uses only generic identifiers.
 /* The new tokens from 1800-2005. */
 %token TK_alias "alias"
 %token TK_always_comb "always_comb"
+%token TK_suggest_parentheses "suggest_parentheses"
```
fangism, line 383: Revert, syntax change isn't needed.

verilog/parser/verilog.lex
```diff
@@ -554,6 +554,7 @@ wone { UpdateLocation(); return TK_wone; }
 uwire { UpdateLocation(); return TK_uwire; }
 alias { UpdateLocation(); return TK_alias; }
 always_comb { UpdateLocation(); return TK_always_comb; }
+suggest_parentheses { UpdateLocation(); return TK_suggest_parentheses; }
```
fangism, line 557: Revert, lexical change isn't needed.

fangism (changes requested):

verilog/analysis/checkers/suggest_parentheses_rule_test.cc
omarahmed1111: I didn't notice this part in this guide. Okay, I will modify that and provide more test cases on this part. :)

verilog/analysis/checkers/suggest_parentheses_rule_test.cc
omarahmed1111: Truth to be said, I was going to ask about this part and how to do that. Thanks for this info. :)

fangism: Glad that helped!  Don't forget to "request review" when ready.  Thanks for working on this again.

verilog/analysis/checkers/suggest_parentheses_rule_test.cc
fangism: Also suggest a few test cases where the true-clause is a different expression, like a shift operation or addition, where no parentheses are recommended (yet).

verilog/analysis/checkers/suggest_parentheses_rule_test.cc
```diff
@@ -0,0 +1,74 @@
...
+#include "verilog/analysis/verilog_analyzer.h"
+
+namespace verilog {
+namespace analysis {
+namespace {
+
+using verible::LintTestCase;
+using verible::RunLintTestCases;
+
+TEST(SuggestParenthesesTest, AcceptTests) {
```
fangism, line 28: Since this contains a mix of positive and negative tests, rename `AcceptTests` to something like `Various`.

verilog/CST/expression.cc
```diff
@@ -72,6 +72,32 @@ bool ConstantIntegerValue(const verible::Symbol& expr, int* value) {
   return absl::SimpleAtoi(text, value);
 }
 
+
+const verible::Symbol* UnwrapExpression(const verible::Symbol& expr) {
+  
+  const auto& node = verible::SymbolCastToNode(expr);
```
fangism, line 78: This will terminate with an assert failure if passed a leaf, so suggest checking `expr.Kind()` first before doing this cast.

verilog/CST/expression.cc
```diff
@@ -72,6 +72,32 @@ bool ConstantIntegerValue(const verible::Symbol& expr, int* value) {
...
+  return GetSubtreeAsSymbol(condition_expr, NodeEnum::kConditionExpression, 0);
+}
+
+const verible::Symbol* GetConditionExpressionTrueCase(const verible::Symbol& condition_expr) {
+  return GetSubtreeAsSymbol(condition_expr, NodeEnum::kConditionExpression, 2);
+}
+
+const verible::Symbol* GetConditionExpressionFalseCase(const verible::Symbol& condition_expr) {
+  return GetSubtreeAsSymbol(condition_expr, NodeEnum::kConditionExpression, 4);
+}
```
fangism, line 99: Let's add unit tests for these, so that should any of these get restructured, we'll know to fix them in one place.

[Example unit tests](https://cs.opensource.google/verible/verible/+/master:verilog/CST/declaration_test.cc).
Or look at many other CST/*_test.cc files.

A typical recipe for unit testing these extractor functions:
1) FindAll of the outer structure of interest (might need to write a quick function for this too), in this case, condition expressions.
2) Write test cases, tagging the substring of text that corresponds to what an accessor function should return: condition, true case, false case.
3) Pass the array of test cases to `TestVerilogSyntaxRangeMatches` (this is where you call FindAll and possibly other functions to refine your search).

fangism (changes requested):
Thanks again for another round, we're almost done.

fangism (approved):
Nice job on the unit tests!  Thank you again for this contribution, others will be able to follow what you've done.

.vscode/settings.json
```diff
@@ -0,0 +1,77 @@
+{
```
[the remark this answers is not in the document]
omarahmed1111: Okay sry, I added them by mistake

omarahmed1111:
> Nice job on the unit tests! Thank you again for this contribution, others will be able to follow what you've done.

Thank you for the great review and help. Glad for my first contribution here :)

.vscode/settings.json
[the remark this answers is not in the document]
omarahmed1111: Do I need to sumbit another update with them deleted or they will be filtered out automatically on merge ?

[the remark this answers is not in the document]
fangism: Your work is done, we took care of it.  (Just waiting for internal approval, but expecting delays over the holiday season.)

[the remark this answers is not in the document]
omarahmed1111: Got it, Thanks

PR #598, opened 2020-11-11
Qualified types

### LINKED ISSUE #476: [kythe] bug in class instances with package name
## Input test case

```systemverilog
pkg::my_class class_name = new();
```

Class type can be some class inside a package and tagged with `kQualifiedId`.
Extend the extractor to cover this case.

MinaToma:
Fixes #476

### DIFF
--- a/verilog/tools/kythe/indexing_facts_tree_extractor.cc
+++ b/verilog/tools/kythe/indexing_facts_tree_extractor.cc
@@ -298,6 +298,10 @@ void IndexingFactsTreeExtractor::Visit(const SyntaxTreeNode& node) {
       ExtractAnonymousScope(node);
       break;
     }
+    case NodeEnum::kUnqualifiedId: {
+      ExtractUnqualifiedId(node);
+      break;
+    }
     default: {
       TreeContextVisitor::Visit(node);
     }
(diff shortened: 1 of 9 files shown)


MinaToma:
@fangism is it possible to have qualified id for module instance? e.g `pkg::my_module foo();`
and if it is possible can you give me an example?

fangism:
> @fangism is it possible to have qualified id for module instance? e.g `pkg::my_module foo();`
> and if it is possible can you give me an example?

Semantically, no, by my reading of the LRM.  However, the grammar is implemented in such a way that it is _syntactically_ possible to handle the grammar conflicts between general data declarations and module instantiations.  So as long as you handle data declarations like `foo_pkg::bar_t var1_name, var2_name, ...;`, that's good enough.

verilog/tools/kythe/indexing_facts_tree_extractor.cc
```diff
@@ -1105,20 +1113,37 @@ void IndexingFactsTreeExtractor::ExtractClassInstances(
...
+    // Then this is an unqualified id node.
+    // e.g "my_class var_name".
+    const SyntaxTreeLeaf& type_identifier =
+        *ABSL_DIE_IF_NULL(AutoUnwrapIdentifier(*type));
+    type_node.Value().AppendAnchor(
+        Anchor(type_identifier.get(), context_.base));
+
+    // Extract parameter list
+    const SyntaxTreeNode* param_list =
+        GetParamListFromDataDeclaration(data_declaration_node);
```
fangism, line 1138: Wouldn't it make more sense to get the ParamList directly from the unqualified id node at this stage, since you've already descended this far?

Also is this functionality duplicated elsewhere that handles arbitrary types (qualified, unqualified), basically anywhere there's a type reference (extends, type parameters, typedef, data declarations)?

fangism (changes requested):

ivan444 (approved):

verilog/tools/kythe/indexing_facts_tree_extractor.cc
MinaToma: I refactored some parts but the remaining parts can't be simply refactored.

fangism (approved):

PR #600, opened 2020-11-12
More Testing For Middle Layer (Kythe Extractor)

MinaToma:
More test to cover `indexing_facts_tree_extractor.cc` and some refactoring.

### DIFF
--- a/verilog/tools/kythe/indexing_facts_tree_extractor.cc
+++ b/verilog/tools/kythe/indexing_facts_tree_extractor.cc
@@ -223,10 +223,6 @@ void IndexingFactsTreeExtractor::Visit(const SyntaxTreeNode& node) {
       ExtractClassDeclaration(node);
       break;
     }
-    case NodeEnum::kSelectVariableDimensionList: {
-      ExtractSelectVariableDimension(node);
-      break;
-    }
     case NodeEnum::kParamDeclaration: {
       ExtractParamDeclaration(node);
       break;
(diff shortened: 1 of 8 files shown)


verilog/tools/kythe/indexing_facts_tree_extractor.cc
```diff
@@ -227,10 +227,6 @@ void IndexingFactsTreeExtractor::Visit(const SyntaxTreeNode& node) {
       ExtractClassDeclaration(node);
       break;
     }
-    case NodeEnum::kSelectVariableDimensionList: {
-      ExtractSelectVariableDimension(node);
-      break;
-    }
```
MinaToma, line 233: @fangism this one is not used. but I was trying to find where this tag appears but failed.
can you give me some snippet where might this appear?

fangism: Anywhere that `base[...]` appears as a part of a reference expression (indexing).  But actually, this was restructured a while ago, and individual dimensions were appended directly to `kReferenceNode`s, rather than inside a node of `kSelectVariableDimensionList`.  However looking at the grammar, I see that it is still reachable through `kEventControl`:

```systemverilog
task t(int y);
  @x[y];
endtask
```

fangism: I'll approve this now, but cover this case in a separate PR.

fangism (approved):
Thanks for updating some test names too.

PR #601, opened 2020-11-13
Test for select variable in event control

MinaToma:
tests for this case: (kythe extractor)
```
//- @t1 defines/binding _
//- @y defines/binding Y
task t1(int y);
    //- @x defines/binding X
    int x[10];
    //- @x ref X
    //- @y ref Y
    @x[y];
endtask

```

### DIFF
--- a/verilog/tools/kythe/testdata/even_control.sv
+++ b/verilog/tools/kythe/testdata/even_control.sv
@@ -1,0 +1,9 @@
+//- @t1 defines/binding _
+//- @y defines/binding Y
+task t1(int y);
+    //- @x defines/binding X
+    int x[10];
+    //- @x ref X
+    //- @y ref Y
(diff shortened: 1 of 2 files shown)


verilog/tools/kythe/testdata/even_control.sv
```diff
@@ -0,0 +1,9 @@
+//- @t1 defines/binding _
```
fangism, line 1: did you mean to name this file "event_control.sv"?  (I'll rename this for you.)

fangism (approved):

verilog/tools/kythe/testdata/even_control.sv
MinaToma: oh yes!!
thanks :)

PR #607, opened 2020-11-16
Add missing "fact_name" to edge JSON

mglb:
Kythe requires "fact_name" to be present. `entrystream` produces an error otherwise:
```
[mglb/fix-json-output 47186d0] Add missing "fact_name" to edge JSON
```
It also fixes invalid JSON syntax (`},}`) introduced when "fact_name" was removed.


fangism (approved):
Thanks for the fix!

PR #613, opened 2020-11-20
Align comments in port list

### LINKED ISSUE #350: Formatter: Align comments in port list
The formatter lines up data types and port names, but doesn't line up comments. This would give slightly easier-to-read code.

https://github.com/lowRISC/ibex/blob/master/rtl/ibex_controller.sv

Expected:

```systemverilog
module something(
    // decoder related signals
    input logic illegal_insn_i,    // decoder has an invalid instr
    input logic ecall_insn_i,      // decoder has ECALL instr
    input logic mret_insn_i,       // decoder has MRET instr
    input logic dret_insn_i,       // decoder has DRET instr
    input logic wfi_insn_i,        // decoder has WFI instr
    input logic ebrk_insn_i,       // decoder has EBREAK instr
    input logic csr_pipe_flush_i,  // do CSR-related pipeline flush
);
endmodule
```

Actual:

```systemverilog
module something(
    // decoder related signals
    input logic illegal_insn_i,  // decoder has an invalid instr
    input logic ecall_insn_i,  // decoder has ECALL instr
    input logic mret_insn_i,  // decoder has MRET instr
    input logic dret_insn_i,  // decoder has DRET instr
    input logic wfi_insn_i,  // decoder has WFI instr
    input logic ebrk_insn_i,  // decoder has EBREAK instr
    input logic csr_pipe_flush_i,  // do CSR-related pipeline flush
);
endmodule
```

wtatarski:
Fixes #350

### DIFF
--- a/common/formatting/align.cc
+++ b/common/formatting/align.cc
@@ -94,6 +94,8 @@ struct AlignmentCell {
   // as a space-only column, usually no more than 1 space wide.
   int left_border_width = 0;
 
+  int TotalWidth() const { return left_border_width + compact_width; }
+
   FormatTokenRange ConstTokensRange() const {
     return FormatTokenRange(tokens.begin(), tokens.end());
   }
@@ -323,14 +325,42 @@ static void ComputeCellWidths(AlignmentMatrix* matrix) {
 typedef std::vector<AlignedColumnConfiguration> AlignedFormattingColumnSchema;
 
 static AlignedFormattingColumnSchema ComputeColumnWidths(
-    const AlignmentMatrix& matrix) {
+    const AlignmentMatrix& matrix,
+    const std::vector<AlignmentColumnProperties>& column_properties) {
   VLOG(2) << __FUNCTION__;
   AlignedFormattingColumnSchema column_configs(matrix.front().size());
+
+  // Check which cell before delimiter is the longest
+  // If this cell is in the last row, the sizes of column with delimiter
+  // must be set to 0
+  int longest_cell_before_delimiter = 0;
+  bool align_to_last_row = false;
+  for (const auto& row : matrix) {
+    auto column_prop_iter = column_properties.begin();
+    for (const auto& cell : row) {
+      if (std::next(column_prop_iter, 1)->contains_delimiter) {
+        if (longest_cell_before_delimiter < cell.TotalWidth()) {
+          longest_cell_before_delimiter = cell.TotalWidth();
+          if (&row == &matrix.back()) align_to_last_row = true;
+        }
+        break;
+      }
+      ++column_prop_iter;
+    }
+  }
+
   for (const auto& row : matrix) {
     auto column_iter = column_configs.begin();
+    auto column_prop_iter = column_properties.begin();
     for (const auto& cell : row) {
-      column_iter->UpdateFromCell(cell);
+      if (column_prop_iter->contains_delimiter && align_to_last_row) {
+        column_iter->width = 0;
+        column_iter->left_border = 0;
+      } else {
+        column_iter->UpdateFromCell(cell);
+      }
       ++column_iter;
+      ++column_prop_iter;
     }
   }
   VLOG(2) << "end of " << __FUNCTION__;
@@ -385,8 +415,13 @@ static std::vector<DeferredTokenAlignment> ComputeAlignedRowSpacings(
       PreFormatToken& ftoken = cell.tokens.front();
       int left_spacing;
       if (properties_iter->flush_left) {
-        left_spacing = accrued_spaces;
-        accrued_spaces = padding;
+        if (properties_iter->contains_delimiter) {
+          left_spacing = 0;
+          accrued_spaces += padding;
+        } else {
+          left_spacing = accrued_spaces;
+          accrued_spaces = padding;
+        }
       } else {  // flush right
         left_spacing = accrued_spaces + padding;
         accrued_spaces = 0;
@@ -460,15 +495,8 @@ static MutableFormatTokenRange GetMutableFormatTokenRange(
   const Symbol* origin = ABSL_DIE_IF_NULL(unwrapped_line.Origin());
   VLOG(2) << "row: " << StringSpanOfSymbol(*origin);
 
-  // Partition may contain text that is outside of the span of the syntax
-  // tree node that was visited, e.g. a trailing comma delimiter.
-  // Exclude those tokens from alignment consideration (for now).
-  const SyntaxTreeLeaf* last_token = GetRightmostLeaf(*origin);
   const auto range_begin = unwrapped_line.TokensRange().begin();
   auto range_end = unwrapped_line.TokensRange().end();
-  // Backwards search is expected to check at most a few tokens.
-  while (!BoundsEqual(std::prev(range_end)->Text(), last_token->get().text()))
-    --range_end;
   CHECK(range_begin <= range_end);
 
   // Scan each token-range for cell boundaries based on syntax,
@@ -585,12 +613,12 @@ AlignablePartitionGroup::CalculateAlignmentSpacings(
   // Compute compact sizes per cell.
   ComputeCellWidths(&result.matrix);
 
-  // Compute max widths per column.
-  AlignedFormattingColumnSchema column_configs(
-      ComputeColumnWidths(result.matrix));
-
   // Extract other non-computed column properties.
   const auto column_properties = column_schema.ColumnProperties();
+
+  // Compute max widths per column.
+  AlignedFormattingColumnSchema column_configs(
+      ComputeColumnWidths(result.matrix, column_properties));
 
   {
     // Total width does not include initial left-indentation.
--- a/common/formatting/align.h
+++ b/common/formatting/align.h
@@ -25,6 +25,8 @@
 #include "common/text/token_info.h"
 #include "common/text/tree_context_visitor.h"
 #include "common/util/logging.h"
+#include "common/text/tree_utils.h" // for GetRightmostLeaf
+#include "common/text/token_stream_view.h" // for TokenRange
 
 namespace verible {
 
@@ -33,6 +35,7 @@ struct AlignmentColumnProperties {
   // If true format cell with padding to the right: |text   |
   // else format cell with padding to the left:     |   text|
   bool flush_left = true;
+  bool contains_delimiter = false;
 
   AlignmentColumnProperties() = default;
   explicit AlignmentColumnProperties(bool flush_left)
@@ -305,6 +308,81 @@ std::vector<ColumnPositionEntry> ScanPartitionForAlignmentCells(
   return scanner.SparseColumns();
 }
 
+// Similarly to the function above this function creates an instance of
+// ScannerType and extracts column alignment information. Firstly it reuses
+// existing column scanner for extracting column alignment information from
+// SyntaxTree (which doesn't contain delimiters and comments) and after that it
+// examines TokenPartitionTree to detect comments and delimiters and extract
+// information for alignment with non_tree_column_scanner function.
+// A 'row' corresponds to a range of format tokens over which spacing is to be
+// adjusted to achieve alignment.
+// A 'non_tree_column_scanner' is a function which takes as an argument
+// TokenRange (containing all tokens excluded from SyntaxTree) and returns
+// data for appropriate alignment.
+// Returns a concatenated sequence of column entries for tokens from SyntaxTree
+// and TokenPartitionTree that will be uniquified and ordered for alignment
+// purposes.
+template <class ScannerType>
+std::vector<ColumnPositionEntry>
+  ScanPartitionForAlignmentCells_WithNonTreeTokens(
+    const TokenPartitionTree& row,
+    const std::function<
+      std::vector<ColumnPositionEntry>
+      (TokenRange, const SyntaxTreeNode&, long unsigned int)> non_tree_column_scanner) {
+  // re-use existing scanner
+  auto column_entries_from_syntax_tree =
+    ScanPartitionForAlignmentCells<ScannerType>(row);
+  const UnwrappedLine& unwrapped_line = row.Value();
+  const Symbol* origin = ABSL_DIE_IF_NULL(unwrapped_line.Origin());
+  // Identify the last token covered by the origin tree.
+  const SyntaxTreeLeaf* last_leaf = GetRightmostLeaf(*origin);
+  const TokenInfo last_tree_token = last_leaf->get();
+
+  // Find maximum number of children amongst every row for
+  // non_tree_column_scanner to set correct SyntaxTreePath for delimiter
+  // and comments
+  static int max_children_size;
+  if (row.IsFirstChild()) {
+    max_children_size = 0;
+    const TokenPartitionTree *row_tmp = &row;
+    do {
+      int children_size = SymbolCastToNode(*row.Value().Origin()).children().size();
+      if (children_size > max_children_size)
+        max_children_size = children_size;
+
+      row_tmp = row_tmp->NextSibling();
+    } while(row_tmp);
+  }
+
+  // Collect tokens excluded from SyntaxTree (delimiters and comments)
+  TokenSequence non_tree_tokens;
+  bool non_tree_flag = false;
+  for (auto i : unwrapped_line.TokensRange()) {
+
+    if (non_tree_flag)
+      non_tree_tokens.push_back(*i.token);
+
+    if (*i.token == last_tree_token)
+      non_tree_flag = true;
+  }
+
+  auto remainder = make_range<TokenSequence::const_iterator>(
+		     non_tree_tokens.begin(), non_tree_tokens.end());
+  const SyntaxTreeNode& node = SymbolCastToNode(*origin);
+  auto trailing_column_entries = non_tree_column_scanner(remainder, node, max_children_size);
+  std::vector<ColumnPositionEntry> all_column_entries;
+  all_column_entries.insert(
+    all_column_entries.begin(),
+    column_entries_from_syntax_tree.begin(),
+    column_entries_from_syntax_tree.end());
+  all_column_entries.insert(
+    all_column_entries.end(),
+    trailing_column_entries.begin(),
+    trailing_column_entries.end());
+
+  return all_column_entries;
+}
+
 // Convenience function for generating alignment cell scanners.
 // This can be useful for constructing maps of scanners based on type.
 //
@@ -321,6 +399,19 @@ template <class ScannerType>
 AlignmentCellScannerFunction AlignmentCellScannerGenerator() {
   return [](const TokenPartitionTree& row) {
     return ScanPartitionForAlignmentCells<ScannerType>(row);
+  };
+}
+
+// Overloaded function for generating alignment cell scanners. This adapter
+// accepts a trailing token scanner function for aligning delimiters and
+// comments.
+template <class ScannerType>
+AlignmentCellScannerFunction AlignmentCellScannerGenerator(
+  const std::function<
+    std::vector<ColumnPositionEntry>
+    (TokenRange, const SyntaxTreeNode&, long unsigned int)> non_tree_column_scanner) {
+  return [non_tree_column_scanner](const TokenPartitionTree& row) {
+    return ScanPartitionForAlignmentCells_WithNonTreeTokens<ScannerType>(row, non_tree_column_scanner);
   };
 }
 
--- a/verilog/formatting/align.cc
+++ b/verilog/formatting/align.cc
@@ -1044,6 +1044,40 @@ std::function<MemberType(const StructType&)> function_from_pointer_to_member(
 using AlignmentHandlerMapType =
     std::map<AlignableSyntaxSubtype, AlignmentGroupHandlers>;
 
+std::vector<verible::ColumnPositionEntry> non_tree_column_scanner(
+    verible::TokenRange token_range, const verible::SyntaxTreeNode& node,
+    long unsigned int max_children_size) {
+  std::vector<verible::ColumnPositionEntry> trailing_column_entries;
+  auto tag = NodeEnum(node.Tag().tag);
+  VLOG(2) << __FUNCTION__ << ", node: " << tag;
+
+  for (auto token : token_range) {
+    switch (token.token_enum()) {
+      case ',': {
+        SyntaxTreePath path{max_children_size};
+        VLOG(2) << "at " << TreePathFormatter(path);
+        AlignmentColumnProperties prop;
+        prop.contains_delimiter = true;
+        verible::ColumnPositionEntry test{path, token, prop};
+        trailing_column_entries.push_back(test);
+        break;
+      }
+      case TK_COMMENT_BLOCK:
+      case TK_EOL_COMMENT: {
+        SyntaxTreePath path{max_children_size + 1};
+        VLOG(2) << "at " << TreePathFormatter(path);
+        verible::ColumnPositionEntry test{path, token, FlushLeft};
+        trailing_column_entries.push_back(test);
+        break;
+      }
+      default:
+        break;
+    }
+  }
+
+  return trailing_column_entries;
+}
+
 // Global registry of all known alignment handlers for Verilog.
 // This organization lets the same handlers be re-used in multiple syntactic
 // contexts, e.g. data declarations can be module items and generate items and
@@ -1068,7 +1102,8 @@ static const AlignmentHandlerMapType& AlignmentHandlerLibrary() {
         function_from_pointer_to_member(
             &FormatStyle::formal_parameters_alignment)}},
       {AlignableSyntaxSubtype::kPortDeclaration,
-       {AlignmentCellScannerGenerator<PortDeclarationColumnSchemaScanner>(),
+       {AlignmentCellScannerGenerator<PortDeclarationColumnSchemaScanner>(
+            non_tree_column_scanner),
         function_from_pointer_to_member(
             &FormatStyle::port_declarations_alignment)}},
       {AlignableSyntaxSubtype::kClassMemberVariables,
@@ -1091,7 +1126,8 @@ static const AlignmentHandlerMapType& AlignmentHandlerLibrary() {
         function_from_pointer_to_member(
             &FormatStyle::assignment_statement_alignment)}},
       {AlignableSyntaxSubtype::kEnumListAssignment,
-       {AlignmentCellScannerGenerator<EnumWithAssignmentsColumnSchemaScanner>(),
+       {AlignmentCellScannerGenerator<EnumWithAssignmentsColumnSchemaScanner>(
+            non_tree_column_scanner),
         function_from_pointer_to_member(
             &FormatStyle::enum_assignment_statement_alignment)}},
       {AlignableSyntaxSubtype::kDistItem,
--- a/verilog/formatting/formatter_test.cc
+++ b/verilog/formatting/formatter_test.cc
@@ -878,10 +878,48 @@ static constexpr FormatterTestCase kFormatterTestCases[] = {
      "output reg yyy //c2\n"
      " ) ;endmodule:foo\n",
      "module foo (\n"
-     "    input  wire x,  //c1\n"   // aligned
-     "    output reg  yyy  //c2\n"  // trailing comments not aligned
+     "    input  wire x,   //c1\n"  // aligned
+     "    output reg  yyy  //c2\n"  // aligned
      ");\n"
      "endmodule : foo\n"},
+    {"module foo(  input wire x  ,/* c1 */\n"
+     "output reg yyy /* c2 */\n"
+     " ) ;endmodule:foo\n",
+     "module foo (\n"
+     "    input  wire x,   /* c1 */\n"  // aligned
+     "    output reg  yyy  /* c2 */\n"  // aligned
+     ");\n"
+     "endmodule : foo\n"},
+    {"module foo(  input wire x  ,/* c1\n"
+     "c2\n"
+     "c3 */\n"
+     "output reg yyy /* c4 */\n"
+     " ) ;endmodule:foo\n",
+     "module foo (\n"
+     "    input  wire x,   /* c1\n"
+     "c2\n"
+     "c3 */\n"  // TODO: align multiline comments
+     "    output reg  yyy  /* c4 */\n"
+     ");\n"
+     "endmodule : foo\n"},
+    {"module foo(  input wire x  ,/* c1 */\n"
+     "output reg yyy,\n"
+     "output z // c2\n"
+     " ) ;endmodule:foo\n",
+     "module foo (\n"
+     "    input  wire x,    /* c1 */\n"  // aligned
+     "    output reg  yyy,\n"
+     "    output      z     // c2\n"  // aligned
+     ");\n"
+     "endmodule : foo\n"},
+    {"module m(input logic [4:0] foo,  // comment\n"
+     "input logic bar // comment\n"
+     " ) ;endmodule:m\n",
+     "module m (\n"
+     "    input logic [4:0] foo,  // comment\n"  // aligned
+     "    input logic       bar   // comment\n"  // aligned
+     ");\n"
+     "endmodule : m\n"},
     {"module foo(  input wire x  , output yy ) ;endmodule:foo\n",
      "module foo (\n"
      "    input  wire x,\n"  // aligned
@@ -3521,12 +3559,12 @@ static constexpr FormatterTestCase kFormatterTestCases[] = {
      "D=3    // baz\n"
      "}foo_t;",
      "typedef enum logic {\n"
-     "  A = 0,  // foo\n"
-     "  B,  // bar\n"
+     "  A = 0,   // foo\n"
+     "  B,       // bar\n"
      "`ifndef DO_PANIC\n"
      "  C = 42,  // answer\n"
      "`endif\n"
-     "  D = 3  // baz\n"
+     "  D = 3    // baz\n"
      "} foo_t;\n"},
     {// with scalar dimensions
      "typedef enum logic[2]\t{ A=0, B=1 }foo_t;",
@@ -9359,7 +9397,17 @@ static constexpr FormatterTestCase kFormatterTestCasesEnumDeclarations[] = {
      "  kA   = 1,  // value kA\n"
      "  // hello world\n"
      "  kAB  = 2,  // value kAB\n"
-     "  kABC = 3  // value kABC"
+     "  kABC = 3   // value kABC"
+     "\n} x;\n"},
+
+    {"typedef enum { kA=1,// value kA\n"
+     "kAB=2,\n"
+     "kABC=3// value kABC\n"
+     "} x;",
+     "typedef enum {\n"
+     "  kA   = 1,  // value kA\n"
+     "  kAB  = 2,\n"
+     "  kABC = 3   // value kABC"
      "\n} x;\n"},
 
     // Numeric constants are currently flushed left, but maybe todo


wtatarski:
@fangism 
It looks like it helps to solve this issue, but let me know if this is a good way to start.

fangism:
> @fangism
> It looks like it helps to solve this issue, but let me know if this is a good way to start.

You probably don't want to touch `KeepSyntaxTreeTokens` because that impacts what the parser sees, and the parser doesn't accept comment tokens.

Take a look at a [high-level tour of the phases of formatting](https://github.com/google/verible/blob/master/doc/formatter.md).
Start by looking at the TokenPartitionTree for a given file.

Example:

```systemverilog
module m(
	input logic foo,  // comment
	input logic faaaaa,  // comment
	output reg z  // comment
);
endmodule
```

```
$ verible-verilog-format --show_token_partition_tree ~/temp/verilog/formatting/align-port.sv
Full token partition tree:
  { ([<auto>], policy: always-expand) @{0}, (origin: "module m(
	...;
endmodule")
    { ([<auto>], policy: always-expand) @{0,0}, (origin: "module m(
	... comment
);")
      { ([module m (], policy: fit-else-expand) }
      { (>>>>[<auto>], policy: tabular-alignment) @{0,0,1}, (origin: "input logic...utput reg z")
        { (>>>>[input logic foo , // comment], policy: fit-else-expand, (origin: "input logic foo")) }
        { (>>>>[input logic faaaaa , // comment], policy: fit-else-expand, (origin: "input logic faaaaa")) }
        { (>>>>[output reg z // comment], policy: fit-else-expand, (origin: "output reg z")) }
      }
      { ([) ;], policy: fit-else-expand) }
    }
    { ([endmodule], policy: always-expand, (origin: "endmodule")) }
  }
}
```

Note these partitions:

```
        { (>>>>[input logic foo , // comment], policy: fit-else-expand, (origin: "input logic foo")) }
        { (>>>>[input logic faaaaa , // comment], policy: fit-else-expand, (origin: "input logic faaaaa")) }
        { (>>>>[output reg z // comment], policy: fit-else-expand, (origin: "output reg z")) }
```

These are the subjects of alignment, where each line corresponds to a row of tokens.
The [entry point to handling alignment for this group of ports is here](https://cs.opensource.google/verible/verible/+/master:verilog/formatting/align.cc;drc=259321c5bb95882d2ac872ca03de6c8d35fca733;l=1239).

To examine how the alignment handler sees this in table-cells form run (page into `less`):

```
$ verible-verilog-format -v=2 --alsologtostderr ~/temp/verilog/formatting/align-port.sv
...
I1120 18:54:28.371143  881696 align.cc:668] including partition: { (>>>>[input logic foo , // comment], policy: fit-else-expand, (origin: "input  logic foo")) }
I1120 18:54:28.371148  881696 align.cc:668] including partition: { (>>>>[input logic faaaaa , // comment], policy: fit-else-expand, (origin: "input  logic faaaaa")) }
I1120 18:54:28.371152  881696 align.cc:668] including partition: { (>>>>[output reg z // comment], policy: fit-else-expand, (origin: "output reg   z")) }
...

I1120 18:54:28.371622  881696 align.cc:569] unique columns: 3
I1120 18:54:28.371629  881696 align.cc:574] Filling dense matrix from sparse representation
I1120 18:54:28.371635  881696 align.cc:234] FillAlignmentRow
I1120 18:54:28.371641  881696 align.cc:286] end of FillAlignmentRow, row: < input | logic | foo >
I1120 18:54:28.371648  881696 align.cc:234] FillAlignmentRow
I1120 18:54:28.371651  881696 align.cc:286] end of FillAlignmentRow, row: < input | logic | faaaaa >
I1120 18:54:28.371654  881696 align.cc:234] FillAlignmentRow
I1120 18:54:28.371660  881696 align.cc:286] end of FillAlignmentRow, row: < output | reg | z >
...
I1120 18:54:28.371769  881696 align.cc:319] end of ComputeCellWidths, cell sizes:
[0+5, 1+5, 1+3]
[0+5, 1+5, 1+6]
[0+6, 1+3, 1+1]
I1120 18:54:28.371781  881696 align.cc:327] ComputeColumnWidths
...
```

The first section shows the same partitions with `--printtree`, which includes the trailing comments.

The second sections shows that only some of those tokens were considered for alignment space adjustment:

```
I1120 18:54:28.371641  881696 align.cc:286] end of FillAlignmentRow, row: < input | logic | foo >
```

Note that `,` is missing and so is `// comment`.  So far this has been ok, because we are left-aligning on the identifier in the 3rd column, so only spacing to the left identifier needs to be accounted for.

The third section shows N+M where N is the minimum spacing before the first token of a cell, and M is the (compressed) minimum total width of all tokens in each cell.  The differences in sums are the space padding amounts needed to align.

Back to comments: you need to **create a 4th column** that starts at the `// comment`.
Once you do that, the rest of the calculation is automatic.

Currently, the function(s) that create the column boundaries only examine the syntax tree which excludes comments, so you'll have to examine the TokenPartitionTree directly to detect trailing comments.

Hope this helps!

wtatarski:
@fangsim
Thank you for your post! 
My question is whether should I try to include  `, // comment`  to origin  `row: < input | logic | foo >` and  then add new column here https://cs.opensource.google/verible/verible/+/master:verilog/formatting/align.cc;l=307;drc=259321c5bb95882d2ac872ca03de6c8d35fca733 ?
Or maybe should I create similar Visit function for TokenPartitionTree?
Did I get it correctly?

fangism:
> @fangsim
> Thank you for your post!
> My question is whether should I try to include `, // comment` to origin `row: < input | logic | foo >` and then add new column here https://cs.opensource.google/verible/verible/+/master:verilog/formatting/align.cc;l=307;drc=259321c5bb95882d2ac872ca03de6c8d35fca733 ?
> Or maybe should I create similar Visit function for TokenPartitionTree?
> Did I get it correctly?

That location may not work because it can only observe tokens that are in the syntax tree, which excludes comments.

Instead, take a look at this [adapter](https://cs.opensource.google/verible/verible/+/master:common/formatting/align.h;drc=75249d04919c38555bc397880cd67d3b59ace358;bpv=1;bpt=1;l=321), that turns a column scanner class into a [column partitioning function](https://cs.opensource.google/verible/verible/+/master:common/formatting/align.h;drc=75249d04919c38555bc397880cd67d3b59ace358;bpv=1;bpt=1;l=182?gsn=AlignmentCellScannerFunction&gs=kythe%3A%3Flang%3Dc%252B%252B%234r2tRfIakuXuHo7f18D7mhhYc3IhGKQUfd5wintmpvU).
That function does have access to the original range of tokens, including comments.  

You could, for instance, create another language-specific adapter that detects trailing comments, and give them their own [ColumnPositionEntry](https://cs.opensource.google/verible/verible/+/master:common/formatting/align.h;drc=75249d04919c38555bc397880cd67d3b59ace358;bpv=1;bpt=1;l=43?gsn=ColumnPositionEntry&gs=kythe%3A%2F%2Fgithub.com%2Fgoogle%2Fverible%3Flang%3Dc%252B%252B%3Fpath%3Dcommon%2Fformatting%2Falign.h%232GE8JcvNdsgSuPs_mAWkykHvzD-YohvIaURcGM90k_A&gs=kythe%3A%2F%2Fgithub.com%2Fgoogle%2Fverible%3Flang%3Dc%252B%252B%3Fpath%3Dcommon%2Fformatting%2Falign.h%23ColumnPositionEntry%253Averible%2523c%2523iONUJzvPWOG&gs=kythe%3A%2F%2Fgithub.com%2Fgoogle%2Fverible%3Flang%3Dc%252B%252B%3Fpath%3Dcommon%2Fformatting%2Falign.h%23JDNhWLCG-zFUDndxWS-zoqm1tsqSqtsUssLfNO5kZMc&gs=kythe%3A%2F%2Fgithub.com%2Fgoogle%2Fverible%3Flang%3Dc%252B%252B%3Fpath%3Dcommon%2Fformatting%2Falign.h%23h1nEo4-RS6VCMcJAowOfaL_7NsdGqxUjCOdsSWT71Wo).

For example, this could be accomplished by scanning forward from the last column/token observed by the original column scanner class.  [GetRightmostLeaf](https://cs.opensource.google/verible/verible/+/master:common/text/tree_utils.h;drc=259321c5bb95882d2ac872ca03de6c8d35fca733;l=41?q=GetRight&ss=verible%2Fverible) might be useful as a starting point for examining tokens from the last leaf.

Hope this helps too.

wtatarski:
@fangism 
I think that I could add new adapter like here: https://github.com/antmicro/verible/blob/align-comments-in-port-list/common/formatting/align.h#L296
which uses `UnwrappedLine` and `TokenRange()` which sees EOL comments. However it should inherits from `PortDeclarationColumnSchemaScanner` field `sparse_columns_` where fourth column needs to be inserted. Of course it should be merged into the flow like the rest of scanners, but this is WiP at the moment.
Anyway my question is whether is it a good approach? I am trying to add a new column out of the flow here: https://github.com/antmicro/verible/blob/align-comments-in-port-list/common/formatting/align.cc#L564
just to check if I am doing it in a good way, but it doesn't work. I think that we would we have to write new function for reserving columns from Tokens, but please let me know if this is a correct approach.

fangism:
> @fangism
> I think that I could add new adapter like here: https://github.com/antmicro/verible/blob/align-comments-in-port-list/common/formatting/align.h#L296
> which uses `UnwrappedLine` and `TokenRange()` which sees EOL comments. However it should inherits from `PortDeclarationColumnSchemaScanner` field `sparse_columns_` where fourth column needs to be inserted. Of course it should be merged into the flow like the rest of scanners, but this is WiP at the moment.
> Anyway my question is whether is it a good approach? I am trying to add a new column out of the flow here: https://github.com/antmicro/verible/blob/align-comments-in-port-list/common/formatting/align.cc#L564
> just to check if I am doing it in a good way, but it doesn't work. I think that we would we have to write new function for reserving columns from Tokens, but please let me know if this is a correct approach.

The `sparse_columns_` field only exists for the sake of [`ScanPartitionForAlignmentCells`](https://cs.opensource.google/verible/verible/+/master:common/formatting/align.h;bpv=1;bpt=1;l=297?gsn=ScanPartitionForAlignmentCells&gs=kythe%3A%2F%2Fgithub.com%2Fgoogle%2Fverible%3Flang%3Dc%252B%252B%3Fpath%3Dcommon%2Fformatting%2Falign.h%23_APH_9Lh54BLCv-i-j_T3-5pK4WQgBlmuYxLOIGhskU), whose sole purposes is to help [`AlignmentCellScannerGenerator`](https://cs.opensource.google/verible/verible/+/master:common/formatting/align.h;drc=e50e3809742930e0ce2a2981596b3572ab985b27;l=323).  `ScanPartitionForAlignmentCells` currently only operates on the syntax tree, but could be extended to accept a handler function to cover the non-tree tokens that follow (e.g. comments), and append columns to the result of [`scanner.SparseColumns()`](https://cs.opensource.google/verible/verible/+/master:common/formatting/align.h;drc=e50e3809742930e0ce2a2981596b3572ab985b27;l=305).  By default, a non-tree token handler would append no columns.  
So as you suggested, we need a `std::function<std::vector<ColumnPositionEntry>(TokenRange)>` for those non-tree tokens.

Example mock-up:

```C++
template <class ScannerType>
std::vector<ColumnPositionEntry> ScanPartitionForAlignmentCells(
    const TokenPartitionTree& row) {
   // same as before, unmodified.  only uses syntax tree.
}

// Also uses a non-tree column scanning function, whose columns are concatenated with those from the above function.
template <class ScannerType>
std::vector<ColumnPositionEntry> ScanPartitionForAlignmentCells_IncludingTrailingNonTreeTokens(
    const TokenPartitionTree& row,
    const std::function<std::vector<ColumnPositionEntry>(TokenRange)>& non_tree_column_scanner) {
  // re-use existing scanner
  const auto column_entries_from_syntax_tree = ScanPartitionForAlignmentCells<ScannerType>(row);
  const UnwrappedLine& unwrapped_line = row.Value();
  const Symbol* origin = ABSL_DIE_IF_NULL(unwrapped_line.Origin());
  // Identify the last token covered by the origin tree.
  // Construct a TokenRange 'remainder' that starts after the last tree token, and spans the rest of the UnwrappedLine.
  const auto trailing_column_entries = non_tree_column_scanner(remainder);
  // concatenate column_entries_from_syntax_tree + trailing_column_entries and return.
}

// Adapter that accepts a tree scanner class and a trailing token scanner function.
// Could be an overload of existing AlignmentCellScannerGenerator or named differently, up to you.
template <class ScannerType>
AlignmentCellScannerFunction AlignmentCellScannerGenerator(const std::function<std::vector<ColumnPositionEntry>(TokenRange)>& non_tree_column_scanner) {
  return [&non_tree_column_scanner](const TokenPartitionTree& row) {
    return ScanPartitionForAlignmentCells_IncludingTrailingNonTreeTokens<ScannerType>(row, non_tree_column_scanner);
  };
}
```

The advantage of this approach is that it doesn't require extending any classes, and re-uses most of the existing code unmodified, with minimal duplication.  You can then use the same comment-token scanner function for the language constructs that want the same handling.

How does this sound?

wtatarski:
Thank you for your advices. However I have one more question how to construct function for creating columns according to tokens. In essence I'm supporting on this function: https://cs.opensource.google/verible/verible/+/master:common/formatting/align.cc;l=148;drc=75249d04919c38555bc397880cd67d3b59ace358;bpv=1;bpt=1 
and I wonder how to calculate `path`. Should it be calculated as a relative to `path` from the last element from `column_entries_from_syntax_tree` ?
Because at this moment I see that I have added fourth column, but it fails later on this CHECK: https://cs.opensource.google/verible/verible/+/master:common/formatting/align.cc;l=260;drc=75249d04919c38555bc397880cd67d3b59ace358;bpv=1;bpt=1

fangism:
> Thank you for your advices. However I have one more question how to construct function for creating columns according to tokens. In essence I'm supporting on this function: https://cs.opensource.google/verible/verible/+/master:common/formatting/align.cc;l=149;drc=75249d04919c38555bc397880cd67d3b59ace358;bpv=1;bpt=1
> and I wonder how to calculate `path`. Should it be calculated as a relative to `path` from the last element from `column_entries_from_syntax_tree` ?
> Because at this moment I see that I have added fourth column, but it fails later on this CHECK: https://cs.opensource.google/verible/verible/+/master:common/formatting/align.cc;l=260;drc=75249d04919c38555bc397880cd67d3b59ace358;bpv=1;bpt=1

Ah yes, Path is only meaningful coming from a syntax tree node, but required for `ColumnPositionEntry`.  
It needs to be:
1) consistent across rows of the same syntax type (e.g. for all port declarations).
2) never collide with real path vectors from the syntax tree.
Since the structure of syntax tree nodes of the same type is intended to be the same, setting the path to `[N]` where N is the number of direct children of the syntax tree node (`Origin()`), will point to one-past-the-last-child should meet the criteria.

Let me know if that works.

wtatarski:
It looks like still it fails on this CHECK which I mentioned in previous post. 
For Origin `kPortDeclaration` I got 6 children and I set `path` in `ColumnPositionEntry` on this value.

fangism:
> It looks like still it fails on this CHECK which I mentioned in previous post.
> For Origin `kPortDeclaration` I got 6 children and I set `path` in `ColumnPositionEntry` on this value.

I traced backwards from `token_end` in that check, it is coming from this function:

https://cs.opensource.google/verible/verible/+/master:common/formatting/align.cc;drc=75249d04919c38555bc397880cd67d3b59ace358;bpv=1;bpt=1;l=456?gsn=GetMutableFormatTokenRange&gs=kythe%3A%2F%2Fgithub.com%2Fgoogle%2Fverible%3Flang%3Dc%252B%252B%3Fpath%3Dcommon%2Fformatting%2Falign.cc%23UWnl6A9IVeyd9G1q6SZ9iu8gRvykIfCm7ZWaYMuyNsU

The comments in that function tell me that I had intentionally shrunk `range_end` to line up with `last_token` which excludes trailing comments.  I did this initially to prevent accidental changing of spaces on text outside the syntax tree Origin().
It might be time to remove that range-shrinking operation, now that we are handling trailing comments.

wtatarski:
Now it works for aligning comments in ports list, but some tests fail, so this needs to be fixed.

fangism:
> Now it works for aligning comments in ports list, but some tests fail, so this needs to be fixed.

Good to hear.  Yes, existing test failures are expected.  Some reasons tests might fail:

* Test cases may have included comments that were not expected to align before
* As a result of additional alignment, the effective table width exceeds the column limit, which may force fall-back behavior to unaligned output for a given section.

wtatarski:
I suspect that the problem is with the last port declaration, which doesn't contain `,`. It changes total aligned column width and some tests fail e.g.
```
module foo (
    input x,
    output y[2:0]
);
endmodule
```
In this example formatter adds additional space between `y` and `[2:0]`

fangism:
> I suspect that the problem is with the last port declaration, which doesn't contain `,`. It changes total aligned column width and some tests fail e.g.
> 
> ```
> module foo (
>     input x,
>     output y[2:0]
> );
> endmodule
> ```
> 
> In this example formatter adds additional space between `y` and `[2:0]`

Previously, commas were considered "trailing text" which all became part of the last column.  We never needed to adjust their spacing or column association, so we were able to get away with being lazy, and still get reasonable looking output.  With the introduction of a comment column, we need to do a little more work.

Consider the following example, which reflect what I think should happen ideally after column detection:

|col[0] |col[1]|col[2]|col[3]|
|---|---|---|---|
| `input` | `x,` | | `//comment` |
| `output` | `z` | `[2:0],` | `//comment` |
| `output` | `y` | `[2:0]` | `//comment` |
| `output` | `v` |  | `//comment` |


(Do you agree?)

Note that in the rows with commas, we attach the comma to the last column, be it identifier ([1]) or dimension ([2]).  To accomplish this it may take a little token range-end adjustment code for the last non-blank cell in each row.

wtatarski:
I think that your requirement which you described in the table is already satisfied: 
```
I common/formatting/align.cc:288] end of FillAlignmentRow, row: < input | x , |  |  |  |  | //c1 >
I common/formatting/align.cc:288] end of FillAlignmentRow, row: < output | y |  | [ | 2 : 0 | ] , | //c2 >
I common/formatting/align.cc:288] end of FillAlignmentRow, row: < output | z |  | [ | 4 : 0 | ] | //c3 >
```
Comma is already in one column with a previous token. I think that there is something wrong with calculating cell width (when dimensions are used):
```
[0+5, 1+2, 0+0, 0+0, 0+0, 0+0, 2+4]
[0+6, 1+1, 0+0, 0+1, 0+3, 0+2, 2+4]
[0+6, 1+1, 0+0, 0+1, 0+3, 0+1, 2+4]
```
and the result is:
```
    input x,  //c1
    output y[2:0],  //c2
    output z[4:0]  //c3
```

fangism:
> I think that your requirement which you described in the table is already satisfied:
> 
> ```
> I common/formatting/align.cc:288] end of FillAlignmentRow, row: < input | x , |  |  |  |  | //c1 >
> I common/formatting/align.cc:288] end of FillAlignmentRow, row: < output | y |  | [ | 2 : 0 | ] , | //c2 >
> I common/formatting/align.cc:288] end of FillAlignmentRow, row: < output | z |  | [ | 4 : 0 | ] | //c3 >
> ```
> 
> Comma is already in one column with a previous token. I think that there is something wrong with calculating cell width (when dimensions are used):
> 
> ```
> [0+5, 1+2, 0+0, 0+0, 0+0, 0+0, 2+4]
> [0+6, 1+1, 0+0, 0+1, 0+3, 0+2, 2+4]
> [0+6, 1+1, 0+0, 0+1, 0+3, 0+1, 2+4]
> ```
> 
> and the result is:
> 
> ```
>     input x,  //c1
>     output y[2:0],  //c2
>     output z[4:0]  //c3
> ```

The `[l+w]` cell widths above look correct to me.

First, verify around [this block](https://cs.opensource.google/verible/verible/+/master:common/formatting/align.cc;drc=75249d04919c38555bc397880cd67d3b59ace358;l=615) that you aren't seeing width-exceeding behavior (when _sum(max-widths)_ exceeds column_limit).

If that looks good, then look inside the [function that calculates pre-token spacing to achieve alignment](https://cs.opensource.google/verible/verible/+/master:common/formatting/align.cc;drc=75249d04919c38555bc397880cd67d3b59ace358;bpv=1;bpt=1;l=366?gsn=ComputeAignedRowSpacings&gs=kythe%3A%2F%2Fgithub.com%2Fgoogle%2Fverible%3Flang%3Dc%252B%252B%3Fpath%3Dcommon%2Fformatting%2Falign.cc%23VzI1IKbaUHLrW4uJtDQGCjxB3u1Ta_1bmWs_A8eKIsY).
Each time a blank/empty cell is encountered in each row (scanning from left to right), it accumulates "spacing debt" which is "paid-off" at the next token in a non-blank cell.

wtatarski:
It looks like that [this block](https://cs.opensource.google/verible/verible/+/master:common/formatting/align.cc;drc=75249d04919c38555bc397880cd67d3b59ace358;l=615) doesn't change anything and there is no width-exceeding behavior.

I wonder whether we should calculate comma to column width. For example [here](https://cs.opensource.google/verible/verible/+/master:common/formatting/align.cc;l=126;drc=75249d04919c38555bc397880cd67d3b59ace358) we can detect if last token in cell is a `,` and decrease width by 1. It helped me get rid of additional space between `y` and `[2:0]` in that case:

```
module foo (
    input  x,
    output y[2:0]
);
endmodule
```

fangism:
> It looks like that [this block](https://cs.opensource.google/verible/verible/+/master:common/formatting/align.cc;drc=75249d04919c38555bc397880cd67d3b59ace358;l=615) doesn't change anything and there is no width-exceeding behavior.
> 
> I wonder whether we should calculate comma to column width. For example [here](https://cs.opensource.google/verible/verible/+/master:common/formatting/align.cc;l=126;drc=75249d04919c38555bc397880cd67d3b59ace358) we can detect if last token in cell is a `,` and decrease width by 1. It helped me get rid of additional space between `y` and `[2:0]` in that case:
> 
> ```
> module foo (
>     input  x,
>     output y[2:0]
> );
> endmodule
> ```

I see what you're saying, if you engulf the `,` with the next column to the left, it counts towards the width and adds an extra space.

To avoid that I think we could adjust the token range boundaries per row such that the  `,` gets added to the first _blank_ cell that follows a non-blank cell, like this:

|col | col | col | col | col |
|----|----|----|----|----|
| `input` | `x` | `,`  |  | `//comment` |
| `output` | `z` | `[2:0]` | `,` | `//comment` |
| `output` | `y` | `[2:0]` |  | `//comment` |
| `output` | `v` |   |  | `//comment` |

Now the `,`s won't interfere with the sizing of the column to the left.  A single character is also short enough that it will not influence the width of the column it shares, _width(`,`) < width(`[2:0]`)_.

One way to do this is to start by letting commas have their own column (`ReserveNewColumn` using their syntax tree path like all the others):

|col | col | col | col | col |
|----|----|----|----|----|
| `input` | `x` | | `,`  | `//comment` |
| `output` | `z` | `[2:0]` | `,` | `//comment` |
| `output` | `y` | `[2:0]` |  | `//comment` |
| `output` | `v` |   |  | `//comment` |

Notice that only the first row is different.
_After_ all of the unique columns have been established for a group of rows, do you then correct the token ranges in the matrix to look like the first table above.
You have to maintain the invariant that the token ranges in a cell span the whole row's tokens, and are gap-less with _end(left) == begin(right)_ for all neighboring cells, even empty ones.

If it helps, you can designate a new column attribute to help with this selection, then the generic code doesn't have to be aware of `,`s or anything language-specific.

Does this sound reasonable?

wtatarski:
I added reserving new column for `,` in a new non_tree_column_scanner. It looks like first part is OK:
```
I common/formatting/align.cc:295] end of FillAlignmentRow, row: < input | x |  |  |  |  | , | // c1 >
I common/formatting/align.cc:295] end of FillAlignmentRow, row: < output | y |  | [ | 2 : 0 | ] | , | // c2 >
I common/formatting/align.cc:295] end of FillAlignmentRow, row: < output | z |  | [ | 4 : 0 | ] |  | // c3 >
```
However now `x` and comments are not aligned:
```
module foo (
    input x,  // c1
    output y[2:0],  // c2
    output z[4:0]  // c3
);
endmodule
```
So should I move `,` to the column right after `x`?

fangism:
> I added reserving new column for `,` in a new non_tree_column_scanner. It looks like first part is OK:
> 
> ```
> I common/formatting/align.cc:295] end of FillAlignmentRow, row: < input | x |  |  |  |  | , | // c1 >
> I common/formatting/align.cc:295] end of FillAlignmentRow, row: < output | y |  | [ | 2 : 0 | ] | , | // c2 >
> I common/formatting/align.cc:295] end of FillAlignmentRow, row: < output | z |  | [ | 4 : 0 | ] |  | // c3 >
> ```
> 
> However now `x` and comments are not aligned:
> 
> ```
> module foo (
>     input x,  // c1
>     output y[2:0],  // c2
>     output z[4:0]  // c3
> );
> endmodule
> ```
> 
> So should I move `,` to the column right after `x`?

Yes, to aim for:


col | col | col | col | col
-- | -- | -- | -- | --
`input` | `x` | `,` |   | `//comment`
`output` | `z` | `[2:0]` | `,` | `//comment`
`output` | `y` | `[2:0]` |   | `//comment`
`output` | `v` |   |   | `//comment`

wtatarski:
I moved `,` to the desired place:
```
end of FillAlignmentRow, row: < input | x | , |  |  |  |  | // c1 >
end of FillAlignmentRow, row: < output | y |  | [ | 2 : 0 | ] | , | // c2 >
end of FillAlignmentRow, row: < output | z |  | [ | 4 : 0 | ] |  | // c3 >
```
but it seems it doesn't changed anything in this particular example. I added moving of comma [at the end of FillAlignmentRow]( https://cs.opensource.google/verible/verible/+/master:common/formatting/align.cc;l=286;drc=75249d04919c38555bc397880cd67d3b59ace358). Maybe should it be somewhere else?

fangism:
> I moved `,` to the desired place:
> 
> ```
> end of FillAlignmentRow, row: < input | x | , |  |  |  |  | // c1 >
> end of FillAlignmentRow, row: < output | y |  | [ | 2 : 0 | ] | , | // c2 >
> end of FillAlignmentRow, row: < output | z |  | [ | 4 : 0 | ] |  | // c3 >
> ```
> 
> but it seems it doesn't changed anything in this particular example. I added moving of comma [at the end of FillAlignmentRow](https://cs.opensource.google/verible/verible/+/master:common/formatting/align.cc;l=286;drc=75249d04919c38555bc397880cd67d3b59ace358). Maybe should it be somewhere else?

**Almost.  Very close.**  We don't want the comma to occupy an otherwise blank column; we want the comma to share a column that is occupied by at least one other non-empty cell (in our working example, the trailing dimensions).  One place to do this would be [after all calls to `FillAlignmentRow` are done](https://cs.opensource.google/verible/verible/+/master:common/formatting/align.cc;drc=75249d04919c38555bc397880cd67d3b59ace358;l=584).

Recall that the matrix is built from only _observed_ "syntax tree paths" of columns as seen by the [column schema](https://cs.opensource.google/verible/verible/+/master:common/formatting/align.cc;drc=75249d04919c38555bc397880cd67d3b59ace358;bpv=1;bpt=1;l=184).
Once the token range matrix is populated, there should not be any empty columns, and at this point the columns are represented by simple indices.  In this table form, moving a token range from cell-to-cell should not create any new columns by accident.

wtatarski:
Hmm I moved comma in the first row to the same column as `]`, but it also doesn't change anything in the output. Actually moving this comma to any other column (like `[` or `2:0`) causes no changes in output. However I see the change in the logs:
```
end of FillAlignmentRow, row: < input | x |  |  |  | , |  | // c1 >
end of FillAlignmentRow, row: < output | y |  | [ | 2 : 0 | ] | , | // c2 >
end of FillAlignmentRow, row: < output | z |  | [ | 4 : 0 | ] |  | // c3 >
```
Some additional modifications are required?

fangism:
> Hmm I moved comma in the first row to the same column as `]`, but it also doesn't change anything in the output. Actually moving this comma to any other column (like `[` or `2:0`) causes no changes in output. However I see the change in the logs:
> 
> ```
> end of FillAlignmentRow, row: < input | x |  |  |  | , |  | // c1 >
> end of FillAlignmentRow, row: < output | y |  | [ | 2 : 0 | ] | , | // c2 >
> end of FillAlignmentRow, row: < output | z |  | [ | 4 : 0 | ] |  | // c3 >
> ```
> 
> Some additional modifications are required?



> Hmm I moved comma in the first row to the same column as `]`, but it also doesn't change anything in the output. Actually moving this comma to any other column (like `[` or `2:0`) causes no changes in output. However I see the change in the logs:
> 
> ```
> end of FillAlignmentRow, row: < input | x |  |  |  | , |  | // c1 >
> end of FillAlignmentRow, row: < output | y |  | [ | 2 : 0 | ] | , | // c2 >
> end of FillAlignmentRow, row: < output | z |  | [ | 4 : 0 | ] |  | // c3 >
> ```
> 
> Some additional modifications are required?

Hm, ok, I'm taking a step back.
I think that we need a new _unaligned_ primitive column type for the `,`s and possible other list-delimiters (let's call it `kUnalignedLeft`).
That is, we want it to hug to the left without affecting spacing calculations and let it overflow into the next column if it must.

[Around here where we decided whether how to pad spacing on the left or right of text](https://cs.opensource.google/verible/verible/+/master:common/formatting/align.cc;drc=75249d04919c38555bc397880cd67d3b59ace358;l=387), we need a third option that is _"print the token(s) in this cell immediately (unaligned), and reduce `accrued_spaces` accordingly (without going below 0)"_.  This means turning the [`flush_left`](https://cs.opensource.google/verible/verible/+/master:common/formatting/align.h;drc=75249d04919c38555bc397880cd67d3b59ace358;l=35) attribute into a 3-way enum.

With this idea, the commas can remain in their own column (using the same syntax tree path), but will print _unaligned_ without affecting other cells' calculations.  By continuing to keep track of `accrued_spaces`, the column of comments that follow commas will become aligned.  I think this change is fairly localized.

Thoughts?

hzeller:
Looks like you guys were deep in thought until the xmas break hit. Were there any new thoughts ?

fangism:
> Looks like you guys were deep in thought until the xmas break hit. Were there any new thoughts ?

Reviewed this thread.  My suggestions from 2020.12.22 still stand.  Summarizing: en route to supporting comment alignment, we need a way for trailing commas to be unaligned-hug-left (a third column style option).  This could be its own PR before handling comments, but doesn't have to be.

wtatarski:
I come back to this issue. I added handling of columns with delimiter by adding flag which matches whether column contains delimiter or not.  It is not exactly the same way which @fangism  wrote, but I think it is good approach for start. 
Now I see 2 errors in tests:

- there is a problem when the latest port (the one which doesn't contain `,`) has the longest width. Then the comments are aligned with 3 spaces.
- after introducing these new changes there is an issue with typedef enum declaration aligning.

common/formatting/align.cc
```diff
@@ -323,14 +325,42 @@ static void ComputeCellWidths(AlignmentMatrix* matrix) {
 typedef std::vector<AlignedColumnConfiguration> AlignedFormattingColumnSchema;
 
 static AlignedFormattingColumnSchema ComputeColumnWidths(
-    const AlignmentMatrix& matrix) {
+    const AlignmentMatrix& matrix,
+    const std::vector<AlignmentColumnProperties> column_properties) {
   VLOG(2) << __FUNCTION__;
   AlignedFormattingColumnSchema column_configs(matrix.front().size());
+
+  // Check which cell before delimiter is the longrest
```
tgorochowik, line 333: longest

verilog/formatting/align.cc
```diff
@@ -1044,6 +1044,40 @@ std::function<MemberType(const StructType&)> function_from_pointer_to_member(
...
+std::vector<verible::ColumnPositionEntry> non_tree_column_scanner(
+    verible::TokenRange token_range, const verible::SyntaxTreeNode& node,
+    long unsigned int max_children_size) {
+  std::vector<verible::ColumnPositionEntry> trailing_column_entries;
+  auto tag = NodeEnum(node.Tag().tag);
+  VLOG(2) << __FUNCTION__ << ", node: " << tag;
+
+  for (auto token : token_range) {
+    switch (token.token_enum()) {
+      case 44: {
```
tgorochowik, line 1056: 44? can we use proper `TK_` define?

common/formatting/align.cc
```diff
@@ -94,6 +94,8 @@ struct AlignmentCell {
   // as a space-only column, usually no more than 1 space wide.
   int left_border_width = 0;
 
+  int TotalWidth() const { return left_border_width + compact_width; }
```
tgorochowik, line 97: I think this could be a static if it's used only as a helper here

tgorochowik (commented):
I think it looks good in general, please see some small comments.

common/formatting/align.cc
wtatarski: Fixed

verilog/formatting/align.cc
wtatarski: Actually I don't see any `TK_` macro for `,`. I added a comment that 44 is `,`

common/formatting/align.cc
wtatarski: Adding static keyword causes an error when compiling.

wtatarski:
@hzeller  Could you take a look at this PR?

verilog/formatting/align.cc
tgorochowik: If no define/enum is generated for this then we need to make sure to add that generation, using a literal here seems pretty fragile

tgorochowik: I just realized 44 is ascii for comma.

It should be okay to use
```
case ',':
```
then

wtatarski: Fixed

verilog/formatting/formatter_test.cc
```diff
@@ -878,8 +878,8 @@ static constexpr FormatterTestCase kFormatterTestCases[] = {
      "output reg yyy //c2\n"
      " ) ;endmodule:foo\n",
      "module foo (\n"
-     "    input  wire x,  //c1\n"   // aligned
-     "    output reg  yyy  //c2\n"  // trailing comments not aligned
+     "    input  wire x,   //c1\n"  // aligned
+     "    output reg  yyy  //c2\n"  // aligned
```
tgorochowik, line 882: Will it also work with `/* these type of comments */`?
Please add regular tests with such comments at the end of line and maybe one test when a multi line comment is actually occupying multiple lines to make sure the formatter doesn't crash.

verilog/formatting/formatter_test.cc
```diff
@@ -9347,7 +9347,7 @@ static constexpr FormatterTestCase kFormatterTestCasesEnumDeclarations[] = {
      "  kA   = 1,  // value kA\n"
      "  // hello world\n"
      "  kAB  = 2,  // value kAB\n"
-     "  kABC = 3  // value kABC"
+     "  kABC = 3   // value kABC"
```
tgorochowik, line 9350: Maybe also add a test case with no comment in between? I mean like:
```
  kAB  = 2, // comment
  kABC = 3,
  kABCD // comment
```

verilog/formatting/formatter_test.cc
wtatarski: Added handling of comment blocks `/* */ ` and tests for them.
Aligning multiline comments doesn't work however it doesn't cause crash.

verilog/formatting/formatter_test.cc
wtatarski: I added above test.

tgorochowik (approved):
LGTM, thanks for the updates!

common/formatting/align.cc
```diff
@@ -323,14 +325,42 @@ static void ComputeCellWidths(AlignmentMatrix* matrix) {
 typedef std::vector<AlignedColumnConfiguration> AlignedFormattingColumnSchema;
 
 static AlignedFormattingColumnSchema ComputeColumnWidths(
-    const AlignmentMatrix& matrix) {
+    const AlignmentMatrix& matrix,
+    const std::vector<AlignmentColumnProperties> column_properties) {
```
hzeller, line 329: make sure to pass vectors by reference so that they are not copied.

verilog/formatting/formatter_test.cc
```diff
@@ -878,8 +878,38 @@ static constexpr FormatterTestCase kFormatterTestCases[] = {
...
+     "endmodule : foo\n"},
+    {"module foo(  input wire x  ,/* c1 */\n"
+     "output reg yyy,\n"
+     "output z // c2\n"
+     " ) ;endmodule:foo\n",
+     "module foo (\n"
+     "    input  wire x,    /* c1 */\n"  // aligned
+     "    output reg  yyy,\n"
+     "    output      z     // c2\n"  // aligned
      ");\n"
```
hzeller, line 913: Maybe also adding a test with a bit-width mixed in ?
```systemverilog
  input logic [4:0] foo,  // comment
  input logic bar, // comment
```

common/formatting/align.h
```diff
@@ -305,6 +308,68 @@ std::vector<ColumnPositionEntry> ScanPartitionForAlignmentCells(
   return scanner.SparseColumns();
 }
 
+// Also uses a non-tree column scanning function, whose columns are
```
hzeller, line 311: Probably a bit more detailed documentation needed so that if we look at it in half a year we still know what this is all about :)

Describe what the parameters are doing similar to above (_a row correspondents ..._) but also of course what the `non_tree_column_scanner` callback is , what it is expected to do etc.

common/formatting/align.h
```diff
@@ -324,6 +389,16 @@ AlignmentCellScannerFunction AlignmentCellScannerGenerator() {
   };
 }
 
+template <class ScannerType>
```
hzeller, line 392: This also wants a comment.

hzeller (commented):
In general, looks good, but I suggest to not miss out on adding comments in header files - this is the public API which anyone modifying the code will go through and it should be clear reading a comment for a method/function what it expects as parameters and what it does.

common/formatting/align.cc
wtatarski: Fixed

verilog/formatting/formatter_test.cc
wtatarski: Added test.

common/formatting/align.h
wtatarski: Added comment

common/formatting/align.h
wtatarski: Fixed

hzeller (approved):
LGTM

PR #620, opened 2020-12-05
Fixing release publishing

### LINKED ISSUE #619: Releases seem to be broken
It seems like the uploading github releases is not working.

mithro:
Fixes #619.

### DIFF
--- a/.github/workflows/success.sh
+++ b/.github/workflows/success.sh
@@ -34,6 +34,7 @@ bin)
     # Create a tag of form v0.0-183-gdf2b162-20191112132344
     rm -rf releasing/out
     git tag "$TAG" || true
+    echo "::set-output name=TAG::$(echo $TAG)"
     ls -l /tmp/releases
     ;;
 
--- a/.github/workflows/success.sh
+++ b/.github/workflows/success.sh
@@ -15,7 +15,7 @@
 
 set -x
 set -e
-export TAG=${TAG:-$(git rev-parse --short "$GITHUB_SHA")}
+export TAG=${TAG:-$(git describe --match=v*)}
 
 git config --local user.name "Deployment Bot"
 git config --local user.email "verible-dev@googlegroups.com"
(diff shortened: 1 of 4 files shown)


fangism (commented):
Thanks for helping to fix this!

.github/workflows/success.sh
```diff
@@ -34,6 +34,7 @@ bin)
     # Create a tag of form v0.0-183-gdf2b162-20191112132344
     rm -rf releasing/out
     git tag "$TAG" || true
+    echo "::set-output name=TAG::$(echo $TAG)"
```
umarcor, line 37: I suggest using `echo "TAG=$TAG" >> $GITHUB_ENV` instead. See https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable.

mithro: Done!

mithro:
@hzeller / @fangism - This should be ready to merge.

.github/workflows/success.sh
mithro: This didn't seem to work - https://github.com/mithro/verible/actions/runs/403055934

fangism:
> @hzeller / @fangism - This should be ready to merge.

Latest comment says 'not quite' :).  Ping me again when ready.

.github/workflows/success.sh
umarcor: See https://github.com/mithro/verible/pull/2

umarcor:
Note that tag 'master' and the corresponding release should be removed from the repo after this is merged.

mithro:
@hzeller / @fangism - I think we should merge this and then @umarcor can send a pull request with the update.

fangism:
> @hzeller / @fangism - I think we should merge this and then @umarcor can send a pull request with the update.

Ok, initiating merge/import...

mithro:
Looks like after #625 was fixed, this change started working! https://github.com/google/verible/releases
Screenshot from 2020-12-08 12-34-57

Issue #621, opened 2020-12-05
Generate the list of GitHub release CI jobs automagically rather than hard coding in the workflow


mithro:
I believe it is possible to have a GitHub Action generate follow up GitHub Action steps. It would be nice if the GitHub Action for building releases just understood the currently supported releases rather than requiring the workflow to be updated.

mithro:
I believe @umarcor might know how to do this.

umarcor:
So, yes and no. It is possible to generate/execute a set of tasks and wrap them in collapsible log blocks. See, for instance, step "*Build and test GHDL in containers*" in https://github.com/ghdl/ghdl/runs/1515138857?check_suite_focus=true or the test script in fomu-workshop: https://github.com/im-tomu/fomu-workshop/runs/1509392137?check_suite_focus=true#step:10:8. However, AFAIAA it is not possible to generate jobs or steps programmatically. That's a purely descriptive thing: what you write in the YAML is what is picked by the web GUI. If that was to be done, some preprocessor/sed would be required as a commit hook.

Overall, I think that listing the release jobs explicitly is not a big deal. That allows better visualisation and it also allows limiting the number of parallel jobs. However, the CI plumbing feels awkward to me. It seems that all the dockerfiles and archives are generated through a shell scrip, in each job; even though only one of the subdirs is used. Then, each job pushes one artifact and attempts to create a release (with one asset).

tgorochowik:
You can generate the matrix programmatically and then use it to generate jobs dynamically:
See this as an example: https://github.com/alainmarcel/uhdm-integration/blob/master/.github/workflows/main.yml

One thing to note is that there is a limit of 256 jobs per matrix.

umarcor:
@tgorochowik I was not aware of that. Thanks a lot for the reference!

PR #622, opened 2020-12-08
[RFC] CI overhaul

umarcor:
Ref #619 #621

This PR shows a different approach to CI and how container images are defined and used.

---

Currently, all the Dockerfiles are generated in a single shell/bash script, using environment variables for setting versions. That feels unidiomatic. In this PR, Dockerfiles are defined in separate files, which can be edited without the need to escape due to bash constraints. Differences in procedures between versions of the OSs are handled through dockerfile stages. At the same time, version variables are handled through build arguments.

---

The current Dockerfile generation strategy cats several sections in a single Dockerfile for each target. Instead, this PR defines two images for each target: 'basic' and 'bazel'. The former contains all the dependencies, except Bazel. The latter is based on the former and includes Bazel.

---

The current build procedure executes a single `docker build` command, where all the dependencies are installed and Verible is built. Then, a dummy container is created just for extracting the artifact. Images and container are not saved/reused. In this PR, two `docker build` commands are executed, one for 'basic' and another one for 'bazel' images. Then, Verible is built in a temporary container.

The rationale is that those 'basic' and 'bazel' images do not need to be built each time. Those can be updated in an scheduled workflow. Then, the regular workflow would only need to pull one image and build Verilator in it. At the same time, any contributor can pull the image and build Verilator, without installing all the dependencies each time.

This was not implemented yet, because a registry needs to be used and I want to have feedback first. Nonetheless, the workflow in this PR shows three steps.

---

Instead of having multiple jobs trying to create a release, in this PR an additional job is defined, which is to be executed after all the build jobs finish. The additional job (named Release) downloads all the artifacts, creates a release and uploads all the assets at once. This job is only executed on branch master, and not for PRs.

/cc @mithro


umarcor:
@googlebot I signed it!

.github/workflows/verible-ci.yml
```diff
@@ -4,99 +4,140 @@ on:
...
+      with:
+       # Download complete repository + tags
+       fetch-depth: 0
+
+    - run: docker pull ${{ matrix.os }}
+
+    - name: Build container image 'basic'
+      run: source ./.github/settings.sh; ./releasing/docker-run.sh basic
+
+    - name: Build container image 'bazel'
```
mithro, line 95: I don't quite get this step?

mithro (commented):
I like a lot of this pull request, need to take a little bit of a closer look however.

.github/workflows/verible-ci.yml
umarcor: Depending on the OS version, the 'basic' image is built with one or other last stage in the multistage dockerfile. Then, depending on the OS version too, Bazel needs to be installed using one or other procedure. Therefore, instead of trying to handle the complexity in a single dockerfile, it is easier to have it in two. In practice, image 'basic' is a temporary cache of image 'bazel', that's why one step is executed right after the other.

My end goal is to execute "Build container image 'basic'" and "Build container image 'bazel'" in a different workflow (probably once a week) and publish 'bazel' images only to a registry (tagged as say `google/verible:ubuntu-bionic_build`). Then, the "regular" CI run would be reduced to building Verible in those pre-built container images.

mithro:
Can you rebase onto master?

.github/workflows/verible-ci.yml
```diff
@@ -4,99 +4,140 @@ on:
...
+    steps:
+
+    - uses: actions/checkout@v2
+      with:
+       # Download complete repository + tags
+       fetch-depth: 0
+
+    - run: docker pull ${{ matrix.os }}
+
+    - name: Build container image 'basic'
```
mithro, line 92: I would call this `base` and maybe `Creating "base" container image for ${{ matrix.os }} (to build Verible inside)` and `Adding "bazel" to "base" container image for ${{ matrix.os }}`?

umarcor:
I can, but not today. It's not straightforward because install and success scripts are moved into the workflow. I'd rather rebase after #623 and #624.

.github/workflows/verible-ci.yml
```diff
@@ -4,99 +4,140 @@ on:
...
+      run: source ./.github/settings.sh; ./releasing/docker-run.sh basic
```
mithro, line 93: `./releasing/docker-run.sh $OS prepare-image-base`

.github/workflows/verible-ci.yml
```diff
@@ -4,99 +4,140 @@ on:
...
+
+    - name: Build container image 'bazel'
+      run: source ./.github/settings.sh; ./releasing/docker-run.sh bazel
```
mithro, line 96: `./releasing/docker-run.sh $OS prepare-image-add-bazel`

.github/workflows/verible-ci.yml
```diff
@@ -4,99 +4,140 @@ on:
...
+
+    - name: Build Verible
+      run: source ./.github/settings.sh; ./releasing/docker-run.sh verible
```
mithro, line 99: `./releasing/docker-run.sh $OS build-verible-release`

releasing/centos.dockerfile
```diff
@@ -0,0 +1,52 @@
...
+ENV BAZEL_LINKOPTS "-static-libstdc++:-lm -static-libstdc++:-lrt"
+ENV BAZEL_LINKLIBS "-l%:libstdc++.a"
+
+# Get a newer GCC version
+RUN yum install -y --nogpgcheck centos-release-scl
+RUN yum install -y --nogpgcheck devtoolset-8
+
+RUN yum install -y bison flex
+
+# Build a newer version of flex
```
mithro, line 41: This is only needed for some versions of CentOS right?

releasing/centos.dockerfile
```diff
@@ -0,0 +1,52 @@
+ARG _VERSION_
```
mithro, line 1: BTW -- Ideally these commands would be in the README and pull into these files in some way (maybe using tuttest?).

mithro (commented):
I think the direction this is going makes sense.

releasing/centos.dockerfile
umarcor: Yes. That dockerfile has theree stages:

- The first stage (L1-L12) is used regardless of the OS version.
- The second stage (L14-L25) is executed on CentOS 8 only.
  - Others use the third stage right below (L27-L53).

.github/workflows/verible-ci.yml
umarcor: Images are now renamed to 'basic' and 'build'. I use 'base' for the images that are retrieved from Docker Hub.

.github/workflows/verible-ci.yml
umarcor: Done.

.github/workflows/verible-ci.yml
umarcor: Done.

releasing/centos.dockerfile
umarcor: The dockerfile was renamed to `centos.basic.dockerfile`.

releasing/centos.dockerfile
umarcor: Now, the list of supported base images is located in `releasing/supported_bases.txt`. Rather than having that content in the README (which requires parsing Markdown; and that's not fancy), I would just add a link from the README to that file.

umarcor:
This PR is now rebased on top of master.

The failure on `centos:8` is expected. It's fixed in #629.

However, I could not track which is the mistake that is making `centos:7` fail.

hzeller:
Did this pull request fall of the radar ? @mithro you were looking at it last, any fresh comments ?

mithro:
Can you rebase this onto master and then I think we can consider merging it?

umarcor:
@mithro, as you see, centos7 is still failing. Let's merge #684 first.

hzeller:
(obsolete PR)

PR #623, opened 2020-12-08
ci: generate Release matrix dynamically

### LINKED ISSUE #621: Generate the list of GitHub release CI jobs automagically rather than hard coding in the workflow

umarcor:
Close #621.

As suggested by @mithro and explained by @tgorochowik, this PR generates the Release matrix dynamically.

### DIFF
--- a/.github/workflows/generate_matrix.sh
+++ b/.github/workflows/generate_matrix.sh
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+# Copyright 2020 The Verible Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+
+matrix='['
+for item in releasing/*/; do
+  matrix+='{"OS": "'"`echo $(basename $item) | cut -d- -f1`"'", "OS_VERSION": "'"`echo $(basename $item) | cut -d- -f2`"'"},'
+done
+matrix+=']'
+
+echo "$matrix"
+
+echo "::set-output name=matrix::$matrix"
--- a/.github/workflows/verible-ci.yml
+++ b/.github/workflows/verible-ci.yml
@@ -5,6 +5,8 @@ on:
   pull_request:
 
 jobs:
+
+
    Check:
       runs-on: ubuntu-16.04
       strategy:
@@ -15,7 +17,7 @@ jobs:
             - MODE: compile
             - MODE: clean
       env: ${{ matrix.env}}
-      name: ${{ matrix.env.MODE}}
+      name: Check · ${{ matrix.env.MODE}}
 
       steps:
         - name: Checkout code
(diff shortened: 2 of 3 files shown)


umarcor:
@googlebot I signed it!

.github/workflows/verible-ci.yml
```diff
@@ -43,23 +45,38 @@ jobs:
...
       strategy:
         fail-fast: false
         matrix:
-          env:
-            - { MODE: bin, OS: ubuntu, OS_VERSION: xenial }
-            - { MODE: bin, OS: ubuntu, OS_VERSION: bionic }
-            - { MODE: bin, OS: ubuntu, OS_VERSION: eoan }
-            - { MODE: bin, OS: ubuntu, OS_VERSION: focal }
-            - { MODE: bin, OS: ubuntu, OS_VERSION: groovy }
-#           - { MODE: bin, OS: centos, OS_VERSION: 6 }
```
tgorochowik, line 57: centos 6 was actually commented out in the previous GHA config - I guess we should delete its dir so its not included in the matrix generation.

umarcor: Done.

.github/workflows/verible-ci.yml
```diff
@@ -51,23 +53,38 @@ jobs:
...
 
+
+   Matrix:
+      runs-on: ubuntu-latest
+      outputs:
+        matrix: ${{ steps.generate.outputs.matrix }}
+      steps:
+      - uses: actions/checkout@v2
+      - id: generate
+        run: |
```
mithro, line 64: I think this would be better as an external script rather than encoded in the `.github/workflows/verible-ci.yml`?

mithro: IE It would be nice to be able to run the script locally to see what the output looks...

.github/workflows/verible-ci.yml
```diff
@@ -51,23 +53,38 @@ jobs:
...
+            matrix+='{"OS": "'"`echo $(basename $item) | cut -d- -f1`"'", "OS_VERSION": "'"`echo $(basename $item) | cut -d- -f2`"'"},'
+          done
+          matrix+=']'
+          echo "$matrix"
+          echo "::set-output name=matrix::$matrix"
+
+
    Release:
+      needs: Matrix
       runs-on: ubuntu-16.04
```
mithro, line 76: We might want to consider not using such an ancient ubuntu here. Let's log an issue to loop back on that?

mithro (approved):
LGTM! Couple of small comments.

.github/workflows/verible-ci.yml
umarcor: Done.

.github/workflows/verible-ci.yml
umarcor: See #626.

.github/workflows/generate_matrix.sh
```diff
@@ -0,0 +1,26 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+
+matrix='['
+for item in releasing/*/; do
```
mithro, line 19: This LGTM. 

I was pondering if this should come from the README.md file instead but decided the directory version makes the most sense. Maybe we should have a GitHub Action check which makes sure the README and directories are kept in sync in some way?

mithro (approved):

.github/workflows/generate_matrix.sh
umarcor: In #622, the way jobs are defined is to be reworked. Instead of reading directories, a list of images will be provided and subdirs will be removed. It will be easier to handle sync after that.

umarcor: Let's continue in https://github.com/google/verible/pull/622#discussion_r538794804.

Issue #606, opened 2020-11-16
always-ff-non-blocking rule fails to flag modifying assignments as well as unary increments & decrements


preusser:
**Summary**

The [always-ff-non-blocking](https://github.com/google/verible/blob/73936a870a7948663cc1009cd8b36a40f3ebe69d/verilog/analysis/checkers/always_ff_non_blocking_rule.cc#L56) rule currently only flags basic assignments using the `=` operator. Blocking assignments using modifying assignment operators or unary increments and decrements evade an enforcement.

** Example **

```systemverilog
always_ff (@posege clk) begin
  Count <= Count + 1;  // valid non-blocking assignment - passing correctly
  Count = Count + 1;   // forbidden blocking assignment - flagged correctly
  Count++;             // forbidden blocking assignment - WRONGLY NOT flagged
  Count += 1;          // forbidden blocking assignment - WRONGLY NOT flagged
end
```

**Proposal**
The rule should flag all forms of blocking assignments including modifying assignments and unary increments and decrements.

Code enforcing the coverage of all types of blocking assignments would be available for review. An update jointly addressing also  #605 appears advisable as the new covered types of blocking assignments are likely frequently used with local variables where they should not be flagged.

fangism:
I concur.  Just needs catch more cases in https://cs.opensource.google/verible/verible/+/master:verilog/analysis/checkers/always_ff_non_blocking_rule.cc.

Issue #605, opened 2020-11-16
always-ff-non-blocking rule wrongly flags assignments to locals


preusser:
**Summary & Example**

With the exception of loop headers, the [always-ff-non-blocking](https://github.com/google/verible/blob/73936a870a7948663cc1009cd8b36a40f3ebe69d/verilog/analysis/checkers/always_ff_non_blocking_rule.cc#L56) currently flags all blocking assignments in an `always_ff` block regardless of the nature of the targeted variable. This disallows structuring non-trivial computations into steps using `automatic` locals, such as:
```systemverilog
always_ff (@posege clk) begin
  automatic type(Count)  count_nxt = Count;
  if(inc_valid)  count_nxt = count_nxt + inc_value;      // Should not be flagged
  if(dec_valid)  count_nxt = count_nxt - dec_value;      // Should not be flagged
  Count <= count_nxt;                                    // Would be flagged if blocking
  Trigger <= count_nxt >= THRESHOLD;                     // Would be flagged if blocking
end
```

**Proposal**
The rule should be waived for assignments to locally declared variables, which cannot be observed by unrelated processes.

The linter can do so when knowing the variable declarations of local, possibly nested blocks of the containing `always_ff` statement. Code tracking the declarations of variables in these local scopes and waiving the rule accordingly would be available for review.

fangism:
Good point!  @snsokolov What do you think?

snsokolov:
> Good point! @snsokolov What do you think?

The style of locally declared automatic variables inside always_ff is rarely used. I think we recommend explicitly declaring these as nets and combinational logic outside of always_ff blocks. Let me ask this internally.

preusser:
This would be a recommendation against tight scoping. And even if so, the proposed modification would not invalidate such a style.

snsokolov:
Hi Thomas,

I checked internally and confirmed that for various reasons (risk of inferred latch is one of them I think) our internal style guide explicitly prohibit mixing any (especially non-trivial) combo logic with sequential logic within the same a sequential block. And I don't think we would change this guidance.

By the way, what about waveforms? Is this okay from debuggability standpoint that `count_nxt` is not visible to waveforms?

Off to David.

fangism:
While we may not be in a good position to change the existing rule, we're not opposed to anyone making a different version of the rule that grants exceptions like the proposed to fit your own style guide.  (Someday making them configurable variants of the same rule.). That being said, the exception requested requires technically:

[...]

PR #634, opened 2020-12-29
Introduced more focused always_ff_{non -> only_local}_blocking_rule.

preusser:
This commit adds an `always_ff_only_local_blocking_rule` as an alternative for the `always_ff_non_blocking_rule` that addresses issues #605 and #606:
- It is stricter as it also triggers on blocking assignments implemented by modifying assignment operators and unary increments and decrements.
- It relaxes the absolute ban on blocking assignments by waiving it for locally scoped variables, for which these assignments are safe.

The rule implementation tracks local declaration scopes _within_ ´always_ff´ blocks in the form of a stack remembering `string_view`s of local identifiers.

Ideally, this rule would replace `always_ff_non_blocking_rule` altogether. In the absence of local declarations, it only sharpens this rule by addressing #606. Keeping both rules in parallel, however, creates an overlapping responsibility for flagging non-modifying blocking assignments to non-local variables. This appears to be an undesirable architectural smell and ultimately also defeats `minus-rule` testing.


verilog/analysis/checkers/always_ff_only_local_blocking_rule.cc
```diff
@@ -0,0 +1,167 @@
...
+      scopes.emplace(depth, locals.size());
+    }
+    // Collect local variable declarations
+    else if(decl_matcher.Matches(symbol, &symbol_man)) {
+      auto& cnt = scopes.top().second;
+      for(auto const& var : SearchSyntaxTree(symbol, var_matcher)) {
+        if(auto const *const  node = dynamic_cast<verible::SyntaxTreeNode const*>(var.match)) {
+          if(auto const *const  ident = dynamic_cast<verible::SyntaxTreeLeaf const*>(node->children()[0].get())) {
+            absl::string_view const  name = ident->get().text();
+            if(DEBUG)  std::cerr << "Registering '" << name << '\'' << std::endl;
```
hzeller, line 111: For these DEBUG outputs, I recommend using VLOG(), which is the common way to create debug messages in this codebase. Check out other places in the code-base for examples.

Here, you could write
```c++
  VLOG(1) << "Registering '" << name << "'";
```
(It will take care of adding a newline automatically).

To output these messages, you can enable the logging, e.g. by setting an environment variable:

```
GLOG_v=1 bazel-bin/verilog/analysis/checkers/always_ff_only_local_blocking_rule_test
```

verilog/analysis/checkers/always_ff_only_local_blocking_rule.cc
```diff
@@ -0,0 +1,167 @@
...
+  static Matcher const  decl_matcher = NodekDataDeclaration();
+  static Matcher const  var_matcher = NodekRegisterVariable();
+  static Matcher const  asgn_blocking_matcher = NodekNetVariableAssignment();
+  static Matcher const  asgn_modify_matcher = NodekAssignModifyStatement();
+  static Matcher const  asgn_incdec_matcher = NodekIncrementDecrementExpression();
+  static Matcher const  ident_matcher = NodekUnqualifiedId();
+
+  // Determine depth in syntax tree and discard state from branches already left
+  int const  depth = context.size();
+  if(depth <= this->inside)  this->inside = 0;
```
hzeller, line 76: Please add a space between if and open parenthesis to [match the style guide](https://github.com/google/verible/blob/master/CONTRIBUTING.md#style). There are various other places I see that would need to have slightly different formatting.

Luckily, this is easy to do automatically, just run `clang-format` on all your files:

```
clang-format -i --style=Google verilog/analysis/checkers/always_ff_only_local_blocking_rule.cc
```

verilog/tools/lint/BUILD
```diff
@@ -109,15 +110,15 @@ _linter_test_configs = [
...
-        # Verifies that all rules *except* this one is lint clean.
-        verilog_style_lint.test(
-            name = testfilebase + "-minusrule-test",
-            srcs = ["testdata/{}.sv".format(testfilebase)],
-            flags = [
-                "--ruleset=all",
-                "--rules=-" + rule,  # exclude one rule
-            ],
-        ),
+#        # Verifies that all rules *except* this one is lint clean.
```
hzeller, line 113: mmh, was this needed due to the overlap of the two always-ff rules ?
Maybe we need to have the files more specific.

verilog/analysis/checkers/always_ff_only_local_blocking_rule.cc
```diff
@@ -0,0 +1,167 @@
...
+                      "blocking assignment to non-locals in sequential logic.");
+}
+
+
+static bool const  DEBUG = false;
+
+void AlwaysFFOnlyLocalBlockingRule::HandleSymbol(verible::Symbol const& symbol,
+                                           SyntaxTreeContext const& context) {
+
+  static Matcher const  always_ff_matcher = NodekAlwaysStatement(AlwaysFFKeyword());
```
hzeller, line 65: Instead of assignment, maybe call constructor directly (the rest of the codebase is using this style, so this keeps it consistent):

```c++
  static const Matcher always_ff_matcher(NodekAlwaysStatement(AlwaysFFKeyword());
```

hzeller (changes requested):
Thanks for your contribution!

Sorry, it took a while after the holidays to get to this, but, getting there :)

I've added a first round of comments, mostly suggesting smallish high-level style considerations (bottom line: run `clang-format`) as well as a suggestion how to simplify debug logging.

One question I had is, maybe we could combine the two `always_ff_...` rules and make them configurable to allow both behaviors; it might be slightly less confusing to a user. I am open both ways, but maybe @fangism has some more refined opinion.

Lastly: please make sure to do the CLA signing, so that we can merge this.

verilog/analysis/checkers/always_ff_only_local_blocking_rule.cc
preusser: ACK. Will do.

verilog/tools/lint/BUILD
preusser: This mechanism ensures that the responsibilities of linter rules are mutually exclusive. As detailed in the above description, this would no longer be the case if both the original `always_ff_non_blocking_rule` and this `always_ff_only_local_blocking_rule` were to be offered as alternatives. Their responsibilities overlap.
From the preceding discussion, I assume that a full replacement of the original rule is not desired. So, I can identify to remaining options to proceed:
1. Relax the exclusiveness requirements, or
2. Merge the rules into one and make its aggressiveness selectable by some parameter (I am not aware of a mechanism to do so.)

preusser:
@googlebot I signed it!

verilog/tools/lint/BUILD
hzeller: I am for merging the rules and make them configurable so that we end up with only one rule. As a first step having something that can be configured so that it behaves as before, but later we might consider to always keep the more stricter form (this is to help transitioning people using the weaker lint rules so far).

There is a simple way to provide configuration to rules. Each rule can override a `Configure(absl::string_view configuration)` method which receives a string that contains configurations.

There are already utilities that help and set configuration parameters, see `common/text/config_utils.h` for this. These allow to directly set a value inside the rule with a very simple set-up.

Here are some examples of rule configurations
  * [LineLengthRule](https://github.com/google/verible/blob/00b99903ae110367e55f8f249d1366cd6c9cf4d6/verilog/analysis/checkers/line_length_rule.cc#L186-L191)
  * [ModuleFilenameRule](https://github.com/google/verible/blob/00b99903ae110367e55f8f249d1366cd6c9cf4d6/verilog/analysis/checkers/module_filename_rule.cc#L133-L137)
  * [ExplicitParameterStorageRule](https://github.com/google/verible/blob/00b99903ae110367e55f8f249d1366cd6c9cf4d6/verilog/analysis/checkers/explicit_parameter_storage_type_rule.cc#L106-L116)
  * [ForbiddenAnonymousStructsUnionsRule](https://github.com/google/verible/blob/00b99903ae110367e55f8f249d1366cd6c9cf4d6/verilog/analysis/checkers/forbidden_anonymous_structs_unions_rule.cc#L69-L75)
  * [ExplicitParameterStorageRule](https://github.com/google/verible/blob/00b99903ae110367e55f8f249d1366cd6c9cf4d6/verilog/analysis/checkers/explicit_parameter_storage_type_rule.cc#L106-L116)

For the `verilog_style_lint.test()`, they would just run in the default configuration, so the existing test should trigger there.

Inside the unit-tests, you can test the behaviour using the [`RunConfiguredLintTestCases<>`](https://github.com/google/verible/blob/00b99903ae110367e55f8f249d1366cd6c9cf4d6/common/analysis/linter_test_utils.h#L91) call. It is easy to see in the existing examples, such as in the [line length rule test](https://github.com/google/verible/blob/00b99903ae110367e55f8f249d1366cd6c9cf4d6/verilog/analysis/checkers/line_length_rule_test.cc#L198-L199)

hzeller (changes requested):
Nice!
Mostly some pedantic but needed to keep the code-base consistent comments....

Also some pointers how to implement configurability and suggestion to move old and new rule.

verilog/analysis/checkers/always_ff_non_blocking_rule.cc
```diff
@@ -48,50 +51,150 @@ absl::string_view AlwaysFFNonBlockingRule::Name() {
...
-    "Use only non-blocking assignments inside \'always_ff\' sequential blocks.";
+    "Use blocking assignments, at most, for locals inside "
+    "'always_ff' sequential blocks.";
 
 std::string AlwaysFFNonBlockingRule::GetDescription(
     DescriptionType description_type) {
-  return absl::StrCat("Checks that there are no occurrences of ",
-                      "blocking assignment in sequential logic.");
+  return "Checks that blocking assignments are, at most, targeting "
+         "locals in sequential logic.";
```
hzeller, line 60: In other rules that have a configuration, we also add some string that explains the parameters; check out the other rules.

These then make it into the automatically generated documentation: https://google.github.io/verible/lint.html

verilog/analysis/checkers/always_ff_non_blocking_rule.h
```diff
@@ -48,7 +50,26 @@ class AlwaysFFNonBlockingRule : public verible::SyntaxTreeLintRule {
...
+  bool catch_modifying_assignments_ = false;
+  bool waive_for_locals_ = false;
+
+  //- Processing State ------------------
+  // Inside an always_ff block
+  int inside_ = 0;
+
+  // Stack of inner begin-end scopes
+  using Scope =
+      std::pair<int, int>;  // depth in syntax tree, number of inherited locals
```
hzeller, line 66: Instead of a pair, it might be as easy to create a struct here with the advantage. Something like:
```
struct Scope { int syntax_tree_depth; int inherited_locals_count; };
```

Self explantory and at the place where it is used, there is no wondering what first or second means.

verilog/analysis/checkers/always_ff_non_blocking_rule.cc
```diff
@@ -48,50 +51,150 @@ absl::string_view AlwaysFFNonBlockingRule::Name() {
...
+  }
+  locals_.resize(scopes_.top().second);
+
+  verible::matcher::BoundSymbolManager symbol_man;
+
+  if (always_ff_matcher.Matches(symbol, &symbol_man)) {
+    // Check for entering an always_ff block
+    VLOG(4) << "always_ff @DEPTH=" << depth << std::endl;
+    inside_ = depth;
+  } else if (inside_) {
```
hzeller, line 108: This `else if (inside_)` goes all the way to the end but it increases the mental burden of keeping that in mind. So I suggest to do something like

```c++
//...
else if (!inside_) {
   return
}
```

Then the lower chain of if/else's is easier to follow.

But even then, things are pretty nested below. Maybe we should have one toplevel if/else and then call into separate private methods ? That would know down the cyclomatic complexity level one more.

```c++
if (block_matcher.Matches(symbol, &symbol_man)) {
    scopes.emplace(depth, locals_.size());
} else if (decl_matcher.Matches(symbol, &symbol_man)) {
   CollectLocalVariableDeclarations(...);
} else if (!context.IsInside(NodeEnum::kLoopHeader)) {
   CheckBlockingAssignmentsOutsieLoopheaders(...);
}
...
```

verilog/analysis/checkers/always_ff_non_blocking_rule.h
```diff
@@ -48,7 +50,26 @@ class AlwaysFFNonBlockingRule : public verible::SyntaxTreeLintRule {
...
+  //- Configuration ---------------------
+  bool catch_modifying_assignments_ = false;
+  bool waive_for_locals_ = false;
+
+  //- Processing State ------------------
+  // Inside an always_ff block
+  int inside_ = 0;
+
+  // Stack of inner begin-end scopes
+  using Scope =
```
hzeller, line 65: If you make this a struct
```c++
struct Scope { int syntax_tree_depth; int inherited_locals; };
```

Then it becomes self-explanatory and also there is no wondering what `.first` or `.second` is.

verilog/analysis/checkers/always_ff_non_blocking_rule.cc
```diff
@@ -48,50 +51,150 @@ absl::string_view AlwaysFFNonBlockingRule::Name() {
...
 const char AlwaysFFNonBlockingRule::kMessage[] =
-    "Use only non-blocking assignments inside \'always_ff\' sequential blocks.";
+    "Use blocking assignments, at most, for locals inside "
+    "'always_ff' sequential blocks.";
 
 std::string AlwaysFFNonBlockingRule::GetDescription(
     DescriptionType description_type) {
-  return absl::StrCat("Checks that there are no occurrences of ",
-                      "blocking assignment in sequential logic.");
+  return "Checks that blocking assignments are, at most, targeting "
```
hzeller, line 59: For rules with parameters, we add them in the description; please check out the other rules that provide configuration how that looks like.

This is then used to generate automatic documentation that then people can use to discover parameters: https://google.github.io/verible/lint.html

hzeller (changes requested):
Nice. The configuration instead of separate rule really makes it easier to deal with.
 
I have some remarks that are now mostly making sure good maintainablility of the code: using a struct instead of a pair in one place, and thoughts about reducing cyclomatic complexity.

verilog/analysis/checkers/always_ff_non_blocking_rule.h
hzeller: (interesting, I wrote this remark not on the final commit, but then github didn't show it in the review overview so I thought it was lost ... so this is why this rough same comment appears twice. Same with the documenting parameters)

hzeller (changes requested):
Starting to look really good.
Sorry for the back and forth - usually the first pull requests are a bit long in the review as you get accustomed to the projects preferred practices. But independent of that, I am actually really happy about this pull request!

verilog/analysis/checkers/always_ff_non_blocking_rule.h
```diff
@@ -36,19 +37,47 @@ class AlwaysFFNonBlockingRule : public verible::SyntaxTreeLintRule {
...
+  bool waive_for_locals_ = false;
+
+  //- Processing State ------------------
+  // Inside an always_ff block
+  int inside_ = 0;
+
+  // Stack of inner begin-end scopes
+  struct Scope {
+    int syntax_tree_depth;
+    int inherited_local_count;
```
hzeller, line 73: We are only assigning from sources of `size_t` (`locals_.size()` below) so let's that type.

```suggestion
    size_t inherited_local_count;
```

hzeller:
Alright, thank you! Will merge once the tests went through.

preusser:
Very good. Thanks!

hzeller:
Thanks, merged!

hzeller:
Here is the updated auto-generated documentation: https://google.github.io/verible/lint.html#always-ff-non-blocking

PR #644, opened 2021-01-13
Verify gcc and g++ version in Kokoro kythe build.

hzeller:
Also remove unused code.

This is an attempt to debug the Kokoro kythe build that is consistently not working [[1]](https://source.cloud.google.com/results/invocations/74ae259f-e375-42c1-a1f5-76d6c99ad49a/targets/foss-fpga-tools%2Fverible_internal%2Fkythe_index_generation/log). It looks like it is using some old compiler or old libstdc++ which results in some of the compile troubles ? At least the compile error messages that reference `usr/include/c++/5/` are suspicious.

So as first step, let's output the gcc version.

Also, while at it, remove some old bazel install code.

### DIFF
--- a/.github/kokoro/kythe.sh
+++ b/.github/kokoro/kythe.sh
@@ -41,6 +41,8 @@ which wget
 which unzip
 which tar
 which "$BAZEL" && "$BAZEL" version
+which gcc && gcc -v
+which g++ && g++ -v
 echo "... done."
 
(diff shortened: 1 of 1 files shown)


hzeller:
(Adding Mithro as reviewer as he most closely has experience with Kokoro)

mithro (approved):
LGTM if CI is green.

PR #648, opened 2021-01-20
disable fork_label rule for linter

### LINKED ISSUE #585: [lint] Flag uses `disable fork_label;`
**Summary**

Flag all uses of `disable` where it is different from `disable fork` as error.

Cadence recommends that disabling forks should be done using a `disable fork;` statement, not the `disable fork_process_label;` construct. The `disable fork_process_label` construct for forks is not defined in the IEEE Std 1800-2017 LRM. See section 9.6.2 of the LRM (2017).

**Test cases and examples**

The more examples you provide, the easier it will be for someone to come along
and implement it.

```systemverilog
// This example should be diagnosed.
disable fork_process_label;
```

```systemverilog
// This variant should NOT be diagnosed (negative test case)
disable fork;
```

**Additional context**

VCS (and likely other simulators) support `disable fork_process_label`, however, Xcelium does not. That's why this construct can go unnoticed if only VCS is used.

See also https://github.com/lowRISC/ibex/pull/1180 for an analysis of the problem, and https://github.com/lowRISC/ibex/pull/1180 for a PR fixing the cases where we had that in Ibex.

CC @PaulOKCrevinn @CathalMCrevinn did I get this description right? Please jump in with comments if needed.

pawelsag:
Fixes  #585


verilog/analysis/checkers/disable_fork_rule_test.cc
```diff
@@ -0,0 +1,60 @@
...
+          "disable fork;\n",
+          "end\nendmodule"},
+      {"module m;\ninitial begin fork\n",
+          "begin\n#6;\nend\n",
+          "begin\n#3;\nend\n",
+          "join_any\n",
+          {kToken, "disable"},
+          " fork_invalid;\n",
+          "end\nendmodule"},
+  };
```
hzeller, line 54: From reading the but, it should be allowed to have something like

```systemverilog
  begin: foobar
  //...
    disable foobar;
  // ...
  end
```

but
```SystemVerilog
  ///
     fork : foobar
     // ...
     disable foobar;
  end
```

is not. So we would need to keep track of the label applied to the last `fork` ?

hzeller (commented):
I just added a clarification question to the bug; I think we need to keep track of the actual label being used that it is not one applied to a fork.

verilog/analysis/checkers/disable_fork_rule_test.cc
hzeller: So yes, the bug clarification indicated that we need to refine this.

hzeller (changes requested):
Looking at the bug again, some more refinements needed (see inline)

verilog/analysis/checkers/disable_fork_rule_test.cc
pawelsag: Hi, thanks for the clarification. 
I have implemented remarks from comments.
Could you verify if the introduced changes solve the issue?

hzeller:
Will have a look at this tomorrow.

i'd like to suggest: Instead of force-push, it is easier to just add more commits to your feature-branch and then push them regularly. After push, they show up as new commit in the pull request and it is easier to review the changes from the previous one. Otherwise it is like looking an entire diff from the start again.

(Don't worry about accumulating commits, once at merge time, I'll squash the whole pull request to one anyway, so only the final result will make it into the repo)

verilog/analysis/checkers/disable_fork_rule.cc
```diff
@@ -0,0 +1,118 @@
...
+                                           const SyntaxTreeContext& context) {
+  verible::matcher::BoundSymbolManager manager;
+  if (DisableMatcher().Matches(symbol, &manager)) {
+    // if no kDisable label, return, nothing to be checked
+    const auto disableLabels = FindAllSymbolIdentifierLeafs(symbol);
+    if (disableLabels.size() == 0) {
+      return;
+    }
+    // validate all kBegin parents
+    // if there is matching label return
```
hzeller, line 82: A brief explanation about the approach you're taking might be useful. e.g. that loop you're going forward and then look backwards from the rcontext are a bit hard to understand.
Using the iterators and assuming they are random access iteratrors (e.g. in `rcontext.begin() + i` or `rcontext.begin() + i + 1)`) is dangerous as you tie yourself to an implementation that happens to exist here. It also makes it hard to read, understand and modify later.

Can you reformulate that without these ? Ideally we only would assume we can forward iterate `for (const auto & : rcontext)`

verilog/analysis/checkers/disable_fork_rule.cc
```diff
@@ -0,0 +1,118 @@
...
+void DisableForkNoLabelsRule::HandleSymbol(const verible::Symbol& symbol,
+                                           const SyntaxTreeContext& context) {
+  verible::matcher::BoundSymbolManager manager;
+  if (DisableMatcher().Matches(symbol, &manager)) {
+    // if no kDisable label, return, nothing to be checked
+    const auto disableLabels = FindAllSymbolIdentifierLeafs(symbol);
+    if (disableLabels.size() == 0) {
+      return;
+    }
+    // validate all kBegin parents
```
hzeller, line 81: In general, did you a what `context` has to offer ? There are all sorts of utility functions like `DirectParentsAre()` or `IsInside()` that can help formulate what you're looking for more specifically.

verilog/analysis/checkers/disable_fork_rule_test.cc
```diff
@@ -0,0 +1,99 @@
...
+      {"module m;\ninitial begin\n fork\n", "begin : foo\n", "begin : foo_2\n",
+       "disable foo;\n", "end\n", "end\n", "join_any\n", "end\nendmodule"},
+      {"module m;\ninitial begin fork\n",
+       "begin\n#6;\nend\n",
+       "begin\n#3;\nend\n",
+       "join_any\n",
+       {kToken, "disable"},
+       " fork_invalid;\n",
+       "end\nendmodule"},
+      {"module m;\n",
```
hzeller, line 56: uhm, the following tests look like they are all the same ?

verilog/analysis/checkers/disable_fork_rule_test.cc
```diff
@@ -0,0 +1,99 @@
...
+       "initial begin:foo;\n",
+       "fork\n",
+       "begin : fo\n",
+       {kToken, "disable"},
```
hzeller, line 60: this should actually be allowed. `disable foo` refers to the begin block above, not a labelled fork.

hzeller (changes requested):

verilog/analysis/checkers/disable_fork_rule_test.cc
pawelsag: But this is not a sequential block since this is a `initial begin` block?
I make such assumption based on the clarification written here https://github.com/google/verible/issues/585#issuecomment-769616018
And in the last paragraph sriyerg wrote:
```
Also, I would assume that attempting to disable process execution threads such as initial, always* and final using their labels:

  initial begin : test
    ...
    if (cond) disable test;
  end

is illegal because these are not considered sequential code-blocks. The LRM makes no mention of validity of this (at least I couldn't find it) . So these will need to be flagged as well.
```
So should they be tried as illegal as well?

verilog/analysis/checkers/disable_fork_rule.cc
pawelsag: Yes, I was there. Unfortunately, I couldn't use any of these functions, cause I need to retrieve the parent nodes to obtain the node's labels

verilog/analysis/checkers/disable_fork_rule.cc
```diff
@@ -0,0 +1,124 @@
...
+            ptag == static_cast<int>(NodeEnum::kFinalStatement) ||
+            ptag == static_cast<int>(NodeEnum::kAlwaysStatement)) {
+          break;
+        }
+        if (beginLabel.get().text() == disableLabel.get().text()) {
+          return;
+        }
+      }
+    }
+    violations_.insert(LintViolation(symbol, kMessage, context));
```
hzeller, line 116: Since we're know what non-sequential statement triggered this, we could possibly adapt the error message ?

verilog/analysis/checkers/disable_fork_rule_test.cc
```diff
@@ -0,0 +1,116 @@
...
+       "end\nendmodule"},
+      {"module m;\ninitial begin\n", "fork\n", "begin : foo\n",
+       "begin : foo_2\n", "disable foo;\n", "end\n", "end\n", "join_any\n",
+       "end\nendmodule"},
+  };
+  RunLintTestCases<VerilogAnalyzer, DisableForkNoLabelsRule>(
+      kDisableStatementTestCases);
+}
+
+TEST(DisableStatementTest, FunctionFailures) {
```
hzeller, line 54: @sriyerg does this look like a good set of test-cases for #585 ?

verilog/analysis/default_rules.h
```diff
@@ -33,6 +33,7 @@ constexpr const char* kDefaultRuleSet[] = {
     "generate-label",
     "generate-label-prefix",
     "always-comb",
+    "disable-statement",
```
hzeller, line 36: That name used to be correct (also in verilog/tools/lint/BUILD).

verilog/analysis/checkers/disable_fork_rule.cc
```diff
@@ -0,0 +1,124 @@
...
+    }
+    // look for every kBegin node starting from kDisableLabel token
+    // the kDisableLabel can be nested in some kBegin nodes
+    // so we're looking for the kBegin node that direct parent
+    // is not one of the initial/final/always statemnts since such blocks are
+    // considered to be invalid. If the label for diable statment is not
+    // found, it means that there is no appropriate label or the label
+    // points to the illegal node such as forked label
+    const auto& rcontext = reversed_view(context);
+    for (auto rc = rcontext.begin(); rc != rcontext.end(); rc++) {
```
hzeller, line 89: I like the new structure with 'early out' continues. It makes the code easy to follow.

hzeller (commented):
Nice, I think we're getting close. Adding @sriyerg to comment if the test-caess look useful (things expanded a bit from the original 'fork' discussion to 'also the other non-sequential statements'. Maybe also @imphil who originally filed the bug.

hzeller (changes requested):
I think we're almost there. Just a little bookkeeping.

verilog/analysis/default_rules.h
```diff
@@ -33,6 +33,7 @@ constexpr const char* kDefaultRuleSet[] = {
     "generate-label",
     "generate-label-prefix",
     "always-comb",
+    "disable-non-sequential",
```
hzeller, line 36: I think we should start with it default off and have people try it before we switch it on by default.

verilog/analysis/checkers/disable_fork_rule_test.cc
```diff
@@ -0,0 +1,116 @@
...
+  constexpr int kToken = TK_disable;
+  const std::initializer_list<LintTestCase> kDisableStatementTestCases = {
+      {"module m;\ninitial begin\n",
+       "fork\n",
+       "begin\n#6;\nend\n",
+       "begin\n#3;\nend\n",
+       "join_any\n",
+       {kToken, "disable"},
+       " fork_invalid;\n",
```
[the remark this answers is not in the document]
pawelsag: Ok, so @hzeller should we change this in the current PR or make a follow-up issue for this bug?

verilog/analysis/checkers/disable_fork_rule_test.cc
```diff
@@ -0,0 +1,116 @@
...
+       "end\n",
+       {kToken, "disable"},
+       " foo;\n",
+       "join_any\n",
+       "end\nendmodule"},
+      {"module m;\n",
+       "initial begin:foo;\n",
+       "fork\n",
+       "begin : fo\n",
+       {kToken, "disable"},
```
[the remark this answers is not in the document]
pawelsag: I changed the tests, also I have added the test case you specified in the comment.
What I wanted to show by these tests is that disabling the non-sequential statements like `always_comb/final/initial` is illegal.

verilog/analysis/checkers/disable_non_sequential_rule.cc
```diff
@@ -0,0 +1,129 @@
...
+}
+
+void DisableForkNoLabelsRule::HandleSymbol(const verible::Symbol& symbol,
+                                           const SyntaxTreeContext& context) {
+  verible::matcher::BoundSymbolManager manager;
+  if (DisableMatcher().Matches(symbol, &manager)) {
+    const char* kMessageFinal = DisableForkNoLabelsRule::kMessage;
+    // if no kDisable label, return, nothing to be checked
+    const auto disableLabels = FindAllSymbolIdentifierLeafs(symbol);
+    if (disableLabels.size() == 0) {
```
hzeller, line 82: for size comparisons against 0, we should use empty() instead (reason: `empty()` is a method that is easier and faster implemented in some containers. By using `empty()` we make sure to not accidentally introduce slowness in case the container we're dealing with is slow on size()).

verilog/analysis/checkers/disable_non_sequential_rule.cc
```diff
@@ -0,0 +1,129 @@
...
+    // found, it means that there is no appropriate label or the label
+    // points to the illegal node such as forked label
+    const auto& rcontext = reversed_view(context);
+    for (auto rc = rcontext.begin(); rc != rcontext.end(); rc++) {
+      const auto& node = *rc;
+      if (node->Tag().tag != static_cast<int>(NodeEnum::kSeqBlock)) {
+        continue;
+      }
+      for (const auto& ch : node->children()) {
+        if (ch.get()->Tag().tag != static_cast<int>(NodeEnum::kBegin)) {
```
hzeller, line 99: this looks like an unnecessary call to `get()`. Just use `->Tag()` direclty.

verilog/analysis/checkers/disable_non_sequential_rule.cc
```diff
@@ -0,0 +1,129 @@
...
+          continue;
+        }
+        const auto& beginLabels = FindAllSymbolIdentifierLeafs(*ch);
+        if (beginLabels.size() == 0) {
```
hzeller, line 103: another empty()-tunity :)

verilog/analysis/checkers/disable_non_sequential_rule.cc
```diff
@@ -0,0 +1,129 @@
...
+  return matcher;
+}
+
+void DisableForkNoLabelsRule::HandleSymbol(const verible::Symbol& symbol,
+                                           const SyntaxTreeContext& context) {
+  verible::matcher::BoundSymbolManager manager;
+  if (DisableMatcher().Matches(symbol, &manager)) {
+    const char* kMessageFinal = DisableForkNoLabelsRule::kMessage;
+    // if no kDisable label, return, nothing to be checked
+    const auto disableLabels = FindAllSymbolIdentifierLeafs(symbol);
```
hzeller, line 81: I'd probably use `const auto &` here, so that we don't have to rely on the compiler to do the right thing.

verilog/analysis/checkers/disable_non_sequential_rule.cc
```diff
@@ -0,0 +1,129 @@
...
+#include "common/analysis/citation.h"
+#include "common/analysis/lint_rule_status.h"
+#include "common/analysis/matcher/bound_symbol_manager.h"
+#include "common/analysis/matcher/core_matchers.h"
+#include "common/analysis/matcher/matcher.h"
+#include "common/text/symbol.h"
+#include "common/text/syntax_tree_context.h"
+#include "verilog/CST/identifier.h"
+#include "verilog/CST/verilog_matchers.h"  // IWYU pragma: keep
+#include "verilog/CST/verilog_nonterminals.h"
```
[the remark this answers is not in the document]
hzeller: make sure to 'resolve conversation' once done with a comment, so that it is easy to see what is left to do.

verilog/analysis/checkers/disable_non_sequential_rule.h
```diff
@@ -0,0 +1,59 @@
...
+// limitations under the License.
+
+#ifndef VERIBLE_VERILOG_ANALYSIS_CHECKERS_DISABLE_STATEMENT_H_
+#define VERIBLE_VERILOG_ANALYSIS_CHECKERS_DISABLE_STATEMENT_H_
+
+#include <set>
+#include <string>
+
+#include "common/analysis/lint_rule_status.h"
+#include "common/analysis/syntax_tree_lint_rule.h"
```
hzeller, line 22: We need to include `common/analysis:syntax_tree_lint_rule` for this in the BUILD-file

verilog/analysis/checkers/disable_non_sequential_rule.cc
pawelsag: That's cool :) Thanks, for this remark

verilog/analysis/checkers/disable_fork_rule_test.cc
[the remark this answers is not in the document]
hzeller: The compile should pass as that does not look yet at semantics. So technically, since this rule _is_ looking at semantics, this is also a good test.

verilog/analysis/checkers/disable_fork_rule_test.cc
```diff
@@ -0,0 +1,116 @@
...
+       "begin : foo\n",
+       "end\n",
+       {kToken, "disable"},
+       " foo;\n",
+       "join_any\n",
+       "end\nendmodule"},
+      {"module m;\n",
+       "initial begin:foo;\n",
+       "fork\n",
+       "begin : fo\n",
```
hzeller: instead of `fo` which looks similar to `foo`, can we use a different name, such as `bar` ?
It will help not less confuse people when they read the test.

verilog/analysis/checkers/disable_fork_rule_test.cc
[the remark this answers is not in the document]
hzeller: Yes, I think we should separate out tests that deal with forks and other non-sequential statements so that we don't mix things in the test-cases and have too many variables tested at once (makes it easier to follow, and _if_ it fails at some point we know exactly what broke and don't have to try to find what aspect was failing).

I think we should have these in separate toplevel test-cases to make it clear to the reader what is being tested.
```c++
TEST(DisableStatementTest, ForkDisableStatements) { ... }
TEST(DisableStatementTest, NonSequentialDisableStatements) { ... }  // Maybe even split this in Initial, ...
```

verilog/analysis/checkers/disable_fork_rule.h
```diff
@@ -0,0 +1,58 @@
...
+#include "common/analysis/lint_rule_status.h"
+#include "common/analysis/syntax_tree_lint_rule.h"
+#include "common/text/symbol.h"
+#include "common/text/syntax_tree_context.h"
+#include "verilog/analysis/descriptions.h"
+
+namespace verilog {
+namespace analysis {
+
+// DisableForkNoLabelsRule finds occurrences of "dsiable fork_label"
```
hzeller: typo: `dsiable`

verilog/analysis/checkers/disable_fork_rule.cc
```diff
@@ -0,0 +1,124 @@
...
+namespace verilog {
+namespace analysis {
+
+using verible::GetStyleGuideCitation;
+using verible::LintRuleStatus;
+using verible::LintViolation;
+using verible::SyntaxTreeContext;
+using verible::matcher::Matcher;
+
+// Register DisableForkNoLabelsRule
```
hzeller: I'd avoid comments that say exactly what the code already says...

hzeller (changes requested):
Given that the scope somewhat widened from fork to also non-sequential statements I think we should
  * separate the tests that test fork vs. other statements
  * rename disable_fork_rule.{h,cc} to disable_statement_rule.{h,cc} maybe ? (Open to suggestions with better name); and then change all the specific references to 'fork'

Other than that, looks pretty good!

verilog/analysis/checkers/disable_fork_rule.h
pawelsag: I believe you are looking on the wrong git revision, I've removed this already.
Also, I changed the file names from disable_fork* to disable_non_sequential*,
but after your last review, I changed them to disable_statement* as suggested.
I believe this is the correct name for this rule.

verilog/analysis/checkers/disable_fork_rule_test.cc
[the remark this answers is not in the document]
pawelsag: I decided that the two separate test cases are looking good just for now.
We are testing two main categories, and tearing them apart into smaller parts will lead to a situation where
there is only one test case for the test suite or we will repeat the tests

verilog/analysis/checkers/disable_statement_rule.cc
```diff
@@ -0,0 +1,128 @@
...
+static const Matcher& DisableMatcher() {
+  static const Matcher matcher(
+      NodekDisableStatement(DisableStatementHasLabel()));
+  return matcher;
+}
+
+void DisableStatementNoLabelsRule::HandleSymbol(
+    const verible::Symbol& symbol, const SyntaxTreeContext& context) {
+  verible::matcher::BoundSymbolManager manager;
+  if (DisableMatcher().Matches(symbol, &manager)) {
```
hzeller, line 77: If this `if` is going over the whole method, consider an early-out here

```c++
 if (!DisableMatcher().Matches(symbol, &manager))
     return;
```

That way, the reader does not have to keep the surrounding branch on their 'reading stack', thus reducing complexity. Also, you have one more indentation level to work with :)

hzeller (commented):
LGTM.
I'll double check tomorrow that the BUILD dependencies are all matching, but looks like then we're ready to merge.
I noticed one opportunity to de-nest something with an early return.

hzeller (approved):

PR #651, opened 2021-01-22
Grant exceptions for custom struct-union-name-style

### LINKED ISSUE #385: Grant exceptions for units in [struct-union-name-style]
Currently, the [struct-union-name-style] is strictly requires lower_snake_case_t-style.

https://github.com/google/verible/blob/2b706555398f8c734bb3fc8ea5dff78ab85460c3/verilog/analysis/checkers/struct_union_name_style_rule.cc#L76

Some teams have expressed a desire to grant exceptions when a component (between underscores) is number followed by a unit that contains a capital letter like `12B` (byte) or `121GW` (gigawatts).  

Task:
Make this rule configurable with any custom set of strings that automatically get a pass when used in the above manner, between `_`s.  Doesn't have to be regex-based, a sequence of plain strings will probably suffice.

pawelsag:
Closes #385

### DIFF
--- a/verilog/analysis/checkers/struct_union_name_style_rule.cc
+++ b/verilog/analysis/checkers/struct_union_name_style_rule.cc
@@ -63,8 +63,9 @@ std::string StructUnionNameStyleRule::GetDescription(
   } else {
     return absl::StrCat(basic_desc,
                         "\n##### Parameters\n"
-                        " * `exceptions` (`String` with exceptions divided by "
-                        "`,`, Default: Empty\n");
+                        " * `exceptions` (Comma-separated list of allowed "
+                        "upper-case elements, such as unit-names"
+                        ". Default: Empty)\n");
   }
 }
 
--- a/verilog/analysis/checkers/struct_union_name_style_rule.cc
+++ b/verilog/analysis/checkers/struct_union_name_style_rule.cc
@@ -53,11 +53,19 @@ const char StructUnionNameStyleRule::kMessageUnion[] = "Union names";
 
 std::string StructUnionNameStyleRule::GetDescription(
     DescriptionType description_type) {
-  return absl::StrCat(
+  static std::string basic_desc = absl::StrCat(
       "Checks that ", Codify("struct", description_type), " and ",
       Codify("union", description_type),
       " names use lower_snake_case naming convention and end with '_t'. See ",
       GetStyleGuideCitation(kTopic), ".");
+  if (description_type == DescriptionType::kHelpRulesFlag) {
+    return absl::StrCat(basic_desc, "exceptions:String with exceptions");
+  } else {
+    return absl::StrCat(basic_desc,
+                        "\n##### Parameters\n"
+                        " * `exceptions` (`String` with exceptions divided by "
+                        "`,`, Default: Empty\n");
+  }
 }
 
 static const Matcher &TypedefMatcher() {
@@ -69,30 +77,24 @@ void StructUnionNameStyleRule::HandleSymbol(const verible::Symbol &symbol,
                                             const SyntaxTreeContext &context) {
   verible::matcher::BoundSymbolManager manager;
   if (TypedefMatcher().Matches(symbol, &manager)) {
-    const char *msg;
     // TODO: This can be changed to checking type of child (by index) when we
     // have consistent shape for all kTypeDeclaration nodes.
-    if (!FindAllStructTypes(symbol).empty()) {
-      msg = kMessageStruct;
-    } else if (!FindAllUnionTypes(symbol).empty()) {
-      msg = kMessageUnion;
-    } else {
-      // Neither a struct nor union definition
-      return;
-    }
+    const bool is_struct = !FindAllStructTypes(symbol).empty();
+    if (!is_struct && FindAllUnionTypes(symbol).empty()) return;
+    const char *const msg = is_struct ? kMessageStruct : kMessageUnion;
+
     const auto *identifier_leaf = GetIdentifierFromTypeDeclaration(symbol);
     const auto name = ABSL_DIE_IF_NULL(identifier_leaf)->get().text();
 
     if (!absl::EndsWith(name, "_t")) {
       violations_.insert(
-          LintViolation(identifier_leaf->get(),
+          LintViolation(*identifier_leaf,
                         absl::StrCat(msg, " have to ends with _t"), context));
       return;
     }
     if (name[0] == '_') {
-      violations_.insert(LintViolation(identifier_leaf->get(),
-                                       absl::StrCat(msg, " can't start with _"),
-                                       context));
+      violations_.insert(LintViolation(
+          *identifier_leaf, absl::StrCat(msg, " can't start with _"), context));
       return;
     }
 
@@ -104,18 +106,18 @@ void StructUnionNameStyleRule::HandleSymbol(const verible::Symbol &symbol,
       }
       if (!absl::ascii_isdigit(*ns.begin())) {
         violations_.insert(LintViolation(
-            identifier_leaf->get(),
+            *identifier_leaf,
             "Section with unit names need to start with digit", context));
         return;
       }
       if (exceptions_.find(std::string(ns)) != exceptions_.end()) {
-        continue;
+        continue;  // number + unit exception found
       }
       const auto &alpha =
           std::find_if(ns.begin(), ns.end(), absl::ascii_isalpha);
-      const auto ns_substr = std::string(alpha, std::end(ns));
+      const auto ns_substr = std::string(alpha, ns.end());
       if (exceptions_.find(ns_substr) == exceptions_.end()) {
-        violations_.insert(LintViolation(identifier_leaf->get(),
+        violations_.insert(LintViolation(*identifier_leaf,
                                          "found digit followed by unit that is "
                                          "not configured as allowed exception",
                                          context));
@@ -131,32 +133,34 @@ absl::Status StructUnionNameStyleRule::Configure(
   std::string raw_tokens;
   auto status = verible::ParseNameValues(
       configuration, {{"exceptions", SetString(&raw_tokens)}});
+  if (!status.ok()) return status;
 
   if (!raw_tokens.empty()) {
     const auto &exceptions = absl::StrSplit(raw_tokens, ',');
     for (const auto &ex : exceptions) {
       const auto &e =
           std::find_if_not(ex.begin(), ex.end(), absl::ascii_isalnum);
-      if (e != ex.end())
+      if (e != ex.end()) {
         return absl::Status(absl::StatusCode::kInvalidArgument,
                             "The exception can be composed of digits and "
                             "alphabetic characters only");
+      }
       const auto &alpha =
           std::find_if(ex.begin(), ex.end(), absl::ascii_isalpha);
-      if (alpha == ex.end())
+      if (alpha == ex.end()) {
         return absl::Status(
             absl::StatusCode::kInvalidArgument,
             "The exception have to contain at least one alphabetic character");
-
+      }
       const auto &digit = std::find_if(alpha, ex.end(), absl::ascii_isdigit);
-      if (digit != ex.end())
+      if (digit != ex.end()) {
         return absl::Status(absl::StatusCode::kInvalidArgument,
-                            "Digits are forbidden when specifying the unit");
-
+                            "Digits after the unit are not allowed");
+      }
       exceptions_.emplace(ex);
     }
   }
-  return status;
+  return absl::OkStatus();
 }
 
 LintRuleStatus StructUnionNameStyleRule::Report() const {
--- a/verilog/analysis/checkers/struct_union_name_style_rule.cc
+++ b/verilog/analysis/checkers/struct_union_name_style_rule.cc
@@ -89,9 +89,7 @@ void StructUnionNameStyleRule::HandleSymbol(const verible::Symbol &symbol,
     if (!exceptions_.empty()) {
       name_filtered.reserve(name.length());
       const auto &underscore_separated_parts = absl::StrSplit(name, '_');
-      const bool starts_with_underscore =
-          underscore_separated_parts.begin()->empty();
-      if (!starts_with_underscore) {
+      if (name[0] != '_') {
         for (const auto &ns : underscore_separated_parts) {
           if (std::find(exceptions_.begin(), exceptions_.end(), ns) ==
               exceptions_.end()) {
--- a/verilog/analysis/checkers/struct_union_name_style_rule.cc
+++ b/verilog/analysis/checkers/struct_union_name_style_rule.cc
@@ -48,10 +48,8 @@ absl::string_view StructUnionNameStyleRule::Name() {
   return "struct-union-name-style";
 }
 const char StructUnionNameStyleRule::kTopic[] = "struct-union-conventions";
-const char StructUnionNameStyleRule::kMessageStruct[] =
-    "Struct names must use lower_snake_case naming convention and end with _t.";
-const char StructUnionNameStyleRule::kMessageUnion[] =
-    "Union names must use lower_snake_case naming convention and end with _t.";
+const char StructUnionNameStyleRule::kMessageStruct[] = "Struct names";
+const char StructUnionNameStyleRule::kMessageUnion[] = "Union names";
 
 std::string StructUnionNameStyleRule::GetDescription(
     DescriptionType description_type) {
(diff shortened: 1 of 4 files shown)


verilog/analysis/checkers/struct_union_name_style_rule.cc
```diff
@@ -81,13 +83,33 @@ void StructUnionNameStyleRule::HandleSymbol(const verible::Symbol& symbol,
     }
     const auto* identifier_leaf = GetIdentifierFromTypeDeclaration(symbol);
     const auto name = ABSL_DIE_IF_NULL(identifier_leaf)->get().text();
-    if (!verible::IsLowerSnakeCaseWithDigits(name) ||
+    // get rid of the exceptions specified by user
+    std::string name_filtered;
+    if(raw_tokens.length()) {
+      name_filtered.reserve(name.length());
+      const auto &name_split = absl::StrSplit(name, '_');
+      const auto &exceptions = absl::StrSplit(raw_tokens, ',');
```
hzeller, line 91: Maybe we can already pre-process the raw_tokens in Configure() to be vector of exceptions ?

verilog/analysis/checkers/struct_union_name_style_rule.cc
```diff
@@ -81,13 +83,33 @@ void StructUnionNameStyleRule::HandleSymbol(const verible::Symbol& symbol,
...
         !absl::EndsWith(name, "_t")) {
       violations_.insert(LintViolation(identifier_leaf->get(), msg, context));
     }
   }
 }
 
+absl::Status StructUnionNameStyleRule::Configure(absl::string_view configuration) {
+  using verible::config::SetString;
+  return verible::ParseNameValues(
+      configuration, {{"exceptions", SetString(&raw_tokens)}});
```
hzeller, line 110: so if `raw_tokens` would be a local variable, we can after `ParseNameValues()` str::Split() these into an array, so we don't have to do this every time, only at the time we configure.

verilog/analysis/checkers/struct_union_name_style_rule.h
```diff
@@ -51,6 +54,8 @@ class StructUnionNameStyleRule : public verible::SyntaxTreeLintRule {
   static const char kMessageStruct[];
   static const char kMessageUnion[];
 
+  std::string raw_tokens;
```
hzeller, line 57: Make sure that instance variables [have a trailing underscore](https://google.github.io/styleguide/cppguide.html#Variable_Names)

hzeller (changes requested):
Oh, looks like I forgot to submit the review last time I looked at it - the 'submit review' box was still open.

Anyway, some smallish stylistic comments, and avoiding-repeated-work performance suggestion

verilog/analysis/checkers/struct_union_name_style_rule.cc
```diff
@@ -81,13 +83,33 @@ void StructUnionNameStyleRule::HandleSymbol(const verible::Symbol& symbol,
     }
     const auto* identifier_leaf = GetIdentifierFromTypeDeclaration(symbol);
     const auto name = ABSL_DIE_IF_NULL(identifier_leaf)->get().text();
-    if (!verible::IsLowerSnakeCaseWithDigits(name) ||
+    // get rid of the exceptions specified by user
+    std::string name_filtered;
+    if(raw_tokens.length()) {
```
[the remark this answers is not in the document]
pawelsag: thanks for this remark, I will keep that in mind in the future

verilog/analysis/checkers/struct_union_name_style_rule.cc
pawelsag: I just reimplemented the code to your suggestions.
Could you take a look at new version?

verilog/analysis/checkers/struct_union_name_style_rule.cc
```diff
@@ -81,13 +83,38 @@ void StructUnionNameStyleRule::HandleSymbol(const verible::Symbol& symbol,
     }
     const auto* identifier_leaf = GetIdentifierFromTypeDeclaration(symbol);
     const auto name = ABSL_DIE_IF_NULL(identifier_leaf)->get().text();
-    if (!verible::IsLowerSnakeCaseWithDigits(name) ||
+    // get rid of the exceptions specified by user
+    std::string name_filtered;
+    if (exceptions_.size()) {
```
hzeller, line 88: Just to figure out if there is something in there, we should use `empty()`. So `if (!exceptions_.empty())` in this case.

verilog/analysis/checkers/struct_union_name_style_rule.h
```diff
@@ -51,6 +54,8 @@ class StructUnionNameStyleRule : public verible::SyntaxTreeLintRule {
   static const char kMessageStruct[];
   static const char kMessageUnion[];
 
+  std::vector<std::string> exceptions_;
```
hzeller, line 57: I think you can make this a `std::set<>`, and the `absl::StrSplit()` called in `Configure()` will automagically assign it to that set.

That will make finding things in that set simple (and fast, though given that there are not many elements, it probably doesn't matter much; so it would mostly emphasize that we're conceptually mean a set here)

verilog/analysis/checkers/struct_union_name_style_rule.cc
```diff
@@ -81,13 +83,38 @@ void StructUnionNameStyleRule::HandleSymbol(const verible::Symbol& symbol,
...
+      name_filtered.reserve(name.length());
+      const auto& name_split = absl::StrSplit(name, '_');
```
hzeller, line 90: `name_split` sounds pretty generic. Maybe `underscore_separated_parts` ? (it is longer but should fit in the 80 chars).

verilog/analysis/checkers/struct_union_name_style_rule.cc
```diff
@@ -81,13 +83,38 @@ void StructUnionNameStyleRule::HandleSymbol(const verible::Symbol& symbol,
...
+      for (const auto& ns : name_split) {
+        if (std::find(exceptions_.begin(), exceptions_.end(), ns) ==
+            exceptions_.end()) {
+          name_filtered.append(ns);
+        }
+      }
+    }
+    if (!verible::IsLowerSnakeCaseWithDigits(
+            name_filtered.length() ? name_filtered : name) ||
```
hzeller, line 99: I'd probably put this in a named variable before used here
```
const absl::string_view relevant_name = name_filtered.empty() ? name : name_filtered;
```

A question: what if the name _only_ contains exceptioned name parts ? In that case, we end up with an empty string, so we choose to use the `name` instead ...
So we can't really use the name length as critterium here; Maybe `exceptions_.empty() ? name : name_filtered;`

But then what happens if we pass an empty string to `IsLowerSnakeCaseWithDigits()` ?

To start, I'd add a test to figure out what happens in that case. Then fix it here.

verilog/analysis/checkers/struct_union_name_style_rule.cc
```diff
@@ -81,13 +83,38 @@ void StructUnionNameStyleRule::HandleSymbol(const verible::Symbol& symbol,
...
+    if (exceptions_.size()) {
+      name_filtered.reserve(name.length());
+      const auto& name_split = absl::StrSplit(name, '_');
+      for (const auto& ns : name_split) {
+        if (std::find(exceptions_.begin(), exceptions_.end(), ns) ==
+            exceptions_.end()) {
+          name_filtered.append(ns);
+        }
+      }
+    }
```
hzeller, line 97: After this, a string `foo_bar_baz` will look like `foobarbaz` as we also removed the underscores.

In particular strings that previously are not allowed, a string with a starting underscore `_foo_bar` will now be `foobar` and becomes allowed.

So I think adding a test-case for this scenario should help figuring out an implementation that covers that.

verilog/analysis/checkers/struct_union_name_style_rule_test.cc
```diff
@@ -44,6 +53,39 @@ TEST(StructUnionNameStyleRuleTest, ValidStructNames) {
...
+  const absl::string_view exceptions = "exceptions:12b,11,14Kj";
+  constexpr int kToken = SymbolIdentifier;
+  const std::initializer_list<LintTestCase> kTestCases = {
+      {""},
+      {"typedef struct baz_t;"},
+      {"typedef struct good_name_t;"},
+      {"typedef struct b_a_z_t;"},
+      {"typedef struct {logic foo; logic bar;}", {kToken, "baz_12B_t"}, ";"},
+      {"typedef struct {logic foo; logic bar;}",
+       {kToken, "good_14kJ_name_t"},
```
hzeller, line 81: 'good' as part of the name is a little confusing here...

hzeller (changes requested):
There might be a few corner-cases that we should make sure are covered.

pawelsag:
I have encountered one problem when I tested the rule.
The struct/union name can't be started with digits, so even when we specify the exception words for this rule,
we have no chance to verify them because it violates another rule.
Should we solve this in the current issue?

verilog/analysis/checkers/struct_union_name_style_rule_test.cc
```diff
@@ -44,6 +53,41 @@ TEST(StructUnionNameStyleRuleTest, ValidStructNames) {
   RunLintTestCases<VerilogAnalyzer, StructUnionNameStyleRule>(kTestCases);
 }
 
+TEST(StructUnionNameStyleRuleTestConfigured, ValidStructNames) {
+  const absl::string_view exceptions = "exceptions:12B,11GB,14kJ,B10,t";
```
snsokolov, line 57: I think the exceptions in the original issue should be about the Unit names only. I.e. this test case needs to be: "exceptions:B,GB,kJ"

hzeller: I think with just using the units and not numbers, your last concern would go away @pawelsag ?

pawelsag: Yes, thanks for this. I am about to update the PR

hzeller:
I don't think the current implementation is capturing the request in #385 yet.

There shall be exceptions (units such as J, GW, KiB...) that follow directly digits that shall be allowed in such structs.
So `foo_16B_bar` is good, but `foo_B_bar` would not. So we also need to look if this is actually after some number.

pawelsag:
Ok, I will reimplement this to the pattern from the original issue to allow only for exceptions where letters are just after the digits.
But there is one thing to consider. The name: `12B_foo_t` of the struct/union will be invalid, cause there is a rule that variables can't start with digits, so the rule which I'm currently implementing wouldn't have a chance to verify it.
Would it be a correct behavior?

hzeller:
I think a type name that starts with a digit will already not pass the lexer/parser, so you can probably ignore that case.

verilog/analysis/checkers/struct_union_name_style_rule.cc
```diff
@@ -79,15 +82,65 @@ void StructUnionNameStyleRule::HandleSymbol(const verible::Symbol& symbol,
...
+          std::find_if_not(ex.begin(), ex.end(), absl::ascii_isalnum);
+      // eliminate exceptions with invalid characters
+      if (e != ex.end()) continue;
+
+      const auto &alpha =
+          std::find_if(ex.begin(), ex.end(), absl::ascii_isalpha);
+      // find first alpha sign
+      // check if string do not start with alpha sign
+      // and if there are any alpha signs
+      if (alpha == ex.begin() || alpha == ex.end()) continue;
```
hzeller, line 132: But isn't a 'numbers followed by a particular unit' is what we'd like to offer ?
If I have a configuration `GiB`, then I'd allow things that contain `foo_12GiB_bar_t`
(#385 states _" number followed by a unit"_ which means not a specific number, just something numeric.)

So the configuration should not (only) explicitly ask for a particular number, just that the exception is something found in the string _prepended_ by something that looks like  a number.

So if someone would like to allow `B`, `GiB` and `J`, then strings such as
`foo_12B_bar_t`, `foo_13B_t`, `baz_42GiB_t` or `quux_2134J_energy_t` would be allowed.
Strings like `foo_B_t` or `baz_GiB_t` or `baz2_GiB_t` would _not_ be allowed (because they are not prepended with a digit).

The exception can be made of course more specific, so `12B` would _only_ allow `foo_12B_t` but not `foo_13B_t`

verilog/analysis/checkers/struct_union_name_style_rule.cc
```diff
@@ -59,16 +62,16 @@ std::string StructUnionNameStyleRule::GetDescription(
       GetStyleGuideCitation(kTopic), ".");
```
hzeller, line 62: We need to add a description about the configurable options here, so that they are shown to the user and show up in the auto generated documentation; here is an [example how a parameter output](https://google.github.io/verible/lint.html#parameter-name-style) looks like.

verilog/analysis/checkers/struct_union_name_style_rule.cc
```diff
@@ -79,15 +82,65 @@ void StructUnionNameStyleRule::HandleSymbol(const verible::Symbol& symbol,
       // Neither a struct nor union definition
       return;
     }
-    const auto* identifier_leaf = GetIdentifierFromTypeDeclaration(symbol);
+    const auto *identifier_leaf = GetIdentifierFromTypeDeclaration(symbol);
     const auto name = ABSL_DIE_IF_NULL(identifier_leaf)->get().text();
-    if (!verible::IsLowerSnakeCaseWithDigits(name) ||
+    // get rid of the exceptions specified by user
+    std::string name_filtered;
+    if (!exceptions_.empty()) {
```
hzeller, line 89: I don't think we can do it this simple (see below about configuration).

We _do_ want to allow configurations that just contain units, such as `GiB`, and want to allow all items that looks like
`<some number>GiB`.

So instead of re-writing the name it is probably better to check this the other way around; this implementation is mostly a remnant from the earlier times when you had to iterate through the configuration string here thus things needed to be in order of that string, but now things are much easier as we have a `std::set` of exceptions, so we can instead look-up exceptions directly in the set. Which also makes it less brittle and much easier to understand (If I didn't know what you try to achieve, I would have to read through this code multiple times to understand what its intend is. Let's make it simpler for us looking at this code in the future...)

So rough implementation suggestion would probably be to go through the name in order, do early-out once we find one of the issues and use the capabilities of directly finding something in a `std::set` to test for the rest.
That way, things are easy to maintain in the future as each step is simple.

Pseudo-code of what I mean:
```
if (!absl::endsWith(name, "_t")) {
   violations.insert(...);
   return;  // detected already first violation: early out.
}
if (name[0] == '_') {
   violations.insert("can not start with underscore...");  // this allows us also to be more specific
   return;   // early out. violation detected, not a problem anymore for our later analysis
}
if (absl::IsLowerCaseSnakeWithDigits(...)) {
  return;  // Name looks cool already. We could leave this out as we essentially implement everyting below.
}
// now, our turn
for (const auto &part : absl::Split(name, '_')) {
   if (alllowercaseordigits(part)) continue;
   if (!isdigit(*part.begin()) {
       violation("sections with unit names need to start with digit");
       return;
   }
   if (exceptions.find(part) != exceptions.end()) continue;  // direct match of things that have digits in exceptions, e.g. 12B
   if (exceptions.find(remove_leading_digits(part)) == exceptions.end()) {
       violation("found digit followed by unit that is not configured as allowed exception");
       return;
   }
}
```

verilog/analysis/checkers/struct_union_name_style_rule.cc
```diff
@@ -79,15 +82,65 @@ void StructUnionNameStyleRule::HandleSymbol(const verible::Symbol& symbol,
...
+      name_filtered.reserve(name.length());
+      const auto &underscore_separated_parts = absl::StrSplit(name, '_');
+      if (name[0] != '_') {
+        for (const auto &ns : underscore_separated_parts) {
+          if (std::find(exceptions_.begin(), exceptions_.end(), ns) ==
```
hzeller, line 94: Note: Finding things in a set is easier: you can do `exceptions.find(...) != exceptions.end()` - which is a fast `O(log N)` tree-lookup (std::set) or `O(1)` hash-lookup (std::unordered_set).

Using `std::find` essentially forces it iterate through the whole set, making it an `O(n)` operation; this is only needed for containers that don't support direct look-ups.

hzeller (changes requested):
So I still think this is not capturing yet fully what is asked for in #385. Comments in-line.

verilog/analysis/checkers/struct_union_name_style_rule.cc
```diff
@@ -79,13 +80,83 @@ void StructUnionNameStyleRule::HandleSymbol(const verible::Symbol& symbol,
...
+        continue;
+      }
+      if (!absl::ascii_isdigit(*ns.begin())) {
+        violations_.insert(LintViolation(
+            identifier_leaf->get(),
+            "Section with unit names need to start with digit", context));
+        return;
+      }
+      if (exceptions_.find(std::string(ns)) != exceptions_.end()) {
+        continue;
```
hzeller, line 112: Maybe add a comment `// explicit number + unit exception found`

hzeller (changes requested):
Some small suggestions which I think will improve readability.

hzeller (changes requested):
LGTM
just a little formulation change in the parameter documentation and we're good to go.

hzeller (approved):

PR #659, opened 2021-01-30
Fix confusing --rules_config flag: it shows a default but doesn't use…

hzeller:
… it.

The --rules_config flag only works if there is a file explicitly set. So
set the default to an empty string to clarify that this is what is
happening.

Could also simplify the logic, as we now can just look at the empty
string instead of having a separate flag containing the information if
the flag was set.

(The default value was a left-over from before there was a choice
between filename and upwards search I think)

Issues #229


tgorochowik (approved):
Makes sense, LGTM!

PR #660, opened 2021-02-02
formatter/doc: Add some info about formatting code

ldalek-antmicro:
Add some info about:
1. TreeUnwrapper code and class inheritance
2. Short info about custom policies
3. Short info about logging with glog library


doc/formatter_simplified_class_diagram.graphml
```diff
@@ -0,0 +1,180 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
hzeller, line 1: Maybe we should document how to regenerate the PNG from from the graphml file.

So personally, I'd probably just suggest using a Makefile right in this directory as BUILD feels like too big of a stone for that bird....

hzeller (approved):
Oh, very nice improvement of the documentation!

I'll leave this open, maybe @fangism is bored and wants to have a look :)

doc/formatter.md
```diff
@@ -67,6 +67,79 @@ subproblem: _Given a box width constraint, what are the best ways to format a
...
+}
+```
+
+```cpp
+void TreeUnwrapper::VisitIndentedSection(const SyntaxTreeNode& node,
+                                         int indentation_delta,
+                                         PartitionPolicyEnum partitioning) {
+  const auto last_ftoken_iter =
+      VisitIndentedChildren(node, indentation_delta, partitioning);
+  // TODO(fangism): do we ever need to remove trailing empty partitions here?
```
fangism, line 126: TODO(me)? Well that's embarrassing!  Probably not a good idea to name authors in documentation, even well-intended ones.  :)

doc/formatter.md
```diff
@@ -67,6 +67,79 @@ subproblem: _Given a box width constraint, what are the best ways to format a
...
+  StartNewUnwrappedLine(PartitionPolicyEnum::kUninitialized, nullptr);
+}
+```
+
+### Custom policies application
+
+The next stage of code formatting is to apply custom functions that would
+reshape the partition tree according to the selected policy.
+An example of that are the [kTabularAlignment](#alignment) and the
+[kAppendFittingSubPartitions](https://cs.opensource.google/verible/verible/+/master:common/formatting/unwrapped_line.h;bpv=1;bpt=1;l=65)
```
fangism, line 140: Try the `symbol:` search atom.

doc/formatter.md
```diff
@@ -92,6 +165,16 @@ InterTokenInfo before each token.
 
 A lot of other details are traced by running with `-v=4 --alsologtostderr`.
 
+Selective output from Verible modules can be controled via [glog] environment
```
fangism, line 168: "controlled"

Also, this might be a better fit somewhere in general developer documentation.

doc/formatter_simplified_class_diagram.graphml
fangism: I found a few markdown enhancements that support more inline rendering:
https://github.com/shd101wyy/markdown-preview-enhanced/blob/master/docs/diagrams.md
Could be worth a look.

I'm also wondering if the open-source codesearch already provides visualization tools.  They have full semantic information to be able to navigate class hierarchies (kythe).  Having the code as source-of-truth again mitigates risk of documentation going stale.

fangism (commented):
Thank you for continuing to improve the documentation for developers!

Other suggestions outside of this PR:
* Find the most trivial PRs from the past that fix one thing (e.g. spacing or indentation) and link to them as examples.  I think I relied too heavily on contributors' search powers to enable blame layers and dig through revision history.

doc/formatter_simplified_class_diagram.graphml
hzeller: Wow, that is an interesting list of things that can be pre-viewed in github.

I was always annoyed that they don't allow viewing SVGs (because it allows potentially dangerous scripting), but this certainly can help in many of the documentation situations.

hzeller: mmh, I tried it,  looks like this is not supported by github. These previews there are done by an add-on for a [particular editor](atom.io) (and that was a very bad 'particle pun').

doc/formatter.md
ldalek-antmicro: I'm keeping it this way as search engine can't find such symbol (which is weird because it does find enum constants in abseil code).

PR #663, opened 2021-02-02
Add a .clang-format file and update CONTRIBUTING.md

hzeller:
Putting the requested style in the clang config file makes
it easiest to refer to and possibly can be automatically
be picked up by the contributors' IDE.

Update CONTRIBUTING file to now include that this is not
done automatically anymore on import but needs to be prepared
by the contributor.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -38,8 +38,8 @@ information on using pull requests.
 
 Our C++ code mostly follows [Google's C++ style guide][google-cpp-style].
 
-The github actions on a pull request will verify this style, so to make
-this smooth, run `clang-format` on your modified files before submitting
+On a pull request, the Continuous Integration (CI) will verify this style, so
+to make this smooth, run `clang-format` on your modified files before submitting
 a pull request:
 
 ```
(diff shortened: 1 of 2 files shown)


mithro (approved):
LGTM!

fangism (approved):
There should be a way to add a `git clang-format` hook from this.  This is a good start!

hzeller:
In #661 I added a clang-format check on pull request. So it will at least not pass CI when things are _not_ formatted correctly.

Last time I checked, a clang-format hook could not be enforced - it was dependent on if the user has installed it or runs it.

umarcor (approved):

CONTRIBUTING.md
```diff
@@ -36,9 +36,15 @@ information on using pull requests.
 
 ### Style
 
-Our C++ code mostly follows [Google's C++ style guide][google-cpp-style]. We can
-take care of automatic code formatting for you when we import your pull
-requests.
+Our C++ code mostly follows [Google's C++ style guide][google-cpp-style].
+
+The github actions on a pull request will verify this style, so to make
```
umarcor, line 41: Negligible suggestion: you might want to use "Continuous Integration (CI)" instead of "github actions", so you/we don't need to update this in the future (as it's happening with many references to Travis in open source projects).

PR #664, opened 2021-02-03
ci: fix missing '|' and indentation issues producing an invalid workflow

umarcor:


hzeller:
Thanks!
I've recently added an `.editorconfig` which hopefully makes sure that the indentation is always with 2 characters.

hzeller:
The indentation seemed to have got undone on the environment modes, which made the CI unhappy https://github.com/google/verible/runs/1826227818

Attempt to fix in #665

umarcor:
That's interesting, because the same commit worked in my fork (https://github.com/umarcor/verible/actions/runs/535250992). Anyway, it's fixed now :D

PR #665, opened 2021-02-03
Fix env-MODE: got unindented in last commit

hzeller:


hzeller:
Matrix failed after #664 https://github.com/google/verible/runs/1826227818
Indentation of the MODE list seems to be the culprit.

Issue #653, opened 2021-01-27
Add clang-format and clang-tidy check in pull requests


hzeller:
Let's add another check to the presubmit checks: run all modified/new c++ files through `clang-format --style=Google` and report back if there is a diff. Similarly `clang-tidy` (need to figure out all useful options).

Show a diff and ask the user to run clang-format/clang-tidy locally on the code and update the pull request in that case.

(We have to report and can probably not update the pull request in-place with clang-format, because that might mess with a WIP pull request.)

hzeller:
But maybe we could just run `clang-format` as a hook at time of merge ? That way, we don't have to bother the contributor with an insignificant technicality (we could still make this a pull request check, but not a merge-blocking one).

In the past operating model we were actually [doing this](https://github.com/google/verible/blob/b343939bb233786cca3469134c414ca40312fdfc/CONTRIBUTING.md#style), but we have to build out these things without the additional round-trip.

mithro:
@hzeller -- I believe there are github actions that can be set up which will add a patch onto people's pull requests with the reformatting (or maybe just a set of suggestions?).

mithro:
@umarcor -- Thoughts?

umarcor:
It is technically possible to have a GitHub Actions workflow/job which interacts with either git or the GitHUb API for:

- Editing the original message in the PR.
- Posting a comment in the PR, or editing an already posted comment by the bot.
- Committing to the PR or to any other repository anywhere.

However, the scope of the default token is limited, and it does not suffice for pushing commits on top of existing PRs from forks. It will only work for PRs created from feature branches in this same repo.

Updating any PR would require the Personal Access Token of some maintainer to be added as a CI secret. That is undesirable, because then the scope is too wide and it's a potential security issue.

I would recommend avoiding having PRs updated automatically through this mechanism. If code style is considered code quality, IMHO it should be a regular job that fails. Contributors should take care of running clang-format and clang-tidy (through a hook, if they want to). Alternatively, if the style is not considered critical, I would not add any CI check, and just let maintainers reformat periodically.

mithro:
BTW -- GitHub has a way to "suggesting" changes in a comment that the owner of the pull request can accept.

umarcor:
It has. However, I think that each suggested change is saved as a different commit. It makes squashing almost a requirement... I find it desirable instead for CI to fail and contributors to run formatting locally.

hzeller:
[...]

PR #669, opened 2021-02-04
Now that we have a .clang-format, use that file as reference.

hzeller:
Previously, we choose a particular clang-format built-in style
('Google'), but since we now have a .clang-format configuration file
use that instead to make sure we always apply the same rules.

Issues #653


hzeller:
(I've added you as writer in the repository @tgorochowik , so should be able to merge this change from the UI)

tgorochowik (approved):

PR #670, opened 2021-02-05
ci: move commands from scripts to workflow

umarcor:
This is another (probably the last) step towards #622.


.github/workflows/verible-ci.yml
```diff
@@ -123,47 +132,70 @@ jobs:
...
       uses: actions/upload-artifact@v2
       with:
         name: verible-binary-${{ matrix.OS }}-${{ matrix.OS_VERSION }}
-        path: /tmp/releases/verible-*.tar.gz
+        path: releasing/out/verible-*.tar.gz
+
+
+  Release:
+    if: github.ref == 'refs/heads/master' && github.event_name != 'pull_request'
+    needs: [ Check, Build]
```
mithro, line 168:
```suggestion
    needs: [ Check, Build ]
```

.github/workflows/script.sh
```diff
@@ -27,13 +27,6 @@ compile|clean)
     bazel build $BAZEL_OPTS //...
     ;;
 
-bin)
```
mithro, line 30: I think it is useful to be able to run this script locally and get a lot of the same behaviour as the CI system.

mithro (commented):
I think there are some good changes in here but I also want to make sure we don't move too much into the GitHub Actions workflow making it hard to run the scripts locally and hard to move to a different CI system.

Thoughts?

hzeller:
Neat emoji characters you introduced for the various steps :)

I'd like to second the importance of being able to run parts locally, so keeping that in mind helps deciding what should be scripts and what should go in the CI (I am not implying that it was good before or is bad now or that there should be any changes, just that this should be considered with this in mind).

A CI yml file can't be executed locally (or is there a runner github provides?), while scripts can. OTOH, it is easy to collect 'cruft' in scripts that make things less declarative. A balance.

umarcor:
> I think there are some good changes in here but I also want to make sure we don't move too much into the GitHub Actions workflow making it hard to run the scripts locally and hard to move to a different CI system.
> 
> Thoughts?

@mithro, I did explicitly try to preserve that. Let me explain:

Currently, most/all helper scripts are located in `.github/workflows/`. Hence, their location suggests that they are to be used in GitHub only and for CI only. If the scripts were to be used either locally or in a different CI system, those should be better located in e.g. `.ci`. That is apparently some negligible distinction, but it illustrates that using these scripts elsewhere is more of a theoretical option than an actually used (supported) feature. If these scripts are/were actually to be used outside of GitHub Actions, other several modifications would be required. At the moment, they are separated files, but not generic enough.

Anyway, these modifications are already done with local execution in mind. See the final proposal (#622), not this intermediate state: https://github.com/umarcor/verible/blob/ci/rework/.github/workflows/verible-ci.yml. E.g., any user will be able to execute the following:

```
source ./.github/settings.sh; ./releasing/docker-run.sh OS:VERSION TASK
```

where TASK is `prepare-image-basic`, `prepare-image-build` or `build-verible-release`.

I think it is desirable to evaluate whether the proposal is constraining execution on other environments, when we do have some other specific use case. 

My main motivation for removing the existing scripts is that they were based on a fixed set of steps (install, before script, script, and after success). That did make sense in Travis, but it is artificial in GitHub Actions. As a result, in several jobs some of those scripts are called for doing nothing. I find that misleading, because it provides some apparent homogeneity, but it is unnecessary verbosity in fact.

Note that I did preserve the helper files which are actually used more than once (e.g. `set-compiler.sh` or `install-bazel.sh`).

.github/workflows/script.sh
umarcor: The relevant part of this snippet is `./docker-run.sh ${OS}-${OS_VERSION}`. That's what users can execute locally (or in a different CI). The motivation for removing it from here is that:

- Tarballs are not generated in `/tmp/releases` anymore, so L33-L34 are not required.
- `docker-run.sh` does not need to be called in-place, so L31 is not required.

umarcor:
> Neat emoji characters you introduced for the various steps :)

@hzeller, I'm glad you liked them! We use emojis/icons a lot in GHDL:

- https://github.com/ghdl/ghdl/actions/runs/538005715
- https://github.com/ghdl/ghdl-cosim/actions/runs/540771169

As a hardware guy, I need to "see" the systems I'm working with, and I find these symbols really helpful for navigating through the multiple parallel and dependent jobs. However, some people might find them not to be "serious" (say professional), so I refrained from adding too many here.

Should you find them useful, I can add symbols for each job name and/or step. The list of available emojis is not specially expressive (https://github.com/ikatyang/emoji-cheat-sheet) so some creativity is required for picking them. But I believe it's a matter of having some consistency in the CI of several projects, until we get used to the meaning.

> I'd like to second the importance of being able to run parts locally, so keeping that in mind helps deciding what should be scripts and what should go in the CI (I am not implying that it was good before or is bad now or that there should be any changes, just that this should be considered with this in mind).

Please, see my previous reply/clarification to Tim.

> A CI yml file can't be executed locally (or is there a runner github provides?), while scripts can. OTOH, it is easy to collect 'cruft' in scripts that make things less declarative. A balance.

Strictly, it is possible to execute GitHub Actions workflow files locally. GitHub provides a runner that you can install/start: https://github.com/actions/runner/. There is also https://github.com/nektos/act.

Nevertheless, I would not recommend executing the YAML file locally. I do agree with both of you: if some script is to be executed locally or in a different CI env, then that script should be a bash file and we should just call it in the workflow. I tried to keep a balance between what is better done declaratively in the workflow, and what is better collected in scripts. Please, take into account that this is WIP towards #622. At the same time, please do discuss/comment specific changes where I might have gone too far.

hzeller:
I agree, I think what is in workflows and what in scripts is actually quite nice and reasonable here.

I'll leave it to @mithro to final review and merge.

mithro (approved):
LGTM.

PR #673, opened 2021-02-09
option to show line with diagnostic and a marker

### LINKED ISSUE #531: option to show line with diagnostic and a marker (clang-style)
**Summary**

from @snsokolov :
Currently `verible-verilog-syntax` and `verible-verilog-lint` display diagnostics in the form:

FILE:LINE:COL: ...

but it might be nice to show the user the whole line with the offending text, plus a marker (clang-style diagnostics):

```
file.sv:5:11: error here...
  foo bar baz;
          ^
```

In other modes when the linter is being used in a service, and is expected to print only (machine parse-able) FILE:LINE:COL:, this would need to be disabled.

pawelsag:
fixes #531

### DIFF
--- a/common/analysis/file_analyzer.h
+++ b/common/analysis/file_analyzer.h
@@ -94,8 +94,14 @@ class FileAnalyzer : public TextStructure {
   std::vector<std::string> TokenErrorMessages() const;
 
   // Diagnostic message for rejected tokens for linter.
+  // Second argument is the show_context option. When enabled
+  // additional diagnostic line is concatenated to an error message 
+  // with marker that points to vulnerable token
   std::string LinterTokenErrorMessage(const RejectedToken&, bool) const;
 
+  // First argument is the show_context option. When enabled
+  // additional diagnostic line is concatenated to an error message 
+  // with marker that points to vulnerable token
   std::vector<std::string> LinterTokenErrorMessages(bool) const;
 
   const std::vector<RejectedToken>& GetRejectedTokens() const {
--- a/common/analysis/file_analyzer.h
+++ b/common/analysis/file_analyzer.h
@@ -94,9 +94,9 @@ class FileAnalyzer : public TextStructure {
   std::vector<std::string> TokenErrorMessages() const;
 
   // Diagnostic message for rejected tokens for linter.
-  std::string LinterTokenErrorMessage(const RejectedToken&) const;
+  std::string LinterTokenErrorMessage(const RejectedToken&, bool) const;
 
-  std::vector<std::string> LinterTokenErrorMessages() const;
+  std::vector<std::string> LinterTokenErrorMessages(bool) const;
 
   const std::vector<RejectedToken>& GetRejectedTokens() const {
     return rejected_tokens_;
--- a/common/analysis/file_analyzer_test.cc
+++ b/common/analysis/file_analyzer_test.cc
@@ -71,6 +71,26 @@ TEST(FileAnalyzerTest, TokenErrorMessageSameLine) {
   }
 }
 
+// Verify that an error token on one line is reported correctly.
+TEST(FileAnalyzerTest, TokenErrorMessageSameLineWithContext) {
+  const std::string text("hello, world\nbye w0rld\n");
+  FakeFileAnalyzer analyzer(text, "hello.txt");
+  const TokenInfo error_token(1, analyzer.Data().Contents().substr(17, 5));
+  {
+    const auto message = analyzer.TokenErrorMessage(error_token);
+    EXPECT_EQ(message, "token: \"w0rld\" at 2:5-9");
+  }
+  {
+    constexpr bool with_diagnostic_contex = true;
+    const auto message = analyzer.LinterTokenErrorMessage(
+        {error_token, AnalysisPhase::kParsePhase}, with_diagnostic_contex);
+    EXPECT_TRUE(
+        absl::StrContains(message,
+                          "hello.txt:2:5: syntax error, rejected \"w0rld\" "
+                          "(syntax-error).\nbye w0rld\n    ^"));
+  }
+}
+
 // Verify that an error token on one character is reported correctly.
 TEST(FileAnalyzerTest, TokenErrorMessageOneChar) {
   const std::string text("hello, world\nbye w0rld\n");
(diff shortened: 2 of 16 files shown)


common/analysis/lint_rule_status.cc
```diff
@@ -82,11 +82,33 @@ void LintStatusFormatter::FormatLintRuleStatuses(
...
+    absl::string_view base, absl::string_view path) const {
+  auto violations = PrepareViolations(statuses);
+  for (auto violation : violations) {
+    FormatViolation(stream, *violation.violation, base, path,
+                    violation.status->url, violation.status->lint_rule_name);
+    *stream << std::endl;
+  }
+}
+
+void LintStatusFormatter::FormatLintRuleStatuses(
```
hzeller, line 99: This duplication of functions is of course increasing the maintenance burden in the future.

I'd prefer it to be only one function.

Maybe we remove the old function and add this new function. And we would only ...

common/analysis/lint_rule_status.cc
```diff
@@ -82,11 +82,33 @@ void LintStatusFormatter::FormatLintRuleStatuses(
...
+    std::ostream* stream, const std::vector<LintRuleStatus>& statuses,
+    absl::string_view base, absl::string_view path,
+    const std::vector<absl::string_view>& lines) const {
+  auto violations = PrepareViolations(statuses);
 
   for (auto violation : violations) {
     FormatViolation(stream, *violation.violation, base, path,
                     violation.status->url, violation.status->lint_rule_name);
     *stream << std::endl;
+    auto cursor = line_column_map_(violation.violation->token.left(base));
```
hzeller, line 109: ... execute this `if (lines.size() > cursor)`.
that way, we can deliberately choose to just pass `{}` in places in which we don't want to show context.

(also slightly more robust in case we have `cursor.line` for whatever reason is outside `lines[]`

The added benefit is, that the compiler will find us all the places where the old function is used, so that we can make a decision at all of these places if we want context or not.

verilog/analysis/verilog_linter.cc
```diff
@@ -299,8 +301,13 @@ absl::Status VerilogLintTextStructure(std::ostream* stream,
     VLOG(1) << "Lint Violations (" << total_violations << "): " << std::endl;
     // Output results to stream using formatter.
     verible::LintStatusFormatter formatter(contents);
-    formatter.FormatLintRuleStatuses(stream, linter_statuses, text_base,
-                                     filename);
+    if (!show_context) {
+      formatter.FormatLintRuleStatuses(stream, linter_statuses, text_base,
```
hzeller, line 305: So here, things would look like
`formatter.FormatLintRuleStatuses(stream, ..., show_context ? text_structure.Lines() : {});`

verilog/tools/lint/lint_tool_test.sh
```diff
@@ -96,6 +96,15 @@ status="$?"
...
 
+echo "=== Test --show_diagnostic_context"
+"$lint_tool" "$TEST_FILE" --show_diagnostic_context > /dev/null 2> "${MY_OUTPUT_FILE}.err"
+
+status="$?"
+[[ $status == 1 ]] || {
+  echo "Expected exit code 1, but got $status"
+  exit 1
+}
+
```
hzeller, line 107: Can we add some sort of shell test that determines that an output is in fact in the particular form 
```
<file>:<line>:<col>: errmsg
```
without the diagnostic context flag and with added ^ marker with ?

hzeller (changes requested):
This only covers the linter so far, right ? Are you planning on verible-verilog-syntax in a separate pull request ?

hzeller:
any update on this one ?

common/analysis/file_analyzer_test.cc
```diff
@@ -63,13 +63,34 @@ TEST(FileAnalyzerTest, TokenErrorMessageSameLine) {
...
+    const auto message = analyzer.TokenErrorMessage(error_token);
+    EXPECT_EQ(message, "token: \"w0rld\" at 2:5-9");
+  }
+  {
+    constexpr bool with_diagnostic_context = true;
+    const auto message = analyzer.LinterTokenErrorMessage(
+        {error_token, AnalysisPhase::kParsePhase}, with_diagnostic_context);
+    EXPECT_TRUE(
+        absl::StrContains(message,
+                          "hello.txt:2:5: syntax error, rejected \"w0rld\" "
```
hzeller, line 89: In general, in this test we want to show that whenever we have an error, we show the context and an arrow that hightlights where things are problematic.
So our tests should try to capture that.

I'd probably break this up so that the newlines align, so that it is easy to see that error message and 'arrow', that should point to "w0rld" above, line up.
If you break after the parenthesis, you should get more space to do so:

```
EXPECT_TRUE(absl::StrContains(
   message,
   ".......\n"
   " ...^...");
```

verilog/analysis/verilog_linter.cc
hzeller: Did you consider my suggestion to use the ?: operator for more compact and readable expression expression here ?

verilog/tools/lint/lint_tool_test.sh
hzeller: An end-to-end test here would be still good; testing if we get some expected error message.

verilog/analysis/verilog_linter.cc
pawelsag: Yes, but as far as I know the initializer_list can't be used inside the ternary operator. At least my compiler doesn't like such construction.
Am I right?

verilog/tools/lint/lint_tool_test.sh
pawelsag: I prepared changes for this issue last Friday, but they somehow vanished...
I haven't checked if the commit was pushed to the branch, sorry for that.
I have updated the PR with the missing tests

common/analysis/file_analyzer.h
```diff
@@ -94,9 +94,15 @@ class FileAnalyzer : public TextStructure {
   std::vector<std::string> TokenErrorMessages() const;
 
   // Diagnostic message for rejected tokens for linter.
-  std::string LinterTokenErrorMessage(const RejectedToken&) const;
-
-  std::vector<std::string> LinterTokenErrorMessages() const;
+  // Second argument is the show_context option. When enabled
+  // additional diagnostic line is concatenated to an error message 
+  // with marker that points to vulnerable token
+  std::string LinterTokenErrorMessage(const RejectedToken&, bool) const;
```
hzeller, line 100: Probably can't harm giving the parameters a name.

common/analysis/file_analyzer_test.cc
hzeller: Nice, now it is easy to see what is marked.

common/analysis/file_analyzer_test.cc
```diff
@@ -80,13 +103,35 @@ TEST(FileAnalyzerTest, TokenErrorMessageOneChar) {
...
+  }
+  {
+    constexpr bool with_diagnostic_context = true;
+    const auto message = analyzer.LinterTokenErrorMessage(
+        {error_token, AnalysisPhase::kParsePhase}, with_diagnostic_context);
+    EXPECT_TRUE(absl::StrContains(message,
+                                  "hello.txt:1:6: syntax error, rejected \",\" "
+                                  "(syntax-error).\n"
+                                  "hello, world\n"
+                                  "     ^"));
```
hzeller, line 131: Nice.

verilog/analysis/verilog_linter.cc
hzeller: I see. Ok, then this is probably best.

hzeller (approved):
LGTM.
I'd probably consider given some method parameters a name (in file_analyzer.h), but that can also be in a follow-up change.

Issue #384, opened 2020-08-06
forbid line continuations inside string-literals


fangism:
The jury has spoken.  
We now want to forbid line-continued string literals like this:

```systemverilog
"this is bad. \
don't to this"
```

and recommend the equivalent (C-style joined string literals):

```systemverilog
"this is better. "
"do this instead."
```

Suggestion: implement as a `TokenStreamLintRule`, scan for all string literals (even inside macro definition bodies), and report offending locations (focusing on the location of "\").

fangism:
b/154548140

pawelsag:
Hi, I wanted to solve this issue, but I have a question related to the example you wrote.
I can't find any information about C-style string in TRM, so this is some kind of SystemVerilog dialect?
Does this also mean that when doing this task the implementations of the parser should be changed?
I prepared the example:
```
module strings; 
string tmp = "Humpty Dumpty sat on a wall."
"Humpty Dumpty had a great fall.";
endmodule
```
and the output was
```
build/strings.sv:4:1: syntax error, rejected ""Humpty Dumpty had a great fall."" (syntax-error).
```

snsokolov:
The C-style string joining above isn't a legal Verilog.  Strings can be concatenated by either by using `{}` or `$sformatf`

```
string tmp = {
    "foo",
    "bar"
};

string tmp2 = $sformatf("%s%s", "foo", "bar");
```

pawelsag:
Thanks for the clarification

PR #674, opened 2021-02-09
forbid line continuations inside string-literals

pawelsag:
Fixes:  #384


verilog/CST/verilog_matchers.h
```diff
@@ -298,6 +298,12 @@ static const auto AlwaysCombKeyword =
 static const auto AlwaysFFKeyword =
     verible::matcher::MakePathMatcher({L(TK_always_ff)});
 
+// Matches occurrence of the 'always' keyword.
```
hzeller, line 301: matches occurence of 'always' keyword ? Sounds like a copy/paste situation here :)

verilog/analysis/checkers/token_stream_lint_rule.cc
```diff
@@ -0,0 +1,81 @@
...
+using verible::matcher::Matcher;
+
+// Register TokenStreamLintRule
+VERILOG_REGISTER_LINT_RULE(TokenStreamLintRule);
+
+absl::string_view TokenStreamLintRule::Name() { return "forbid-line-continuations"; }
+const char TokenStreamLintRule::kTopic[] = "forbid-line-continuations";
+const char TokenStreamLintRule::kMessage[] =
+    "The lines can't be continued with \'\\\', use C-style multiline "
+    "cocatenation";
```
hzeller, line 48: concatenation

verilog/analysis/checkers/token_stream_lint_rule.h
```diff
@@ -0,0 +1,60 @@
...
+#include "common/analysis/lint_rule_status.h"
+#include "common/analysis/syntax_tree_lint_rule.h"
+#include "common/text/symbol.h"
+#include "common/text/syntax_tree_context.h"
+#include "verilog/analysis/descriptions.h"
+
+namespace verilog {
+namespace analysis {
+
+// AlwaysCombRule finds occurrences of "always @*" or "always @(*)".
```
hzeller, line 30: This also looks like a copy/paste leftover...

verilog/analysis/checkers/token_stream_lint_rule_test.cc
```diff
@@ -0,0 +1,56 @@
...
+      {""},
+      {"module m;\nendmodule\n"},
+      {"module m;\n", "string tmp = {\"Humpty Dumpty sat on a wall.\",",
+       "\"Humpty Dumpty had a great fall.\"};", "\nendmodule"},
+      {"module m;\n",
+       "string tmp=",
+       {kToken,
+        "\"Humpty Dumpty sat on a wall. \\\nHumpty Dumpty had a great fall.\""},
+       ";",
+       "\nendmodule"},
```
hzeller, line 47: Might be probably good to have one test case of a regular backslash followed by something in the string.
a \n for instance or a \\. Which of course would mean they would be encoded as \\n or \\\\ here in the string encoding a string :)

hzeller (changes requested):
Mostly looks like there are some copy/paste leftovers. Other than that LGTM

verilog/analysis/checkers/token_stream_lint_rule.cc
```diff
@@ -0,0 +1,81 @@
...
+using verible::SyntaxTreeContext;
+using verible::matcher::Matcher;
+
+// Register TokenStreamLintRule
+VERILOG_REGISTER_LINT_RULE(TokenStreamLintRule);
+
+absl::string_view TokenStreamLintRule::Name() { return "forbid-line-continuations"; }
+const char TokenStreamLintRule::kTopic[] = "forbid-line-continuations";
+const char TokenStreamLintRule::kMessage[] =
+    "The lines can't be continued with \'\\\', use C-style multiline "
```
tgorochowik, line 47: @hzeller should we call this `C-style concatenation` even though it isn't exactly the same? Shouldn't we just recommend to use the `concatenation operator` or explicitly say to concatenate using braces?

hzeller: I agree, I think `concatenation operator with braces` would be easier to understand in this context.

hzeller (approved):
LGTM, after the change of the message as suggested by @tgorochowik

hzeller:
Thanks, merged!

PR #680, opened 2021-02-18
ci: fix path to artifacts in Release

umarcor:
#670 broke the Release job (which is not executed for PRs). This PR fixes and updates it.


hzeller:
Thanks!

PR #684, opened 2021-02-20
ci: add 'releasing/supported_bases.txt'

umarcor:
This is one additional step towards #622. I hoped to merge that one already, but I/we are not ready yet.

In this PR, the subdirs and `.keepme` files are removed from `releasing`. Instead, a `supported_bases.text` file is added, where the target distributions/environments (container images) are listed.

At the same time, script `docker-generate.sh` is merged into `docker-run.sh`. Instead of generating multiple dockerfiles, the one for the target task is generated and used only.

The purpose is to not build Verible in a dockerfile, which is unidiomatic. Instead, create a container image with all the dependencies, but then build Verible in a temporary container. For that reason, script `build.sh` is created. In `docker-run.sh`, the dockerfile is generated and the image is built. Then, a container is started and the build script is executed inside.

I could successfully move the execution of `github-pages-setup.sh` and `github-releases-setup.sh` into the new `build.sh`. However, moving `bazel build` produces an error on centos7, due to `BAZEL_CXXOPTS`. I could not find the problem there yet.

I propose to have this merged now, and I will then rebase #622, which will make it easier to analyze the centos7 issue.


.github/workflows/generate_matrix.sh
```diff
@@ -16,8 +16,9 @@
 set -e
```
mithro, line 16: Should we replace this shell script with a Python script?

.github/workflows/verible-ci.yml
```diff
@@ -5,6 +5,7 @@ on:
   pull_request:
 
 env:
+  DOCKER_BUILDKIT: 1
```
mithro, line 8: This might need a comment about why BUILDKIT is being used?

.github/workflows/verible-ci.yml
```diff
@@ -116,20 +118,23 @@ jobs:
...
       with:
-        name: verible-kythe-index
         path: kythe_output/*.kzip
 
 
   Matrix:
     runs-on: ubuntu-latest
+    name: Generate Build matrix
     outputs:
       matrix: ${{ steps.generate.outputs.matrix }}
```
mithro, line 131:
```suggestion
      matrix: ${{ steps.generate.outputs.matrix }}

```

releasing/docker-run.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
mithro, line 1: I wonder if we should replace this bash script with a Python version?

mithro (commented):
I _think_ I generally like the direction this is going but I worry about making these shell scripts more complicated (shell scripts quickly become unmaintainable).  Do you think we should rewrite them in Python?

.github/workflows/generate_matrix.sh
umarcor: Yes.

.github/workflows/verible-ci.yml
umarcor: There is no specific reason to do so. It's just the second generation. It provides advantages but there is no specific new feature which we are using (yet). On some platforms, buildkit is enabled by default.

.github/workflows/verible-ci.yml
umarcor: I don't see what's the change here...

releasing/docker-run.sh
umarcor: Yes, but not yet. This is going towards https://github.com/umarcor/verible/blob/ci/rework/releasing/docker-run.sh. I think we can change to Python when we achieve that.

.github/workflows/generate_matrix.sh
umarcor: Done

.github/workflows/verible-ci.yml
mithro: Adding a newline after matrix.....

.github/workflows/generate-matrix
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env python3
```
mithro, line 1: Let's make this file end in `.py`.

releasing/docker-run.sh
mithro: Okay, SGTM.

mithro (approved):
If you fix the script name to end in `.py` we can merge!

mithro:
@googlebot I consent.

PR #685, opened 2021-02-22
Add License and continuous integration badge.

hzeller:


mithro:
FYI - Badges are generally positioned under the first header before the first paragraph. See image below,
Screenshot from 2021-02-22 19-42-18

PR #687, opened 2021-02-25
releasing: move stages to dockerfiles

umarcor:
Instead of hardcoding the content of the dockerfiles in `releasing/docker-run.sh`, in this PR multiple `*.dockerstage` files are created. Yet, the stages are merged through `cat` to produce a single Dockerfile. Therefore, this PR does not involve any functional modification.

Moreover, scripts are moved from `.github/workflows` to `.github/bin`.

Ref: #622


hzeller:
This looks good to me; @mithro ?

hzeller:
Ping @mithro

mithro (approved):
LGTM!

mithro:
Looks like you might need to rebase here.

mithro:
FYI @umarcor - I rebased this pull request and then did a small refactoring of how the dockerstage files work at https://github.com/mithro/verible/tree/ci/dockerstages -- take a look and see what you think....

umarcor:
> FYI @umarcor - I rebased this pull request and then did a small refactoring of how the dockerstage files work at https://github.com/mithro/verible/tree/ci/dockerstages -- take a look and see what you think....

I'm not sure about that refactoring being completely correct. I picked the directory structure from it. However, it feels to me that the for loop is missing some logic. I don't see where is the selection between `common` and `TARGET_VERSION` done. See https://github.com/umarcor/verible/commits/ci/dockerstages-dir and https://github.com/umarcor/verible/actions/runs/707392711.

mithro:
@umarcor -- The general idea was, "look for most specific file, then fall back to a less specific file, then fall back to a general file".

mithro (approved):
LGTM. I'm going to merge this in the current state. We can continue to improve it later.

mithro:
> @umarcor -- The general idea was, "look for most specific file, then fall back to a less specific file, then fall back to a general file".

IE For example,

 - 1. `<OS>`/`<OS version>`/`xxx.dockerfile`
 - 2. `<OS>`/`xxx.dockerfile`
 - 3. `xxx.dockerfile`

For example for `compilers` on `centos-6` use `$TOP/centos/6/compilers.dockerstage` if found. If not found use the OS specific version at `$TOP/centos/compilers.dockerstage`. If that isn't found use the generic file found at `$TOP/compilers.dockerstage`.

Does that make sense?

I didn't actually test that my code works however.

umarcor:
> LGTM. I'm going to merge this in the current state. We can continue to improve it later.

Yes, please, and thanks :D

> I didn't actually test that my code works however.

By default, the changes you push are not executed in CI. You need to add a temporary commit such as https://github.com/umarcor/verible/commit/584f39e82b8fc62af24db2d8d84e38524c3edb2b

> > @umarcor -- The general idea was, "look for most specific file, then fall back to a less specific file, then fall back to a general file".
> 
> IE For example,
> 
> * 1. `<OS>`/`<OS version>`/`xxx.dockerfile`
> 
> * 1. `<OS>`/`xxx.dockerfile`
> 
> * 1. `xxx.dockerfile`
> 
> 
> For example for `compilers` on `centos-6` use `$TOP/centos/6/compilers.dockerstage` if found. If not found use the OS specific version at `$TOP/centos/compilers.dockerstage`. If that isn't found use the generic file found at `$TOP/compilers.dockerstage`.
> 
> Does that make sense?

Thanks for explaining. It does make sense as a general approach given the current state. However, this is work in progress towards #622. The idea is to merge these partial dockerfiles for each "OS version" into a single one for each distribution, which uses `--build-args`. So, there will only be 2 levels instead of three. Let me rebase and rework #622 after this is merged, and I will make a proposal taking your thoughts into account.

Issue #693, opened 2021-03-03
Syntax errors due to streaming `with` operator


udinator:
The  SV linter is throwing unexpected errors on the `with` construct used in streaming operations, even though this construct is part of the LRM.
Example error message:
`../src/lowrisc_dv_mem_bkdr_if_0/sram_scrambler_pkg.sv:58:28: syntax error, rejected "with" (https://github.com/google/verible).`

Example test case:
```systemverilog
module test;
  initial begin
    logic unpacked_arr [16] = '{0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1};
    logic [3:0] nibble_arr[4];
    for (int i = 0; i < 4; i++) begin
      nibble_arr[i] = {<< {unpacked_arr with [i*4 +: 4]}};
    end
  end
endmodule
```

udinator:
b/181792412

hzeller:
Context: this is LRM 14.4.14
And yes, confirmed, it is indeed missing.

hzeller:
Also caught by [sv-tests](https://symbiflow.github.io/sv-tests-results/#Verible|11.4.14.4|dynamic_array_unpack_stream_with)

hzeller:
More context: this was [indeed thought of](https://github.com/google/verible/blob/35680ebb3eed3ebc0200610252f857fa8e381510/verilog/parser/verilog.y#L3223-L3225) but not added initially due to parser generation conflicts.
(just collecting this here as context for whoever will be working on this)

PR #694, opened 2021-03-08
Work-in-progress #693.

hzeller:
Work-in-progress: First step
  * enable the rule to allow streaming operator with 'with'
  * temporarily disable the rules that will create conflicts.
This will accept the streaming operator with 'with' syntax
at expense of other constructs.

Next step: resolve shift-reduce conflicts.

### DIFF
--- a/verilog/parser/verilog.lex
+++ b/verilog/parser/verilog.lex
@@ -672,6 +672,7 @@ wait_order { UpdateLocation(); return TK_wait_order; }
 wildcard { UpdateLocation(); return TK_wildcard; }
 <COVERGROUP>with { UpdateLocation(); return TK_with__covergroup; }
 with { UpdateLocation(); return TK_with; }
+"with"[ \t]*"[" { UpdateLocation(); return TK_with_followed_by_bracket; }
 within { UpdateLocation(); return TK_within; }
 timeprecision_check { UpdateLocation(); return TK_timeprecision_check; }
 timeunit_check { UpdateLocation(); return TK_timeunit_check; }
--- a/verilog/parser/verilog.lex
+++ b/verilog/parser/verilog.lex
@@ -672,7 +672,12 @@ wait_order { UpdateLocation(); return TK_wait_order; }
 wildcard { UpdateLocation(); return TK_wildcard; }
 <COVERGROUP>with { UpdateLocation(); return TK_with__covergroup; }
 with { UpdateLocation(); return TK_with; }
-"with"[ \t]*"[" { UpdateLocation(); return TK_with_followed_by_bracket; }
+"with"[ \t]*"[" {
+  /* TODO: formulate the ws sequence as any whitespace+comment sequence*/
+  yyless(4);
+  UpdateLocation();
+  return TK_with__followed_by_bracket;
+}
 within { UpdateLocation(); return TK_within; }
 timeprecision_check { UpdateLocation(); return TK_timeprecision_check; }
 timeunit_check { UpdateLocation(); return TK_timeunit_check; }
--- a/verilog/parser/verilog.y
+++ b/verilog/parser/verilog.y
@@ -480,6 +480,7 @@ is not locally defined, so the grammar here uses only generic identifiers.
 %token TK_wait_order "wait_order"
 %token TK_wildcard "wildcard"
 %token TK_with "with"
+%token TK_with_followed_by_bracket
 %token TK_with__covergroup "with(covergroup)"
 %token TK_within "within"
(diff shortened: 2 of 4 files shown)


hzeller:
@fangism , can you check if this approach is sound ? Description of approach in CL description of 5764608e600d2affb83f5bdc049707e43b2ff247

verilog/parser/verilog.lex
```diff
@@ -672,6 +672,7 @@ wait_order { UpdateLocation(); return TK_wait_order; }
 wildcard { UpdateLocation(); return TK_wildcard; }
 <COVERGROUP>with { UpdateLocation(); return TK_with__covergroup; }
 with { UpdateLocation(); return TK_with; }
+"with"[ \t]*"[" { UpdateLocation(); return TK_with_followed_by_bracket; }
```
fangism, line 675: I would avoid merging this as a single token, however you can achieve context-sensitivity to give variants of the same textual token by using a combination of lexer states and `yyless` to rewind.  Keep in mind that when you create variants of tokens like `[` and `(` (and yes, even `;`) that the formatter needs to be handle these similarly.

verilog/parser/verilog.y
```diff
@@ -3217,12 +3218,18 @@ block_item_or_statement_or_null_list_opt
...
+  | expression TK_PO_NEG expression
+  ;
+
 stream_expression
   : expression
     { $$ = move($1); }
-  /* TODO(fangism):
-  | expression TK_with select_variable_dimension
-   */
+  | expression TK_with_followed_by_bracket array_range_expression ']'
```
fangism, line 3232: Related to earlier comment, separate the tokens, but you may give `[` its own flavor.

fangism (commented):
General question: what was the type of conflict when you tried to uncomment the TODO regarding `with`?  Can you summarize the conflict state, with depth=1 inbound states?

You're thinking of the right kind of tactics.  Also if there's any context-sensitivity you cannot implement in the lexer alone, there is lexical_context.{h,cc} available.

hzeller:
> General question: what was the type of conflict when you tried to uncomment the TODO regarding `with`? Can you summarize the conflict state, with depth=1 inbound states?

Not entirely sure what you mean that goes beyond the change description in 5764608e600d2affb83f5bdc049707e43b2ff247 (is there a particular output that I should be looking at from bison and extract that from there ?), but the problem arose when seeing `with` in this context:

```yacc
stream_expression
   : expression TK_with '[' array_range_expression ']'
```

Then the parser could shift assuming that this is an expression that has an optional `with` at the end (like expressions with constraint block), or reduce the previous expression and move on to the stream expression with-clause. So only seeing one token ahead (`TK_with`), created this conflict because only then the next token would decide what that is (which could either be `(` or `[`).

Solved this by inspecting that all the other cases are followed with something not `[`, so this could be disambiguated by just peeking ahead in the lexer.

verilog/parser/verilog.lex
hzeller: Thanks, I think I'll reformulate it like that.

fangism: Another thing to keep in mind is *comments*.  You'll notice that in various state blocks in the lexer, I had to explicitly add comment handling to push state.

fangism:
> > General question: what was the type of conflict when you tried to uncomment the TODO regarding `with`? Can you summarize the conflict state, with depth=1 inbound states?
> 
> Not entirely sure what you mean that goes beyond the change description in [5764608](https://github.com/google/verible/commit/5764608e600d2affb83f5bdc049707e43b2ff247) (is there a particular output that I should be looking at from bison and extract that from there ?), but the problem arose when seeing `with` in this context:
> 
> ```yacc
> stream_expression
>    : expression TK_with '[' array_range_expression ']'
> ```
> 
> Then the parser could shift assuming that this is an expression that has an optional `with` at the end (like expressions with constraint block), or reduce the previous expression and move on to the stream expression with-clause. So only seeing one token ahead (`TK_with`), created this conflict because only then the next token would decide what that is (which could either be `(` or `[`).
> 
> Solved this by inspecting that all the other cases are followed with something not `[`, so this could be disambiguated by just peeking ahead in the lexer.

If one token look-ahead is enough then this might work.  If it handling comments (e.g. after `TK_with`) directly in the lexer starts to get messy, this might be done as simply in the lexical context class, where you'd add a tiny-state machine that ignores comments.

hzeller:
mmh, I rebased the branch, but that looks like it created a mess.
I'll create a separate pull request.

PR #695, opened 2021-03-10
Improve Actions hygiene

sethvargo:
👋 hello there! I'm a fellow Googler who works on projects that leverage GitHub Actions for CI/CD. Recently I noticed a large increase in our queue time, and I've tracked it down to the [limit of 180 concurrent jobs](https://docs.github.com/en/actions/reference/usage-limits-billing-and-administration) for an organization. To help be better citizens, I'm proposing changes across a few repositories that will reduce GitHub Actions hours and consumption. I hope these changes are reasonable and I'm happy to talk through them in more detail.

- Only run GitHub Actions for pushes and PRs against the main branch of the repository. If your team uses a forking model, this change will not affect you. If your team pushes branches to the repository directly, this changes actions to only run against the primary branches or if you open a Pull Request against a primary branch.

- For long-running jobs (especially tests), I added the "Cancel previous" workflow. This is very helpful to prevent a large queue backlog when you are doing rapid development and pushing multiple commits. Without this, GitHub Actions' default behavior is to run all actions on all commits.

There are other changes you could make, depending on your project (but I'm not an expert):

- If you have tests that should only run when a subset of code changes, consider gating your workflow to particular file paths. For example, we have some jobs that do Terraform linting, but [they only run when Terraform files are changed](https://github.com/google/exposure-notifications-verification-server/blob/c4f59fee71042cf668747e599e7c769fca736554/.github/workflows/terraform.yml#L3-L11).

Hopefully these changes are not too controversial and also hopefully you can see how this would reduce actions consumption to be good citizens to fellow Googlers. If you have any questions, feel free to respond here or ping me on chat. Thank you!

### DIFF
--- a/.github/workflows/verible-ci.yml
+++ b/.github/workflows/verible-ci.yml
@@ -2,7 +2,11 @@ name: verible-ci
 
 on:
   push:
+    branches:
+    - master
   pull_request:
+    branches:
+    - master
 
 env:
   DOCKER_BUILDKIT: 1
@@ -16,6 +20,11 @@ jobs:
     runs-on: ubuntu-20.04
 
     steps:
+
+    - name: Cancel previous
+      uses: styfle/cancel-workflow-action@0.7.0
+      with:
+        access_token: ${{ github.token }}
 
     - name: Checkout code
       uses: actions/checkout@v2
@@ -45,6 +54,11 @@ jobs:
     name: Check · ${{ matrix.mode }}
 
     steps:
+    
+    - name: Cancel previous
+      uses: styfle/cancel-workflow-action@0.7.0
+      with:
+        access_token: ${{ github.token }}
 
     - name: Checkout code
       uses: actions/checkout@v2
--- a/.github/workflows/verible-ci.yml
+++ b/.github/workflows/verible-ci.yml
@@ -193,7 +193,7 @@ jobs:
     needs: [ Check, Build ]
     runs-on: ubuntu-20.04
     name: 📦 Release
-    if: ${{github.ref == 'refs/heads/main' || github.ref == 'refs/head/master'}}
+    if: ${{github.event_name == 'push' && (github.ref == 'refs/heads/main' || github.ref == 'refs/head/master')}}
 
     steps:
 
@@ -219,7 +219,6 @@ jobs:
         ls -lah artifact
 
     - name: Deploy release and assets
-      if: github.ref == 'refs/heads/master' && github.event_name != 'pull_request'
       uses: svenstaro/upload-release-action@v2
       with:
         repo_token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/verible-ci.yml
+++ b/.github/workflows/verible-ci.yml
@@ -24,7 +24,7 @@ jobs:
     steps:
 
     - name: Cancel previous
-      uses: styfle/cancel-workflow-action@0.7.0
+      uses: styfle/cancel-workflow-action@0.8.0
       with:
         access_token: ${{ github.token }}
 
@@ -58,7 +58,7 @@ jobs:
     steps:
     
     - name: Cancel previous
-      uses: styfle/cancel-workflow-action@0.7.0
+      uses: styfle/cancel-workflow-action@0.8.0
       with:
         access_token: ${{ github.token }}
 
@@ -193,6 +193,7 @@ jobs:
     needs: [ Check, Build ]
     runs-on: ubuntu-20.04
     name: 📦 Release
+    if: ${{github.ref == 'refs/heads/main' || github.ref == 'refs/head/master'}}
 
     steps:
 
--- a/.github/workflows/verible-ci.yml
+++ b/.github/workflows/verible-ci.yml
@@ -4,9 +4,11 @@ on:
   push:
     branches:
     - master
+    - main
   pull_request:
     branches:
     - master
+    - main
 
 env:
   DOCKER_BUILDKIT: 1


hzeller:
Thanks for bringing this to our attention; in general I like the direction.
Since @mithro was reviewing a lot of CI actions recently and also it authoring a few in other projects, I'll add him to the review.

I think the push and pull_request branches will probably be a no-op as we're developing and pull-requesting against master anyway. I might be in slight favor of _not_ actually specifying the branch (at this point), as we consider renaming `master` to `main`, and the less things to keep track of that needs to be changed in that process, the better.

For the cancel workflow, you added the `with access_token: ${{ github.token }}` constraint - is there anything we have to set-up for this ? It sounds like an auth token of sorts.

Finally, I think the largest impact could be to _not_ build all the releases in every pull request. This is stricly only needed on merges with master as we're building [releases from head](https://github.com/google/verible/releases).
(and there, the cancel workflow will actually help a lot independently, so that only the last of a string of merged pull requests will build a fresh release).
In a pull request, running the formatting verification and running all the tests is sufficient (until we find some platform specific deviations I guess, but that will be a concern for later times)

So my suggestions for this CL`^H^H`pull request would be
  * Don't specify `master` for now on the pull/merge request action running to make it easier to move to `main` later.
  * For a reduced impact on pull requests on using up Github Actions quota, skip the release buliding for pull requests

sethvargo:
Hey @hzeller 

Thank you for the reply. 

> For the cancel workflow, you added the with access_token: ${{ github.token }} constraint - is there anything we have to set-up for this ? It sounds like an auth token of sorts.

Nope, it's automatically-injected by the runner.

> Don't specify master for now on the pull/merge request action running to make it easier to move to main later.

Where do you do _development_? If you're pushing branches to the main repo, not specifying the branches to build will always cause a double build (once for the branch, once for the ff-merge on the PR). We can specify two branches in there, master and main, and that would make it easy to migrate in the future.

> For a reduced impact on pull requests on using up Github Actions quota, skip the release buliding for pull requests

Yea, unfortunately I'm not an expert on _this_ project, so I can't provide a good recommendation here on whether that's a good idea. If you do go that route, I'm happy to help you configure it!

hzeller:
> Where do you do _development_? If you're pushing branches to the main repo, not specifying the branches to build will always cause a double build (once for the branch, once for the ff-merge on the PR). We can specify two branches in there, master and main, and that would make it easy to migrate in the future.

So development is typically from someones' fork sending a pull request to verible master (E.g. if you look at #694 - that is a pull request to master, but the branch originates from my personal [hzeller:fixing-693](https://github.com/hzeller/verible/tree/fixing-693) branch. That means the github actions quota is actually used on the `hzeller`'s user quota, not this repositories' quota (unless of course it accounts that to the same organization as I am in, but given that I am in multiple organizations, I don't know how it could differentiate).
Alas, in this case, the actions actually don't even complete successfully.

Anyway, these are probably wasted github actions resources, no matter where it is accounted to. So given these circumstances, it is probably better to not waste resources on the originating PR branch (I suppose nobody will look there anyway ?), and limit it to `master` and `main`.

> 
> > For a reduced impact on pull requests on using up Github Actions quota, skip the release buliding for pull requests
> 
> Yea, unfortunately I'm not an expert on _this_ project, so I can't provide a good recommendation here on whether that's a good idea. If you do go that route, I'm happy to help you configure it!

Yes, I think we should _not_ run the `verible-ci / Release` in pull requests only on merges to master. They use up resources including the time of the engineers waiting for a pull request to go green. If you look at any recent open pull request with completed actions, the `verible-ci / Release` tasks accumulate about 3h of computing time every time...

So while at it, if you could configure this out, that would be awesome.

sethvargo:
> So while at it, if you could configure this out, that would be awesome.

Done

.github/workflows/verible-ci.yml
```diff
@@ -193,6 +193,7 @@ jobs:
     needs: [ Check, Build ]
     runs-on: ubuntu-20.04
     name: 📦 Release
+    if: ${{github.ref == 'refs/heads/main' || github.ref == 'refs/head/master'}}
```
hzeller, line 196: (still waiting for the CI to finish, but question:) Wouldn't this also need to be qualified that this only happens on `push` ?

sethvargo: Don't wait up - it could easily be many hours (that's what prompted this endeavour 😄).

I'll add it for clarity, but`push` is implied. From [here](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context):

> `github.ref` - The branch or tag ref that triggered the workflow run. For branches this in the format refs/heads/<branch_name>, and for tags it is refs/tags/<tag_name>.

The only way for the ref to be main/master would be for main/master to _trigger_ the workflow. I guess _technically_ you could open a PR from main->main, but that seems like an odd thing to do 😄

sethvargo: Done, although I did find, scrolling down a bit, a similar guard already existed on the _step_. I've removed that since it's redundant now.

hzeller:
Thanks, merged!

hzeller:
Mmh, now that the CI is finished (after a little wait :)) after the merge to master https://github.com/google/verible/actions/runs/641059261

... it seems like it actually skipped the release, even though this is the place where we do want the release.
 https://github.com/google/verible/runs/2090072118

Is maybe the `github.event_name == 'push'` condition not doing what we think it should ?

@sethvargo do you have some idea what is going on ?

(Once that is figured out, we should also probably move the `if` condition to the `Build:` instead of the `Release:`, so that we don't waste time building artifacts that are not released. They would be 'nice-to-have' to find platform dependent problems, but given the current crunch for resources, I am happy go forgo these in the context of pull requests)

sethvargo:
Hmmmm. I don't see anything obvious on why it would skip. The best way to debug would be to drop the conditional for now and add an echo step like:

```yaml
name: vars
run: |-
  echo "event_name: ${{github.event_name}}, ref: ${{github.ref}}"
```

And check the values. The queue is still 20h+ though, so we might just want to remove the conditional for now if it's blocking and experiment later.

.github/workflows/verible-ci.yml
```diff
@@ -177,6 +193,7 @@ jobs:
     needs: [ Check, Build ]
     runs-on: ubuntu-20.04
     name: 📦 Release
+    if: ${{github.event_name == 'push' && (github.ref == 'refs/heads/main' || github.ref == 'refs/head/master')}}
```
hzeller, line 196: Could the `refs/head/master` instead of `refs/heads/master` be the culprit ? Looks like a typo...

sethvargo: Yes, that could certainly be it 😄 https://github.com/google/verible/pull/698

hzeller: Oh, also I created #699 in parallel (with the skipping already moved to the build step)

PR #697, opened 2021-03-11
fix parsing of 'with' inside streaming operator #693

### LINKED ISSUE #693: Syntax errors due to streaming `with` operator

hzeller:
The stream expression has an expression followed by
an optional 'with' '[' array_range_expression ']'
(LRM 11.4.14)

There are a few other cases where an expression can
be trailed with 'with',
 * with_constraint_block: there we expect an
   identifier list in parenthesis after 'with'
 * similar with array methods with predicate
   with '(' ...)

Since the parser with one look-ahead can't see beyond
the 'with', it runs into a shift/reduce conflict
as it does not know if '(' or '[' is to follow.

Disambiguate that in the lexer by creating a
separate token TK_with__followed_by_bracket which is
a TK_with, where the lexer already looked ahead and
knows that it will be followed by '['.

After seeing the '[', everything but the "with" is put
back into the stream to be processed separately.

Changed lexer and parser as well as associated unit tests including formatter.

Fixing #693 

### DIFF
--- a/verilog/parser/verilog.lex
+++ b/verilog/parser/verilog.lex
@@ -672,9 +672,8 @@ wait_order { UpdateLocation(); return TK_wait_order; }
 wildcard { UpdateLocation(); return TK_wildcard; }
 <COVERGROUP>with { UpdateLocation(); return TK_with__covergroup; }
 with { UpdateLocation(); return TK_with; }
-"with"[ \t]*"[" {
-  /* TODO: formulate the ws sequence as any whitespace+comment sequence*/
-  yyless(4);
+"with"({TraditionalCommentOrSpace}|{EndOfLineComment})*"[" {
+  yyless(4);    // only consume 'with', push rest back to input
   UpdateLocation();
   return TK_with__followed_by_bracket;
 }
--- a/verilog/parser/verilog.lex
+++ b/verilog/parser/verilog.lex
@@ -672,6 +672,12 @@ wait_order { UpdateLocation(); return TK_wait_order; }
 wildcard { UpdateLocation(); return TK_wildcard; }
 <COVERGROUP>with { UpdateLocation(); return TK_with__covergroup; }
 with { UpdateLocation(); return TK_with; }
+"with"[ \t]*"[" {
+  /* TODO: formulate the ws sequence as any whitespace+comment sequence*/
+  yyless(4);
+  UpdateLocation();
+  return TK_with__followed_by_bracket;
+}
 within { UpdateLocation(); return TK_within; }
 timeprecision_check { UpdateLocation(); return TK_timeprecision_check; }
 timeunit_check { UpdateLocation(); return TK_timeunit_check; }
--- a/verilog/parser/verilog_parser_unittest.cc
+++ b/verilog/parser/verilog_parser_unittest.cc
@@ -627,7 +627,8 @@ static const ParserTestCaseArray kFunctionTests = {
     "{>>8 {foo, bar with \t [ a +: b]}} = nums;\n"
     "endfunction",
     "function void unpack_id_with(int nums);\n"
-    "{>>8 {foo, bar with /* some comment */ [ a +: b]}} = nums;\n"
+    "{>>8 {foo, bar with /* some comment */ /* another one */ // more\n"
+    "[ a +: b]}} = nums;\n"
     "endfunction",
     "function void unpack_id_with(int nums);\n"
     "{>>8 {foo, bar with // eol comment\n"
--- a/verilog/parser/verilog_parser_unittest.cc
+++ b/verilog/parser/verilog_parser_unittest.cc
@@ -624,7 +624,14 @@ static const ParserTestCaseArray kFunctionTests = {
     "{<< byte {this.reserved, this.id}} = bytes;\n"
     "endfunction",
     "function void unpack_id_with(int nums);\n"
-    "{>>8 {foo, bar with [ a +: b]}} = nums;\n"
+    "{>>8 {foo, bar with \t [ a +: b]}} = nums;\n"
+    "endfunction",
+    "function void unpack_id_with(int nums);\n"
+    "{>>8 {foo, bar with /* some comment */ [ a +: b]}} = nums;\n"
+    "endfunction",
+    "function void unpack_id_with(int nums);\n"
+    "{>>8 {foo, bar with // eol comment\n"
+    "      [ a +: b]}} = nums;\n"
     "endfunction",
     // concatenation lvalue
     "module foo;\n"
--- a/verilog/parser/verilog_parser_unittest.cc
+++ b/verilog/parser/verilog_parser_unittest.cc
@@ -623,6 +623,9 @@ static const ParserTestCaseArray kFunctionTests = {
     "function void unpack_id(utils_pkg::bytestream_t bytes);\n"
     "{<< byte {this.reserved, this.id}} = bytes;\n"
     "endfunction",
+    "function void unpack_id_with(int nums);\n"
+    "{>>8 {foo, bar with [ a +: b]}} = nums;\n"
+    "endfunction",
     // concatenation lvalue
     "module foo;\n"
     "  initial begin\n"
--- a/verilog/formatting/formatter_test.cc
+++ b/verilog/formatting/formatter_test.cc
@@ -775,6 +775,10 @@ static constexpr FormatterTestCase kFormatterTestCases[] = {
         "   parameter  int  b={ << 4  {{ << 2 { a }  }} } ;",
         "parameter int b = {<<4{{<<2{a}}}};\n",
     },
+    {
+      "   parameter  int  b={ << {foo with [ 1 +: 4 ] } } ;",
+      "parameter int b = {<<{foo with [1+:4]}};\n",
+    },
 
     // basic module test cases
     {"module foo;endmodule:foo\n",
--- a/verilog/formatting/token_annotator.cc
+++ b/verilog/formatting/token_annotator.cc
@@ -208,6 +208,9 @@ static WithReason<int> SpacesRequiredBetween(
   if (right_context.IsInsideFirst({NodeEnum::kStreamingConcatenation}, {})) {
     if (left.TokenEnum() == TK_LS || left.TokenEnum() == TK_RS) {
       return {0, "No space around streaming operators"};
+    } else if (left.TokenEnum() == TK_with__followed_by_bracket ||
+               right.TokenEnum() == TK_with__followed_by_bracket) {
+      return {1, "Space around 'with' keyword in streaming concatenation."};
     } else if (left.format_token_enum == FormatTokenType::numeric_literal ||
                left.format_token_enum == FormatTokenType::identifier ||
                left.format_token_enum == FormatTokenType::keyword) {
--- a/verilog/formatting/verilog_token.cc
+++ b/verilog/formatting/verilog_token.cc
@@ -298,6 +298,7 @@ static const auto* FormatTokenTypeMap =
         {verilog_tokentype::TK_wait_order, FTT::keyword},
         {verilog_tokentype::TK_wildcard, FTT::keyword},
         {verilog_tokentype::TK_with, FTT::keyword},
+        {verilog_tokentype::TK_with__followed_by_bracket, FTT::keyword},
         {verilog_tokentype::TK_with__covergroup, FTT::keyword},
         {verilog_tokentype::TK_within, FTT::keyword},
         {verilog_tokentype::TK_timeprecision_check, FTT::keyword},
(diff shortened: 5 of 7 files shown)


hzeller:
Implemented the suggestion to `yyless` back to only consume `with` for the rest to be consumed later.

Next step is to see if it is easy to implement in flex to have everything between `with` and `[` just be some arbitrary whitespace including comments (Currently simplistically implemented as `[ \t]*`).

Also the formatter needs to be able to deal with this token sequence it seems ... right now it removes all whitespace.

hzeller:
Upcoming: Fix formatting.

Right now, for

```verilog
module test;
  initial begin
    logic unpacked_arr [16] = '{0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1};
    logic [3:0] nibble_arr[4];
    for (int i = 0; i < 4; i++) begin
      nibble_arr[i] = {<< {unpacked_arr with [i*4 +: 4]}};
    end
  end
endmodule
```

... the token partition tree looks like:
```
Full token partition tree:
  { ([<auto>], policy: always-expand) @{0}, (origin: "module test...d
endmodule")
    { ([module test ;], policy: fit-else-expand, (origin: "module test;")) }
    { (>>[<auto>], policy: fit-else-expand) @{0,1}, (origin: "initial beg...  end
  end")
      { (>>[initial begin], policy: fit-else-expand) }
      { (>>>>[<auto>], policy: tabular-alignment) @{0,1,1}, (origin: "logic unpac...}};
    end")
        { (>>>>[logic unpacked_arr [ 16 ] = '{ 0 , 1 , 0 , 1 , 0 , 1 , 0 , 1 , 0 , 1 , 0 , 1 , 0 , 1 , 0 , 1 } ;], policy: fit-else-expand, (origin: "logic unpac..., 1, 0, 1};")) }
        { (>>>>[logic [ 3 : 0 ] nibble_arr [ 4 ] ;], policy: fit-else-expand, (origin: "logic [3:0]...ble_arr[4];")) }
        { (>>>>[<auto>], policy: fit-else-expand) @{0,1,1,2}, (origin: "for (int i ...}};
    end")
          { (>>>>[<auto>], policy: fit-else-expand) @{0,1,1,2,0}, (origin: "for (int i ...i < 4; i++)")
            { (>>>>[for (], policy: fit-else-expand) }
            { (>>>>>>>>[<auto>], policy: fit-else-expand) @{0,1,1,2,0,1}, (origin: "int i = 0; i < 4; i++")
              { (>>>>>>>>[int i = 0 ;], policy: fit-else-expand, (origin: "int i = 0")) }
              { (>>>>>>>>[i < 4 ;], policy: fit-else-expand, (origin: "i < 4")) }
              { (>>>>>>>>[i ++], policy: fit-else-expand, (origin: "i++")) }
            }
            { (>>>>[) begin], policy: uninitialized) }
          }
          { (>>>>>>[nibble_arr [ i ] = { << { unpacked_arr with [ i * 4 +: 4 ] } } ;], policy: fit-else-expand, (origin: "nibble_arr[...*4 +: 4]}};")) }
          { (>>>>[end], policy: always-expand, (origin: "end")) }
        }
      }
      { (>>[end], policy: always-expand, (origin: "end")) }
    }
    { ([endmodule], policy: always-expand, (origin: "endmodule")) }
  }
}
```

With that, the formatting is not adding any spaces in the relevant line; resulting in the following faulty formatting:

```verilog
module test;
  initial begin
    logic unpacked_arr[16] = '{0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1};
    logic [3:0] nibble_arr[4];
    for (int i = 0; i < 4; i++) begin
      nibble_arr[i] = {<<{unpacked_arrwith[i*4+:4]}};
    end
  end
endmodule
```

verilog/parser/verilog.lex
```diff
@@ -672,9 +672,8 @@ wait_order { UpdateLocation(); return TK_wait_order; }
 wildcard { UpdateLocation(); return TK_wildcard; }
 <COVERGROUP>with { UpdateLocation(); return TK_with__covergroup; }
 with { UpdateLocation(); return TK_with; }
-"with"[ \t]*"[" {
-  /* TODO: formulate the ws sequence as any whitespace+comment sequence*/
-  yyless(4);
+"with"({TraditionalCommentOrSpace}|{EndOfLineComment})*"[" {
```
fangism, line 675: This might be easier with an `<AFTER_WITH>` state, where `with` returns a `TK_width` after pushing a new state.
It will be easier to re-use rules inside states than glue them together like this in monolithic regular expressions.
Inside this new state, you handle whitespace, comments, and `[`, and for everything else, you can pop the state and rewind, to defer to default `<INITIAL>` state handling (or whatever was previously on the state stack).

fangism: You will also want lexer unit-tests for this.  :)

verilog/parser/verilog_parser_unittest.cc
```diff
@@ -624,7 +624,14 @@ static const ParserTestCaseArray kFunctionTests = {
     "{<< byte {this.reserved, this.id}} = bytes;\n"
     "endfunction",
     "function void unpack_id_with(int nums);\n"
-    "{>>8 {foo, bar with [ a +: b]}} = nums;\n"
+    "{>>8 {foo, bar with \t [ a +: b]}} = nums;\n"
+    "endfunction",
+    "function void unpack_id_with(int nums);\n"
+    "{>>8 {foo, bar with /* some comment */ [ a +: b]}} = nums;\n"
```
fangism, line 630: More test case ideas:
* multiple comments in between `with` and `[`.

fangism:
> Upcoming: Fix formatting.
> 
> Right now, for
> 
> ```verilog
> module test;
>   initial begin
>     logic unpacked_arr [16] = '{0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1};
>     logic [3:0] nibble_arr[4];
>     for (int i = 0; i < 4; i++) begin
>       nibble_arr[i] = {<< {unpacked_arr with [i*4 +: 4]}};
>     end
>   end
> endmodule
> ```
> 
> ... the token partition tree looks like:
> 
> ```
> Full token partition tree:
> { ([<auto>], policy: always-expand) @{}
>   { ([<auto>], policy: always-expand) @{0}, (origin: "module test...d
> endmodule")
>     { ([module test ;], policy: fit-else-expand, (origin: "module test;")) }
>     { (>>[<auto>], policy: fit-else-expand) @{0,1}, (origin: "initial beg...  end
>   end")
>       { (>>[initial begin], policy: fit-else-expand) }
>       { (>>>>[<auto>], policy: tabular-alignment) @{0,1,1}, (origin: "logic unpac...}};
>     end")
>         { (>>>>[logic unpacked_arr [ 16 ] = '{ 0 , 1 , 0 , 1 , 0 , 1 , 0 , 1 , 0 , 1 , 0 , 1 , 0 , 1 , 0 , 1 } ;], policy: fit-else-expand, (origin: "logic unpac..., 1, 0, 1};")) }
>         { (>>>>[logic [ 3 : 0 ] nibble_arr [ 4 ] ;], policy: fit-else-expand, (origin: "logic [3:0]...ble_arr[4];")) }
>         { (>>>>[<auto>], policy: fit-else-expand) @{0,1,1,2}, (origin: "for (int i ...}};
>     end")
>           { (>>>>[<auto>], policy: fit-else-expand) @{0,1,1,2,0}, (origin: "for (int i ...i < 4; i++)")
>             { (>>>>[for (], policy: fit-else-expand) }
>             { (>>>>>>>>[<auto>], policy: fit-else-expand) @{0,1,1,2,0,1}, (origin: "int i = 0; i < 4; i++")
>               { (>>>>>>>>[int i = 0 ;], policy: fit-else-expand, (origin: "int i = 0")) }
>               { (>>>>>>>>[i < 4 ;], policy: fit-else-expand, (origin: "i < 4")) }
>               { (>>>>>>>>[i ++], policy: fit-else-expand, (origin: "i++")) }
>             }
>             { (>>>>[) begin], policy: uninitialized) }
>           }
>           { (>>>>>>[nibble_arr [ i ] = { << { unpacked_arr with [ i * 4 +: 4 ] } } ;], policy: fit-else-expand, (origin: "nibble_arr[...*4 +: 4]}};")) }
>           { (>>>>[end], policy: always-expand, (origin: "end")) }
>         }
>       }
>       { (>>[end], policy: always-expand, (origin: "end")) }
>     }
>     { ([endmodule], policy: always-expand, (origin: "endmodule")) }
>   }
> }
> ```
> 
> With that, the formatting is not adding any spaces in the relevant line; resulting in the following faulty formatting:
> 
> ```verilog
> module test;
>   initial begin
>     logic unpacked_arr[16] = '{0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1};
>     logic [3:0] nibble_arr[4];
>     for (int i = 0; i < 4; i++) begin
>       nibble_arr[i] = {<<{unpacked_arrwith[i*4+:4]}};
>     end
>   end
> endmodule
> ```

When the tokens mistakenly fused, did the formatter catch it as data corruption and bail out?  (If not, that itself is a bug.)

This is controlled by inter-token spacing (token_annotator).  There is a general rule that keywords and identifiers required a minimum 1 space separation.  You will likely need a classification entry to make the special `[` look and feel like a regular `[`.

verilog/parser/verilog.lex
hzeller: Ok, will need to familiarize myself how states work. They did look somewhat complicated.

The beauty if this is, that I don't have to handle anything special, including '['; I just look ahead, find that this is the next token and TK_with needs to be a new token.

verilog/parser/verilog_parser_unittest.cc
hzeller: Done.

verilog/parser/verilog.lex
hzeller: (lexer unit test added)

hzeller: I looked at the AFTER_DOT example, which is somewhat involved. I have to play with it to see if it is more advantageous.

hzeller:
Yes, the formatter was bailing out correctly, finding that it created a syntax error after reparsing.

The issue in this particular case in the annotator was, that there was a special case in the annotator that removed spaces around streaming operators components. But we do have to keep them around the 'with' keyword. So that needed to be special-special cased.

Anyway, I think this is ready to review. It still uses the lexer without extra state machine though, will have to investigate it that would feel better.
The bug however would be fixed with the current state already.

hzeller:
I have created https://github.com/google/verible/issues/749 to follow-up on the implementation improvement suggestion by @fangism . I'd like to submit this rather now to be able to deal with the immediate parse situations, but then look for the nicer implementation later when dust in other corners of the project settled.

hzeller:
@tgorochowik I handed the review baton on to you as David focuses on other things currently.

tgorochowik (approved):
To be honest I agree with David - this solution looks pretty hacky and I think it would be best to avoid doing it like this.

However it does solve the issue, so I guess it's okay to merge it and revisit at a later stage. So let me approve this in hope it won't start the "broken window theory" thing.

fangism:
Sorry, it's been a while since looking at this.  I should have prefaced my initial comment with: if it can be encoded using standard techniques for expressing precedence and associativity in LR grammars, do that first.  Because `with` appears in different contexts, you might have to create an auxiliary rule like `expression_followed_by_with` as one way to avoid/delay the conflict, but this is not always easy.  Based on early conversations, I believe Henner attempted some of these approaches?

verilog/parser/verilog.lex
fangism: Looking over this again, I still stand by this first comment.  Generally, the options are:

1) track state within the lexer to alter the enumeration of the _latter_ token
2) track state inside the lexical context class to alter the enumeration of the _earlier_ token (effectively achieving lookahead) -- this approach is pretty localized, you're defining a tiny state-machine that can run along side other independent state machines.

hzeller:
Ok, will not submit now and revisit when there is a bit more time.

PR #699, opened 2021-03-11
Fix typo in conditional action and exclude Build in pull requests.

hzeller:
* fix typo refs/heads/master vs. refs/heads/master
 * Don't build all matrix artifacts in pull requests. Actions is
   currently limited in parallel runners, so don't do unnecesary
   work, even if it was nice-to-have.


hzeller:
cc: @sethvargo

hzeller:
I guess question will be if actiions will work nicely and skipping Release, because one of its preconditions has been skipped ? Or would the condition also be repeated in Release ?

Issue #689, opened 2021-02-26
TokenStreamLintRule crashes on ternary expression.


hzeller:
**Describe the bug**

A simple file with a ternary expression assigning a string

```
$ cat > /tmp/a.sv << EOF
module a();
    string x = 1 ? "foo" : "bar";
endmodule
EOF
```
... will crash the linter in the `TokenStreamLintRule`

```
$ bazel run verilog/tools/lint:verible-verilog-lint -- /tmp/a.sv
F0225 16:39:42.290365 506207 tree_utils.cc:125] Check failed: symbol.Kind() == SymbolKind::kLeaf (SymbolKind::kNode vs. SymbolKind::kLeaf) got: Node @0 (tag: 92) {
  Leaf @0 (#300: "1")
}
*** Check failure stack trace: ***
    @     0x55d3f8650b20  google::LogMessage::Fail()
    @     0x55d3f8650a6b  google::LogMessage::SendToLog()
    @     0x55d3f86503bc  google::LogMessage::Flush()
    @     0x55d3f865331c  google::LogMessageFatal::~LogMessageFatal()
    @     0x55d3f86443fd  verible::SymbolCastToLeaf()
    @     0x55d3f85ff49e  verilog::analysis::TokenStreamLintRule::HandleSymbol()
    @     0x55d3f854ebb5  verible::SyntaxTreeLinter::Visit()
 ...
```

Disabled the rule in #688 for now, so it will not crash at head. Pawel, since you implemented it, can you have a look, at a test-case and fix ?

PR #701, opened 2021-03-12
Fixing TokenStreamLintRule crashes on ternary expression

pawelsag:
Fixing #689


verilog/analysis/checkers/token_stream_lint_rule.cc
```diff
@@ -69,7 +70,16 @@ void TokenStreamLintRule::HandleSymbol(const verible::Symbol& symbol,
   verible::matcher::BoundSymbolManager manager;
   if (StringLiteralMatcher().Matches(symbol, &manager)) {
     const auto& string_node = SymbolCastToNode(symbol);
-    const auto& string_literal = SymbolCastToLeaf(*string_node.children()[0]);
+    const auto& node_children = string_node.children();
+    const auto& literal =
```
hzeller, line 74: So could it be that we actually can have more than one string literal in this node ?
I suspect in our ternary situation we might have some TK_StringLiteral ':' TK_StringLiteral ?
Here, we're only looking for the first occurence.

So if there are multiple, should we check all of the children's string-literals ?
(might actually simplify the code as we can simply loop through the node-children and `continue` on non-`TK_StringLiteral`-ness).

verilog/analysis/checkers/token_stream_lint_rule.cc
```diff
@@ -69,7 +70,16 @@ void TokenStreamLintRule::HandleSymbol(const verible::Symbol& symbol,
   verible::matcher::BoundSymbolManager manager;
   if (StringLiteralMatcher().Matches(symbol, &manager)) {
```
hzeller, line 71: To keep the indentation level flat, we can formulate this first test with an early return
```
  if (!StringLiteralMatcher().Matches(symbol, &manager))
    return;
```

verilog/analysis/checkers/token_stream_lint_rule_test.cc
```diff
@@ -41,6 +41,7 @@ TEST(StringLiteralConcatenationTest, FunctionFailures) {
        "\"Humpty Dumpty had a great fall.\"};", "\nendmodule"},
       {"module m;\n", "string tmp = {\"Humpty Dumpty \\ sat on a wall.\",",
        "\"Humpty Dumpty had a great fall.\"};", "\nendmodule"},
+      {"module a();\nstring x = 1 ? \"foo\" : \"bar\";\nendmodule"},
```
hzeller, line 44: Are there otherr situations where the previous code would've stumbled upon ? If you are aware of some, we should probably add them.

Also since we now know that `?:` is actually showing up as more complicated node-structure, we should test that we can detect line-continuation attempts in each branch (line continuation attempted in`"foo"` or in `"bar"` ?)

hzeller (commented):
Some suggestions for additional tests and possibly better coverage of corner-cases.

verilog/analysis/checkers/token_stream_lint_rule.cc
pawelsag: In our ternary example when I printed the tree I found such structure:
```
Node @0 (tag: kConditionExpression) {
   Node @0 (tag: kNumber) {
      Leaf @0 (#TK_DecNumber @27-28: "1")
   }
   Leaf @1 (#'?' @29-30: "?")
   Node @2 (tag: kExpression) {
         Leaf @0 (#TK_StringLiteral @31-36: ""foo"")
   }
   Leaf @3 (#':' @37-38: ":")
   Leaf @4 (#TK_StringLiteral @39-44: ""bar"")
}
```
And as we can see we have two string literals. So the symbols passed to checker are: `kConditionExpression` and `kExpression`.  So for the first node, I'm looking for  `TK_StringLiteral` within its children, and the same is done for the second obtained symbol.
In the previous solution, I assumed that the string literal will be always in the first position in the children's leaf-list.

I believe that the later `if` validation can be removed cause the checker symbol is guaranteed to points to a node with TK_StringLiteral and there is no chance that there wouldn't be the desired string token.

So shouldn't I look for the string literal for each given symbol within the symbol's children? (but only leaves are considered, I do not dive into the nodes)
Do I understand it correctly?

verilog/analysis/checkers/token_stream_lint_rule_test.cc
pawelsag: I have split the test structure, and few more test cases were added, but just for now I have no more ideas

verilog/analysis/checkers/token_stream_lint_rule_test.cc
```diff
@@ -55,6 +67,15 @@ TEST(StringLiteralConcatenationTest, FunctionFailures) {
         "fall.\""},
        ";",
        "\nendmodule"},
+      {"module m;\nstring x = 1 ?",
+       {kToken, "\"foo\\\nincorrect\""},
+       " : \"bar\";\nendmodule"},
```
hzeller, line 72: Another test where the second branch of the ?:  `"\"bar\\\nis incorrect\""` maybe ?

hzeller (changes requested):
Nice. One more test-case suggestion.

verilog/analysis/checkers/token_stream_lint_rule_test.cc
pawelsag: Thanks for this test, I found one more bug when I applied the suggestion. The bug was related to incorrectly pointing token position.

hzeller (approved):
LGTM, Thanks!

PR #703, opened 2021-03-18
Don't violate the 'no news are good news' principle.

hzeller:
The formatter was quite noisy when it didn't have to change
anything. Make that only happen when --verbose is chosen.


hzeller:
All (Unix or in general command line) tools are commonly designed around this principle so it is a strong expectation that nothing is reported unless there is an actual oddity that occurred.

fangism (approved):

PR #708, opened 2021-03-23
Fix a couple of smallish clang format findings.

hzeller:
For some reason the presubmit did not pick these up.


tgorochowik (approved):
LGTM

PR #711, opened 2021-03-24
Testing clang format presubmit.

hzeller:
DO NOT SUBMIT

### DIFF
--- a/.github/workflows/run-clang-format.sh
+++ b/.github/workflows/run-clang-format.sh
@@ -12,7 +12,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-set -x
 FILE_LIST_OUT=${TMPDIR:-/tmp}/file-list.$$.txt
 FORMAT_OUT=${TMPDIR:-/tmp}/clang-format-diff.$$.out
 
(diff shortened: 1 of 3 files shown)


.github/workflows/run-clang-format.sh
```diff
@@ -12,19 +12,30 @@
...
+}
+trap cleanup EXIT
 
 # Run on all the files that are affected
-clang-format -i --style=file $(git diff --name-only --diff-filter=AM -r origin/master | grep '\(\.cc\|\.h\)$') 2> /dev/null
 
-# Check if we got any diff
-git diff > ${FORMAT_OUT}
+# Finding all affected files and list them for inspection in the log output
+git diff --name-only --diff-filter=AM -r origin/master | grep '\(\.cc\|\.h\)$' | tee "${FILE_LIST_OUT}"
```
mithro, line 26: Take a look at `GITHUB_BASE_REF` and `GITHUB_HEAD_REF` -- see the full amount at https://docs.github.com/en/actions/reference/environment-variables

mithro (commented):
BTW The way this was set up to work is that you format all the files and then check if git sees any changes to the files.

hzeller:
Yes. currently I try to figure out why it does not work with `--style=file` which used to work before.
So all the changes you see currently are mostly decomposing the problem.

PR #712, opened 2021-03-24
clang-format does not work with --style=file currently.

hzeller:
Use --style=Google

More investigation needed.


.github/workflows/run-clang-format.sh
```diff
@@ -16,7 +16,7 @@
 FORMAT_OUT=${TMPDIR:-/tmp}/clang-format-diff.out
 
 # Run on all the files that are affected
```
mithro, line 18: Maybe add a `clang-format --version` here?

mithro: Looks like this is done in an earlier step....

mithro: `clang-format version 10.0.0-4ubuntu1`

.github/workflows/run-clang-format.sh
```diff
@@ -16,7 +16,7 @@
...
-clang-format -i --style=file $(git diff --name-only --diff-filter=AM -r origin/master | grep '\(\.cc\|\.h\)$') 2> /dev/null
```
mithro, line 19: It's probably a good idea to make sure clang-format is being run in the root directory.

.github/workflows/run-clang-format.sh
mithro:
```
  --dump-config              - Dump configuration options to stdout and exit.
                               Can be used with -style option.
  --fallback-style=<string>  - The name of the predefined style used as a
                               fallback in case clang-format is invoked with
                               -style=file, but can not find the .clang-format
                               file to use.
                               Use -fallback-style=none to skip formatting.
```
and
```
  --verbose                  - If set, shows the list of processed files
```
Could be useful too....

hzeller: `--fallback-style` I tried. It then does not fall back, which indicates that it found the file, but then didn't act properly.
`strace` was not installed, but that will be one of my next experiments in `testing-clang-format` to see if it picked up the file.

.github/workflows/run-clang-format.sh
hzeller: `clang-format` is run from the root directory (the script is invoked as `./.github/workflows/run-clang-format.sh`). But even if it was executed in a different directory, `git diff` would give out relative names (such as `../src/foo.cc`), so everything would still work (and the `.clang-format` file is looked up relative to the source file it is processing not the location the execution. So that als should work).

.github/workflows/run-clang-format.sh
```diff
@@ -16,7 +16,7 @@
...
+clang-format -i --style=Google $(git diff --name-only --diff-filter=AM -r origin/master | grep '\(\.cc\|\.h\)$') 2> /dev/null
```
mithro, line 19: `git diff --name-only --diff-filter=AM -r origin/master` <-- Using `origin/master` there is most certainly wrong....

hzeller: the `GITHUB_BASE_REF` environment variable is also `master` (which is our main branch from which we branch off in a pull request). So it would be exactly the same (but we have to prepend origin/).

So we could do something like `BASE_BRANCH=origin/${GITHUB_BASE_REF:-master}`, but the invocation of `git diff` is exactly the same.

hzeller: I am now explicitly determining the point from where it forked off:
```
FORK_POINT=$(git merge-base --fork-point ${BASE_BRANCH})
```

hzeller:
Good to merge ? I'd like to get this in soon as some recent pull requests were not properly format checked.

hzeller:
CI will pass after #719 is merged

.github/workflows/run-clang-format.sh
```diff
@@ -15,8 +15,9 @@
 
 FORMAT_OUT=${TMPDIR:-/tmp}/clang-format-diff.out
```
mithro, line 17: Potential future improvement -- I wonder if it is worth checking the clang-format version is correct?

mithro (approved):
LGTM!

mithro:
Did we ever figure out *why* `clang-format` doesn't seem to be using the style file?

hzeller:
No, style-file mystery not solved yet. I have in plan to strace it on actions to see if it is actually opening the file. But first things first.

clang-format version testing I was considering, but I'd actually rather have whatever is latest in CI, then if we happen to get a newer one that is more 'strict' fail with these.

hzeller:
Oh dear, rebasing to master looks like it was not a good idea.

PR #714, opened 2021-03-24
Testing towards a build on MacOS

hzeller:
DO NOT SUBMIT
Testing towards a build on MacOS
(we expect M4 to fail compile, but let's see)


mithro:
BTW @per-gron and @QuantamHD both have a lot of experience with bazel rules (through https://github.com/hdl/bazel_rules_hdl)  so might be able to provide some advice here....

hzeller:
Currently trying to work around some MacOS issues...
Since compiling flex and m4 locally was not working so well so far, currently on the path to see what happens if we use local versions of the same (but of course MacOS has _ancient_ versions of these installed, so the brew install puts it in some other directory ... causing all kinds of additional problems).
Once that is exhausted, will try to go back to see if I can make it get these things compiled hermetically.

So at this point, it is less of an issue with bazel rules, just trying to compile something on a Unix that hasn't been updated since 2001...

PR #721, opened 2021-03-30
clang-format: use style Google; always format all files.

hzeller:
* -style=file does not seem to work in CI; more investigation
   needed. For now: use Google style explicitly.
 * Always format all files in repo to avoid corner-cases
   with git branches.


hzeller:
Git messed-up #712 ... so this is the same in fresh :)

mithro (approved):
LGTM.

BTW You could have just force pushed to the old branch to update that....

PR #722, opened 2021-03-30
Run CI with -Werror

hzeller:
* Enable all warnings .. and disable the subset that are
   already in the code-base.
 * Disable one specific warning that popped up directly
   in the BUILD file.


hzeller:
FYI @mithro
I pressed 'merge' before requesting Review.

.github/workflows/script.sh
```diff
@@ -18,6 +18,39 @@ set -e
 
 source ./.github/settings.sh
 
+# TODO(b/171679296): re-enable c++11 support
```
mithro, line 21: Why wouldn't we want these in `.github/settings.sh`?

PR #723, opened 2021-03-30
Move Bazel opts back into settings.sh

hzeller:
moving the settings for bazel builds back to settings.sh


mithro (approved):
I'm pretty sure those `--cxxopt` can be set in the bazel rules in some way, but for now LGTM....

hzeller:
You can set them individually per target (which is what I am doing for things under our control, mostly), but there are other sources that we indirectly import from the `WORKSPACE` file that introduce a lot of warnings that we ideally would not have to deal with. I have to figure out how we can do that.

Issue #307, opened 2020-05-25
Support for Windows


corco:
I have been looking to add support for Windows but hit a few brick walls. I'm looking for advice on the best way to do that for the project, assuming supporting Windows is desired of course!

I looked at the current setup using `cc_configure_make`. `cc_configure_make` does not support Windows and besides, I don't believe m4/flex/bison will compile anyways.

I also looked [rules_m4](https://github.com/jmillikin/rules_m4)/[rules_flex](https://github.com/jmillikin/rules_flex)/[rules_bison](https://github.com/jmillikin/rules_bison). m4/bison compiles (bison has a small issue at running) but flex does not compile on Windows and it doesn't look like it will anytime soon. However, the three repos works on OSX, so if you're interested I can do a pull request to replace `cc_configure_make` with these repos.

Finally, to support Windows I see three ways forward:

1. Commit flex/bison generated files in the repository, removing the requirements on flex/bison. We could keep the bazel rules to generate the files and a test, for CI, to make sure the committed files match the sources.
2. Compile for Windows on a LInux machine with MinGW. m4/flex/bison would be compiled with the host compiler, and verible on MinGW. It would not be possible to build on a Windows machine though.
3. Look to use [winflexbison](https://github.com/lexxmark/winflexbison) on Windows.

fangism:
Support for building on Windows is welcome, but I have zero experience developing on Windows myself.  (We've been having trouble getting m4 to build on Mac OS X.)

1. is not a bad idea, and is compatible with automake-style `dist_` files.  Perhaps this could be automated on branch that generates source distributions.  @mithro Does this seem practical?

fangism:
FYI, In #370 , @muhdmud [1] was able to follow instructions at https://docs.microsoft.com/en-us/windows/wsl/install-win10 before bazel build instructions to successfully build Verible.

[1] https://github.com/google/verible/issues/370#issuecomment-670547189

corco:
WSL allows to run Linux application on Windows, but it is not the same as a native Windows application. For starters, you need to have WSL installed on your Windows machine and I don't think you can invoke WSL application from Windows command line and vice-versa. I don't use Windows so I may be wrong there.

I managed to build a native Windows verible by using generated files from flex/bison instead of building them. I wanted to keep the file generation as a bazel test that would run only on Linux.

[...]

PR #728, opened 2021-04-01
Use github.com/jmillikin/rules_{flex,bison} bazel rules.

hzeller:
Instead of using the configure build, use the external
flex bazel rule that is reported to be more portable
to other platforms (at least MacOS #260, probably not
Windows #307 yet).


tgorochowik (commented):
@hzeller kythe build seems to be broken after these changes

hzeller:
Yes, this is still a draft. Reason is known (not having access to FlexLexer.h), but I need to figure out if this can be done without changing the bazel rules or if I need to file a feature request.

verilog/tools/syntax/verilog_syntax_test.sh
```diff
@@ -171,7 +171,7 @@ All lexed tokens:
 (#"endmodule" @11-20: "endmodule")
 (#"<<\\\\n>>" @20-21: "
 ")
-(#"end of file" @21-21: "")
+(#\$end @21-21: "")
```
hzeller, line 174: This looks like an artifact from scaling back from bison 3.6.2 to 3.3.2.
I looked if it was simple to actually upgrade the rules_bazel to 3.6.2 and it seems there is a new dependency ('textstyle') starting from 3.4 which makes things harder to port.

Apart from obvious issues integrating it in environments that are using a newer version (possibly the `end of file` string can be changed for the test), are you aware of any issues performance or feature-wise which would still warrant investigating going to the 3.6.x release @fangism ?

hzeller: Fixed in https://github.com/google/verible/pull/728/commits/70db525c5c50120f121b464f52c893cb456afa0a

mithro:
BTW You want to use the upstream source of https://github.com/hdl/bazel_rules_hdl

mithro:
Sorry, I think I got confused between github.com/jmillikin/rules_{flex,bison} and https://github.com/jesec/bazel_rules_hdl

bazel/bison.bzl
```diff
@@ -1,3 +1,4 @@
+# -*- Python -*-
```
fangism, line 1: Does this have to be "Python" (other files too)?

bazel/flex.bzl
```diff
@@ -1,3 +1,4 @@
+# -*- Python -*-
```
fangism, line 1: Python?

fangism (commented):
Looks ok so far, untested.

bazel/bison.bzl
hzeller: That makes a common editor that doesn't know about bazel files that this is essentially a Python file (which it is).

bazel/flex.bzl
hzeller: Yep, intentional.

bazel/bison.bzl
tgorochowik: Makes me think.. should we try to add a CI job that checks formatting of bazel files too? We could probably re-use the formatter scripts from sv-test (which use yapf)

(Not as a part of this PR of course)

tgorochowik (approved):
LGTM, works fine on Linux and the code is much cleaner now.

One question though: the repo with the rules does not seem to be actively maintained, I guess we don't always need the latest versions so that's fine, but maybe we should keep a fork of that repo somewhere just in case in disappears?

hzeller:
For bazel rules formatting, there is actually the [buildifier](https://github.com/bazelbuild/buildtools) tool. I have it somewhere on my TODO list, but should probably file an issue for it that it is not getting lost.

Maintainance of the [flex](https://github.com/jmillikin/rules_flex) repo: last change December 2020, so that looks good and actively maintained.
The [bison](https://github.com/jmillikin/rules_bison) was last updated in October 2019, so that looks a bit outdated, but there seems to be a good reason: bison added a bunch of very specific dependencies that would make it harder to port (I tried to change to a newer version of bison and it requires some work).

I will try to see if I can go to the latest version of bison and upstream that, and watch the responsiveness of the author. Right now I am not worried too much about it going away.

PR #733, opened 2021-04-01
[kythe] Extend VerilogProject to support adding virtual files

ivan444:
Virtual file: file for which we have the path & content, but not the physical file. This is required for running the indexing server


hzeller:
Do you know the reason for the failed build ?

ivan444:
Looking. It was working fine locally.

ivan444:
The build should be fixed now. I limited my build & test to kythe part, instead of the whole project. Forgot that `VeribleProject` is used elsewhere.

hzeller (approved):

PR #735, opened 2021-04-01
Bazel test mac

hzeller:
testing if the new bazel rules will compile on mac.


hzeller:
Down to one failure, captured in #739

Issue #310, opened 2020-05-28
Support for naming conventions


zegervdv:
Are there any plans to add linting rules for naming conventions?

For example module inputs should start with `i_` or should end with `_i`.
Preferably this would be configurable using some kind of template.

fangism:
I'm hearing similar requests coming from some teams.  I acknowledge not everyone will share the same convention, but fortunately, we've introduced rule configurability so it should be possible to pass regular expressions to the rules that check conventions.

@msfschaffer I believe you wanted this too, right?

To point to one example:
https://github.com/google/verible/blob/3968647730731ee8533b75097245efc23a9271a8/verilog/analysis/checkers/struct_union_name_style_rule.cc#L77

The currently hardcoded style checks could come from rule configurations.

Here's an example of a configurable rule:
https://github.com/google/verible/blob/3968647730731ee8533b75097245efc23a9271a8/verilog/analysis/checkers/explicit_parameter_storage_type_rule.cc#L104

pawelsag:
Hi, so if I understand it correctly we would like to have the whole name to be configured with regex or just only some parts of the gathered name should be verified with regex?

What regex should I use, the one from the cpp standard?

zegervdv:
I think the whole name would allow for most flexibility. In the example I gave we want to check that input ports of the module always start with `i_`, but other organizations use the rule that inputs should end on `_i`. 
So, if possible, a full name regex match would allow for both, and many other rules you could think of. 

The configuration for that could then be something like: `\bi_.*` to match the start of a word. 

I'm using PCRE style regexes here but, for me at least, as long as it supports the basics and is well documented, the exact syntax is not that important. 

I guess that is up to @fangism (or other maintainers) to decide.

pawelsag:
Ok, thanks for the clarification.

PR #737, opened 2021-04-02
WIP: Naming conventions with regex

pawelsag:
Fixes: #310

But first, the https://github.com/google/verible/pull/651 need to be merged


hzeller:
mmh, I have to think about this. I don't like `std::regex` at all as it uses exceptions to indicate issues with parsing expressions, which is the wrong choice of error reporting, because it is not exceptional (besides the std:: implementation being very slow).

OTOH, we already use it a one place in the [`lint_waiver`](https://github.com/google/verible/blob/7c994b1af0ef6c83439d00186abdb0c36c2320ef/common/analysis/lint_waiver.cc#L334-L341), for which we then explicitly [enabled exceptions](https://github.com/google/verible/blob/7c994b1af0ef6c83439d00186abdb0c36c2320ef/common/analysis/BUILD#L99).

The reason why we used `std::regex` in lint_waiver was to be compatible in C++11, C++14 _and_ C++17, which was not very possible with the preferred choice, [re2](https://github.com/google/re2) (due to some compatibility issues with StringPiece == std::string_view or not). Given that we are now [somewhat tied to C++17](https://github.com/google/verible/blob/master/.github/settings.sh#L31), this might be worthwhile revisting.

For this change, this should not hold us back for now, and we should proceed for now assuming we use `std::regex`.

verilog/analysis/checkers/constraint_name_style_rule.h
```diff
@@ -15,6 +15,8 @@
 #ifndef VERIBLE_VERILOG_ANALYSIS_CHECKERS_CONSTRAINT_NAME_STYLE_RULE_H_
 #define VERIBLE_VERILOG_ANALYSIS_CHECKERS_CONSTRAINT_NAME_STYLE_RULE_H_
 
+#include <optional>
+#include <regex>
```
hzeller, line 19: Make sure to enable exceptions for this library (example in [lint_waiver]( https://github.com/google/verible/blob/7c994b1af0ef6c83439d00186abdb0c36c2320ef/common/analysis/BUILD#L99)).

We usually prefer to disable exceptions for everything _except_ for the files that absolutely need them.

verilog/analysis/checkers/enum_name_style_rule.cc
```diff
@@ -49,10 +50,19 @@ const char EnumNameStyleRule::kMessage[] =
...
 std::string EnumNameStyleRule::GetDescription(
     DescriptionType description_type) {
-  return absl::StrCat("Checks that ", Codify("enum", description_type),
-                      " names use lower_snake_case naming convention"
-                      " and end with '_t' or '_e'. See ",
-                      GetStyleGuideCitation(kTopic), ".");
+  static std::string basic_desc =
+      absl::StrCat("Checks that ", Codify("enum", description_type),
+                   " names use lower_snake_case naming convention"
+                   " and end with '_t' or '_e'. See ",
```
hzeller, line 56: .. `and also the additional optional regular expression.`

Otherwise the explanation looks a bit confusing to the user: tests for_t and _e ... and then there is a parameter with regular expressions - it is unclear from seeing that if that is replacing the _t, _e rule or augmenting it.

Similar in the other rules you added regexp: we should update the documentation to include the optional regexp capability.

verilog/analysis/checkers/enum_name_style_rule.cc
```diff
@@ -49,10 +50,19 @@ const char EnumNameStyleRule::kMessage[] =
...
+                   GetStyleGuideCitation(kTopic), ".");
+  if (description_type == DescriptionType::kHelpRulesFlag) {
+    return absl::StrCat(basic_desc, "Paramteres: name_regex:regex rule");
```
hzeller, line 59: Typo: `Paramteres`

hzeller (commented):
Some initial comments; mostly I think we have to make it clear to the user that the regex are in addition to the thing that is already tested.

In the meantime I'll explore using [RE2](https://github.com/google/re2) instead of std::regex as I like to avoid exceptions (long story). But nothing to worry about for this PR, this is fairly straightforward to change.

pawelsag:
Ok so if I understand you correctly when regex is enabled we should validate the regex, if the regex does not match, we should fall back to the already implemented rules, right?

hzeller:
I was mostly thinking: we always check the built-in rule, and then, in addition, also the regexp. IIRC, that is how you implemented it.

Mostly it was important to me that we describe what we do in the description.

We could of course think of an either/or approach (if regexp defined, use that, otherwise use built-in rule).

verilog/analysis/checkers/enum_name_style_rule.cc
pawelsag: Ok so I changed the description as suggested, but I change the phrase to: `or also the additional optional regular expression.`, cause we can use either regex or the original rule, but when regex is not enabled we do not validate the old rules.

PR #740, opened 2021-04-03
Upgrade to latest absl.

hzeller:
And use absl::StrContains() with a char, that wasn't available
before.


mithro (approved):
LGTM.

mithro:
We should probably set up some type of GitHub action to automatically send PRs to roll things like this forward....

PR #741, opened 2021-04-03
CI: make sure all libraries that use exceptions declare so.

hzeller:

### DIFF
--- a/.github/bin/script.sh
+++ b/.github/bin/script.sh
@@ -23,7 +23,7 @@ source ./.github/settings.sh
 # as possible to compile without exceptions.
 # ... and unfortunately we have to disable this for now as the external
 # kythe dependency does use exceptions.
-# TODO(hzeller): Uncomment after  #747 is flying (excuse the pun)
+# TODO(hzeller): Uncomment after https://github.com/google/verible/issues/747 is fixed.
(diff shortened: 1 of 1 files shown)


hzeller:
So essentially, this pull request is a no-op as it just ends up adding a comment. Once we have reduced the Kythe dependencies, we can uncomment that line. For now it is essentially just a documentation.

.github/bin/script.sh
```diff
@@ -18,6 +18,13 @@ set -e
 
 source ./.github/settings.sh
 
+# Make sure we don't have cc_library rules that use exceptions but do not
```
mithro, line 21: This should reference a github issue which points to these lines to be uncommented when the issue is resolved.

hzeller: Done.

.github/bin/script.sh
```diff
@@ -18,6 +18,14 @@ set -e
...
+# declare copts = ["-fexceptions"] in the rule. We want to make it as simple
+# as possible to compile without exceptions.
+# ... and unfortunately we have to disable this for now as the external
+# kythe dependency does use exceptions.
+# TODO(hzeller): Uncomment after  #747 is flying (excuse the pun)
```
mithro, line 26:
```suggestion
# TODO(hzeller): Uncomment after https://github.com/google/verible/issues/747 is fixed.
```

PR #745, opened 2021-04-05
verilog/analysis: Add ParsedVerilogSourceFile class

ldalek-antmicro:
This class is an extended version of InMemoryVerilogSourceFile.
It wraps a TextStructureView object and allows to skip loading and
parsing phase.

### DIFF
--- a/verilog/analysis/verilog_project.h
+++ b/verilog/analysis/verilog_project.h
@@ -149,7 +149,7 @@ std::ostream& operator<<(std::ostream&, const VerilogSourceFile&);
 
 // An in-memory source file that doesn't require file-system access,
 // nor create temporary files.
-class InMemoryVerilogSourceFile : public VerilogSourceFile {
+class InMemoryVerilogSourceFile final : public VerilogSourceFile {
  public:
   // filename can be fake, it is not used to open any file.
   InMemoryVerilogSourceFile(absl::string_view filename,
--- a/verilog/analysis/verilog_project.h
+++ b/verilog/analysis/verilog_project.h
@@ -60,12 +60,12 @@ class VerilogSourceFile {
   // Attempts to lex and parse the file (without preprocessing).
   // Will Open() if the file is not already opened.
   // Depending on context, not all files are suitable for standalone parsing.
-  absl::Status Parse();
+  virtual absl::Status Parse();
 
   // After Open(), the underlying text structure contains at least the file's
   // contents.  After Parse(), it may contain other analyzed structural forms.
   // Before Open(), this returns nullptr.
-  const verible::TextStructureView* GetTextStructure() const;
+  virtual const verible::TextStructureView* GetTextStructure() const;
 
   // Returns the first non-Ok status if there is one, else OkStatus().
   absl::Status Status() const { return status_; }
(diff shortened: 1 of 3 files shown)


verilog/analysis/verilog_project.h
```diff
@@ -165,6 +165,30 @@ class InMemoryVerilogSourceFile : public VerilogSourceFile {
   const absl::string_view contents_for_open_;
 };
 
+// Source file that was already parsed and got its own TextStructure.
+// Doesn't require file-system access, nor create temporary files.
+class ParsedVerilogSourceFile : public VerilogSourceFile {
+ public:
+  // filename can be fake, it is not used to open any file.
+  ParsedVerilogSourceFile(absl::string_view filename,
+                          const verible::TextStructureView& text_structure,
```
hzeller, line 174: the ownership of this probably needs to be described in the constructor description, as this expects the text_structure be  available for the lifetime of this object.

I'd suggest to make this a `const verible::TextStructuredView*`, also text_structure_. Passing a pointer makes it more obvious that the object is keeping a reference to it.

(usually, when we pass const references in constructors, a copy is made, so this is what the programmer expects. If we keep a reference to the thing we can make it more visible by passing a pointer)

verilog/analysis/verilog_project.h
```diff
@@ -165,6 +165,30 @@ class InMemoryVerilogSourceFile : public VerilogSourceFile {
...
+                          absl::string_view corpus = "")
+      : VerilogSourceFile(filename, filename, corpus),
+        text_structure_(text_structure) {}
+
+  // Do nothing (file contents already loaded)
+  absl::Status Open() override;
```
hzeller, line 180: Given that we don't expect to derive further, we can make the `override` a `final` (same below; same in `InMemoryVerilogSourceFile`). Might need a `class ParsedVerilogSourceFile final : ...` above.

hzeller (changes requested):
(This looks like a preparatory change for some other change. What are you going to use it for ?)

Looks mostly good, but I think using a const reference in the constructor and holding on to it might need either some specific explanation or, better, passing a pointer to keep the reader on the right track.

verilog/analysis/verilog_project.h
ldalek-antmicro: I made the whole class final (same with InMemoryVerilogSourceFile class).

verilog/analysis/verilog_project.h
ldalek-antmicro: Done, done & done

ldalek-antmicro:
> (This looks like a preparatory change for some other change. What are you going to use it for ?)
> 
It is used in #217 (forbid implicit declarations, PR: https://github.com/google/verible/pull/244).

verilog/analysis/verilog_project.h
hzeller: Nice. At that point, the compiler knows that once it can proof that there can only be a ParsedVerilogSourceFile in a particular instance, it can replace calls to the virtual methods with direkt calls and does not have to go through the virtual method table. While here the performance win is probably not noticeable, it is a win for the reader of the code.

So for readability, I'd then also suggest to replace `override` with `final` in all the methods so that it is easier to see for the reader at first glance without having looked at the class declaration above.

Not critical for this review, can be changed later.

hzeller (approved):
LGTM

PR #746, opened 2021-04-05
[SymbolTable]: initial support for implicit declarations

ghost:
Support implicit declarations as defined in SystemVerilog LRM `6.10: Implicit declarations`

### DIFF
--- a/verilog/analysis/symbol_table.cc
+++ b/verilog/analysis/symbol_table.cc
@@ -161,6 +161,9 @@ static absl::Status DiagnoseMemberSymbolResolutionFailure(
                    context_name, "."));
 }
 
+static const SymbolTableNode* LookupSymbolUpwards(
+    const SymbolTableNode& context, absl::string_view symbol);
+
 class SymbolTable::Builder : public TreeContextVisitor {
  public:
   Builder(const VerilogSourceFile& source, SymbolTable* symbol_table,
@@ -265,6 +268,9 @@ class SymbolTable::Builder : public TreeContextVisitor {
         break;
       case NodeEnum::kEnumType:
         DescendEnumType(node);
+        break;
+      case NodeEnum::kLPValue:
+        HandlePossibleImplicitDeclaration(node);
         break;
       default:
         Descend(node);
@@ -521,6 +527,26 @@ class SymbolTable::Builder : public TreeContextVisitor {
     }
   }
 
+  void HandlePossibleImplicitDeclaration(const SyntaxTreeNode& node) {
+    VLOG(2) << __FUNCTION__;
+
+    // Only left-hand side of continuous assignment statements are allowed to
+    // implicitly declare nets (LRM 6.10: Implicit declarations).
+    if (Context().DirectParentsAre(
+            {NodeEnum::kNetVariableAssignment, NodeEnum::kAssignmentList,
+             NodeEnum::kContinuousAssignmentStatement})) {
+      CHECK(node.MatchesTag(NodeEnum::kLPValue));
+
+      DeclarationTypeInfo decl_type_info;
+      const ValueSaver<DeclarationTypeInfo*> save_type(&declaration_type_info_,
+                                                       &decl_type_info);
+      declaration_type_info_->implicit = true;
+      Descend(node);
+    } else {
+      Descend(node);
+    }
+  }
+
   void HandleIdentifier(const SyntaxTreeLeaf& leaf) {
     const absl::string_view text = leaf.get().text();
     VLOG(2) << __FUNCTION__ << ": " << text;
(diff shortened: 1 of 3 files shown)


verilog/analysis/symbol_table.cc
```diff
@@ -582,6 +608,29 @@ class SymbolTable::Builder : public TreeContextVisitor {
...
+
+        const ReferenceComponent implicit_ref{
+            .identifier = text,
+            .ref_type = InferReferenceType(),
+            .required_metatype = InferMetaType(),
+            // pre-resolve
+            .resolved_symbol = &implicit_declaration,
+        };
+
+        ref.PushReferenceComponent(implicit_ref);
```
fangism, line 629: Is this push needed?  When a symbol is implicitly declared, do we ever need to build a hierarchical reference from it with the net/variable as a parent?

verilog/analysis/symbol_table_test.cc
```diff
@@ -2183,6 +2183,138 @@ TEST(BuildSymbolTableTest, ModuleDeclarationLocalsDependOnParameter) {
...
+  EXPECT_EQ(module_m_info.file_origin, &src);
+  EXPECT_EQ(module_m_info.declared_type.syntax_origin,
+            nullptr);  // there is no module meta-type
+
+  MUST_ASSIGN_LOOKUP_SYMBOL(a_variable, module_m, "a");
+  EXPECT_EQ(a_variable_info.metatype, SymbolMetaType::kDataNetVariableInstance);
+  const ReferenceComponentNode* a_type_ref =
+      a_variable_info.declared_type.user_defined_type;
+  EXPECT_EQ(a_type_ref, nullptr);  // implicit type is primitive type
+
```
fangism, line 2210: also `EXPECT_TRUE` on `.implicit`?

verilog/analysis/symbol_table_test.cc
```diff
@@ -2183,6 +2183,138 @@ TEST(BuildSymbolTableTest, ModuleDeclarationLocalsDependOnParameter) {
...
+  EXPECT_EQ(module_m_info.local_references_to_bind.size(), 2);
+  const auto ref_map(module_m_info.LocalReferencesMapViewForTesting());
+
+  ASSIGN_MUST_FIND(a_refs, ref_map, "a");
+  ASSERT_EQ(a_refs.size(), 2);  // all references to "N" parameter
```
fangism, line 2269: reference to "a"

verilog/analysis/symbol_table_test.cc
```diff
@@ -2183,6 +2183,138 @@ TEST(BuildSymbolTableTest, ModuleDeclarationLocalsDependOnParameter) {
...
+  EXPECT_EQ(a_variable_info.metatype, SymbolMetaType::kDataNetVariableInstance);
+  const ReferenceComponentNode* a_type_ref =
+      a_variable_info.declared_type.user_defined_type;
+  EXPECT_EQ(a_type_ref, nullptr);  // implicit type is primitive type
+
+  EXPECT_EQ(module_m_info.local_references_to_bind.size(), 1);
+  const auto ref_map(module_m_info.LocalReferencesMapViewForTesting());
+
+  ASSIGN_MUST_FIND(a_refs, ref_map, "a");
+  ASSERT_EQ(a_refs.size(), 1);  // all references to "N" parameter
```
fangism, line 2215: references to "a"

verilog/analysis/symbol_table_test.cc
```diff
@@ -2183,6 +2183,138 @@ TEST(BuildSymbolTableTest, ModuleDeclarationLocalsDependOnParameter) {
...
+                            "assign a = 1'b0;"
+                            "assign a = 1'b1;"
+                            "endmodule\n");
+  const auto status = src.Parse();
+  ASSERT_TRUE(status.ok()) << status.message();
+  SymbolTable symbol_table(nullptr);
+  const SymbolTableNode& root_symbol(symbol_table.Root());
+
+  const auto build_diagnostics = BuildSymbolTable(src, &symbol_table);
+  EXPECT_EMPTY_STATUSES(build_diagnostics);
```
fangism, line 2251: Should we expect a diagnostic on an attempt to re-define `a` in the same scope?

verilog/analysis/symbol_table_test.cc
ldalek-antmicro: Yes. Good idea. Thanks!

verilog/analysis/symbol_table.cc
ldalek-antmicro: I think we do. Implicit declaration is also an reference to that symbol and we want to keep that reference.

verilog/analysis/symbol_table_test.cc
ldalek-antmicro: We do have such diagnostic (test case covering that is here: https://github.com/google/verible/pull/746/files#diff-45eb889cefc594107d8a95e50aa16ed073b4a7bf296ea244e2c20daa9762fe9cR2558)

I think that you meant an attempt to reference an implictly declared variable.
I thought about that and I think it would be nice to have an extra flag/switch that would report such references but I think that exceeds scope of this PR.

verilog/analysis/symbol_table_test.cc
```diff
@@ -2183,6 +2183,395 @@ TEST(BuildSymbolTableTest, ModuleDeclarationLocalsDependOnParameter) {
...
+    for (const auto& a_ref : a_refs) {
+      const ReferenceComponent& a_ref_comp(a_ref->components->Value());
+      EXPECT_EQ(a_ref_comp.resolved_symbol,
+                &a_variable);  // resolved successfully
+    }
+
+    for (const auto& n_a_ref : n_a_refs) {
+      const ReferenceComponent& n_a_ref_comp(n_a_ref->components->Value());
+      EXPECT_EQ(n_a_ref_comp.resolved_symbol,
+                &a_variable);  // resolved successfully
```
fangism, line 2381: I might need a language lawyer to explain to me.  Is the second declaration of "a" in the nested module supposed to resolve to that in the outer module?  Maybe add a comment to this test case that cites the LRM passage about this.

fangism (approved):
Great work on this!  Just minor comments and question.

verilog/analysis/symbol_table_test.cc
hzeller: +1 for adding comment.

hzeller (approved):
LGTM. Please double-check that you've gone through fangisms comments as well.

hzeller:
This will be relevant for finally implementing #217 (An implementation should make use of the testing from the initial approach in #244 )

PR #757, opened 2021-04-14
Add MacOS platform compatibility check CI

hzeller:
Issues: #260

### DIFF
--- a/.github/workflows/macos-compile.yml
+++ b/.github/workflows/macos-compile.yml
@@ -1,0 +1,29 @@
+# Platform specific MacOS test. See verible-ci for main continuous integration.
+name: macos-compile
+
+on:
+  push:
+    branches:
+    - master
+    - main
+  pull_request:
+    branches:
+    - master
+    - main
+
+jobs:
+  MacOsBuild:
+    runs-on: macos-latest
+    steps:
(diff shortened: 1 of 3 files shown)


mithro (commented):
Woot!

hzeller:
Aaaand: it passes :)
(had to revert to older version of ABSL, there is something I probably have to file separately with the absl team)

.github/workflows/macos-compile.yml
```diff
@@ -0,0 +1,29 @@
...
+    - master
+    - main
+  pull_request:
+    branches:
+    - master
+    - main
+
+jobs:
+  MacOsBuild:
+    runs-on: macos-latest
```
fangism, line 16: comment on what version OS this happens to be?  other known versions of OS X Developer Tools?

fangism (approved):
Fantastic work!

.github/workflows/macos-compile.yml
hzeller: Version is unknown. Whatever github choses to be 'latest'.
Given that this is just meant as a litmus test, we actually don't really want to know, we just want to make sure that whatever typical latest macos there is, it will compile.

PR #759, opened 2021-04-14
Bump version number of used bazel to 4.0.0

hzeller:
After a couple of months transition period, this is now
the dominant version of bazel used.


hzeller:
Need to wait for #760 to resolve for this to proceed.

mithro (approved):
If CI says yes, go ahead!

hzeller:
Yes, won't work yet until @ivan444 has removed the dependency on whole of kythe as there is a broken *.bzl file in the kythe version we use, which trips the new bazel.

Issue #730, opened 2021-04-01
[kythe] Indexer's JSON format isn't proper JSON


ivan444:
Running the Kythe indexer with `--print_kythe_facts json` will produce a stream of Kythe facts in JSON format. One fact in JSON format per line (that's what Kythe's entrystream binary accepts, and we need to keep it). The overall output isn't well formatted JSON.

`json_debug` format suffers from the same issue. This breaks any consumer of this output that expects a well formatted JSON.

Change `json_debug` to output a proper JSON format (and consider changing `json_debug` to something that better reflects the purpose if the clients start using it).

Context: https://groups.google.com/g/verible-users/c/iir_WASxkak

PR #763, opened 2021-04-16
[kythe] Make --json print output single line. For human readable JSON…

ivan444:
… output use --json_debug

Related to #730

### DIFF
--- a/verilog/tools/kythe/kythe_facts.cc
+++ b/verilog/tools/kythe/kythe_facts.cc
@@ -68,20 +68,22 @@ bool VName::operator<(const VName& other) const {
 
 std::ostream& VName::FormatJSON(std::ostream& stream, bool debug,
                                 int indentation) const {
-  stream << "{\n";
+  // Output new line only in debug mode.
+  const std::string new_line = debug ? "\n" : "";
+  stream << "{" << new_line;
   const verible::Spacer idt(indentation + 2);
   stream << idt << "\"signature\": \""
-         << (debug ? signature.ToString() : signature.ToBase64()) << "\",\n"  //
-         << idt << "\"path\": \"" << path << "\",\n"                          //
-         << idt << "\"language\": \"" << language << "\",\n"                  //
-         << idt << "\"root\": \"" << root << "\",\n"                          //
-         << idt << "\"corpus\": \"" << corpus << "\"\n"                       //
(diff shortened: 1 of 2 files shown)


verilog/tools/kythe/kythe_facts.cc
```diff
@@ -68,20 +68,22 @@ bool VName::operator<(const VName& other) const {
 
 std::ostream& VName::FormatJSON(std::ostream& stream, bool debug,
                                 int indentation) const {
-  stream << "{\n";
+  // Output new line only in debug mode.
+  const std::string new_line = debug ? "\n" : "";
```
hzeller, line 72: Since it is only sometimes newline, the variable name is only sometimes right, I'd call it `separator` or `field_separator`.

(the following question is mostly showing my ignorance of how JSON is typically consumed)
Is it intentional to have an empty string instead of a 0x20 space for more even more compactness or easier shell-parsing (e.g. `awk -F,`) ?
Otherwise I'd expect that we have a 'comma space next-thing' instead of 'comma next-thing'.

verilog/tools/kythe/kythe_facts.cc
```diff
@@ -96,22 +98,26 @@ bool Fact::operator<(const Fact& other) const {
 
 std::ostream& Fact::FormatJSON(std::ostream& stream, bool debug,
                                int indentation) const {
-  const int indent_more = indentation + 2;
+  // Output new line only in debug mode.
+  const std::string new_line = debug ? "\n" : "";
```
hzeller, line 102: new_line -> separator ?

hzeller (changes requested):
Mostly LGTM, but variable name suggestion and one question.

verilog/tools/kythe/kythe_facts.cc
ivan444: Done.

verilog/tools/kythe/kythe_facts.cc
ivan444: Renamed.

It's unlikely that many people consume JSON using `awk`. It's inconvenient & generally inapplicable as field separator (e.g., you could have a list). Also, there's `jq` which is breeze to use.

hzeller (approved):

Issue #760, opened 2021-04-14
Testing later version of Kythe: new build problems


hzeller:
Resolution in #625 was to clamp 0.0.49 to some older version of Kythe. Did a quick test today to see if the latest version `0.0.52` would work, but different issues emerge:

Using this in WORKSPACE
```
http_archive(
    name = "io_kythe",
    sha256 = "dde86b6fa277486665dc83c66aaf9f5f961427e3e417538a09498494c0be0390",
    strip_prefix = "kythe-0.0.52",
    urls = [
        "https://github.com/kythe/kythe/archive/refs/tags/v0.0.52.tar.gz"
    ],
)
```
Stumbles upon some ruby dependency.
<details><summary>bazel build log output</summary>

```
$ bazel build ...
INFO: Repository bazelruby_rules_ruby instantiated at:
  /home/hzeller/src/my/verible/WORKSPACE:152:24: in <toplevel>
  /home/hzeller/.cache/bazel/_bazel_hzeller/fde5431bd58d2dc68f037724f34a8e77/external/io_kythe/setup.bzl:120:10: in kythe_rule_repositories
  /home/hzeller/.cache/bazel/_bazel_hzeller/fde5431bd58d2dc68f037724f34a8e77/external/io_kythe/setup.bzl:7:18: in maybe
Repository rule http_archive defined at:
  /home/hzeller/.cache/bazel/_bazel_hzeller/fde5431bd58d2dc68f037724f34a8e77/external/bazel_tools/tools/build_defs/repo/http.bzl:336:31: in <toplevel>
INFO: Repository 'bazelruby_rules_ruby' used the following cache hits instead of downloading the corresponding file.
 * Hash 'abfc2758cc379e0ff0eb9824e3b507c1633d4c8f99f24735aef63c7180be50f0' for https://github.com/bazelruby/rules_ruby/archive/v0.4.1.zip
If the definition of 'bazelruby_rules_ruby' was updated, verify that the hashes were also updated.
ERROR: An error occurred during the fetch of repository 'bazelruby_rules_ruby':
   Traceback (most recent call last):
        File "/home/hzeller/.cache/bazel/_bazel_hzeller/fde5431bd58d2dc68f037724f34a8e77/external/bazel_tools/tools/build_defs/repo/http.bzl", line 121, column 10, in _http_archive_impl
                patch(ctx)
[...]

PR #764, opened 2021-04-16
[kythe] Remove the dependency on linked Kythe verification binary and

ivan444:
io_bazel.

Instead of linking the Kythe verification binary, use precompiled one
from Github Workflow.
Remove Kythe verification test from the general list of tests and
introduce it as an explicit Github workflow step.

Replace bash unit test library from io_bazel with poor man's simple test
suite (remove large dependency -- if we need bash xUnit test framework,
there are smaller ones available).

Related to #760

### DIFF
--- a/.github/bin/verify-kythe-extraction.sh
+++ b/.github/bin/verify-kythe-extraction.sh
@@ -15,9 +15,8 @@
 
 set -e
 
-KYTHE_DIRNAME="kythe-${KYTHE_VERSION}"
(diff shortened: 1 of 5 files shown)


hzeller (approved):
LGTM. Nice detangling!

mithro:
Great work! 

As a thank you, if you are interested in a receiving a small "FPGA care package" for doing this work, send me an email to me@mith.ro with your address + phone number and I'll get one sent to you.

PR #768, opened 2021-04-19
Rename 'script.sh' to 'build-and-test.sh'

hzeller:
script.sh was from a very early stage of our github actions
when this was all we did. These days, the name does not make
sense anymore.


mithro (approved):
Makes total sense!

PR #771, opened 2021-04-20
Added support for win_flex_bison on Windows

corco:
As a first step towards compiling on Windows (#307), use [WinFlexBison](https://github.com/lexxmark/winflexbison) on Windows instead of compiling m4/flex/bison (which doesn't work)

### DIFF
--- a/.github/workflows/windows-compile.yml
+++ b/.github/workflows/windows-compile.yml
@@ -28,8 +28,8 @@ jobs:
     # fully compiled on Windows (yet!)
     - name: Build flex/bison files
       run: |
-        $env:FLEX = 'C:/ProgramData/Chocolatey/lib/winflexbison/tools/win_flex.exe'
-        $env:BISON = 'C:/ProgramData/Chocolatey/lib/winflexbison/tools/win_bison.exe'
+        $env:FLEX = ($env:ChocolateyInstall + '/lib/winflexbison/tools/win_flex.exe') -replace '\\', '/'
+        $env:BISON = ($env:ChocolateyInstall + '/lib/winflexbison/tools/win_bison.exe') -replace '\\', '/'
         bazel build //common/analysis:command_file_lex
         bazel build //verilog/parser:verilog_y
         bazel build //verilog/parser:verilog_lex
--- a/.github/workflows/windows-compile.yml
+++ b/.github/workflows/windows-compile.yml
@@ -24,6 +24,8 @@ jobs:
       uses: actions/checkout@v2
       with:
         fetch-depth: 0
+    # Only check flex/bison targets at the moment, since Verible cannot be
+    # fully compiled on Windows (yet!)
     - name: Build flex/bison files
       run: |
         $env:FLEX = 'C:/ProgramData/Chocolatey/lib/winflexbison/tools/win_flex.exe'
--- a/.github/workflows/windows-compile.yml
+++ b/.github/workflows/windows-compile.yml
@@ -1,0 +1,33 @@
+# Platform specific Windows test. See verible-ci for main continuous integration.
+name: windows-compile
+
+on:
+  push:
+    branches:
+    - master
+    - main
+  pull_request:
+    branches:
+    - master
+    - main
+
+jobs:
+  WindowsBuild:
+    runs-on: windows-latest
+    steps:
+    # Need winflexbison for Kythe, bazel fails if environment
+    # FLEX/BISON is not set to a valid program
+    - name: Install deps
+      run: |
+        choco install winflexbison
+    - name: Checkout code
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+    - name: Build flex/bison files
+      run: |
+        $env:FLEX = 'C:/ProgramData/Chocolatey/lib/winflexbison/tools/win_flex.exe'
+        $env:BISON = 'C:/ProgramData/Chocolatey/lib/winflexbison/tools/win_bison.exe'
+        bazel build //common/analysis:command_file_lex
+        bazel build //verilog/parser:verilog_y
+        bazel build //verilog/parser:verilog_lex
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -78,7 +78,15 @@ load("@com_github_google_rules_install//:setup.bzl", "install_rules_setup")
 
 install_rules_setup()
 
-all_content = """filegroup(name = "all", srcs = glob(["**"]), visibility = ["//visibility:public"])"""
+# Need to load before rules_flex/rules_bison to make sure
+# win_flex_bison is the chosen toolchain on Windows
+load("//bazel:win_flex_bison.bzl", "win_flex_configure")
+
+win_flex_configure(
+    name = "win_flex_bison",
+    sha256 = "39c6086ce211d5415500acc5ed2d8939861ca1696aee48909c7f6daf5122b505",
+    url = "https://github.com/lexxmark/winflexbison/releases/download/v2.5.24/win_flex_bison-2.5.24.zip",
+)
 
 http_archive(
     name = "rules_m4",
--- a/bazel/bison.bzl
+++ b/bazel/bison.bzl
@@ -1,5 +1,5 @@
 # -*- Python -*-
-# Copyright 2017-2020 The Verible Authors.
+# Copyright 2017-2021 The Verible Authors.
 #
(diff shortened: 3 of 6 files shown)


corco:
@googlebot I signed it!

mithro:
Is there a way to verify this is working on GitHub Actions?

.github/workflows/windows-compile.yml
```diff
@@ -0,0 +1,29 @@
+# Platform specific Windows test. See verible-ci for main continuous integration.
+name: macos-compile
```
hzeller, line 2: Sounds like you also want to change the name :)

WORKSPACE
```diff
@@ -78,7 +78,15 @@ load("@com_github_google_rules_install//:setup.bzl", "install_rules_setup")
...
 
-all_content = """filegroup(name = "all", srcs = glob(["**"]), visibility = ["//visibility:public"])"""
+# Need to load before rules_flex/rules_bison to make sure
+# win_flex_bison is the chosen toolchain on Windows
+load("//bazel:win_flex_bison.bzl", "win_flex_configure")
+
+win_flex_configure(
+    name = "win_flex_bison",
+    sha256 = "39c6086ce211d5415500acc5ed2d8939861ca1696aee48909c7f6daf5122b505",
+    url = "https://github.com/lexxmark/winflexbison/releases/download/v2.5.24/win_flex_bison-2.5.24.zip",
```
hzeller, line 88: The repository there doesn't have a `COPYING` or `LICENSE` file in the toplevel.
They are implicitly and unambiguously GPL, as these are derivatives of bison and flex and their build-instructions, but maybe we should poke them and making sure that this in their repo marked in the toplevel.

bazel/bison.bzl
```diff
@@ -30,9 +30,15 @@ def genyacc(
         name = name,
         srcs = [src],
         outs = [header_out, source_out] + extra_outs,
-        cmd = "M4=$(M4) $(BISON) --defines=$(location " + header_out + ") --output-file=$(location " + source_out + ") " + " ".join(extra_options) + " $<",
-        toolchains = [
-            "@rules_bison//bison:current_bison_toolchain",
-            "@rules_m4//m4:current_m4_toolchain",
-        ],
+        cmd = select({
```
hzeller, line 33: Nice. #TIL about selecting depending on platform.

hzeller (commented):
Very cool!
I mostly only have one comment (changing the name of the workflow from mac to windows :) ).

Also I suggest if you can and have not already, run the changed bzl and added BUILD file through [`buildifier`](https://github.com/bazelbuild/buildtools/blob/master/buildifier/README.md)  (from https://github.com/bazelbuild/buildtools ) which creates a uniform formatting (I don't have any complaints about the formatting on these, but this will make sure it is uniform).

Other than that, we're good to final review and merge once the CI is passing.

.github/workflows/windows-compile.yml
```diff
@@ -0,0 +1,29 @@
...
+    - name: Install deps
+      run: |
+        choco install winflexbison
+    - name: Checkout code
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+    - name: Build flex/bison files
+      run: |
+        bazel build //common/analysis:command_file_lex
```
hzeller, line 27: I'd probably add a comment saying that these are the first targets ported to work on Windows until Verible is fully platform independent.

WORKSPACE
corco: I raised an issue [on the repository](https://github.com/lexxmark/winflexbison/issues/75). It should be GPLv3

corco: Should be GPLv3. I raised [an issue](https://github.com/lexxmark/winflexbison/issues/75)

.github/workflows/windows-compile.yml
```diff
@@ -0,0 +1,35 @@
...
+        choco install winflexbison
+    - name: Checkout code
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+    # Only check flex/bison targets at the moment, since Verible cannot be
+    # fully compiled on Windows (yet!)
+    - name: Build flex/bison files
+      run: |
+        $env:FLEX = 'C:/ProgramData/Chocolatey/lib/winflexbison/tools/win_flex.exe'
```
hzeller, line 31: This sounds oddly specific and I am worried that this is brittle and might break easily.

Is there a way to find the path where the binaries end up in some other way ? I could imagine that the `C:/ProgramData/Chocolatey/` is the base-path where `choco` stores things, maybe it can be queried what its base-path is, so that we can assemble the final path from there ?

(On Mac-OS for instance, we ask the `brew` package manager [where it installed binaries]( https://github.com/google/verible/blob/6b827debc9dbd7da2d1a4adad8a16113e0083ac0/.github/workflows/macos-compile.yml#L27) )

hzeller: Question for my understanding: the rules don't actually build flex and bison in the hermetic build from source (like they do on the Unixes), but rely on the chocolatey system-installed ones, this is why we need this in the first place ?

(Similar to what Kythe is doing for its building, which requires us to install flex and bison for the Unixy builds currently)

corco: For Verible, the rules I added download the github released binary and uses that in the genlex/genyacc. There is no need for a system installed flex/yacc.

However, the WORKSPACE calls [lexyacc_configure from Kythe](https://github.com/kythe/kythe/blob/160b28af24a477fd75458e033523cf9ce81733ee/tools/build_rules/lexyacc/lexyacc.bzl#L156), which expects to find flex/bison on the path or in the environment variables FLEX/BISON, so a valid flex/bison must be installed to satisfy Kythe (otherwise, bazel fails instantaneously). On Windows, we have to use the environment variable because the name is different than what Kythe expects (win_flex.exe vs flex...).

When building the Kythe's lex/yacc targets, the system installed flex/bison gets used (same as Unix).

There is now a way to [forgo installing flex/bison](https://github.com/kythe/kythe/pull/4880) on the system and use a bazel label instead. If you want to update to latest Kythe, I could provide a pull-request for that, which remove usage of the system flex/bison.

hzeller: Ok cool. We are soon actually not needing Kythe direct dependency anymore (one step in #772) , so we might be able to reduce the dependency soon here as well.

hzeller:
Thanks, merged!

bazel/win_flex_bison.bzl
```diff
@@ -0,0 +1,119 @@
...
+    implementation = _remote_win_flex_bison,
+    local = True,
+    attrs = {
+        "url": attr.string(mandatory = True),
+        "sha256": attr.string(default = ""),
+    },
+)
+
+def win_flex_configure(name, url, sha256 = ""):
+    remote_win_flex_bison(
```
hzeller, line 110: Could this `win_flex_configure()` be made to return early if we're not on Windows ?

There seems to be an issue with pre-fetching all the needed dependencies vs. what is then actually needed in download later in environments that separate out the fetching and building like on nixos; this is why it needs this patch:
https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/tools/language-servers/verible/remove-unused-deps.patch

I suspect what is happening: the `win_flex_configure()` called from the `WORKSPACE` unconditionally downloads the dependencies , but then on a non-Windows system, the `select()` in the `{bison,flex}.bzl` realizes that it doesn't actually needs it, so it then leaves it out of the prepeared download blob.
Later, after disconnect from the network, on `build`, the unconditional download attempt will fail. At least this is what I _think_ is happening.

Anyway, if here, we could just skip calling the `remote_win_flex_bison()` and register toolchain part if we're not on windows, that should probably help. Do you think you can add that @corco ?

corco: I'll will look at it at the end of the week

corco: I took a look. It seems that [this issue](https://github.com/bazelbuild/bazel/issues/13493) is part of the problem. There may be a way to fix it in either nix or bazel, but I am not comfortable with either...

I would suggest to either wait until bazel fix the issue, or simply remove downloading win_flex_bison and require the user to install it prior with chocolatey, like the use_local_flex_bison flag. I would move toward that second solution if @hzeller agrees.

hzeller: Oh, sorry, indeed I didn't see your previous comment.
I think removing the download and let the user `use_local_flex_bison` sounds good to me.

PR #776, opened 2021-04-20
CI: don't use system installed flex/bison anymore.

hzeller:
After #772, we don't need to build Kythe in its entirety
anymore. It required a system installed flex/bison, don't
need that anymore now.

### DIFF
--- a/.github/workflows/macos-compile.yml
+++ b/.github/workflows/macos-compile.yml
@@ -15,15 +15,10 @@ jobs:
   MacOsBuild:
     runs-on: macos-latest
     steps:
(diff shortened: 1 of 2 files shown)


mithro (approved):
If CI says yes, merge away!

corco:
Shouldn't it be removed from the releasing folder as well?

hzeller:
Yes, good point; someone moved the files to `releasing/` so my quick grep in `.github/` didn't find it.
Let's await this CI finishing, then I'll clean up the files in releasing (though might consider making it a separate pull request)

Issue #765, opened 2021-04-19
Move to use C++17 and with it, embrace std::filesystem


hzeller:
- [ ] Evaluate if we can move to c++17, by checking all the (sometimes pretty old) distributions we support when building.  **No** : all compilers can do c++17, but not all of them provide the `<filesystem>` header yet.
- [x] Are there inconveniences that we need to factor in ? For instance do we need to tell users to add `--cxxopts=-std=c++17`, or is that the default already with using a current bazel (4.0.0). If it is the default with 4.0.0 (current version bazel) but not 3.7 which we use for CI currently, then we might first need to await the Kythe dependency reduction (which now pins us on 3.7) (_inconveniences already managed in `.bazelrc`_; Independent, using Bazel 4.0.0 now #794  )
  - [x] Adapt CI to use C++17 enabled compiler if not already _already happening, but release docker gcc's can't deal all with `<filesystem>`_
  - [x] Add possible documentation updates if special `cxxopts` are needed when building (hopefully not). _not needed_
- [x] Replace all unix-specific code with std::filesystem (Done in _#789_, but currently only enabled for Windows)
   - [x] There is at least one function [`UpwardFileSearch()`](https://github.com/google/verible/blob/6ef4938d0a800b93ebf9fa7fdf692c4ed06c0471/common/util/file_util.h#L56-L60) that needs a unit test first. Possibly more.
   - [x] To use `std::filesystem` in the past, there was the inconvenience to have to add `-lstdc++fs`. This might probably be possible to add to the `copts` in the BUILD file, but we need to make sure this works on all `gcc` and `clang` a well as on all used compiler versions of these (because it seems to be impossible in bazel to change `copts` in BUILD files depending on toolchain used).  _for all gcc versions that do support `<filesystem>` nothing has to be added to link options_

This issue will probably the aggregation point for multiple submits until we have replaced

hzeller:
In #789 an implementation to used `std::filesystem` has been added, but _only_ enabled for Windows right now as the release builds (in [`releasing/`](https://github.com/google/verible/tree/master/releasing)) sometimes run on such old platforms that they don't have new enough compilers yet.

TODO: install the necessary compilers there (backports?) and always [enable std::filesystem]( https://github.com/google/verible/blob/495dda72923d1a13422a6f3a3fd131ed3b1fff37/common/util/file_util.cc#L44) for all platforms.

Once that is done, the legacy implementation can be removed.

hzeller:
[...]

PR #781, opened 2021-04-21
Towards compiling on windows

corco:
This is a first step towards compiling on Windows (#307). In summary:

- Add Windows compile flags in bazelrc. I used --enable_platform_specific_config, so in theory everything should work as before on Linux/MacOS.
- ~~Updated glog to latest version since the current version does not compile on Windows~~ (now in #784)
- Forces the use of Abseil's string_view implementation. On C++17, Abseil's string_view is an alias for the STL's string_view. However, on MSVC, the string_view::iterator cannot be cast to a char*, which is expected by Verible.
- ~~Remove process.cc. The file doesn't compile on Windows and is unused, so easy fix!~~ (now in #785)

I will also note that MSVC's compiler is more pedantic than gcc/clang and refuse to use designated initializers notation, which won't be supported until C++20. As a consequence, I did my test with clang-cl, the MSVC-compatible compiler from LLVM. I suggest to only support clang-cl for the moment.

The biggest issue to proceed testing on Windows is #765. common/util/file_util.cc is a transitive dependency to almost everything and thus make running test on Windows impossible.


common/util/BUILD
```diff
@@ -29,13 +29,6 @@ cc_library(
     hdrs = ["forward.h"],
 )
 
-cc_library(
-    name = "process",
```
hzeller, line 33: I think we needed process to invoke the external `diff` binary, but that is now done internally in verible, so not needed anymore.

I suggest to break 'removing process.cc' out as a separate pull request, so that we can get the codebase cleaner towards platform independence with each step.

bazel/absl.patch
```diff
@@ -0,0 +1,12 @@
+diff -uw -r a/absl/base/options.h b/absl/base/options.h
+--- a/absl/base/options.h	2021-04-02 23:00:31.996584600 -0400
++++ b/absl/base/options.h	2021-04-02 23:01:07.506208600 -0400
+@@ -154,7 +154,7 @@
+ // absl::string_view is a typedef of std::string_view, use the feature macro
+ // ABSL_USES_STD_STRING_VIEW.
+
+-#define ABSL_OPTION_USE_STD_STRING_VIEW 2
```
hzeller, line 8: is this also an issue with `clang-cl` ? (I'd assume that clang might come with its own std:: library, though it is probably more likely that they use the Windows std:: headers for ABI compatibility).

WORKSPACE
```diff
@@ -34,11 +39,9 @@ http_archive(
 
 http_archive(
     name = "com_github_google_glog",
-    # Using the same version as in kythe
-    # https://github.com/kythe/kythe/blob/410f69c5bcb69fabcb78a5200b7631a1bffabd31/external.bzl#L155
-    sha256 = "9b4867ab66c33c41e2672b5de7e3133d38411cdb75eeb0d2b72c88bb10375c71",
-    strip_prefix = "glog-ba8a9f6952d04d1403b97df24e6836227751454e",
-    urls = ["https://github.com/google/glog/archive/ba8a9f6952d04d1403b97df24e6836227751454e.zip"],
+    sha256 = "62efeb57ff70db9ea2129a16d0f908941e355d09d6d83c9f7b18557c0a7ab59e",
```
hzeller, line 42: Good. With Kythe out of the way, we can upgrade.

I suggest to send this glog update as separate pull request.

WORKSPACE
```diff
@@ -4,6 +4,11 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
 http_archive(
     name = "com_google_absl",
+    # On MSVC's STL implementation, string_view cannot be constructed from
+    # a string_view::iterator. This patch forces the use of absl's string_view
```
hzeller, line 8: Is this a bug in the MSVC STL implementation that we should file with Microsoft, or is this within the frame of interpretation of the standard ?
Do you have an example of an error message this produces ?

verilog/parser/BUILD
```diff
@@ -117,11 +122,17 @@ record_recovered_syntax_errors(
 cc_library(
     name = "verilog_y_cc",
     srcs = ["verilog-final.tab.cc"],
-    hdrs = ["verilog_parse_interface.h"],
-    copts = [
-        "-Wno-implicit-fallthrough",
-        "-Wno-type-limits",
-    ],
+    hdrs = ["verilog_parse_interface.h"] + select({
+        "@platforms//os:windows": ["verilog.tab.hh"],
```
hzeller, line 126: So this is intentional to not include verilog.tab.hh, but the verilog_parse_interface which is generated.
What was the compile trouble you had seen when _not_ adding `verilog.tab.hh` here ?

hzeller (commented):
A few questions and some encouragement to break out some of the changes we can do immediately as separate pull requests, so that we accumulate the trace of what had to be done in #307 

Where are we using designated initializers ?  I have not found an option in gcc or clang to warn about designated initializers so couldn't find all the places easily.

Maybe we can just comment uses of `.foo = bar` as `/* .foo = */ bar` for now until we're on C++20.

WORKSPACE
hzeller: Maybe we can use the almost-next-release version https://github.com/google/glog/archive/refs/tags/v0.5.0-rc2.zip

That way, it provides us a visual hint that we should update to v0.5.0 once it is out.

corco: Have you considered using renovatebot to keep dependencies up-to-date?

hzeller: Only partially. Reason is, that almost always when there is a dependency changed, it turns out it requires some human intervention, so I like to do it as a deliberate action every now and then (example here: the particular version was chosen because of implicit Kythe dependency).

WORKSPACE
corco: That appears to be a deliberate strict interpretation of the standard. I found [this Reddit](https://www.reddit.com/r/cpp/comments/bg7a52/stdstring_view_has_no_constructor_taking_its_own/) about it.

It can be fixed in the code by using replacing `.begin()` with `.data()` and `.end()` with `.data() + .length()`, but since it is used at several places in Verible's codebase, I think it is simpler to use absl's implementation, for now at least.

To be fair, I am not a programmer, especially if an OS is involved, so likely someone more used to `string_view` can find an easier fix.

WORKSPACE
corco: > Where are we using designated initializers ? I have not found an option in gcc or clang to warn about designated initializers so couldn't find all the places easily.
> 
> Maybe we can just comment uses of `.foo = bar` as `/* .foo = */ bar` for now until we're on C++20.

There is [verilog_linter_configuration_test](https://github.com/google/verible/blob/713e510250ba0f18e3fd634e27c7c4a182647411/verilog/analysis/verilog_linter_configuration_test.cc#L782), [kythe_facts_extractor](https://github.com/google/verible/blob/713e510250ba0f18e3fd634e27c7c4a182647411/verilog/tools/kythe/kythe_facts_extractor.cc#L678), [symbol_table](https://github.com/google/verible/blob/713e510250ba0f18e3fd634e27c7c4a182647411/verilog/analysis/symbol_table.cc#L448) and at least 5 other files.

There are other files that won't compile with MSVC (notably `editscript.h`). For the moment, I suggest we go with `clang-cl`. Once we have a version working on Windows, we can update the project to compile on MSVC.

hzeller: Sounds good. One step at a time.

verilog/parser/BUILD
corco: Without `verilog.tab.hh`, bazel complains `verilog-final.tab.cc` needs it and refuse to compile it.

I checked the generated `verilog.tab.cc` on Linux and it has this wonderful comment:

``` 
/* In a future release of Bison, this section will be replaced
   by #include "verilog.tab.hh".  */
```

And, of course, `win_flex_bison` has a "future release of Bison". I suggest I change the `win_flex_bison` release to match `rules_bison` version?

hzeller: Yes, I think for now matching the versions if possible would be best.
We then can later see how we can deal with this particular issue separately.

.bazelrc
```diff
@@ -9,8 +9,18 @@ build --workspace_status_command=bazel/build-version.sh
...
-run --cxxopt=-std=c++17 --host_cxxopt=-std=c++17 --client_env=BAZEL_CXXOPTS=-std=c++17
-test --cxxopt=-std=c++17 --host_cxxopt=-std=c++17 --client_env=BAZEL_CXXOPTS=-std=c++17
+build:linux --cxxopt=-std=c++17 --host_cxxopt=-std=c++17 --client_env=BAZEL_CXXOPTS=-std=c++17
+run:linux --cxxopt=-std=c++17 --host_cxxopt=-std=c++17 --client_env=BAZEL_CXXOPTS=-std=c++17
+test:linux --cxxopt=-std=c++17 --host_cxxopt=-std=c++17 --client_env=BAZEL_CXXOPTS=-std=c++17
+
+build:macos --cxxopt=-std=c++17 --host_cxxopt=-std=c++17 --client_env=BAZEL_CXXOPTS=-std=c++17
+run:macos --cxxopt=-std=c++17 --host_cxxopt=-std=c++17 --client_env=BAZEL_CXXOPTS=-std=c++17
+test:macos --cxxopt=-std=c++17 --host_cxxopt=-std=c++17 --client_env=BAZEL_CXXOPTS=-std=c++17
+
```
hzeller, line 23: Maybe adding a comment that we use `clang-cl` as there are still too many compatibility issues with msvc.

hzeller:
With these changes, are there some tests that already compile and test successfully ?
If you run
```
bazel test --keep_going ...
```

It will compile and test and also not stop at the first error, so we will get a list of tests that make it to the end and compile and test successfully. Probably a lot of these are indirectly dependent on the file-util, but the others we then list them in the windows-CI, something like
```
bazel test //common/strings:range_test \
           //common/strings:split_test \
           //verilog/parser:verilog_lexer_unittest
```

Or if there is a whole range working we can say `bazel test //common/strings/...`.

With that, we can make sure that future changes to the code-base don't break the tests that you already got working.

corco:
Rebased on master (to grab #784 & #785) and added passing test. Not a pretty commit, but it is encouraging to know 171 test passes. A single test failed (gonna look at it) and 449 fail to build, mostly because of file_util!

hzeller (approved):
LGTM, very cool.
The number of tests that pass is indeed very encouraging!

hzeller:
Alright, merged!

PR #782, opened 2021-04-21
Upgrade googletest to github HEAD.

hzeller:
The last released version does not use TEST_TMPDIR set by
bazel and always falls back to /tmp for running tests. That
makes them less hermetic.
(Unfortunately, that only fixes it on Linux, for MacOS I
have an upstream patch open).

The last release was in Oct 2019, so taking github HEAD.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -9,11 +9,15 @@ http_archive(
     urls = ["https://github.com/abseil/abseil-cpp/archive/20200923.3.zip"],
 )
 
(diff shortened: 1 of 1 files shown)


hzeller:
CC @johanvdhaegen I noticed that even on Linux the `TEST_TMPDIR` environment variable is not used with the latest release of googletest. Changing googletest to HEAD fixes it on Linux, but upstream patch to googletest to also fix it for MacOS is still in review.

mithro (approved):
If CI says yes, go ahead and merge!

PR #783, opened 2021-04-21
Add unit-test for UpwardFileSearch().

hzeller:
Precondition to change the implementation to use
std::filesystem.

Issues: #765 #307

### DIFF
--- a/common/util/file_util.cc
+++ b/common/util/file_util.cc
@@ -174,7 +174,16 @@ absl::Status SetContents(absl::string_view filename,
 }
 
 std::string JoinPath(absl::string_view base, absl::string_view name) {
-  return absl::StrCat(base, "/", name);
+  // TODO: this will certainly be different on Windows
+  static constexpr std::string_view kFileSeparator = "/";
+
+  // Make sure that we don't concatenation multiple superfluous separators.
+  // Note, since we're using string_view, the substr() operations are cheap.
+  while (base.length() && base[base.length()-1] == kFileSeparator[0])
+    base = base.substr(0, base.length() - 1);
+  while (name.length() && name[0] == kFileSeparator[0])
+    name = name.substr(1);
+  return absl::StrCat(base, kFileSeparator, name);
 }
 
 absl::Status CreateDir(absl::string_view dir) {
--- a/common/util/file_util_test.cc
+++ b/common/util/file_util_test.cc
@@ -19,6 +19,7 @@
 #include <vector>
 
 #include "absl/status/status.h"
+#include "absl/strings/match.h"
 #include "absl/strings/string_view.h"
 #include "gmock/gmock.h"
 #include "gtest/gtest.h"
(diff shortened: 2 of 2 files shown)


common/util/file_util.cc
```diff
@@ -174,7 +174,18 @@ absl::Status SetContents(absl::string_view filename,
 }
 
 std::string JoinPath(absl::string_view base, absl::string_view name) {
-  return absl::StrCat(base, "/", name);
+  // TODO: this will certainly be different on Windows
```
tgorochowik, line 177: I assume the idea is not to worry about that for now as it will be fixed when switching to c++17?

common/util/file_util.cc
```diff
@@ -174,7 +174,18 @@ absl::Status SetContents(absl::string_view filename,
...
+  static constexpr absl::string_view kFileSeparator = "/";
+
+  // Make sure that we don't concatenation multiple superfluous separators.
+  // Note, since we're using string_view, the substr() operations are cheap.
+  while (!base.empty() && base[base.length() - 1] == kFileSeparator[0]) {
+    base = base.substr(0, base.length() - 1);
+  }
```
tgorochowik, line 184: It is not immediately clear that this will work if base is just `"/"` - but I guess it will as the separator will be the leading `/` then - maybe we could add a comment saying why it's okay?

common/util/file_util.cc
```diff
@@ -174,7 +174,18 @@ absl::Status SetContents(absl::string_view filename,
 }
 
 std::string JoinPath(absl::string_view base, absl::string_view name) {
-  return absl::StrCat(base, "/", name);
+  // TODO: this will certainly be different on Windows
+  static constexpr absl::string_view kFileSeparator = "/";
+
+  // Make sure that we don't concatenation multiple superfluous separators.
```
tgorochowik, line 180: `concatenate`

tgorochowik (approved):
LGTM! Thank you for working on this!

common/util/file_util_test.cc
```diff
@@ -61,6 +64,18 @@ TEST(FileUtil, Stem) {
   EXPECT_EQ(file::Stem("/foo/bar"), "/foo/bar");
 }
 
+TEST(FileUtil, JoinPath) {
+  EXPECT_EQ(file::JoinPath("foo", ""), "foo/");
+  EXPECT_EQ(file::JoinPath("", "bar"), "/bar");
```
tgorochowik, line 69: Could we also add
`EXPECT_EQ(file::JoinPath("/", "bar"), "/bar");`

common/util/file_util.cc
hzeller: Yes. I'll work on using `std::filesystem` soon, so this is mostly a reminder.

common/util/file_util.cc
hzeller: Done.

common/util/file_util.cc
hzeller: I solved it by adding a test to make sure that this corner-case is covered, since anyone wondering about the implementation will look there first.

common/util/file_util_test.cc
hzeller: Done, also with a few more variants.

PR #784, opened 2021-04-21
Updated glog to latest glog library

corco:
Split commit from #781. Current glog version does not compile on Windows, the new one does.


hzeller (approved):
As soon as the CI turns green, I'll merge.

PR #785, opened 2021-04-21
Removed unused process.cc file/test

corco:
Split from #781. The file does not compile on Windows. Since it is unused, better to clean up and remove!


hzeller (approved):
LGTM.
As soon as CI is through and happy, I'll merge.

PR #789, opened 2021-04-23
Reformulate unix-specific file operations with std::filesystem

hzeller:
Issues #307 #765

### DIFF
--- a/.github/workflows/verible-ci.yml
+++ b/.github/workflows/verible-ci.yml
@@ -155,7 +155,7 @@ jobs:
     runs-on: ubuntu-20.04
     # Github actions resources are limited; don't build artifacts for all
     # platforms on every pull request, only on push.
-    if: ${{github.event_name == 'push'}}
+    #if: ${{github.event_name == 'push'}}
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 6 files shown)


mithro (approved):

mithro:
If CI says yes, go ahead.

hzeller:
I might need some help with the `releasing/` stuff @mithro : the c++17 `<filesystem>` header is only available with gcc-8 (I think) onwards.
Can these compilers easily be installed/used in the release docker files ?

hzeller:
CC: @corco once the issue with releasing on older platforms is resolved, this will bring a `file_util` that compile on windows (notice the windows CI passing, which now also includes the `file_util_test`).

So if you want to get unblocked until this pull request can be merged, feel free to add the changes to your local working branch for a new round of finding the next issues to tackle.

corco:
@mithro @hzeller In the spirit of unblocking #307, I suggest to keep the old file separate and use it on non-Windows platform. Getting rid of the old file should be a separate issue (unless the fix is quick & easy of course). As much as I don't like having platform-specific files, it should be the only one to support Windows.

Something like this in the BUILD file:

```bazel
srcs = select({
        "@platforms//os:windows": ["file_util_c++17.cc"],
        "//conditions:default": ["file_util_unix.cc"],
    }),
```

hzeller:
Another thing I considered was maybe using some compatibility header such as https://github.com/gulrak/filesystem

I plan to make it work to the end of this day whatever it takes. If in the next few hours it turns out the releases are a bit more work, I'll probably submit your suggestion @corco , then we have time to get it nice later.

hzeller:
Alright @corco this is merged. So hopefully it helps in further unlocking more Windows tests.

Issue #247, opened 2020-03-27
Flag declarations that shadow other declared names


fangism:
Every time a symbol is declared, we should check whether or not a pre-existing declaration in scope bears the same name.  In some instances, this is illegal in Verilog, other instances, it is permitted.  We would like to catch all cases regardless.

```systemverilog
module foo;
...
endmodule

module bar;
// 'foo' here refers to above module type
wire foo;
// 'foo' may now resolve to something else, or perhaps be context-dependent.
endmodule
```

```systemverilog
function f;
  for (int i=0; ...; ) begin
    for (int i=k; ...; ) begin
      // 'i' in here refers to the inner loop
    end
  end
endfunction
```

This was inspired in part by the `-Wshadow` diagnostics of gcc/clang.

fangism:
b/69064083

PR #790, opened 2021-04-23
Variable shadowing rule

pawelsag:
Fixing https://github.com/google/verible/issues/247  
I believe I've eliminated most of the annoying cases that shouldn't be detected.


verilog/analysis/checkers/instance_shadow_rule.cc
```diff
@@ -0,0 +1,134 @@
...
+  return ctx.IsInside(NodeEnum::kSeqBlock) ||
+         ctx.IsInside(NodeEnum::kGenvarDeclaration) ||
+         ctx.IsInside(NodeEnum::kReference);
+}
+
+void InstanceShadowRule::HandleSymbol(const verible::Symbol& symbol,
+                                      const SyntaxTreeContext& context) {
+  verible::matcher::BoundSymbolManager manager;
+  bool omit_node = false;
+  if (InstanceShadowMatcher().Matches(symbol, &manager)) {
```
hzeller, line 74: Like in suggestions in other reviews, make use of early-out if possible to reduce reading complexity.

```c++
if (!InstanceShadowMatcher().Matches(symbol, &manager))
  return;
```

verilog/analysis/checkers/instance_shadow_rule.cc
```diff
@@ -0,0 +1,134 @@
...
+
+    const auto& rcontext = reversed_view(context);
+    const auto& rdirectParent = *std::next(rcontext.begin());
+    // we are looking for the potential labels that might overlap the considered
+    // declaration. We are searching all the labels within the visible scope
+    // until we find the node or we reach the top of the scope
+    for (auto rc = rcontext.begin(); rc != rcontext.end(); rc++) {
+      const auto& node = *rc;
+
+      for (const auto& ch : node->children()) {
```
hzeller, line 88: If we have space to write out the full type, we should use that instead of `auto`.

The abbreviation `ch` sounds like 'char'. Let's call it `child`.

verilog/analysis/default_rules.h
```diff
@@ -24,6 +24,7 @@ namespace analysis {
 // LINT.IfChange
 constexpr const char* kDefaultRuleSet[] = {
     "invalid-system-task-function",
+    "instance-shadowing",
```
hzeller, line 27: Like all new rules first comment this out (see list at the bottom of this array). Once implemented, we can inform everyone on the verible mailing list that this will be enabled and then go from there.
That way, people can adapt to potentially newly triggering lint rules.

verilog/tools/lint/testdata/shadow_parameter.sv
```diff
@@ -0,0 +1,8 @@
+class jumbo_packet;
```
hzeller, line 1: Since the example is called shadow parameter: should we have a comment-documented shadowed parameter ?

hzeller (commented):
First round of comments.

verilog/analysis/checkers/instance_shadow_rule.cc
```diff
@@ -0,0 +1,135 @@
...
+        }
+        // if considered label is the last node
+        // this is the ending node with label
+        if (rdirectParent->children().back().get() ==
+            node->children().back().get()) {
+          return;
+        }
+
+        if (overlappingLabel.get().text() == label.get().text()) {
+          violations_.insert(LintViolation(symbol, kMessage, context));
```
hzeller, line 119: I think we should make sure to provide some actionable information. In particular in larger designs it might be hard to find what has been shadowed. I am thinking of an error message of the form
```
systemverilog-file.sv:12:15: Symbol `foo` is shadowing symbol `foo` defined at systemverilog-file:3:7:
```

hzeller (commented):
I think this is mostly LGTM. We do have the information about the shadowed symbol, so it should also be simple to make the error message a bit more useful and show the exact location of where that has been defined.

verilog/analysis/checkers/instance_shadow_rule.cc
pawelsag: I have added the new mechanism for injecting the location of the tokens into the vulnerability description.
What do you think about this solution?
This allows us to put some additional tokens and mark the location in the text,
then substitute the markers `\@` with tokens location.

common/analysis/lint_rule_status.h
```diff
@@ -221,6 +240,15 @@ class LintStatusFormatter {
                        absl::string_view url,
                        absl::string_view rule_name) const;
 
+  // Substitute the markers `@` with tokens location
+  // this allows us to create custom reason msg
+  // with different token location that are related to found
+  // vulnerable token
+  std::string ReplaceWithHelperTokens(const std::vector<TokenInfo>& tokens,
```
hzeller, line 247: I'd probably call this `FormatWithRelatedTokens()` or similar to emphasize that this is a string formatting functionality.

Also, this needs documentation about the other parameters (And `reason` should probably be `message` to emphasize the formatting part).

The fact that backslash escapes the at-symbol `\@` is missing in the documentation.

Finally, since this is not touching anything in the struct, I'd make it a static method, which makes it easier to identify as a helper function.

common/analysis/lint_rule_status.cc
```diff
@@ -105,6 +108,36 @@ struct LintViolationWithStatus {
...
 };
 
+std::string LintStatusFormatter::ReplaceWithHelperTokens(
+    const std::vector<verible::TokenInfo>& tokens, absl::string_view reason,
+    absl::string_view path, absl::string_view base) const {
+  if (tokens.empty()) {
+    return std::string(reason);
+  }
+
+  size_t end_pos = reason.find("@", 0), beg_pos = 0;
```
hzeller, line 118: make these two separate lines; it is easy to overlook the beg_pos declaration. Also, in that case you can use `beg_pos` already

```c++
size_t beg_pos = 0;
size_t end_pos = reason.find("@", beg_pos);
```

common/analysis/lint_rule_status.cc
```diff
@@ -105,6 +108,36 @@ struct LintViolationWithStatus {
...
+  std::ostringstream s;
+  for (const auto& token : tokens) {
+    if (end_pos == std::string_view::npos) {
+      s << reason.substr(beg_pos);
+      break;
+    }
+
+    if (!(end_pos != 0 && reason[end_pos - 1] == '\\')) {
+      s << reason.substr(beg_pos, end_pos - beg_pos);
+      s << path << ":";
```
hzeller, line 128: We should probably write in the header documentation of this function that the tokens all must come from the same filename (right now, this is always the case we look file by file, but we might parse multiple files and thus error messages that span over multiple files later)

verilog/analysis/checkers/instance_shadow_rule.cc
```diff
@@ -0,0 +1,140 @@
...
+  if (labels.empty()) return;
+  // if the considered symbol name is not a declaration
+  if (context.IsInside(NodeEnum::kReference)) return;
+
+  const auto& rcontext = reversed_view(context);
+  const auto& rdirectParent = *std::next(rcontext.begin());
+  // we are looking for the potential labels that might overlap the considered
+  // declaration. We are searching all the labels within the visible scope
+  // until we find the node or we reach the top of the scope
+  for (auto rc = rcontext.begin(); rc != rcontext.end(); rc++) {
```
hzeller, line 87: Might be interesting to measure what performance impact going through nested loops for each symbol has in practice on a larger project.

hzeller (approved):
In general, I like the idea of the `@`-formatting.
Naming and documentation of a function needs improving, otherwise LGTM.

hzeller:
Looking back, why was this not merged back then ? It looks like there were just a few comments left.

hzeller:
Should we get this finalized @pawelsag @kgugala ?

common/analysis/lint_rule_status.h
jbylicki: I have applied the naming changes, but there was a problem with making the method static. Well, it requires the access to `line_column_map` which is not static, causing a problem.

common/analysis/lint_rule_status.cc
jbylicki: Done

verilog/analysis/checkers/instance_shadow_rule_test.cc
```diff
@@ -0,0 +1,142 @@
...
+       ";\n",
+       "endfunction\n"},
+      {"function automatic int foo (input bit in, input bit ",
+       {kToken, "in"},
+       ");\n",
+       "bit out;\n",
+       "endfunction\n"},
+  };
+  RunLintTestCases<VerilogAnalyzer, InstanceShadowRule>(
+      kInstanceShadowingTestCases);
```
hzeller, line 137: Can we also test that the correct variable location of the previous definition is mentioned in the output message ?

hzeller (commented):
LGTM, but I think we also should test that we actually get the desired mentioning of the previous reference in the shadow parameter rule.

verilog/analysis/checkers/instance_shadow_rule_test.cc
jbylicki: Done

verilog/analysis/checkers/instance_shadow_rule.cc
```diff
@@ -0,0 +1,139 @@
...
+}
+
+void InstanceShadowRule::HandleSymbol(const verible::Symbol& symbol,
+                                      const SyntaxTreeContext& context) {
+  verible::matcher::BoundSymbolManager manager;
+  bool omit_node = false;
+  if (!InstanceShadowMatcher().Matches(symbol, &manager)) {
+    return;
+  }
+  const auto& labels = FindAllSymbolIdentifierLeafs(symbol);
```
mglb, line 78: Not a ref
```suggestion
  const auto labels = FindAllSymbolIdentifierLeafs(symbol);
```

mglb (changes requested):

mglb (approved):

hzeller:
Rebased in #2133 (with a couple of fixes to compile with head) where final review/merge should happen.
Closing this one.

PR #792, opened 2021-04-23
Fixed common/formatting/align.cc on Windows

corco:
Please verify if the `.base()` has a purpose. On Windows, compilation failed with:

```
common/formatting/align.cc(598,42): error: no member named 'base' in 'std::_Vector_iterator<std::_Vector_val<std::_Simple_types<verible::VectorTree<verible::UnwrappedLine>>>>'
    const int indentation = rows.front().base()->Value().IndentationSpaces();
                            ~~~~~~~~~~~~ ^
```


hzeller:
Interesting, I thought `base()` was only defined for some sort of reverse iterator. Paging @fangism - do you know of there was a reason to use `base()` in the past here, and what its intention was ?

hzeller:
Thanks, merged.

PR #796, opened 2021-04-23
Compiling on Windows

corco:
Verible now compiles on Windows! Work is not over, as not all test succeeds yet. The changes to make it compile is minimal, only replacing `unistd.h` with the appropriate Windows equivalent.

Making testing works, which is very much a WIP, revolves around 2 cookie-cutter fix that needs review:

1. On Windows, bazel doesn't symlink runfiles by default (you must have a recent Windows and enable developer mode for it) and instead create a `MANIFEST` file with the requested runfiles (refer to this [issue](https://github.com/bazelbuild/bazel/issues/8843)). Using the `--enable_runfiles` flag caused other issues with *very* long path that made some test fail. I couldn't fix that long path issue, it would require renaming some tests that have excessively long names.

    Instead, I modified the `sh_test` targets to include a bazel script allowing to get the fullpath of the desired resource, and modified the script accordingly.

2. The Verible tools output CRLF instead of the test-expected LF. I decided to ignore line endings (thus making test pass) for now. I suggest to open an issue to discuss how to handle line endings and modify/add tests accordingly at a later stage. IMHO, using CRLF on Windows is fine, `git` by default converts line endings with the `core.autocrlf` settings and there is `dos2unix` that can be added to scripts if it is an issue. I think the issue merits more thought, but shouldn't be a blocker to proceed with Windows support.

### DIFF
--- a/.github/workflows/windows-compile.yml
+++ b/.github/workflows/windows-compile.yml
@@ -39,5 +39,6 @@ jobs:
                     //verilog/formatting/... `
                     //verilog/parser/... `
                     //verilog/preprocessor/... `
+                    //verilog/tools/diff/... `
                     //verilog/tools/kythe/... `
                     //verilog/transform/...
--- a/.github/workflows/windows-compile.yml
+++ b/.github/workflows/windows-compile.yml
@@ -21,13 +21,9 @@ jobs:
         fetch-depth: 0
     - name: Install dependencies
       run: choco install llvm
-    # Only check flex/bison targets at the moment, since Verible cannot be
-    # fully compiled on Windows (yet!)
-    - name: Build flex/bison files
+    - name: Build Verible
       run: |
-        bazel build //common/analysis:command_file_lex
-        bazel build //verilog/parser:verilog_y
-        bazel build //verilog/parser:verilog_lex
+        bazel build -c opt //...
     # Windows support is in progress, only run "working" tests
     - name: Run known-to-work tests
       run: |
--- a/.github/workflows/windows-compile.yml
+++ b/.github/workflows/windows-compile.yml
@@ -32,49 +32,7 @@ jobs:
     - name: Run known-to-work tests
       run: |
         bazel test  --test_output=errors `
-                    //common/util:file_util_test `
-                    //common/analysis:command_file_lexer_test `
-                    //common/analysis:file_analyzer_test `
-                    //common/analysis:line_linter_test `
-                    //common/analysis:lint_rule_status_test `
-                    //common/analysis:linter_test_utils_test `
-                    //common/analysis:syntax_tree_linter_test `
-                    //common/analysis:syntax_tree_search_test `
-                    //common/analysis:syntax_tree_search_test_utils_test `
-                    //common/analysis:text_structure_linter_test `
-                    //common/analysis:token_stream_linter_test `
-                    //common/analysis/matcher:core_matchers_test `
-                    //common/analysis/matcher:descent_path_test `
-                    //common/analysis/matcher:matcher_builders_test `
-                    //common/analysis/matcher:matcher_test `
-                    //common/formatting/... `
-                    //common/lexer/... `
-                    //common/parser/... `
-                    //common/strings/... `
-                    //common/text/... `
-                    //common/tools:patch_tool_test
-                    //common/util:algorithm_test `
-                    //common/util:auto_iterator_test `
-                    //common/util:auto_pop_stack_test `
-                    //common/util:bijective_map_test `
-                    //common/util:container_iterator_range_test `
-                    //common/util:enum_flags_test `
-                    //common/util:expandable_tree_view_test `
-                    //common/util:forward_test `
-                    //common/util:interval_map_test `
-                    //common/util:interval_set_test `
-                    //common/util:interval_test `
-                    //common/util:iterator_adaptors_test `
-                    //common/util:iterator_range_test `
-                    //common/util:map_tree_test `
-                    //common/util:range_test `
-                    //common/util:spacer_test `
-                    //common/util:subcommand_test `
-                    //common/util:top_n_test `
-                    //common/util:type_traits_test `
-                    //common/util:value_saver_test `
-                    //common/util:vector_tree_test `
-                    //common/util:with_reason_test `
+                    //common/... `
                     //external_libs:editscript_test `
                     //verilog/analysis:descriptions_test `
                     //verilog/analysis:extractors_test `
--- a/.github/workflows/windows-compile.yml
+++ b/.github/workflows/windows-compile.yml
@@ -50,20 +50,7 @@ jobs:
                     //common/formatting/... `
                     //common/lexer/... `
                     //common/parser/... `
-                    //common/strings:comment_utils_test `
-                    //common/strings:compare_test `
-                    //common/strings:diff_test `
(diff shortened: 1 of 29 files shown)


common/tools/patch_tool_test.sh
```diff
@@ -26,7 +40,12 @@ readonly MY_EXPECT_FILE
   echo "Expecting 1 positional argument, verible-patch-tool path."
   exit 1
 }
-patch_tool="$1"
+
+# On Windows, runfiles are not symlinked (unless --enable_runfiles is active,
+# which causes other issues on bazel 3.7.0), so use rlocation to find the path
+# to the executable. On Unix, you could simply use $1, but using rlocation
+# works too
```
hzeller, line 47: Ok to understand the `$(location :verible-patch-tool)` we use in the BUILD rule only outputs the relative path from the start of toplevel `bazel-bin`, but since that is in the Windows case not in the PATH, we need to use `rlocation`  to resolve ?

(Does it make a difference with Bazel 4.0 ?)

common/tools/patch_tool_test.sh
```diff
@@ -13,6 +13,20 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# --- begin runfiles.bash initialization v2 ---
+# Copy-pasted from the Bazel Bash runfiles library v2.
```
hzeller, line 17: Since we need this snippet in all tests can we put that in one place (maybe `bazel:runfile_support`) somewhere and then just source in all the tests where we need it ?

verilog/analysis/symbol_table_test.cc
```diff
@@ -7597,7 +7597,12 @@ TEST(BuildSymbolTableTest, ModuleInstancesFromProjectMissingFile) {
   symbol_table.BuildSingleTranslationUnit("file/not/found.txt",
                                           &build_diagnostics);
   ASSERT_FALSE(build_diagnostics.empty());
+#ifdef _WIN32
```
hzeller, line 7600: If we change this code to compare the status code instead of the message it should hopefully be platform independent:

```
EXPECT_EQ(build_diagnostics.front().code(), absl::StatusCode::kNotFound);
```

verilog/tools/kythe/kythe_proto_output.cc
```diff
@@ -12,15 +12,19 @@
...
+#include <stdio.h>
+#define STDOUT_FILENO _fileno(stdout)
+#endif
 
 #include "google/protobuf/io/coded_stream.h"
 #include "google/protobuf/io/zero_copy_stream_impl.h"
 #include "third_party/proto/kythe/storage.pb.h"
 #include "verilog/tools/kythe/kythe_facts.h"
 #include "verilog/tools/kythe/kythe_facts_extractor.h"
+#include "verilog/tools/kythe/kythe_proto_output.h"
```
hzeller, line 27: We place the header of the particular implementation always at the top (because then it is easy to see if there are some hidden dependencies), so the original location around line 14 was correct.

hzeller (commented):
Very exciting!

w.r.t. to th CRLF I agree that we should not worry about it right now. It is expected to get them in the Windows environment.

I think mostly I'd like to see the repetetive runfile handling snippet put in one place somewhere so that we just need a one-liner to source in our tests. The code in question should probalby have some comments what it does and why or a link to some external page that explains the things it does.

(Note, we recently started using bazel 4.0.0 by default (vs. 3.7.0) if that makes a difference or could make things simpler)

.github/workflows/windows-compile.yml
```diff
@@ -21,135 +21,21 @@ jobs:
...
-                    //verilog/analysis/checkers:signal_name_style_rule_test `
-                    //verilog/analysis/checkers:struct_union_name_style_rule_test `
-                    //verilog/analysis/checkers:suggest_parentheses_rule_test `
-                    //verilog/analysis/checkers:token_stream_lint_rule_test `
-                    //verilog/analysis/checkers:undersized_binary_literal_rule_test `
-                    //verilog/analysis/checkers:unpacked_dimensions_rule_test `
-                    //verilog/analysis/checkers:uvm_macro_semicolon_rule_test `
-                    //verilog/analysis/checkers:v2001_generate_begin_rule_test `
-                    //verilog/analysis/checkers:void_cast_rule_test `
+                    //common/... `
```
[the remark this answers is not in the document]
corco: All files *compiles* but many test still fails. I expect most of the failing tests can be fixed with the same 2 ideas (fix runfiles and ignore crlf). I will continue fixing other files, but I want to make sure this is the right approach.

common/tools/patch_tool_test.sh
corco: I'm not a bazel expert, but I think bazel creates a folder to run the script (kinda like a sandbox) and make `deps` files available using symlinks. So the `$(location :verible-patch-tool)` is the location of the symlinked script from the sandbox folder.

Except on Windows it doesn't create the symlink, but instead create a `MANIFEST` file (also available on Unix) that has a mapping of all available runfiles to its fullpath. `rlocation` is a convenience function that look inside the `MANIFEST` file for a relative runfile path and returns the fullpath.

And no, bazel 4.0 doesn't make a difference.

common/tools/patch_tool_test.sh
corco: The problem is how to find that snippet on Windows? The answer is `rlocation`, which the snippet import... We could implement our own version, but that is kinda what the snippet does already. The snippet is the way [Bazel recommends](https://github.com/bazelbuild/bazel/blob/9ad5aeddb24b8c6ca45ec3fbe07b00d80cd8fb59/tools/bash/runfiles/runfiles.bash) it AFAIK.

There is probably a way using a genrule that would wrap the script to run around some initialization. I will give it a try.

verilog/tools/kythe/kythe_proto_output.cc
corco: The formatting was [suggested by the CI server](https://github.com/google/verible/runs/2423970246#step:5:32), am I understanding something the wrong way?

hzeller: It is important to have it as the very first include, then the formatter recognizes it. I suppose it was shifting under the the include changes you made, so it was not recognized as such, and thus suggested to be sorted alphabetically with the rest.

So if you move it to the top as first include, it should work.

common/tools/patch_tool.cc
```diff
@@ -12,7 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#ifndef _WIN32
```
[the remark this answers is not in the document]
corco: There is also the `#define STDOUT_FILENO _fileno(stdout)` from `kythe_proto_output.cc`.

I vote for later... I am more confortable in SystemVerilog or embedded C after all!

common/tools/patch_tool_test.sh
hzeller: ha, yes, it wouldn't be found, good point.
But if we find something to reduce the repetetive boilderplate would probably good.

corco: I added a convenience `sh_test` wrapper that loads the bash runfiles library before calling the test. That reduces boilerplate code significantly.

hzeller (approved):
LGTM
Cool, having the boilerplate at one place is now much nicer!

PR #798, opened 2021-04-26
Provide IsInteractiveTerminalSession().

hzeller:
This encapsulates the isatty() call and the different
includes needed on Unix and Windows.

Issues #307


hzeller:
CC: @corco

PR #799, opened 2021-04-27
Expanding Windows tests running in CI 

hzeller:
First step: run all to see which pass.


hzeller:
FYI @corco

corco:
Cool! I hit the wall of linter, but it's nice to see that everything else passes.

The problem with the linter is that bazel rules marked `test = true` (like the one generated in the linter) needs to create an executable. The rule wright now creates a bash script, which is not executable on Windows.

I started working on a fix, which mostly replaces the current rule with `sh_test_with_runfiles_lib`

Issue #445, opened 2020-09-11
Wrong formatting of constraints blocks


msfschaffner:
Verible erroneously reformats the following blocks into one-liners.

before:
```systemverilog
constraint c_iv {
    if (fixed_iv_en) {
      aes_iv == fixed_iv
    };
  }

  constraint c_operation {
     if (fixed_operation_en) {
         aes_operation == fixed_operation
     };
  }
```

after:
```systemverilog
constraint c_iv {if (fixed_iv_en) {aes_iv == fixed_iv};}

constraint c_operation {if (fixed_operation_en) {aes_operation == fixed_operation};}
```

msfschaffner:
b/168312418

vasantteja:
Can I assign this to myself?

fangism:
Yes, thank you for volunteering!

Tips: examine `verible-verilog-syntax --printtree` to see the internal syntax tree structure.
`verible-verilog-format --show_token_partition_tree` (optionally `--show_inter_token_info`) will show you the formatter internal representation as a hierarchical tree of token ranges.

Feel free to ask any questions to help get you going.

fangism:
@vasantteja Checking in, need any help or have any questions?

fangism:
@vasantteja Checking in, need any help or have any questions?

vasantteja:
@fangism sorry, I was traveling and I couldn't take a look at this. I will reach out to you if I have any doubts. Thanks for the patience.

ldalek-antmicro:
Hi @msfschaffner,
Just for clarification: all constraints blocks should be expanded?

For example I've found in an Ibex source code such line:
```
constraint only_vec_instr_c {soft only_vec_instr == 0;}
```
here: https://github.com/lowRISC/opentitan/blob/master/hw/vendor/lowrisc_ibex/vendor/google_riscv-dv/src/riscv_vector_cfg.sv#L32

and I'm not sure how to procceed.

Current PR expands all blocks except some simple one like that above:
https://github.com/google/verible/pull/706/files#diff-631e76d6700a6682c3bdc7f0aad5aa02a4f117aa2e8ae3e6b96c2ddaf34c92f1R9164

msfschaffner:
Hey @ldalek-antmicro,

How about we standardize on something that only expands if the expression inside the outermost `{}` brackets contains more `{}` blocks or does not fit onto one single line?

ldalek-antmicro:
Thanks for an answer @msfschaffner 

[...]

PR #802, opened 2021-04-28
Fix indentation on named arguments of functions

### LINKED ISSUE #335: Missing indentation on named arguments of functions
When formatting function calls with 'named arguments' (not sure what they are called) the formatter does not indent the arguments in the same way as for unnamed arguments.

Example:
```systemverilog
  always_comb begin
    value = function_name(
    .long_parameter(8'hA),
    .another_parameter(signal_123),
    .parameter_three(foobar)
    );
  end
```

For unnamed arguments (if the line were long enough) this would be:
```systemverilog
  always_comb begin
    value = function_name(8'hA, signal_123,
                          foobar);
  end
```

Adding some indentation before the arguments seems more readable to me:
```systemverilog
  always_comb begin
    value = function_name(
      .long_parameter(8'hA),
      .another_parameter(signal_123),
      .parameter_three(foobar)
    );
  end
```

wtatarski:
Fixes #335


tgorochowik (commented):
LGTM in general.

But I think we should add a test that mixes named and positional arguments (note that positional arguments must come fist, see 13.5.4 in LRM)

wtatarski:
Added test with mixed named and positional arguments.

tgorochowik (approved):

hzeller (commented):
Generally looks good though I see that there might be some corner cases, which you should ask about and discuss on #335 so that some decisions of how to behave in each of these cases can be made there and referred to later in case there are questions. Pull in some of the OpenTitan folks who can give good guidance (See #445 for some such discussion about corner cases).

Otherwise we might flip-flop between competing implementations later, which is probably better to clear out first.

wtatarski:
@hzeller 
I think that all cases which you mentioned are fixed now.

hzeller (approved):
LGTM

hzeller:
Please test this with OpenTitan formatting and gather potential feedback from that team in #335 so that we can potentially refine.

verilog/formatting/formatter.cc
```diff
@@ -379,6 +396,45 @@ static void DeterminePartitionExpansion(
         node_view.Expand();
       }
     }
+
```
hzeller, line 399: Is there a break missing here @wtatarski (or: ABSL_FALLTHROUGH_INTENDED) ?
(also a test should trigger if the uintended situation is happening)

Please add a follow-up CL to clarify this part of the code (and possibly add a test that makes sure that this is caught)

PR #805, opened 2021-05-01
Use std::filesystem & GCC 9

### LINKED ISSUE #765: Move to use C++17 and with it, embrace std::filesystem

corco:
Closes #765. Use `std::filesystem` on all platforms (and with it C++17).

To achieve this, I updated the compiler to GCC 9 on all platforms, which does not need an extra link flag to bring `std::filesystem`. It could be done with GCC 8 by adding the link flag `-lstdc++fs`. However, of all the platforms that need a "non default compiler", only CentOS 8 has gcc 8, so the cost to bring GCC 9 is small. Since GCC 9 is 2 years old, I opted for the newer compiler and leaving the compile flag as is.


hzeller:
Ah, very cool!

To test the new compiler settings in the releases, can you comment out the `f: ${{github.event_name == 'push'}}` around here:
https://github.com/google/verible/blob/master/.github/workflows/verible-ci.yml#L158

.. that way, the releases are also built during the pull request.

releasing/centos/8/compiler.dockerstage
```diff
@@ -1,6 +1,9 @@
-# Install C++ compiler
-RUN yum -y group install --nogpgcheck "Development Tools" || \
-    yum -y groupinstall --nogpgcheck "Development Tools"
+# Link libstdc++ statically so people don't have to install devtoolset-8
```
mithro, line 1:
```suggestion
# Link libstdc++ statically so people don't have to install devtoolset-9
```

mithro (approved):
LGTM from the CI side (please drop the temp+revert - use `git rebase -i` for an easy way).

@hzeller will need to review the C++ side.

hzeller:
c++ looks good.
One We should run some intermediate commit with the CI buliding the artifacts ( https://github.com/google/verible/pull/805#issuecomment-830489798 ).

Once that passes, re-enable `if: ${{github.event_name == 'push'}}` again.

corco:
> c++ looks good.
> One We should run some intermediate commit with the CI buliding the artifacts ( [#805 (comment)](https://github.com/google/verible/pull/805#issuecomment-830489798) ).
> 
> Once that passes, re-enable `if: ${{github.event_name == 'push'}}` again.

I did (and it passed), then removed that commit as suggested by [@mithro](https://github.com/google/verible/pull/805#pullrequestreview-649800796)

hzeller:
Oh, sorry I missed that. The github pull reqeust page didn't show that change, but I now found it in the email.

Another pull request also modified `.bazelrc` so you might need to do a quick rebase upstream/master of your branch.

hzeller:
Thanks, merged!

alessandroaimar:
On CentOS8, after installing gcc version 9.2.1 20191120 (Red Hat 9.2.1-2) (GCC) the build was not working until I modified in .bazelrc line 14 the following adding the -lstdc++fs flag

common:linux --cxxopt=-std=c++17 --cxxopt=-lstdc++fs --host_cxxopt=-std=c++17 --host_cxxopt=-lstdc++fs --client_env=BAZEL_CXXOPTS=-std=c++17

corco:
That option is not necessary on GCC-9. Are you sure the build used gcc-9 and not the default CentOS compiler?

If you installed GCC-9 using `yum install -y --nogpgcheck gcc-toolset-9-toolchain`, then you need to use `scl` to enable it.

I believe something like `scl enable gcc-toolset-9 bash` should provide a bash shell where gcc-9 is the default compiler, from which you can run `bazel build ...`

I will see to update the README

alessandroaimar:
Using `scl enable gcc-toolset-9 bash ` did not produce any effect. Is the build flow actually using the shell default compiler or is it creating a shell of its own?

corco:
Yes, the build flow uses the shell default compiler. You can compel bazel to use another compiler with environment variables `CC` and `CXX`, you can refer to bazel's documentation for that effect. I believe using `scl` is the recommended way on CentOS to use another toolchain though, I don't use CentOS for development however.

What do you mean `scl enable gcc-toolset-9 bash` did not produce any effect? This works for me:

```bash
docker run -it --rm centos:8
yum install -y gcc-c++ gcc-toolset-9
g++ --version               # 8.3.1
scl enable gcc-toolset-9 bash
g++ --version               # 9.2.1
```

From that shell I can build `verible` without issue, once git & bazel are installed of course.

PR #807, opened 2021-05-01
Use `common` for platform-specific bazel configs:

johanvdhaegen:
- Use `common` options instead of separate `build`, `run`, `test` command options.

- Prevents invalidation of bazel cache due to changed compiler options.
  Reduces compile time of consecutive bazel build, test, run commands.

- A single `build` option line per platform works also since `test` and
  `run` command options inherit options from `build`, according to
  https://docs.bazel.build/versions/master/guide.html#bazelrc-syntax-and-semantics

- Extends abseil compile fix (--features=-supports_dynamic_linker) from macos:test to macos:build and macos:run.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -13,20 +13,13 @@ build --enable_platform_specific_config
 
 # Use C++17 (bazel defaults to C++0x (pre-C++11))
 # TODO(b/171679296): Lower the requirement, go back to c++11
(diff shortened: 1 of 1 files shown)


hzeller (approved):

hzeller:
Thanks!

Issue #519, opened 2020-10-01
Override style-guide citation links and references


fangism:
**Summary**

Currently, style guide citation links are hard-coded into the rule implementations themselves.
In the future, users and rule-set maintainers may wish to redirect these to point to their own style guides.

**Proposal**

Have a way at run-time to re-map each rule's citation to a new link, for example loading such a map through a command-line option that reads an external configuration.

fangism:
b/169867142

pawelsag:
I thinking about loading the custom style guide citation from the JSON file. What do you think about this idea?

mithro:
I would suggest that we follow whatever other tools do in this area? What do clang-format and pylint do?

pawelsag:
Most of the tools are using just plain text configuration in a format
```
key1: value
key2: value
```
I will do the same for our config:
```
rule 1: style guide citation 1
rule 2: style guide citation 2
```

PR #817, opened 2021-05-15
Override style guide citations

pawelsag:
Fixes: Override style-guide citation links and references #519


verilog/analysis/verilog_linter.cc
```diff
@@ -332,10 +333,51 @@ absl::Status PrintRuleInfo(std::ostream* os,
   return absl::OkStatus();
 }
 
+static void AppendCitation(CustomCitationMap& citations,
+                           const absl::string_view& rule_raw) {
+  const size_t eq_pos = rule_raw.find(':');
```
hzeller, line 338: `eq_pos` sounds like it was re-used from some other code where we have an equal.... here this should probably be more `colon_pos`.

verilog/analysis/verilog_linter.cc
```diff
@@ -332,10 +333,51 @@ absl::Status PrintRuleInfo(std::ostream* os,
...
+  if (!eq_pos || eq_pos == absl::string_view::npos) return;
```
hzeller, line 339: I'd explicitly compare `eq_pos == 0` because that is what we mean. `!eq_pos` evaluates to the same thing, but doesn't reflect what we mean.

verilog/analysis/verilog_linter.cc
```diff
@@ -332,10 +333,51 @@ absl::Status PrintRuleInfo(std::ostream* os,
   return absl::OkStatus();
 }
 
+static void AppendCitation(CustomCitationMap& citations,
```
hzeller, line 336: Output parameters should [usually be at the end](https://google.github.io/styleguide/cppguide.html#Inputs_and_Outputs);

Also it is custom in this code-base to have output parameters pointers instead of references, but given that this is a static helper function used right afterwards, the confusion about it being a reference is probably not that big, so if you _really_ prefer the reference, keep it :)

verilog/analysis/verilog_linter.h
```diff
@@ -125,10 +127,18 @@ absl::Status PrintRuleInfo(std::ostream*,
...
+                                     const CustomCitationMap&);
 
 // Outputs the descriptions for every rule, formatted for markdown.
-void GetLintRuleDescriptionsMarkdown(std::ostream*);
+// When custom citations are delivered substitute chosen rule description
+// with citations specified by user
+void GetLintRuleDescriptionsMarkdown(std::ostream*, const CustomCitationMap&);
+
+// Parse the file with custom citations
+CustomCitationMap ParseCitations(absl::string_view);
```
hzeller, line 141: The parameter should have a name and describe what it is. E.g. it is not quite clear from reading this if the parameter is the filename or the file content. It 'sounds' like the filename, but looking at the implementation, it is actually the content.
So naming it and adding things to the function header comment will help.

verilog/analysis/verilog_linter_test.cc
```diff
@@ -383,25 +383,69 @@ TEST_F(VerilogLinterTest, MultiByteUTF8CharactersAreOnlyCountedOnce) {
   EXPECT_EQ(diagnostics.second, "");
 }
 
+TEST(VerilogLinterDocumentationTest, ParseCustomCitations) {
+  absl::string_view citations_raw =
+      "struct-union-name-style:one line citation is ok\n"
+      "signal-name-style:multi line citation\\\n"
```
hzeller, line 389: More test suggestions (and possibly code adaption needs)
  - what if there is a space before and/or after the colon: will it be properly stripped ?
  - what if we have \r\n as newlines (verible also runs on windows, so we need to expect that)

hzeller (commented):
First quick round of comments.

verilog/analysis/verilog_linter.cc
```diff
@@ -332,10 +333,51 @@ absl::Status PrintRuleInfo(std::ostream* os,
...
-                                     absl::string_view flag_value) {
+                                     absl::string_view flag_value,
+                                     const CustomCitationMap& citations) {
   // Set up the map.
   auto rule_map = analysis::GetAllRuleDescriptionsHelpFlag();
+  if (!citations.empty()) {
+    for (const auto& [rule_id, citation] : citations) {
+      auto rule = rule_map.find(rule_id);
+      if (rule != rule_map.end()) {
+        // we can move here, cause process
```
hzeller, line 375: I'd just plainly copy it. There is no good reason to save a few instructions and a string-copy at the expense of having a justifying comment clarifying that this will all be good.
(Also moving from a const std::string, how does that work ?)

verilog/analysis/verilog_linter.cc
```diff
@@ -332,10 +333,51 @@ absl::Status PrintRuleInfo(std::ostream* os,
...
+  return citations;
+}
+
 void GetLintRuleDescriptionsHelpFlag(std::ostream* os,
-                                     absl::string_view flag_value) {
+                                     absl::string_view flag_value,
+                                     const CustomCitationMap& citations) {
   // Set up the map.
   auto rule_map = analysis::GetAllRuleDescriptionsHelpFlag();
+  if (!citations.empty()) {
```
hzeller, line 371: if citations is empty, then the loop will not do anything, so no real need to wrap this in an if (and thus just making it more complex with an added indentation level)

verilog/analysis/verilog_linter.cc
```diff
@@ -356,12 +398,24 @@ void GetLintRuleDescriptionsHelpFlag(std::ostream* os,
...
 
-void GetLintRuleDescriptionsMarkdown(std::ostream* os) {
+void GetLintRuleDescriptionsMarkdown(std::ostream* os,
+                                     const CustomCitationMap& citations) {
   auto rule_map = analysis::GetAllRuleDescriptionsMarkdown();
   for (const auto& rule_id : analysis::kDefaultRuleSet) {
     rule_map[rule_id].default_enabled = true;
   }
 
+  if (!citations.empty()) {
```
hzeller, line 408: Same here with the loop: if 'citations' is empty, the loop will finish immediately anyway, so no extra wrapping in !empty() needed.

verilog/analysis/verilog_linter.cc
```diff
@@ -356,12 +398,24 @@ void GetLintRuleDescriptionsHelpFlag(std::ostream* os,
...
+    for (const auto& [rule_id, citation] : citations) {
+      auto rule = rule_map.find(rule_id);
+      if (rule != rule_map.end()) {
+        // we can move here, cause process
+        // exits after this function
+        rule->second.description = std::move(citation);
```
hzeller, line 414: Same std::move avoidance argument as above.

verilog/analysis/verilog_linter.cc
pawelsag: At first, I decided to use the equal sign, but in the end, I changed my mind to colon and forgot about changing the variable names.

PR #819, opened 2021-05-17
Align parameters in exported/imported function

### LINKED ISSUE #433: Align function/task formal parameters
Example from: https://github.com/msfschaffner/opentitan/pull/7/files/2bf9354bef93019b3778564ea746aa51584e8d2d#diff-f7facda2840d21c08bab1b5280a8d485

Desirable output:

```systemverilog
  import "DPI-C" context function void c_dpi_aes_crypt_message(
    input  bit              impl_i,
    input  bit              op_i,
    input  bit        [5:0] mode_i,
    input  bit  [3:0][31:0] iv_i,
    input  bit        [2:0] key_len_i,
    input  bit  [7:0][31:0] key_i,
    input  bit        [7:0] data_i[],
    output bit        [7:0] data_o[]
  );
```

However, I sense that forcing parameters onto their own line may not always be desirable.  We'll need to come up with a way to sense the author's intent.

wtatarski:
Fixes #433

### DIFF
--- a/verilog/CST/DPI.h
+++ b/verilog/CST/DPI.h
@@ -57,10 +57,10 @@ verible::SymbolPtr MakeDPIImport(T0&& keyword, T1&& spec, T2&& property,
 // Partial specialization provided as a workaround to passing nullptr
 // in positions 3 and 4 (optional symbols).  Compiler is not guaranteed
 // to deduce to that some paths are not reachble/applicable.
-template <typename T0, typename T1, typename T2, typename T3>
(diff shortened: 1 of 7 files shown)


tgorochowik (approved):
LGTM in general. One question though - could we add the configuration option mentioned in the original issue? (I think it's okay to add that separately, but let's not forget about it)

hzeller (approved):
LGTM

w.r.t. the configurability of alignment of array dimensions: I'd probably tie that together with whatever happens in module ports for consistency. But agreed, can happen later.

PR #821, opened 2021-05-19
Enable warning about switch/case fallthrough.

hzeller:


hzeller:
Hey Wojciech, this will be the change that enables switch/case fallthrough detection. Right now, the CI will fail with the missing break or intended fallthrough annotation in [the formatter.cc](https://github.com/google/verible/pull/802/files#diff-360ef463a0e0f2086036e2b52e34a09d8d141a4e8b3c86cfcd1bdfdddc2ff2a2R399).

So we can use this as a test to see if the CI can detect such cases in the future automatically.

PR #822, opened 2021-05-19
Fix missing break

wtatarski:
Related #821 #820 #802


wtatarski:
I think that it was only a missing break. Now it seems all tests passed.

tgorochowik (approved):

hzeller (approved):

PR #823, opened 2021-05-19
Add version/feature check to verible_verilog_syntax.py

mglb:
Makes `VeribleVerilogSyntax` constructor raise an exception when `verible-verilog-syntax` executable version is too old. It checks the version printed by `--version` switch. It fallbacks to checking `--helpfull` output for `--export_json` string when version string is not available.

The executable's version is additionally stored in the `version` attribute.

Depends on https://github.com/google/verible/pull/832 to pass Windows build in CI.

### DIFF
--- a/verilog/tools/syntax/export_json_examples/verible_verilog_syntax.py
+++ b/verilog/tools/syntax/export_json_examples/verible_verilog_syntax.py
@@ -387,8 +387,8 @@ class VeribleVerilogSyntax:
     if isinstance(executable, str):
       executable = [executable]
 
-    proc = subprocess.run([*executable, "--version"], capture_output=True,
-        encoding="utf-8")
+    proc = subprocess.run([*executable, "--version"], stdout=subprocess.PIPE,
+        stderr=subprocess.STDOUT, encoding="utf-8")
 
     ver_match = self._VERSION_RE.match(proc.stdout)
     if ver_match:
@@ -413,10 +413,10 @@ class VeribleVerilogSyntax:
 
     else:
       # Version not available; check help message for `--export_json` flag
-      proc = subprocess.run([*executable, "--helpfull"], capture_output=True,
-          encoding="utf-8")
+      proc = subprocess.run([*executable, "--helpfull"], stdout=subprocess.PIPE,
+          stderr=subprocess.STDOUT, encoding="utf-8")
 
-      if not self._EXPORT_JSON_RE.search(proc.stdout + proc.stderr):
+      if not self._EXPORT_JSON_RE.search(proc.stdout):
         raise Exception("Unsupported verible-verilog-syntax version."
             f"Minimum required version: {self._MIN_VERSION}.")
 
--- a/verilog/tools/syntax/export_json_examples/verible_verilog_syntax.py
+++ b/verilog/tools/syntax/export_json_examples/verible_verilog_syntax.py
@@ -17,7 +17,7 @@ import collections
 import json
 import re
 import subprocess
-from typing import Any, Callable, Dict, Iterable, List, Optional, Union
+from typing import Any, Callable, Dict, Iterable, List, Optional, Sequence, Union
 
 import anytree
 import dataclasses
@@ -374,7 +374,10 @@ class VeribleVerilogSyntax:
     executable: path to ``verible-verilog-syntax`` binary.
   """
 
-  def __init__(self, executable: str = "verible-verilog-syntax"):
+  def __init__(self, executable: Union[str, List[str]] = "verible-verilog-syntax"):
+    if isinstance(executable, str):
+      executable = [executable]
+
     self.executable = executable
 
   @staticmethod
@@ -436,7 +439,7 @@ class VeribleVerilogSyntax:
     if options["gen_rawtokens"]:
       args.append("-printrawtokens")
 
-    proc = subprocess.run([self.executable, *args , *paths],
+    proc = subprocess.run([*self.executable, *args , *paths],
         stdout=subprocess.PIPE,
         input=input_,
         encoding="utf-8",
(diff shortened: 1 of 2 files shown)


hzeller:
I think in general, we can't rely on the git outputs available, because if the binary has been compiled from a tar archive source, there is no git info.

So the only thing to look for is probably the built time. It doesn't really help to tell if something is too old as it only regards the build time not the commit time, but that is as close as it gets.

mithro:
We should be embedding the git info in any tar archives we produce?

hzeller:
Yes, the binary archives we generate should contain that compiled in.

But if someone just downloads a source archive [generated by github](https://github.com/google/verible/archive/refs/heads/master.tar.gz), then this will not be included @mithro.

mglb:
> I think in general, we can't rely on the git outputs available, because if the binary has been compiled from a tar archive source, there is no git info.

Yeah, I've added a fallback that just checks `--helpfull` output for `--export_json`.

mithro:
> Yes, the binary archives we generate should contain that compiled in.
> 
> But if someone just downloads a source archive [generated by github](https://github.com/google/verible/archive/refs/heads/master.tar.gz), then this will not be included @mithro.

I think we should either refuse to compile or output 'UNKNOWN' version from that?

verilog/tools/syntax/export_json_examples/verible_verilog_syntax.py
```diff
@@ -370,11 +370,58 @@ class VeribleVerilogSyntax:
...
+  _VERSION_RE = re.compile(r"^v([0-9]+)\.([0-9]+)-([0-9]+)-g[0-9a-fA-F]+$",
+      re.MULTILINE)
+  _EXPORT_JSON_RE = re.compile(r"(^|\s)--export_json\b")
+  _MIN_VERSION = "v0.0-925-gc1a388a"
+
+  def __init__(self, executable: Union[str, List[str]] = "verible-verilog-syntax"):
+    if isinstance(executable, str):
+      executable = [executable]
+
+    proc = subprocess.run([*executable, "--version"], capture_output=True,
```
hzeller, line 390: when I run this in an environment with a 3.6.13 Python, I get the following error
```
TypeError: __init__() got an unexpected keyword argument 'capture_output'
```

Is this a feature that is only available in newer Python versions ?

mglb: Yes, it has been introduced in 3.7. I changed it to `stdout`/`stderr` - should work with 3.6.

verilog/tools/syntax/export_json_examples/verible_verilog_syntax.py
```diff
@@ -370,11 +370,58 @@ class VeribleVerilogSyntax:
...
+
+      self.version = ver_match.group()
+
+    else:
+      # Version not available; check help message for `--export_json` flag
+      proc = subprocess.run([*executable, "--helpfull"], stdout=subprocess.PIPE,
+          stderr=subprocess.STDOUT, encoding="utf-8")
+
+      if not self._EXPORT_JSON_RE.search(proc.stdout):
+        raise Exception("Unsupported verible-verilog-syntax version."
```
hzeller, line 420: Maybe we should print what version we observed, so that it is easy to manually compare ?

hzeller (commented):
I am not convinced yet that this is a good solution; parsing version numbers assuming it coming from a particular formatted git string output from `--version` or even `--helpfull` will always be fragile (In our company internal repo we're compiling verible tools with a more detailed version number that represents the state in our mono-repo, so it will not match the regular expression at all ... and in facts fails the unit test here).

Taking a step back - what is the problem we try to solve ? I suppose we want to make sure that the json format emitted is compatible with the format being parsed in the Python script. So we should formulate it as such.

So to make this more explicit, I'd consider having version numbers in the json output, maybe some toplevel field `version: 1` or so which is then checked in the Python script.
Whenever the format changes in an incompatible way (which will be a deliberate choice), we should increase that when emitting the Json format (and adapt the script to understand the new format and compare against version==2). That way, we make versioning and checks a deliberate choice, but also have very discrete points when output changes (and hopefully this number will only increase slowly).

mglb:
> Taking a step back - what is the problem we try to solve ? I suppose we want to make sure that the json format emitted is compatible with the format being parsed in the Python script. So we should formulate it as such.

Oh, I see I didn't mention this. Originally it was a request from OpenTitan to check whether Verible supports JSON output or not. If it is supported, Verible should be used for parsing.
Eventually it has been done so that the script always tries to use Verible and fallbacks to non-Verible parsing when Verible failed for any reason.

Since the original reason for this PR is no longer valid, I think it can be closed for now. Should I create an issue to make this attempt easily discoverable in case the need for version checking appear in the future?


> So to make this more explicit, I'd consider having version numbers in the json output, maybe some toplevel field `version: 1` or so which is then checked in the Python script.
> Whenever the format changes in an incompatible way (which will be a deliberate choice), we should increase that when emitting the Json format (and adapt the script to understand the new format and compare against version==2). That way, we make versioning and checks a deliberate choice, but also have very discrete points when output changes (and hopefully this number will only increase slowly).

This is a good idea. It could be introduced when the JSON version changes to "2", so that "no version field" == "version 1".

hzeller:
> Since the original reason for this PR is no longer valid, I think it can be closed for now. Should I create an issue to make this attempt easily discoverable in case the need for version checking appear in the future?

Yes, I think an issue where we also collect the potential ideas (marking version numbers in json) sounds good.
 
> This is a good idea. It could be introduced when the JSON version changes to "2", so that "no version field" == "version 1".

Sounds good. After all, it will only the version > 1 json that introduces the new feature called version number :)

PR #827, opened 2021-05-20
Replaced build-version.sh with build-version.py

### LINKED ISSUE #826: WindowsBuild: `verible-* --version` doesn't print version
This happens at least in CI (example: https://github.com/google/verible/pull/823/checks?check_run_id=2620255630#step:4:658 ):
```
$ verible-verilog-syntax.exe --version
Built   2021-05-19T12:49:01Z
$
```

Output on linux for comparison:
```
$ verible-verilog-syntax --version
v0.0-1217-gb5506e2
Commit  2021-05-13
Built   2021-05-13T14:55:29Z
$
```
I guess https://github.com/google/verible/blob/master/bazel/build-version.sh fails to get info from git.

corco:
Fixes #826

The bazel option `--workspace_status_command` must point to an executable file, and on Windows bash shell script are not executable. Fortunately, Python scripts are executable and I opted to replace the current bash script with a Python one.

The alternative would be to have separate scripts for Linux/MacOS/Windows.


bazel/build-version.py
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env python3
+"""
```
hzeller, line 2: Make sure to add the Verible authors Apache license header (like in the previous bash script)

hzeller (commented):
Cool!
LGTM. Just make sure to add the license header.

PR #835, opened 2021-05-25
Fix README badge

tgorochowik:
The relative link does not work here:
https://google.github.io/verible/

So it has to be an absolute link to work both in the README and on github pages.


mithro (approved):
LGTM.

PR #839, opened 2021-05-31
formatter: fix formatting of bind statements

### LINKED ISSUE #432: Formatting of bind statements should behave the same as module/interface instantiations
The following bind statements
```systemverilog
module test;
  bind entropy_src tlul_assert #(
    .EndpointType("Device")
  ) tlul_assert_device (
    .clk_i,
    .rst_ni,
    .h2d  (tl_i),
    .d2h  (tl_o)
  );
endmodule
```
gets reformatted as 
```systemverilog
module test;
  bind entropy_src tlul_assert #(
      .EndpointType("Device")
  ) tlul_assert_device (.clk_i, .rst_ni, .h2d(tl_i), .d2h(tl_o));
endmodule
```
which is a weird choice. I.e., the formatter prefers one parameter assignment per line, and then puts all port connections onto a single line. I would expect it to leave the port connections one per line also...

wtatarski:
Fixes #432

### DIFF
--- a/verilog/formatting/formatter_test.cc
+++ b/verilog/formatting/formatter_test.cc
@@ -1605,7 +1605,53 @@ static constexpr FormatterTestCase kFormatterTestCases[] = {
      "      .in (ex_input_pins.in)\n"  // aligned
      "  );\n"
      "endmodule\n"},
-
+    {"module test;\n"
+     "bind entropy_src tlul_assert #(.EndpointType(\"Device\"))\n"
+     "tlul_assert_device (.clk_i, .rst_ni, .h2d(tl_i), .d2h(tl_o));\n"
+     "endmodule\n",
+     "module test;\n"
+     "  bind entropy_src tlul_assert #(\n"
+     "      .EndpointType(\"Device\")\n"
+     "  ) tlul_assert_device (\n"
+     "      .clk_i,\n"
+     "      .rst_ni,\n"
+     "      .h2d(tl_i),\n"
+     "      .d2h(tl_o)\n"
+     "  );\n"
+     "endmodule\n"},
+    {"module test;\n"
+     "bind entropy_src tlul_assert #(.EndpointType(\"Device\"))\n"
+     "tlul_assert_device (.clk_i, .rst_ni,\n\n .h2d(tl_i),\n\n .d2h(tl_o));\n"
+     "endmodule\n",
+     "module test;\n"
+     "  bind entropy_src tlul_assert #(\n"
+     "      .EndpointType(\"Device\")\n"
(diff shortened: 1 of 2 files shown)


verilog/formatting/formatter_test.cc
```diff
@@ -1574,6 +1574,20 @@ static constexpr FormatterTestCase kFormatterTestCases[] = {
...
+     "tlul_assert_device (.clk_i, .rst_ni, .h2d(tl_i), .d2h(tl_o));\n"
+     "endmodule\n",
+     "module test;\n"
+     "  bind entropy_src tlul_assert #(\n"
+     "      .EndpointType(\"Device\")\n"
+     "  ) tlul_assert_device (\n"
+     "      .clk_i,\n"
+     "      .rst_ni,\n"
+     "      .h2d(tl_i),\n"
+     "      .d2h(tl_o)\n"
```
tgorochowik, line 1588: Could we add some tests with:
* blank line in between
* comments in between

tgorochowik (commented):
The code looks good. I guess in this case we don't need any switches to enable this dynamically. One thing we could add is more tests to make sure it is in line with other alignment requirements

verilog/formatting/formatter_test.cc
hzeller: +1

hzeller (commented):
LGTM after test added Tomasz suggested.

verilog/formatting/formatter_test.cc
```diff
@@ -1605,7 +1605,54 @@ static constexpr FormatterTestCase kFormatterTestCases[] = {
...
+     "      .EndpointType(\"Device\")\n"
+     "  ) tlul_assert_device (\n"
+     "      //c1\n"
+     "      .clk_i,\n"
+     "      .rst_ni,\n"
+     "      //c2\n"
+     "      .h2d(tl_i),\n"
+     "      //c3\n"
+     "      .d2h(tl_o)\n"
+     "  );\n"
```
tgorochowik, line 1654: That was an unfortunate choice of code to test as we don't see what is happening to the alignment here.

Could you add tests with comments and empty lines with ports with names of different sizes? This way we will actually see if they are getting aligned or not

tgorochowik: BTW what happens if there is a port without `()`? Can we add such test?

And shouldn't we always align to those anyway ports anyway? (this is just a question, please don't change the implementation without any ACK that we should be doing this - or maybe there is something similar in the style guide?)

tgorochowik (approved):
LGTM, thanks for the additional test cases

PR #842, opened 2021-06-02
For testing container if empty, use empty(), not size().

hzeller:


mithro (approved):
LGTM.

PR #844, opened 2021-06-03
Fix indentation and style.

hzeller:


bazel/build-version.py
```diff
@@ -12,9 +12,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""
-Invoke bazel with --workspace_status_command=bazel/build-version.py to
-get this invoked and populate bazel-out/volatile-status.txt
+"""Invoke bazel with --workspace_status_command=bazel/build-version.py to get this invoked and populate bazel-out/volatile-status.txt
```
mithro, line 15: This wrapping looks wrong?

hzeller: It was auto-generated with the Google-internal formatter.

mithro:
We should at Python file format checking....

PR #848, opened 2021-06-04
Make asan clean

hzeller:
Compiling with address sanitizer options revealed a memory overrun. Fix the problem and also add address sanitizer compilation to CI.


mithro (approved):
LGTM!

(Kinda embarrassing that we hadn't already set this up!)

PR #857, opened 2021-06-14
Fix indenatation of initial begin with EOL comment

### LINKED ISSUE #499: incorrect indentation of `initial begin...end` blocks with //comment
**Test case**

```systemverilog
initial // clock generation
  begin
    clk = 0;
    forever begin
      #4ns clk = !clk;
    end
  end
```

**Actual output**

```systemverilog
initial // clock generation
    begin
  clk = 0;
  forever begin
    #4ns clk = !clk;
  end
end
```

The formatter should not touch this input block at all as it is already formatted correctly.

```systemverilog
initial // clock generation
  begin
    clk = 0;
    forever begin
      #4ns clk = !clk;
    end
  end
```

Citations to published style guides would help.

wtatarski:
Fixes #499

### DIFF
--- a/verilog/formatting/formatter_test.cc
+++ b/verilog/formatting/formatter_test.cc
@@ -1652,6 +1652,15 @@ static constexpr FormatterTestCase kFormatterTestCases[] = {
      "    .out1234(out),\n"
      "    .clk1234(clk),\n"
      ");\n"},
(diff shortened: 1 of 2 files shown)


tgorochowik (approved):
LGTM

Issue #360, opened 2020-07-08
Formatter: Align nested ternaries


GregAC:
A common construct in verilog is using a nested ternary to express a priority selection structure, the output produced by the formatter for these isn't hugely readable, as an example from: https://github.com/lowRISC/ibex/blob/master/rtl/ibex_icache.sv

Before formatting:
```systemverilog
assign tag_index_ic0 = inval_prog_q   ? inval_index_q :
                       ecc_write_req  ? ecc_write_index :
                       fill_grant_ic0 ? fill_index_ic0 :
                                        lookup_index_ic0;
```

After formatting:
```systemverilog
assign tag_index_ic0 = inval_prog_q ? inval_index_q :
    ecc_write_req ? ecc_write_index : fill_grant_ic0 ? fill_index_ic0 : lookup_index_ic0;
```

In the before version it's very clear what output is associated with what condition, the formatted version is harder to read.

fangism:
This makes for an interesting case of shape-detection: looking for a syntax tree shape that is actually linearly nested but formatted in a flat strategy.  It turns out this is what happens already with cascaded if-else-if constructs in 

https://github.com/google/verible/blob/19dade32b36684002ea4f88a9b8786d5d3402b92/verilog/formatting/tree_unwrapper.cc#L588

and

https://github.com/google/verible/blob/19dade32b36684002ea4f88a9b8786d5d3402b92/verilog/formatting/tree_unwrapper.cc#L1176

The handling for this reshapes partitions from the bottom up (last else-if first) and incrementally in the latter function.  

To solve this problem, I would start by tackling token partition tree shaping, and then alignment of those partitions next, using the existing alignment framework.

Since were dealing with expressions that can appear nearly anywhere, in the long run, we should look into bottom-up (dynamic programming) strategies to make this generally applicable in many places.

fangism:
@msfschaffner FYI, you mentioned this one

fangism:
Also worth noting is that this is potentially a nested alignment problem.  `assign` statements themselves may desire a particular alignment around `=`.

tgorochowik:
Continued on #953, closing this ticket

PR #860, opened 2021-06-23
Keep '?' with condition in ternaries

### LINKED ISSUE #359: Formatter: Ensure condition is on the same line as ? in ternaries
Where the formatting is adding new lines to a ternary expression it should ensure the `?` part of the expression is kept on the same line as the condition. As an example from: https://github.com/lowRISC/ibex/blob/master/rtl/ibex_icache.sv

Before formatting:
```systemverilog
assign prefetch_addr_d     =
    lookup_grant_ic0 ? (lookup_addr_aligned + {{ADDR_W-LINE_W-1{1'b0}},1'b1,{LINE_W{1'b0}}}) :
                       addr_i;
```

After formatting:
```systemverilog
  assign prefetch_addr_d = lookup_grant_ic0
      ? (lookup_addr_aligned + {{ADDR_W - LINE_W - 1{1'b0}}, 1'b1, {LINE_W{1'b0}}}) : addr_i;
```

### LINKED ISSUE #431: Wrong hang/wrap indentation in ternaries with comments
the following testcase 

```systemverilog
module test;
  assign cnt_next = (foo) ? '0          : // clear 
                    (bar) ? cnt + 1'b1  : // count 
                            cnt;
endmodule
```
gets formatted as follows:

```systemverilog
module test;
  assign cnt_next = (foo) ? '0 :  // clear
  (bar) ? cnt + 1'b1 :  // count
  cnt;
endmodule

```
 
this seems to have the wrong hang/wrap indentation.

it could be that this is due to the comments at the end of the lines.
without the comments this gets formatted as:

```systemverilog
module test;
  assign cnt_next = (foo) ? '0 : (bar) ? cnt + 1'b1 : cnt;
endmodule
```
and if we make the expression longer, it gets wrapped correctly:
```systemverilog
module test;
  assign cnt_next = (bar) ?
      '0 : (baz) ? cnt + 1'b1 : reallylongnamethatistoolongtofitontoalineandshouldbewrapped;
endmodule
```

rkapuscik:
Fixes #359.
Also fixes #431 - only adds test cases with comments, it seems the indentation issue was resolved some time ago.

### DIFF
--- a/verilog/formatting/formatter_test.cc
+++ b/verilog/formatting/formatter_test.cc
@@ -780,6 +780,186 @@ static constexpr FormatterTestCase kFormatterTestCases[] = {
         "assign prefetch_d = lookup_grant_ic0 ?\n"
         "    (lookup_addr + 1) : addr_i;\n",
     },
+    {
+        "module test;\n"
+        " assign next = // EOL\n"
+        "  foo ? '0 :\n"
+        "  cnt;\n"
+        "endmodule\n",
+        "module test;\n"
+        "  assign next =  // EOL\n"
+        "      foo ? '0 : cnt;\n"
+        "endmodule\n",
+    },
+    {
+        "module test;\n"
+        " assign next = foo // EOL\n"
+        "  ? '0 :\n"
+        "  cnt;\n"
+        "endmodule\n",
+        "module test;\n"
+        "  assign next = foo  // EOL\n"
+        "      ? '0 : cnt;\n"
+        "endmodule\n",
+    },
+    {
+        "module test;\n"
+        " assign next = foo ? // EOL\n"
+        "  '0 :\n"
+        "  cnt;\n"
+        "endmodule\n",
+        "module test;\n"
+        "  assign next = foo ?  // EOL\n"
+        "      '0 : cnt;\n"
+        "endmodule\n",
+    },
+    {
+        "module test;\n"
+        " assign next = foo ? '0 // EOL\n"
+        "  :\n"
+        "  cnt;\n"
+        "endmodule\n",
+        "module test;\n"
+        "  assign next = foo ? '0  // EOL\n"
+        "      : cnt;\n"
+        "endmodule\n",
+    },
+    {
+        "module test;\n"
+        " assign next = foo ? '0 : // EOL\n"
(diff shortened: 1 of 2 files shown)


hzeller:
#359 requests that question marks is on the _same_ line as the expression, but the tests here all confirm that it doesn't do that yet (except with `//EOL` comments forcing it). What am I missing ?

rkapuscik:
After reading the issue again, I think the way it is worded is confusing - I understood that as "keep the question mark part of the expression with the condition". This seems to be in line with:
* the issue name ("Ensure condition is on the same line as ? in ternaries")
* [this comment](https://github.com/lowRISC/ibex/issues/687#issue-579469182) ("Keep condition and ? together in ternaries")
* [this comment](https://github.com/lowRISC/ibex/pull/1014#discussion_r451413088) ("I feel the ? should be contained on the same line as the actual condition").

Style guide itself is not clear on the matter:
* an example [here](https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md#expressions) is:
```
assign a = ((addr & mask) == `MY_ADDRESS) ?
           matches_value :
           doesnt_match_value;
```
* on the other hand [here](https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md#parentheses) both the `?` and the following expression are kept on the same line:
```
assign foo = condition_a ? a :
             condition_b ? b : not_a_nor_b;
```

One point that seems clear across all these examples is that `?` should be kept on the same line as the condition and this is the purpose of this PR (except when separated by EOL comment).

Formatting of nested ternaries will be handled with #360.

If the intent was to keep the whole expression after the `?`, this would possibly require larger changes than penalty tuning (and possibly clarifying the contrary example in the style guide). This could be handled by repartitioning the ternaries (nested or not).

hzeller (approved):
The changes to document the current behavior in the test are good, but we should clarify in the original bug what the exact behavior should be; given the ambiguity the implementation might need adaption.

tgorochowik:
@hzeller we've got a response from Greg: https://github.com/chipsalliance/verible/issues/359#issuecomment-889141313 - I think this can be merged

tgorochowik:
Let's merge this

PR #879, opened 2021-08-06
Provide build-environment set-up for nix.

hzeller:
Make it simple and hermetic to set-up the build environment
on systems with the nix package manager.
Inspired by https://github.com/chipsalliance/Surelog/pull/1606


hzeller:
CC @sjkelly

PR #888, opened 2021-08-19
Also run coverage tests.

hzeller:
In preparation to send it to codecov


hzeller:
I _think_ this should be ready, but codecov does not show anything yet https://app.codecov.io/gh/chipsalliance/verible

From the confusing error messages it sounds like that it doesn't have a 'base report', which would be plausible as this is the first push, but other than that the CI output does not show any error. So I suggest that we submit this and then see if there are corrections needed later.

hzeller:
It shouldn't need a token, but I'll try that now as well

PR #890, opened 2021-08-23
Link the version of gcov relevant to current compiler.

hzeller:


hzeller:
Trying to get to the bottom of why the coverage turns out to be empty.

hzeller:
Looks like linking the right gcov made it work.

tgorochowik (approved):

PR #901, opened 2021-09-01
Extend the undersized binary literal rule to also check Hex and Oct.

hzeller:
The numbers to check are configuration parameters; available are
the boolean parameters 'bin' (default on, pre-existing rule),
'oct' and 'hex' (both default off).

### DIFF
--- a/verilog/analysis/checkers/undersized_binary_literal_rule.cc
+++ b/verilog/analysis/checkers/undersized_binary_literal_rule.cc
@@ -1,4 +1,4 @@
-// Copyright 2017-2020 The Verible Authors.
+// Copyright 2017-2021 The Verible Authors.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -26,6 +26,7 @@
 #include "common/analysis/matcher/bound_symbol_manager.h"
 #include "common/analysis/matcher/matcher.h"
 #include "common/text/concrete_syntax_leaf.h"
+#include "common/text/config_utils.h"
 #include "common/text/symbol.h"
 #include "common/text/syntax_tree_context.h"
 #include "common/text/token_info.h"
@@ -38,11 +39,13 @@
 namespace verilog {
 namespace analysis {
 
+using verible::down_cast;
 using verible::GetStyleGuideCitation;
 using verible::LintRuleStatus;
 using verible::LintViolation;
 using verible::SyntaxTreeContext;
 using verible::SyntaxTreeLeaf;
+using verible::SyntaxTreeNode;
 using verible::matcher::Matcher;
 
 // Register UndersizedBinaryLiteralRule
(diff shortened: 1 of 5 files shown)


hzeller:
Removing the bazel cache for Kythe in a separate commit was because the cached kzip and some new kzip generated conflicts apparently (the kzips are stored in the bazel output directory).

Error message was something like this:
```
Error merging archives: unit already exists
```

So, let Kythe be run full without bazel cache.

hzeller:
I am wondering if this should be default 'on' for the new types ? It might be dependent on the style what people use.

So question for people in the 'field'; would you like the linter complain about situations where there are not enough digits when providing explicitly sized numerals:
```
  assign a = 32'hAB;   // This would be flagged
  assign b = 32'h000000AB;  // This would be ok
  assign c = 32'h00000AB;  // typical case that happens: manually miscounted while typing. Would be flagged.

```

msfschaffner:
> I am wondering if this should be default 'on' for the new types ? It might be dependent on the style what people use.
> 
> So question for people in the 'field'; would you like the linter complain about situations where there are not enough digits when providing explicitly sized numerals:
> 
> ```
>   assign a = 32'hAB;   // This would be flagged
>   assign b = 32'h000000AB;  // This would be ok
>   assign c = 32'h00000AB;  // typical case that happens: manually miscounted while typing. Would be flagged.
> ```

How about setting the default to match the status quo (not enabled)?
As a side note: in OpenTitan, we don't mandate `hex` and `oct` literals to be full size.

hzeller:
Thanks @msfschaffner that is what I thought.
The current defaults as implemented in this PR are only enabling for binary but not hex and octal.

verilog/analysis/checkers/undersized_binary_literal_rule.cc
```diff
@@ -55,65 +58,108 @@ const char UndersizedBinaryLiteralRule::kTopic[] = "number-literals";
...
-              violations_.insert(LintViolation(
-                  digits_leaf->get(),
-                  FormatReason(width_text, base_text, digits_text), context));
-            }
-          }  // else width is not constant, so ignore
-        }
-      }
-    }
+  if (!NumberMatcher().Matches(symbol, &manager)) return;
+  const auto width_leaf = manager.GetAs<SyntaxTreeLeaf>("width");
```
fangism, line 90: Thank you for flattening the flow control structure!

verilog/analysis/checkers/undersized_binary_literal_rule.cc
```diff
@@ -55,65 +58,108 @@ const char UndersizedBinaryLiteralRule::kTopic[] = "number-literals";
...
+    absl::string_view width, absl::string_view base_text, char base,
     absl::string_view literal) {
-  return absl::StrCat("Binary literal ", width, base, literal,
-                      " is shorter than its declared width: ", width, ".");
+  absl::string_view base_describe;
+  switch (base) {
+    case 'b':
+      base_describe = "Binary";
+      break;
+    case 'h':
```
fangism, line 145: do we need case 'x' as well?

fangism (approved):
Nice set of test cases!

verilog/analysis/checkers/undersized_binary_literal_rule.cc
hzeller: Is `x` a valid base ? I only know about **b**inary, **d**ecimal, **o**ctal, **h**ex (which is also the set of values that `BasedNumber` can do.

PR #903, opened 2021-09-01
Provide AutoFixes in binary size literal rule.

hzeller:


hzeller:
This allows the linter to output a patch what to do to fix:

```
echo "localparam x = 32'hAB;" | bazel run verilog/tools/lint:verible-verilog-lint -- --rules=undersized-binary-literal=hex:true --autofix_output_file=/dev/stdout --autofix=yes -
```

Results in 
```patch
--- a/-
+++ b/-
@@ -1 +1 @@
-localparam x = 32'hAB;
+localparam x = 32'h000000AB;
```

verilog/analysis/checkers/undersized_binary_literal_rule_test.cc
```diff
@@ -169,6 +169,46 @@ TEST(UndersizedBinaryLiteralTest, DecimalNumbersNeverCare) {
...
+}
+
+static std::string UndersizeApplyFix(absl::string_view input) {
+  VerilogAnalyzer analyzer(input, "");
+  CHECK(analyzer.Analyze().ok()) << "Parse issue " << input;
+  using verible::LintRunner;
+  LintRunner<verible::SyntaxTreeLintRule> runner(CreateConfiguredRule());
+  const verible::LintRuleStatus rule_status = runner.Run(analyzer.Data(), "");
+  EXPECT_EQ(rule_status.violations.size(), 1);
+  EXPECT_EQ(rule_status.violations.begin()->autofixes.size(), 1);
```
fangism, line 189: Would ASSERT_EQ be more appropriate, to stop execution immediately?

verilog/analysis/checkers/undersized_binary_literal_rule_test.cc
```diff
@@ -169,6 +169,46 @@ TEST(UndersizedBinaryLiteralTest, DecimalNumbersNeverCare) {
...
+
+static std::unique_ptr<UndersizedBinaryLiteralRule> CreateConfiguredRule() {
+  auto result = std::make_unique<UndersizedBinaryLiteralRule>();
+  CHECK(result->Configure("bin:true;hex:true;oct:true").ok());
+  return result;
+}
+
+static std::string UndersizeApplyFix(absl::string_view input) {
+  VerilogAnalyzer analyzer(input, "");
+  CHECK(analyzer.Analyze().ok()) << "Parse issue " << input;
```
fangism, line 184: ASSERT_EQ, to allow other tests to run.

fangism (approved):
Love it, suggest-fix!

verilog/analysis/checkers/undersized_binary_literal_rule_test.cc
hzeller: It would, but for whatever reason the gtest.h chokes on that, possibly because the `ASSERT_EQ()` is not used inside a `TEST()` macro. Replacing it with `CHECK_EQ()` now.

verilog/analysis/checkers/undersized_binary_literal_rule_test.cc
hzeller: Same problem here, `ASSERT_EQ()` does not work inside helper functions.

PR #904, opened 2021-09-02
Provide more detailed choices for linter autofix.

hzeller:
The autofix choices were easy to set in a way that
it would accidentally overwrite the input file (setting
--autofix=yes and forgetting output file). This patch
makes the choice explicit and in particular requires
to choose 'inplace' option to modify output.

  * no
  * patch-interactive
  * patch
  * inplace-interactive
  * inplace

The `yes` option does not exist anymore, as it is replaced
by the more detailed choices. Hopefully that will not break
too many users at this point. The github action though
is probably affected.

While at it, make the default output stdout for the patch
choices (and move error messages to stderr). That way, the
tool is behaving like other command line tools.

### DIFF
--- a/verilog/analysis/verilog_linter.cc
+++ b/verilog/analysis/verilog_linter.cc
@@ -156,7 +156,7 @@ void ViolationFixer::HandleViolation(
   std::stringstream violation_message;
   formatter.FormatViolation(&violation_message, violation, base, path, url,
                             rule_name);
-  (*stream_) << violation_message.str() << std::endl;
+  (*message_stream_) << violation_message.str() << std::endl;
 
   if (violation.autofixes.empty()) {
     return;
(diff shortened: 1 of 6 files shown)


hzeller:
@mglb @tgorochowik this might have implications of what parameters the github action has to use when invoking the tool. Can we pin a particular version there, so that we don't run into trouble with:

  1. Old version needing --autofix=yes and writing errors to stdout
  2. new version needing --autofix=patch and writing errors to stderr

If there might be trouble I could add a 'compatiblity' --autofix=yes option with deprecation message, but if we don't have to, I'd prefer that.

tgorochowik:
The action with the autofixer integrated does not yet have any release associated with it so we can just delay it a little bit nore

wsipak:
It's OK, the action will use `--autofix=patch` before its new version. LGTM

tgorochowik (approved):

PR #906, opened 2021-09-02
Augment unit-tests of AutoFix rules to test for fixes.

hzeller:
Add the boilderplate and test for typical fixes.

### DIFF
--- a/verilog/analysis/checkers/endif_comment_rule.cc
+++ b/verilog/analysis/checkers/endif_comment_rule.cc
@@ -120,14 +120,35 @@ void EndifCommentRule::HandleToken(const TokenInfo& token) {
           auto endif_end =
               last_endif_.text().substr(last_endif_.text().size(), 0);
 
+          // The whitespace before the comment is something that should come
+          // from the formatter configuration so that we don't emit code that
+          // then would be complained about by the formatter.
+          // Or, make this a configuration option in the lint rule (probably
+          // best, as we don't intermingle formatter/linter).
+          // TODO(hzeller) Until that is happening just comment here.
+          constexpr int kSpacesBeforeComment = 2;
+          std::string ws(kSpacesBeforeComment, ' ');
+
+          // TOOD(hzeller): rethink the 'alternative fixes' thing. Problem is,
+          // that it makes it impossible to run this unattended as that choice
+          // would've to be made at apply time (and the chooser doesn't provide
+          // that yet, and it would probably complicate the UI a lot when it
+          // will be added).
+          //
+          // Maybe better: only have _one_ alternative and pre-configure the
+          // rule to do that.
+          // So with the above our rule needs two configurations.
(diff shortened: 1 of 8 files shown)


verilog/analysis/checkers/endif_comment_rule.cc
```diff
@@ -120,14 +120,35 @@ void EndifCommentRule::HandleToken(const TokenInfo& token) {
...
+          // The whitespace before the comment is something that should come
+          // from the formatter configuration so that we don't emit code that
+          // then would be complained about by the formatter.
+          // Or, make this a configuration option in the lint rule (probably
+          // best, as we don't intermingle formatter/linter).
+          // TODO(hzeller) Until that is happening just comment here.
+          constexpr int kSpacesBeforeComment = 2;
+          std::string ws(kSpacesBeforeComment, ' ');
+
+          // TOOD(hzeller): rethink the 'alternative fixes' thing. Problem is,
```
hzeller, line 132: @mglb The alternatives are probably hard to surface in the UI (right now, we just choose the first one), and in particular for unattended patch generation, this is also not working well.

So maybe we should just add a configuration to this lint rule (configuration infrastructure is there), to pre-choose how fixes are generated. That should simplify things. WDYT ?

verilog/analysis/checkers/endif_comment_rule_test.cc
```diff
@@ -106,6 +107,19 @@ TEST(EndifCommentRuleTest, RejectsEndifWithoutComment) {
   RunLintTestCases<VerilogAnalyzer, EndifCommentRule>(kTestCases);
 }
 
+TEST(EndifCommentRuleTest, ApplyAutoFix) {
+  // Alternatives the auto fix offers
+  constexpr int kEOLComment = 0;
+  constexpr int kBlocComment = 1;
```
mglb, line 113: kBloc**k**Comment ?

verilog/analysis/checkers/endif_comment_rule.cc
mglb: We can call this configuration option "default choice". This will work well even with (hypothetical) UIs that offer choice picking (language server or other IDE integration comes to mind).

verilog/analysis/checkers/endif_comment_rule_test.cc
hzeller: Done, called it `kBlkComment` so that it fits in the line below.

verilog/analysis/checkers/endif_comment_rule.cc
hzeller: mmh, ok that could be a way. I'd still see that a single choice, chosen in the configuration file of the project as part of the style-guide, would simplify the workflow, as the IDE's task is to reduce decision fatigue.

Anyway, something for another change in the future.

PR #907, opened 2021-09-03
Add a lint rule that detects truncated number literals.

hzeller:
Detects number literals that have a shorter bit-width than the
digits provided, e.g. 16'h1FFFF.

Decimal numbers are only dealt with correctly up to 128 bits,
beyond that, only some lower-bound estimates are applied so
it might miss complain opportunity there. Given that super-long
constants are very likely not given as decimal, this
tradeoff is accepted.

### DIFF
--- a/verilog/analysis/default_rules.h
+++ b/verilog/analysis/default_rules.h
@@ -65,6 +65,7 @@ constexpr const char* kDefaultRuleSet[] = {
     "positive-meaning-parameter-name",
     "constraint-name-style",
     "suggest-parentheses",
+    // TODO(hzeller) enable after #907 "truncated-numeric-literal",
     // TODO(fangism): enable in production:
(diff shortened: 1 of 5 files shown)


verilog/analysis/default_rules.h
```diff
@@ -65,6 +65,7 @@ constexpr const char* kDefaultRuleSet[] = {
     "positive-meaning-parameter-name",
     "constraint-name-style",
     "suggest-parentheses",
+    "truncated-numeric-literal",
```
fangism, line 68: I used to recommend flipping the default switch separately to make it easily revertible.

fangism (approved):
Awesome test coverage of corner cases!

verilog/analysis/default_rules.h
hzeller: sounds good, will comment here and make separate PR with just that flip.

PR #909, opened 2021-09-03
Make the test util violation messages more intuitive to read.

hzeller:


hzeller:
ok, now that the test is adapted, it will also pass CI :)

PR #914, opened 2021-09-05
Add clang-tidy script, to be run from the CI.

hzeller:
* Use bazel rule to generate compilation database.
  * Run clang-tidy with some initial useful warnings enabled.
    The output file is currently just displayed in the CI log,
    but once that is clean, we can fail-on-warning

Fixes: #653


fangism (approved):
Sweet!

PR #921, opened 2021-09-08
Document in development doc how to run coverage locally.

hzeller:
Also provide script to go along with it for ease-of-use.

### DIFF
--- a/.github/bin/generate-coverage-html.sh
+++ b/.github/bin/generate-coverage-html.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+# Copyright 2021 The Verible Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


.github/bin/generate-coverage-html.sh
```diff
@@ -0,0 +1,38 @@
...
+#
+# Precondition, having run bazel coverage with coverage data collection, e.g.
+#   MODE=coverage .github/bin/build-and-test.sh
+#
+
+PROJECT_ROOT=$(dirname $0)/../../
+OUTPUT_DIR=coverage-html
+
+cd $PROJECT_ROOT
+genhtml -o ${OUTPUT_DIR} bazel-out/_coverage/_coverage_report.dat
```
fangism, line 37: genhtml is a tool from another required package?  if so, include in comment above.

fangism (approved):
😍

.github/bin/generate-coverage-html.sh
hzeller: It is mentioned above in the readme: `gcov` is the package to install.
That is also the package to install to allow bazel coverage to work in the first place.

hzeller: mmh, maybe it is lcov, let me check.

hzeller: It was `lcov`. Fixed.

PR #922, opened 2021-09-09
Add tests for --json_export in verilog syntax shell test.

### LINKED ISSUE #891: Create output tests for `verible-verilog-syntax --export_json`
Create tests for JSON emitted by `verible-verilog-syntax --export_json`.

It should use some JSON parser to compare the data instead of output text.

C++ tests that can be used as a reference:
https://github.com/chipsalliance/verible/blob/master/verilog/analysis/json_diagnostics_test.cc
https://github.com/chipsalliance/verible/blob/master/common/text/token_info_json_test.cc
https://github.com/chipsalliance/verible/blob/master/verilog/CST/verilog_tree_json_test.cc

Tests should be added to [verilog/tools/syntax/verilog_syntax_test.sh](https://github.com/chipsalliance/verible/blob/master/verilog/tools/syntax/verilog_syntax_test.sh) or separate `.sh` script.

hzeller:
To be self-contained, avoid pulling in any external json parser but utilize shell capabilities to implement a simple json canonicalizer.

This allows to compare tool-output with an expected output ignoring the actual formatting of the json data.

Fixes #891

### DIFF
--- a/verilog/tools/syntax/verilog_syntax_test.sh
+++ b/verilog/tools/syntax/verilog_syntax_test.sh
@@ -34,6 +34,34 @@ function strip_error() {
   sed -e 's| (syntax-error).||'
 }
 
+# Takes some json input from stdin and creates a string that has all
+# dictionaries sorted and all whitespace canonicalized.
+# Very simplistic, and output is not necessarily parseable as json, but goal
+# is to have some comparable string (will break if comma in strings)
+function json_canonicalize() {
+  local nest=()  # Stack of nested {} and [] regions.
+  local depth=-1
+  while read -n1 c ; do
+    case $c in
+      "{"|"[")  # Open next nested level
+	      nest[$((++depth))]=""
+	      ;;
+	    "}")  # Post-process dictionaries: sort by the keys for canonicalization.
+	      sorted=$(echo "${nest[${depth}]}" | tr ',' '\n' | sort | tr '\n' '%')
+	      nest[$((--depth))]+="{${sorted}}"
+	      ;;
+	    "]")  # Array sequence is kept as-is.
+	      no_comma=$(echo "${nest[${depth}]}" | tr ',' '%')
+	      nest[$((--depth))]+="[${no_comma}]"
+	      ;;
+	    *)
+	      nest[${depth}]+="$c"
+	      ;;
+    esac
+  done
+  echo ${nest[0]} | tr '%' ','
+}
+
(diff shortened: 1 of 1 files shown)


verilog/tools/syntax/verilog_syntax_test.sh
```diff
@@ -94,6 +122,35 @@ EOF
...
+json_canonicalize > "$MY_EXPECT_FILE" <<EOF
+{"-":{
+  "errors": [
+      {"line": 0, "column": 7, "phase": "parse", "text": "1" },
+      {"line": 1, "column": 0, "phase": "parse", "text": "endmodule" },
+      {"line": 3, "column": 0, "phase": "parse", "text": "endmodule" }
+  ]}}
+EOF
+
+tr '\r\n' '\n' < $MY_OUTPUT_FILE | json_canonicalize > "${MY_OUTPUT_FILE}.1"
```
fangism, line 150: FYI, there's a too called dos2unix that helps with newline conversion as well.

verilog/tools/syntax/verilog_syntax_test.sh
```diff
@@ -34,6 +34,34 @@ function strip_error() {
   sed -e 's| (syntax-error).||'
 }
 
+# Takes some json input from stdin and creates a string that has all
+# dictionaries sorted and all whitespace canonicalized.
+# Very simplistic, and output is not necessarily parseable as json, but goal
+# is to have some comparable string (will break if comma in strings)
+function json_canonicalize() {
```
fangism, line 41: Could be in its own standalone shell script?  Seems very useful.

fangism (approved):
Seems very useful, could there be an implementation in C++?  maybe even pre-existing in some JSON lib?

verilog/tools/syntax/verilog_syntax_test.sh
hzeller: I wanted to keep the number of different tools used as little as possible, as this would run on various operating systems.

verilog/tools/syntax/verilog_syntax_test.sh
hzeller: I guess once it is used in more than one place, we can move it out.

hzeller:
I didn't want to use the currently used C++ library for a quick tool as my intent is to actually replace that very library. So the tests should not depend on that being changed as it could hide subtle issues.

PR #924, opened 2021-09-10
Switch to https://github.com/nlohmann/json JSON library.

hzeller:
This is a more modern JSON library, slightly
faster and also usually more supported these
days.


fangism (approved):

hzeller:
CC: @mglb <- FYI, as you have been working on JSON output. This library is almost source compatible, so readability is essentially the same.

PR #929, opened 2021-09-13
For undersized literals with value one: also suggest decimal.

hzeller:
e.g. 32'h1 now provides two suggestions
  * 32'd1
  * 32'h00000001


hzeller:
Question is: should this be extended to not only `1` but anything that is single-digit _and_ also is a decimal decimal ? So for instance `32'o1 … 32'o7` and `32'h1 … 32'h9`

fangism (approved):
nice test cases!  Sometimes, someone will want an `_ in the literal for readability.  Style may vary.

PR #931, opened 2021-09-14
Fix formatting of ternaries

mglb:
Fixes https://github.com/chipsalliance/verible/issues/887

Formatted test cases from the issue:

```systemverilog
module X ();
  assign myvar = check ? {10'b0, foo} : (bar && (baz == '0)) ? hello_longlonglong :
      world_longlonglong ? temp1_longlonglong : temp2_longlonglong;
endmodule : X
```

```systemverilog
module X ();
  assign {dc_wrap, dc_htbt_d} = !(htbt_ctr_q == blink_param_x_i) ?
      {1'b0, dc_htbt_q} : ((dc_htbt_q == duty_cycle_a_i) && pattern_repeat) ?
      {1'b0, duty_cycle_a_i} : (htbt_direction) ? {1'b0, dc_htbt_q} - {1'b0, blink_param_y_i} -
      1'b1 : {1'b0, dc_htbt_q} + {1'b0, blink_param_y_i} + 1'b1;
endmodule : X
```

```systemverilog
module X ();
  assign {phase_wrap, off_phase} = {1'b0, phase_delay_scaled[15:0]} +
      {1'b0, duty_cycle_scaled[15:0]};
endmodule : X
```

Todo:
* unit tests

Ideas for further work:
* It would be nice to implement tabular alignment of nested ternary operators.

### DIFF
--- a/verilog/formatting/formatter_test.cc
+++ b/verilog/formatting/formatter_test.cc
@@ -978,6 +978,28 @@ static constexpr FormatterTestCase kFormatterTestCases[] = {
         "    ((e == f) && g) ? {1'b0, f} : (h) ?\n"
         "    {1'b0, e} - 1'b1 : {1'b0, e} + 1'b1;\n",
     },
+    {
+        "assign {aaaaaaaaaa, bbbbbbbbb} = {1'b0, cccccccccccccccccc[15:0]} +\n"
+        "                                 {1'b0, ddddddddddddddddd[15:0]};\n",
+        "assign {aaaaaaaaaa, bbbbbbbbb} =\n"
+        "    {1'b0, cccccccccccccccccc[15:0]} + {\n"
+        "    1'b0, ddddddddddddddddd[15:0]};\n",
+    },
+    {
+        "covergroup a(string b);\n"
+        "foobar: cross foo, bar {"
+        "ignore_bins baz = binsof(qux) intersect {1, 2, 3, 4, 5, 6, 7};"
+        "}\n"
+        "endgroup : a\n",
+        "covergroup a(string b);\n"
+        "  foobar: cross foo, bar{\n"
+        "    ignore_bins baz =\n"
+        "        binsof (qux) intersect {\n"
+        "      1, 2, 3, 4, 5, 6, 7\n"
+        "    };\n"
+        "  }\n"
+        "endgroup : a\n",
+    },
 
     // streaming operators
     {
--- a/verilog/formatting/formatter_test.cc
+++ b/verilog/formatting/formatter_test.cc
@@ -395,8 +395,8 @@ static constexpr FormatterTestCase kFormatterTestCases[] = {
      " `AINIT(S, (D == 4 && K inside {0, 1}) ||"
      " (D == 3 && K== 4))\n",
      "`AINIT(S,\n"
-     "       (D == 4 && K inside {0, 1}) || (\n"
-     "           D == 3 && K == 4))\n"},
+     "       (D == 4 && K inside {0, 1}) ||\n"
+     "           (D == 3 && K == 4))\n"},
     {// long macro call breaking
      " `ASSERT_INIT(S, D == 4 && K inside {0, 1})\n",
      "`ASSERT_INIT(S,\n"
@@ -982,8 +982,8 @@ static constexpr FormatterTestCase kFormatterTestCases[] = {
         "assign {aaaaaaaaaa, bbbbbbbbb} = {1'b0, cccccccccccccccccc[15:0]} +\n"
         "                                 {1'b0, ddddddddddddddddd[15:0]};\n",
         "assign {aaaaaaaaaa, bbbbbbbbb} =\n"
-        "    {1'b0, cccccccccccccccccc[15:0]} + {\n"
-        "    1'b0, ddddddddddddddddd[15:0]};\n",
+        "    {1'b0, cccccccccccccccccc[15:0]} +\n"
+        "    {1'b0, ddddddddddddddddd[15:0]};\n",
     },
     {
         "covergroup a(string b);\n"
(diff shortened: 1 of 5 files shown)


verilog/formatting/formatter_test.cc
```diff
@@ -395,8 +395,8 @@ static constexpr FormatterTestCase kFormatterTestCases[] = {
      " `AINIT(S, (D == 4 && K inside {0, 1}) ||"
      " (D == 3 && K== 4))\n",
      "`AINIT(S,\n"
-     "       (D == 4 && K inside {0, 1}) || (\n"
-     "           D == 3 && K == 4))\n"},
+     "       (D == 4 && K inside {0, 1}) ||\n"
+     "           (D == 3 && K == 4))\n"},
```
mglb, line 399: FYI: this has been added as a test for https://github.com/chipsalliance/verible/issues/351. The new behavior seems to be better.

hzeller: +1

hzeller:
+1 for fixing the curly-brace concatenation compactness.

I think tabular alignment is something we should strive for, as there are many places where this pattern is used in SystemVerilog code. In a first step, might need to recognize if we would actually destroying an existing artistically aligned `?:` chain and don't verschlimmbessern it; because changing 

```Verilog
assign myvar = check                ? {10'b0, foo}       :
               (bar && (baz == '0)) ? hello_longlonglong :
               world_longlonglong   ? temp1_longlonglong : temp2_longlonglong;
```
to
```Verilog
myvar = check ? {10'b0, foo} : (bar && (baz == '0)) ? hello_longlonglong :
      world_longlonglong ? temp1_longlonglong : temp2_longlonglong;
```

Would not necessarily improve readability and make things worse. So not sure if we should take on tabular alignment in tihs PR as well ? Though might be a good timely follow-up change.

There are possibly different styles as well.

I've seen something like this (a)

```verilog
  cond_a ? value_a :
  cond_b ? value_b :
  cond_c ? value_c :
           value_d;
```

Though sometimes the last condition is just added at the end (b)
```verilog
  cond_a ? value_a :
  cond_b ? value_b :
  cond_c ? value_c : value_d;
```

But I could imagine if we stick to (a) rarely someone would complain if we change (b) to (a) (nicer ,because more space in the line for complex values; worse, as it violates the 'use less vertical space' goal).

Though for just one condition, I'd probably line up the the `:` with the `?`

```verilog
  cond_a ? value_a
         : value_b;
```

Looking at existing code-artwork could give some ideas.

Adding @msfschaffner for further comments.

msfschaffner:
Likewise, thanks for taking a stab at this @mglb!

Tabular alignment would be very desirable for ternaries indeed.

In general I like @hzeller's suggestions, see below for a couple of comments.

--------------------------
As for whether to use (a) or (b), I think would have a slight preference for (b) since it reduces the amount of lines used. I've seen this in RTL code from other designers as well, e.g.:
https://github.com/lowRISC/opentitan/blob/882f9d24b6b03c8247b9b68edc266801237c9964/hw/ip/keymgr/rtl/keymgr.sv#L558-L559

I would say, go for (b) if it fits, otherwise try (a).

--------------------------
I think I would break a single condition as follows:
```systemverilog
  cond_a ? value_a :
           value_b;
```
rather than
```systemverilog
  cond_a ? value_a
         : value_b;
```
to keep it consistent with the general case. This is what several other designers on OpenTitan also prefer.
See the following examples taken from various blocks (written by different designers) in the OpenTitan code base:

https://github.com/lowRISC/opentitan/blob/882f9d24b6b03c8247b9b68edc266801237c9964/hw/ip/csrng/rtl/csrng_core.sv#L1049-L1050

https://github.com/lowRISC/opentitan/blob/882f9d24b6b03c8247b9b68edc266801237c9964/hw/ip/aes/rtl/aes_sbox_dom.sv#L869-L876

https://github.com/lowRISC/opentitan/blob/882f9d24b6b03c8247b9b68edc266801237c9964/hw/ip/aes/rtl/aes_sbox_dom.sv#L896-L897

https://github.com/lowRISC/opentitan/blob/882f9d24b6b03c8247b9b68edc266801237c9964/hw/ip/keymgr/rtl/keymgr.sv#L494-L495

--------------------------
Now, there are also cases where the conditions and selected expressions are excessingly long. 
In such cases, I would just default to something like
```systemverilog
assign out = (veeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeery_long_cond) ? 
             very_loooooooooooooooooooooooong_expr0          :
             very_long_eeeeeeeeeeeeeeeeeeeeeeeeeexpr1;
```


Example from the code base:
https://github.com/lowRISC/opentitan/blob/882f9d24b6b03c8247b9b68edc266801237c9964/hw/ip/otp_ctrl/rtl/otp_ctrl_part_unbuf.sv#L317-L320

mglb:
> +1 for fixing the curly-brace concatenation compactness.
> 
> I think tabular alignment is something we should strive for, as there are many places where this pattern is used in SystemVerilog code. In a first step, might need to recognize if we would actually destroying an existing artistically aligned `?:` chain and don't verschlimmbessern it; because changing
> 
> ```verilog
> assign myvar = check                ? {10'b0, foo}       :
>                (bar && (baz == '0)) ? hello_longlonglong :
>                world_longlonglong   ? temp1_longlonglong : temp2_longlonglong;
> ```
> 
> to
> 
> ```verilog
> myvar = check ? {10'b0, foo} : (bar && (baz == '0)) ? hello_longlonglong :
>       world_longlonglong ? temp1_longlonglong : temp2_longlonglong;
> ```
> 
> Would not necessarily improve readability and make things worse. So not sure if we should take on tabular alignment in tihs PR as well ? Though might be a good timely follow-up change.

Tabular alignment would handle detection problem to some degree. More generic detection could be based on continuation comments detection introduced in https://github.com/chipsalliance/verible/pull/858, but implementing it reliably could require as much work as just implementing tabular alignment for multiple styles.

Anyway, it can be introduced in additional PR. As I understand, this PR in its current form (well, ignoring lack of some tests) doesn't make things worse than they currently are, so there's no point in holding this code until other features are completed.

---

As for multiple style choices for tabular alignment:

I would try implementing possibility to register multiple CellScanner classes and partitioner functions for a single syntax structure (e.g. nested ternary operator, struct members, etc). Each class/function would format the code using different style. When the first one fails to format the code (e.g. due to column limit), a next one would be used, and so on.

hzeller (approved):
LGTM as first step. Next step should then be the tabular alignment as discussed
(maybe create a separate issue out of the extracted disucssion)

PR #934, opened 2021-09-17
Use isblank instead of isspace for trailing spaces detection

wsipak:
`no-trailing-spaces` rule reports issues in every line of files which use `\r\n` line endings.
Example file: https://github.com/chipsalliance/UHDM-integration-tests/blob/master/tests/AssignmentPattern/top.sv

~~The solution uses `std::isblank` which means only spaces and tabs in the end of lines will cause errors.~~

~~A useful comparison of `isblank` and `isspace` can be seen in the table in the bottom of this page:
https://en.cppreference.com/w/cpp/string/byte/isspace~~

EDIT:
The suggested test case has shown that the idea of solving the problem with `isblank` won't work.
Right now the code drops one `\r` character from the end of the line if it exists.

### DIFF
--- a/verilog/analysis/checkers/no_trailing_spaces_rule.cc
+++ b/verilog/analysis/checkers/no_trailing_spaces_rule.cc
@@ -24,6 +24,7 @@
 
 #include "absl/strings/str_cat.h"
 #include "absl/strings/string_view.h"
+#include "absl/strings/strip.h"
 #include "common/analysis/lint_rule_status.h"
 #include "common/text/token_info.h"
 #include "verilog/analysis/descriptions.h"
@@ -53,11 +54,14 @@ const LintRuleDescriptor& NoTrailingSpacesRule::GetDescriptor() {
 }
 
 void NoTrailingSpacesRule::HandleLine(absl::string_view line) {
+  // Lines may end with \n or \r\n. '\n' is already excluded.
+  // Exclude '\r'
+  absl::ConsumeSuffix(&line, "\r");
+  // Now any line endings (either \n or \r\n) are excluded.
   // Searches each line in reverse for spaces.
   const auto rbegin = line.crbegin();
   const auto rend = line.crend();
   if (rbegin != rend) {
-    // Lines already exclude \n, so we can check using std::isspace.
     const auto reverse_iter =
         std::find_if(rbegin, rend, [](char c) { return !std::isspace(c); });
     const int trailing = std::distance(rbegin, reverse_iter);
--- a/verilog/analysis/checkers/no_trailing_spaces_rule_test.cc
+++ b/verilog/analysis/checkers/no_trailing_spaces_rule_test.cc
@@ -35,6 +35,7 @@ using verible::RunLintTestCases;
 TEST(NoTrailingSpacesRuleTest, AcceptsText) {
   const std::initializer_list<LintTestCase> kTestCases = {
       {""},
+      {"\r\n"},
       {"\n"},
       {"\n\n\n"},
       {"\naaa"},
@@ -42,6 +43,8 @@ TEST(NoTrailingSpacesRuleTest, AcceptsText) {
       {"  xxx\n\n"},
       {"\txxx\n\n"},
       {"  xxx\n  yyy\n"},
+      {"  xxx\n  yyy\r\n"},
+      {"  xxx\n  yyy\r\n\r\n"},
       {"module foo;\nendmodule\n"},
   };
   RunLintTestCases<VerilogAnalyzer, NoTrailingSpacesRule>(kTestCases);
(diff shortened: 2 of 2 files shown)


wsipak:
@googlebot I signed it!

hzeller:
Thanks!
You also want to add a test to the rejects trailing spaces test, such as
```
 {"module foo;\nendmodule", {kToken, " "}, "\r\n"},
```

verilog/analysis/checkers/no_trailing_spaces_rule.cc
```diff
@@ -59,7 +59,7 @@ void NoTrailingSpacesRule::HandleLine(absl::string_view line) {
   if (rbegin != rend) {
     // Lines already exclude \n, so we can check using std::isspace.
```
hzeller, line 60: this comment is now outdated.

verilog/analysis/checkers/no_trailing_spaces_rule.cc
```diff
@@ -53,13 +53,18 @@ const LintRuleDescriptor& NoTrailingSpacesRule::GetDescriptor() {
 }
 
 void NoTrailingSpacesRule::HandleLine(absl::string_view line) {
+  // Lines may end with \n or \r\n. '\n' is already excluded.
+  // Exclude '\r'
+  if (!line.empty() && line.back() == '\r') {
+    line.remove_suffix(1);
```
hzeller, line 59: Check out the functionality provided by absl. This looks like a good canidate for `absl::ConsumeSuffix(&line, "\r")`

verilog/analysis/checkers/no_trailing_spaces_rule.cc
```diff
@@ -53,13 +53,18 @@ const LintRuleDescriptor& NoTrailingSpacesRule::GetDescriptor() {
...
+  }
+  // Now any line endings (either \n or \r\n) are excluded.
   // Searches each line in reverse for spaces.
   const auto rbegin = line.crbegin();
   const auto rend = line.crend();
   if (rbegin != rend) {
-    // Lines already exclude \n, so we can check using std::isspace.
     const auto reverse_iter =
-        std::find_if(rbegin, rend, [](char c) { return !std::isspace(c); });
+        std::find_if(rbegin, rend, [](char c) { return !std::isblank(c); });
```
hzeller, line 67: After you've removed the `\r` above, do you still need the change from isspace to isblank ? I suspect now the original code will just work.

verilog/analysis/checkers/no_trailing_spaces_rule_test.cc
```diff
@@ -42,6 +42,7 @@ TEST(NoTrailingSpacesRuleTest, AcceptsText) {
       {"  xxx\n\n"},
       {"\txxx\n\n"},
       {"  xxx\n  yyy\n"},
+      {"  xxx\n  yyy\r\n"},
```
hzeller, line 45: Now that the end of line is two characters long, another good corner case to consider would be to check that two consecutive `\r\n` don't create trouble as an additional test-case.
`{"  xxx\n  yyy\r\n\r\n"},`

verilog/analysis/checkers/no_trailing_spaces_rule_test.cc
```diff
@@ -62,6 +63,7 @@ TEST(NoTrailingSpacesRuleTest, RejectsTrailingSpaces) {
       {"a  b", {kToken, "   "}, "\n\n1  2", {kToken, "  "}, "\n"},
       {"module foo;", {kToken, " "}, "\nendmodule\n"},
       {"module foo;\nendmodule", {kToken, " "}, "\n"},
+      {"module foo;\nendmodule", {kToken, "  "}, "\r\n"},
```
hzeller, line 66: Similar here. We essentially need to check that the newline tokenization works as it should (which happens somewhere in the lexer, but we need to check that it doesn't get confused with `\r\n`. We expect not, but adding a test-case will help us make sure.
`{"module foo;\nendmodule", {kToken, "  "}, "\r\n\r\n"},`

hzeller (changes requested):
Nice. Maybe now we can even revert the original change now that the actual root cause, fixing the potential additional `\r` at the end, is fixed.
Also added some suggestions for potential additional test cases.

verilog/analysis/checkers/no_trailing_spaces_rule_test.cc
wsipak: Thank you, I've added the test case.

verilog/analysis/checkers/no_trailing_spaces_rule_test.cc
wsipak: I've added this to the test, thank you for suggesting the cases.

verilog/analysis/checkers/no_trailing_spaces_rule.cc
wsipak: Yes, definitely the previous one will be better as it matches a narrower set of characters. It's `isspace` again.

verilog/analysis/checkers/no_trailing_spaces_rule.cc
wsipak: Thank you! Good to know, especially that `ends_with` from `std` comes in C++20, so It wouldn't fit here.
Right now this code uses `absl::ConsumeSuffix`.

verilog/analysis/checkers/no_trailing_spaces_rule.cc
wsipak: The comments have been updated so that they match what actually happens in the code.

wsipak:
@hzeller  Thank you so much for the review. I think the current state of the branch includes all the requested changes. I have also updated the description of the PR as the approach I used initially is no longer in effect.

hzeller:
Thanks, merged!

Issue #902, opened 2021-09-01
Internal: Consider ways to extend and rename/alias rules


hzeller:
In #901 the `undersized-binary-literal` rule was extended to not only understand binary but also octal and hex numbers.

While in this case, the naming still somewhat makes sense (after all, we provide binary numbers), there might be other cases in which renaming a rule to reflect the wider range might be desirable (`undersized-numeric-literal` ?), while also having a backward compatible 'old' name, specifying an alias 
```
undersized-binary-literal ::= undersized-numeric-literal with configuration="bin:true"
```

There should be an easy, low-boilerplate overhead in the code to do that.

PR #935, opened 2021-09-17
Allow rule aliasing

wsipak:
This is a proposal of a solution for #902.
With these changes, we can use an alternative name for a rule.
This is how aliases can be defined:
```
VERILOG_REGISTER_LINT_RULE(UndersizedBinaryLiteralRule);
VERILOG_ALIAS_LINT_RULE(UndersizedBinaryLiteralRule, "undersized-numeric-literal", "some-other-alias");
```

`--help_rules=all` prints the aliases for each rule where they exist.
When `--help_rules` option is used with a specific name, the result is the same regardless of if the original name or alias was used.


hzeller:
We might also need configuration.

The issue in #902 is not completely hashed out yet. The problem we'd like to solve is

  * we start out with a rule with a particular name that does something
  * Later, some more functionality is added as it is closely related ,and this can then be switched and configured using the configuration.
  * Now, the original name does not necessarily fit anymore, but we want to keep it around as people might use it. Also, we probably have to keep most of the configuration disabled to not confuse existing users
  * Ideally, we now can define different rules with a different name that are essentially just call the old rule, but with a potentially different configuration.
  * There might be all kinds of corner-cases that have to be thought of: in messages for instance, we'd need the new name to display. Maybe by having a different `LintRuleDescriptor` ? So from that perspective, it just looks like a different rule, but goal of #902 should be to reduce the necessary boilerplate to do so.

So before we can come to a definitive implementation conclusion, we should discuss if there are other corner cases to consider, how would we like the such aliases define etc.

mglb:
>     * There might be all kinds of corner-cases that have to be thought of: in messages for instance, we'd need the new name to display.

"New name" is always the primary name of the rule implementation, right? So assuming `undersized-binary-literal ::= undersized-numeric-literal with configuration="bin:true"`, do both `+undersized-binary-literal` and `+undersized-numeric-literal=bin:true` would display `undersized-numeric-literal`?

If so, translating aliases in a configuration reading code should cover all cases. The rest of codebase wouldn't even know that alias has been used.

> Maybe by having a different `LintRuleDescriptor`?

Or something like:

```c++
struct LintRuleAliasDescriptor {
  LintRuleId name;             // ID/name of the alias.
  absl::string_view topic;     // section in style-guide
  absl::string_view dv_topic;  // section in design verification style-guide
  std::string desc;            // Detailed description
  std::vector<std::pair<absl::string_view, absl::string_view>>
      param_defaults;  // List of param values set by the alias
};

// ...

const std::vector<LintRuleAliasDescriptor>&
UndersizedNumericLiteralRule::GetAliasDescriptors() {
  static const std::vector<LintRuleAliasDescriptor> d{
      {
          .name = "undersized-binary-literal",
          // Empty .topic -> the same as the original rule's
          // Empty .desc -> something like:
          //   "Alias for <original_rule_name> with config=<param>:<value>"
          .param_defaults =
              {
                  {"bin", "true"},
              },
      },
  };
  return d;
}

```
`VERILOG_REGISTER_LINT_RULE` could even use SFINAE to detect presence of `GetAliasDescriptors` function, limiting registration to just one macro call.

wsipak:
1. The current state solves the problem of configuration and doesn't yetsolve the problem of printing messages with the right names. Rules can be aliased without changing the way they're registered. Following @mglb's ideas, aliasing a rule requires adding a function that returns a vector of alias descriptors.

2. (about mixing original names with aliases in configuration)

Right now it's legal to use the same rule several times in configuration. You can disable it, enable, disable again, then apply addiontional changes on top of it using the comandline arguments. This is processed in order, the defaults first, then the config file, then the arguments.

Should we disallow this (or warn about this) when we introduce aliases?

Example case:
```
+rule1         // rule1 is enabled
-rule1-alias   // rule1 is disabled
+rule1         // rule1 is enabled
```
It's okay, would work the same as without aliases.

Another case:
```
+rule1-alias   // rule1 is enabled, with default parameters of the alias
-rule1         // rule1 is disabled
```
Now, if we enable the rule again, either by its original name or an alias, should we care about the previously set parameter values? Should we delete them? Of course this case may make no sense in practical view, but as Verible allows to enable and disable rules multiple times, do we need to address this additionally?

3. (about printing messages with the right name)

Aliases are translated to the original names early, so that there's no need to create double rules and keep track of whether the original one, or the aliased one is enabled or disabled.

The name of a rule (printed when Verible finds violations) is stored in a const structure, like this one:
https://cs.opensource.google/verible/verible/+/master:verilog/analysis/checkers/undersized_binary_literal_rule.cc;l=56

In order to implement printing the right name (alias, if alias was used, the original name otherwise) we could change this to variable and overwrite the `.name` field, but I expect that It would be better to keep it as a constant.

We could add a field for the name that would actually be printed in logs (like `printed_name`) and assign the value of either `.name` from the descriptor of the rule, or descriptor of an alias, depending on which was used in configuration.

Alternatively, we can store information about which name shall be printed in an additional container, (not in each rule).

hzeller:
I like this implementation so far. I'll play around with it a bit and report back.

PR #939, opened 2021-09-23
Provide UI choice if there are multiple autofix alternatives.

hzeller:
Most lint rules with autofix have exactly one choice of fixes,
but there are some that have 2 or even three alternatives.

Provide a way to choose if there are multiple options.
The short choices are then including the digits
'1', '2', etc. The default 'y' is equivalent to choosing
choice 1.

If there are multiple options, print all of these and
mark as "1. alternative", "2. alternative", etc.

Also while at it: improve interactive sessions by always
printing the available patch first (or patches, if there are
alternatives), as the user typically will not
choose blindly, so make the initial 'print' automatic.

### DIFF
--- a/verilog/analysis/verilog_linter.cc
+++ b/verilog/analysis/verilog_linter.cc
@@ -114,6 +114,18 @@ static void PrintFix(std::ostream& stream, absl::string_view text,
   verible::LineDiffsToUnifiedDiff(stream, diff, 1);
 }
 
+static void PrintFixAlternatives(std::ostream& stream, absl::string_view text,
+                                 const std::vector<verible::AutoFix>& fixes) {
+  const bool print_alternative_number = fixes.size() > 1;
+  for (size_t i = 0; i < fixes.size(); ++i) {
+    if (print_alternative_number) {
+      stream << "=== " << (i + 1)
+             << ". alternative =================================\n";
+    }
+    PrintFix(stream, text, fixes[i]);
+  }
+}
+
 void ViolationFixer::CommitFixes(absl::string_view source_content,
                                  absl::string_view source_path,
                                  const verible::AutoFix& fix) const {
@@ -162,14 +174,21 @@ void ViolationFixer::HandleViolation(
     return;
   }
 
+  static std::string_view previous_fix_conflict =
+      "The fix conflicts with "
+      "previously applied fixes, rejecting.\n";
+
   AnswerChoice answer;
-  for (;;) {
+  for (bool first_round = true; /**/; first_round = false) {
     if (ultimate_answer_ != AnswerChoice::kUnknown) {
       answer = ultimate_answer_;
     } else if (auto found = rule_answers_.find(rule_name);
                found != rule_answers_.end()) {
       answer = found->second;
     } else {
+      if (is_interactive_ && first_round) {  // Show the user what is available.
+        PrintFixAlternatives(*message_stream_, base, violation.autofixes);
+      }
       answer = answer_chooser_(violation, rule_name);
     }
 
@@ -180,13 +199,23 @@ void ViolationFixer::HandleViolation(
       case AnswerChoice::kApplyAllForRule:
         rule_answers_[rule_name] = AnswerChoice::kApply;
         [[fallthrough]];
-      case AnswerChoice::kApply:
-        // Apply first available fix
+      case AnswerChoice::kApply:  // Apply first available fix
         if (!fix->AddEdits(violation.autofixes[0].Edits())) {
-          VLOG(2) << "The fix conflicts with previously applied fixes, "
-                     "rejecting.";
-        } else {
-          (*message_stream_) << "(fixed)" << std::endl;
+          *message_stream_ << previous_fix_conflict;
+        }
+        break;
+
+      case AnswerChoice::kApplySecondAlternative:
+        if (violation.autofixes.size() > 1 &&
+            !fix->AddEdits(violation.autofixes[1].Edits())) {
+          *message_stream_ << previous_fix_conflict;
+        }
+        break;
+
+      case AnswerChoice::kApplyThirdAlternative:
+        if (violation.autofixes.size() > 2 &&
+            !fix->AddEdits(violation.autofixes[2].Edits())) {
+          *message_stream_ << previous_fix_conflict;
         }
         break;
 
@@ -200,8 +229,7 @@ void ViolationFixer::HandleViolation(
         return;
 
       case AnswerChoice::kPrintFix:
-        // Print first available fix
-        PrintFix(*message_stream_, base, violation.autofixes[0]);
+        PrintFixAlternatives(*message_stream_, base, violation.autofixes);
         continue;
       case AnswerChoice::kPrintAppliedFixes:
         PrintFix(*message_stream_, base, *fix);
@@ -218,7 +246,7 @@ void ViolationFixer::HandleViolation(
 ViolationFixer::AnswerChoice ViolationFixer::InteractiveAnswerChooser(
     const verible::LintViolation& violation, absl::string_view rule_name) {
   static absl::string_view help_message =
-      "y - apply fix\n"
+      "y - apply first fix [or type number of alternative e.g. '2']\n"
       "n - reject fix\n"
       "a - apply this and all remaining fixes for violations of this rule\n"
       "d - reject this and all remaining fixes for violations of this rule\n"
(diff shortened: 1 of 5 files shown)


verilog/analysis/verilog_linter.cc
```diff
@@ -228,19 +256,36 @@ ViolationFixer::AnswerChoice ViolationFixer::InteractiveAnswerChooser(
...
+
   for (;;) {
     const char c = verible::ReadCharFromUser(
         std::cin, std::cerr, verible::IsInteractiveTerminalSession(),
-        "Autofix is available. Apply? [y,n,a,d,A,D,p,P,?] ");
+        "Autofix is available. Apply? [" + alternative_list +
+            "y,n,a,d,A,D,p,P,?] ");
 
     switch (c) {
+      case '1':
```
fangism, line 273: I think it might be cleaner to handle numeric inputs in a separate construct, outside of (after) this case statement.  It feels awkward to have to enumerate and limit choices to 3 even if that is sufficient in practice.

fangism (changes requested):

verilog/analysis/verilog_linter.cc
hzeller: I wanted to avoid expanding the notion of [AnswerChoice](https://github.com/chipsalliance/verible/pull/939/files#diff-1a62db1c3e48ce6fae3dbe0110a88c70bdbb2cd3e5b71fdf4e4769a26b6440d5R174-R177) from an enum to something else though I could be convinced otherwise (but it would be overkill at this point).

I could imagine something like the following to return from this chooser
```
  struct Answer {
     AnswerChoice choice;
     int alternative = 0;
 }
```

It _could_ be useful for the 'apply all' situation to also convey the potential alternatives.

Then, when we have such a struct, it would also warrant making this switch/case more specific if it sees some digit.

hzeller: Alright, changed the Answer to be a more complex struct like suggested above to convey the alternative.

However, this switch/case didn't change much, as we still handl just single-character interactions (though added TODO to potentially reconsider that for an 'apply all always the second alternative' kind-of situations)

hzeller:
In a follow-up change I plan to make some basic highlighting of the session (not color of diff yet, but better visibility when going through the lint outputs). After this is submitted.

interactive-lint-apply

verilog/analysis/verilog_linter.cc
```diff
@@ -215,10 +234,10 @@ void ViolationFixer::HandleViolation(
   }
 }
 
-ViolationFixer::AnswerChoice ViolationFixer::InteractiveAnswerChooser(
+ViolationFixer::Answer ViolationFixer::InteractiveAnswerChooser(
     const verible::LintViolation& violation, absl::string_view rule_name) {
   static absl::string_view help_message =
-      "y - apply fix\n"
+      "y - apply first fix [or type number of alternative e.g. '2']\n"
```
fangism, line 240: maybe show help on a separate line:

`[2-9]` apply alternative fix

verilog/analysis/verilog_linter.cc
```diff
@@ -228,36 +247,54 @@ ViolationFixer::AnswerChoice ViolationFixer::InteractiveAnswerChooser(
       "P - show fixes applied in this file so far\n"
       "? - print this help and prompt again\n";
 
+  // If we have alternatives, show this in the short-menu.
+  std::string alternative_list;
+  if (size_t fix_count = violation.autofixes.size(); fix_count > 1) {
+    for (size_t i = 0; i < fix_count; ++i)
+      alternative_list += std::to_string(i + 1) + ",";
```
fangism, line 254: https://abseil.io/tips/3 -- maybe pass through StrCat, instead of operator+?

verilog/analysis/verilog_linter.cc
```diff
@@ -228,36 +247,54 @@ ViolationFixer::AnswerChoice ViolationFixer::InteractiveAnswerChooser(
...
+  }
+
   for (;;) {
     const char c = verible::ReadCharFromUser(
         std::cin, std::cerr, verible::IsInteractiveTerminalSession(),
-        "Autofix is available. Apply? [y,n,a,d,A,D,p,P,?] ");
+        "Autofix is available. Apply? [" + alternative_list +
+            "y,n,a,d,A,D,p,P,?] ");
```
fangism, line 261: maybe insert `1-9` among the choices?

verilog/tools/lint/lint_tool_test.sh
```diff
@@ -887,4 +887,60 @@ interactive_autofix_test "@^(y***D" \
...
 ################################################################################
 
+echo "=== Test --autofix=inplace-interactive: Choose alternative fix"
+
+# Choose alternatives with key '1' or '2'
+
+# Files with alternatives in autofixes
+ORIGINAL_ALT_AUTO_FIX="${TEST_TMPDIR}/orig-autofix-alternative.sv"
+>"${ORIGINAL_ALT_AUTO_FIX}"
+echo -en "module AlternativeAutoFix;\n" >>"${ORIGINAL_ALT_AUTO_FIX}"
```
fangism, line 897: For multi-line echo sequences, I recommend the here-doc form:

```
cat > OUTPUT <<EOF
...
EOF
```

verilog/tools/lint/lint_tool_test.sh
```diff
@@ -887,4 +887,60 @@ interactive_autofix_test "@^(y***D" \
...
+
+# Choosing first non-existing alternative, then an existing one.
+cp ${ORIGINAL_ALT_AUTO_FIX} ${TEST_FILE}
+"$lint_tool" --ruleset=none --rules="undersized-binary-literal=hex:true" \
+             --autofix=inplace-interactive \
+             "${TEST_FILE}" > /dev/null 2>&1 \
+             <<< "41"
+
+check_diff "${REFERENCE_ALT_AUTO_FIX_1}" "${TEST_FILE}" "${DIFF_FILE}" \
+    "First alternative not coosen."
```
fangism, line 932: chosen

verilog/tools/lint/lint_tool_test.sh
```diff
@@ -887,4 +887,60 @@ interactive_autofix_test "@^(y***D" \
...
+
+failure=0
+cp ${ORIGINAL_ALT_AUTO_FIX} ${TEST_FILE}
+"$lint_tool" --ruleset=none --rules="undersized-binary-literal=hex:true" \
+             --autofix=inplace-interactive \
+             "${TEST_FILE}" > /dev/null 2>&1 \
+             <<< "1"
+
+check_diff "${REFERENCE_ALT_AUTO_FIX_1}" "${TEST_FILE}" "${DIFF_FILE}" \
+    "First alternative not coosen."
```
fangism, line 921: chosen

verilog/tools/lint/lint_tool_test.sh
```diff
@@ -887,4 +887,60 @@ interactive_autofix_test "@^(y***D" \
...
+(( failure|="$?" ))
+
+cp ${ORIGINAL_ALT_AUTO_FIX} ${TEST_FILE}
+"$lint_tool" --ruleset=none --rules="undersized-binary-literal=hex:true" \
+             --autofix=inplace-interactive \
+             "${TEST_FILE}" > /dev/null 2>&1 \
+             <<< "2"
+check_diff "${REFERENCE_ALT_AUTO_FIX_2}" "${TEST_FILE}" "${DIFF_FILE}" \
+    "Second alternative not coosen."
```
fangism, line 941: chosen

fangism (approved):

verilog/analysis/verilog_linter.cc
hzeller: The available choices are in there. So if if there are two choices, it will `[1,2,y,n...]`

verilog/analysis/verilog_linter.cc
hzeller: Fixed, using `absl::StrAppend()`

verilog/tools/lint/lint_tool_test.sh
hzeller: Done (originally, I used the other way as it was commonly used in this file, mostly because the other examples were using various whitespace-relevant examples; but not needed in this case)

verilog/analysis/verilog_linter.cc
hzeller: Done. Help message will now be generated depending on if there is an alternate fix and only inlcude such line if it is.

hzeller:
FYI @mglb : added UI choice for autofix alternatives.

tgorochowik:
@wsipak is this something we need to take into account in the github action?

hzeller:
I imagine for the action we should probably emit the sequence of Autofixes including nested alternatives as JSON, so that they are easy to parse and process and be independent of a hard-to-manually-parse terminal session.

wsipak:
The syntax of JSON-based Reviewdog Diagnostic Format (RDFormat) can handle multiple suggestions, those are stored in a list, and this part of the syntax is described here:
https://github.com/reviewdog/reviewdog/blob/master/proto/rdf/reviewdog.proto#L60
So, having multiple suggestions won't collide with our present solution.
However, we'll need to make sure that the action doesn't run Verible interactively, as there will be no input from `stdin`.
I think that's all we need to consider about this topic.

mithro:
@hzeller - Outputting results in a format compatible with Reviewdog seems like a great idea.

PR #942, opened 2021-09-23
Add autofix-hints to all autofixes.

hzeller:
Since we have a chance to show a description of a lint autofix
in the potential UI or github pull, add descriptions to all of
them.

Remove the AutoFix constructors that allow to skip adding a
description, thus AutoFix implementors are always required
to add a description.

### DIFF
--- a/verilog/tools/lint/README.md
+++ b/verilog/tools/lint/README.md
@@ -1,7 +1,7 @@
 # SystemVerilog Style Linter
 
 <!--*
-freshness: { owner: 'hzeller' reviewed: '2020-10-07' }
+freshness: { owner: 'hzeller' reviewed: '2021-09-23' }
 *-->
 
 The `verible-verilog-lint` SV style linter analyzes code for patterns and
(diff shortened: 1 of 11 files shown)


verilog/tools/lint/README.md
```diff
@@ -124,17 +124,30 @@ The `--rules` flag allows to enable/disable rules as well as pass configuration
 to rules that accept them. It accepts a comma-separated list
 [rule names][lint-rule-list]. If prefixed with a `-` (minus), the rule is
 disabled. No prefix or a '+' (plus) prefix enables the rule. An optional
-configuration can be passed after an `=` assignment.
+configuration can be passed after an `=` assignment. Each name/value is
+
+So rule configurations with parameters looks like this
+```
+  --rules=[+-]rule-name=<param>:<paramvalue>;<param2>:<value>,[+-]next-rule...
```
fangism, line 131: Possible suggestion: `;` is a shell token, so the flag may need to be single- or double-quote protected, as you have an in example below.

fangism (approved):

verilog/tools/lint/README.md
hzeller: Good idea. Uupdated.

PR #945, opened 2021-09-23
Fix some low-hanging fruit reducing global static initialization.

hzeller:
Fixed some constructors to be constexpr so that static const
objects can be changed to static constexpr.

However, the biggest culprits are still ABSL_FLAG and
VERILOG_REGISTER_LINT_RULE.


fangism (approved):
Nice!!!

PR #946, opened 2021-09-23
Apply some clang-tidy performance findings.

hzeller:
Most notably it complained that return values that are
'const' can't be returned by an automatic move, so
would have to be copied. Typically, that happens
with absl::Status in our code base.

Changed common pattern

```
const absl::Status status = foo();
if (status.ok()) return status;
```

To something that does not const-qualify, but limit
the scope

```
if (absl::Status status = foo(); !status.ok()) {
  return status;
}
```


fangism (approved):
Good catches with the non-const auto status, and noexcept!

PR #969, opened 2021-10-04
Add protocol base support for the language server protocol.

hzeller:
* JsonRpcDispatcher is implementing the needed functionality
     for JSON-RPC [1] calling notifications and RPC-calls.
   * MessageStreamSplitter is needed in the LSP [2] specifically
     as the transport layer separates each message with a
     header (containing a Content-Length: header) and a body
     (very similar to HTTP).
    
[1] https://www.jsonrpc.org/specification
[2] https://microsoft.github.io/language-server-protocol/specification

### DIFF
--- a/.github/bin/run-clang-tidy.sh
+++ b/.github/bin/run-clang-tidy.sh
@@ -34,9 +34,12 @@ cat bazel-bin/compile_commands.json \
 # for now.
 # Also, exclude kythe for now, as it is somehwat noisy and should be
 # addressed separately.
+# common/lsp excluded as the compilation database generation gets confused
+# with pure libraries not yet wired up to a binary (it is otherwise tidy-clean)
 find . -name "*.cc" -and -not -name "*test*.cc" \
      -or -name "*.h" -and -not -name "*test*.h" \
   | grep -v "verilog/tools/kythe" \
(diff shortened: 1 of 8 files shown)


hzeller:
Not sure why codecov has not commented yet, maybe they do rate limiting as there was another pull request ongoing ?
Anyway, we have 100% incremental coverage.

hzeller:
Since this is libraries only right now, here is an example how both of these would be wired up to make a working LSP service:

https://github.com/hzeller/bare-lsp/blob/main/main.cc#L76-L91

hzeller:
An example how it is possible to wire up handlers that have properly typed parameters is seen in the [next pull request](https://github.com/chipsalliance/verible/pull/971/files/e0adfec547ed5840420ee9e9f82bf046be0de9b9..72afa8eb8a28e92c40d4cf48119799a366bfaf7a#diff-58c1a3cb6e15d9f2c16ff455ecc72692c403d6dfa3a8266358986dd0f24feec6R125-R133) - the types there are generated with all the necessary boilerplate to be auto-converted from `nlohmann::json`, so the register handler function that take a `nlohmann::json` parameter also accept functions that take a convertible parameter. Very neat way to use auto-conversion in C++ for a change.

PR #970, opened 2021-10-04
Skip permission test if tests are run as root.

### LINKED ISSUE #963: test fail when I use sudo bazel test -c opt //...
//common/util:file_util_test                                             FAILED in 0.1s
  /home/banana/.cache/bazel/_bazel_root/ef4da2825d7ad34a78c6af14a6b3ef53/execroot/com_google_verible/bazel-out/k8-opt/testlogs/common/util/file_util_test/test.log

Executed 1 out of 551 tests: 550 tests pass and 1 fails locally.
There were tests whose specified size is too big. Use the --test_verbose_timeout_warnings command line option to see which ones these areINFO: Build completed, 1 test FAILED, 2 total actions

hzeller:
With root permissions, even a non-readable file will be readable.
So if for whatever reason the bazel test is run with root permission,
skip the test.

Fixes #963

### DIFF
--- a/common/util/file_util_test.cc
+++ b/common/util/file_util_test.cc
@@ -130,17 +130,16 @@ TEST(FileUtil, StatusErrorReporting) {
   // this test here.
   // TODO: Can we make permission-denied test that works on Windows ?
 
-  // Enforce a permission denied situation
-  // TODO: Issue #963 - if this turns out to not always succeed to chmod
-  // we should just skip the test.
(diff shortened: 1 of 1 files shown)


mithro:
Can we just fail with a "Don't run the tests as root" type message?

hzeller:
The test is sortof doing that right now - it fails when running as root :)
But it should not fail due to that, as our tests only want to test the functionality of the library but not the broken environment.

fangism (approved):

PR #972, opened 2021-10-06
line length rule: add a parameter to check comments

wsipak:
This addresses https://github.com/chipsalliance/verible/issues/941
There's a new parameter ~~`allow_comments`~~ `ignore_comments`.
When it's set to `false`, exceeding the line length limit in comments won't be forgiven.
The default value is `true`, so that it doesn't change the way it works by default now.
It works with single line comments and block comments, but those are not distinguished in configuration right now (Should we consider them separately?).

Also, would it be preferred to rename the parameter to something like `check_comments` so that it has the opposite meaning, and set to `false` would result in **not** linting comments?

Would it be better to have two separate values of length limit for non-comment and comment lines?
This probably could be addressed and implemented in another PR, but I'm noting it here as a thing to consider.

### DIFF
--- a/verilog/analysis/checkers/line_length_rule.cc
+++ b/verilog/analysis/checkers/line_length_rule.cc
@@ -189,11 +189,12 @@ static verible::TokenRange SeekLastNonNewlineToken(
   // text_begin is the frontal limiter that shall not be overrun.
   // If such a token isn't found, the input range is returned
   auto found_it = std::find_if(
-      std::reverse_iterator(token_range.begin()),
-      std::reverse_iterator(text_begin),
+      std::reverse_iterator<TokenSequence::const_iterator>(token_range.begin()),
+      std::reverse_iterator<TokenSequence::const_iterator>(text_begin),
       [](const TokenInfo& t) { return t.token_enum() != TK_NEWLINE; });
 
-  if (found_it != std::reverse_iterator(text_begin)) {
+  if (found_it !=
+      std::reverse_iterator<TokenSequence::const_iterator>(text_begin)) {
     // Use the last non-newline token as a new beginning.
     auto new_begin = found_it.base() - 1;
     // We move the front backwards, so now it's possible that there are


hzeller:
Yes, I think for the user's point of view a parameter `check_comments` would be easiest to understand.

wsipak:
@hzeller Thank you for your feedback. The parameter is now named `check_comments`.

verilog/analysis/checkers/line_length_rule.cc
```diff
@@ -70,16 +70,21 @@ const LintRuleDescriptor& LineLengthRule::GetDescriptor() {
       .desc =
           "Checks that all lines do not exceed the maximum allowed "
           "length. ",
-      .param = {{"length", absl::StrCat(kDefaultLineLength),
-                 "Desired line length"}},
+      .param =
+          {
+              {"length", absl::StrCat(kDefaultLineLength),
+               "Desired line length"},
+              {"check_comments", "false", "Check comments."},
```
hzeller, line 77: Did we previously check the length in comments ? Then maybe this should default to 'true'.

hzeller (commented):
What happened to this PR, looks like we were close ?
@wsipak do you still have the branch to finish ?

verilog/analysis/checkers/line_length_rule.cc
```diff
@@ -158,16 +165,67 @@ static bool AllowLongLineException(TokenSequence::const_iterator token_begin,
...
     const int observed_line_length = verible::utf8_len(line);
     if (observed_line_length > line_length_limit_) {
-      const auto token_range = text_structure.TokenRangeOnLine(lineno);
+      // range of tokens that *begin* in this line.
+      auto token_range = text_structure.TokenRangeOnLine(lineno);
       // Recall that token_range is *unfiltered* and may contain non-essential
       // whitespace 'tokens'.
+      // This shrinks the range if there are leading newline tokens
+      token_range = StripNewlineTokens(token_range);
+      if (token_range.begin() == token_range.end()) {
```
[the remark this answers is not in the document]
jbylicki: Done

verilog/analysis/checkers/line_length_rule.cc
```diff
@@ -158,16 +165,67 @@ static bool AllowLongLineException(TokenSequence::const_iterator token_begin,
...
+  auto reversed = reversed_view(token_range);
+  auto last_non_newline =
+      std::find_if(reversed.begin(), reversed.end(), [](const TokenInfo& t) {
+        return t.token_enum() != TK_NEWLINE;
+      }).base();
+
+  return make_range(token_range.begin(), last_non_newline);
+}
+
+static verible::TokenRange SeekTokensBackwards(
```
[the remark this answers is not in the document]
jbylicki: Good idea, changed it

verilog/analysis/checkers/line_length_rule.cc
```diff
@@ -165,15 +172,75 @@ static bool AllowLongLineException(TokenSequence::const_iterator token_begin,
...
+}
+
+verible::TokenRange& CheckTokenRangeForTrailingNewlines(
+    verible::TokenRange* orig_range,
+    TokenSequence::const_iterator text_structure) {
+  // Recall that token_range is *unfiltered* and may contain non-essential
+  // whitespace 'tokens'.
+  // This shrinks the range if there are leading newline tokens
+  *orig_range = StripNewlineTokens(*orig_range);
+  if (orig_range->begin() == orig_range->end()) {
```
kbieganski, line 214:
```suggestion
  if (!std::empty(*orig_range)) return *orig_range; 
```
Maybe something like this? Then just return `StripNewlineTokens(...)` in the other branch.

kbieganski (changes requested):

verilog/analysis/checkers/line_length_rule.cc
jbylicki: It does not implement enough methods to call either `std::empty` or `TokenRange.empty()` and your proposal (along with variations I have tried) does not compile. The implementation is at `common/util/iterator_range.h`.

PR #983, opened 2021-10-10
Remove file event dispatcher.

hzeller:
This is hard to get to work on Windows as that OS doesn't support
that concept. Instead, this shall be solved differently.


hzeller:
This will also fix the centos build-failures for the master-branch CI.

PR #998, opened 2021-10-16
Add baseline language server.

hzeller:
This adds a basic language server, based on the dummy language
server in common/lsp.

The basic language server keeps track of the file content
with EditTextBuffer, this adds a corresponding family of
classes lsp-parse-buffer.{h,cc}, tied to the first with event
listeners, that updates an internal state with parsed and linted
options, to be used by the services on top of that.

The parse buffers are note directly tested in unit-tests
(they are essentially just delegating to well-tested
verible internals), but are thoroughly covered by the
verible-verilog-ls_test.sh.


hzeller:
it looks like a lot of lines, but the language server and its test is mostly the same as the one in common/lsp

PR #999, opened 2021-10-17
Put the verible logo into the headline of the toplevel README.

hzeller:


hzeller:
There is a [view file for the README](https://github.com/chipsalliance/verible/blob/ff4bc072c972cde56ac84ee2ded62a578a84eb6c/README.md) option that can show how it looks.

I considred removeing the `V` from the headline to have just that letter speak for itself, but it would not be accessible and also the Kerning is totaly wrong.

PR #1019, opened 2021-10-18
smoke-test: print out the issues that might be fixed now.

hzeller:
From all the tool/file issues that were not encountered
anymore, create a unique set of referenced bugtracker
issues and print their links for quick navigation.


hzeller:
Looking at https://github.com/chipsalliance/verible/pull/933/checks?check_run_id=3928387313#step:6:1142 I noticed that it would actually be nice if the referenced bug numbers were displayed.

PR #1020, opened 2021-10-18
Multiple compiler test: add a compilation with clang

hzeller:
.. as this could've found the accidental dead code issue
that made it into the repository, then fixed in #1001


hzeller:
(Draft for now until clang versions have been figured out with ubuntu 18 and until the test-code to trigger the warning has been removed)

hzeller:
I verified that the issue fixed in #1001 would in fact have been caught by this.

PR #1067, opened 2021-10-29
Make windows bazel cache work.

hzeller:
Do a blaze shutdown so that the cache storage can access files on
Windows.
    
Possibly due to non-posix filesystem semantics, it seems to be
impossible to package files that are still open on Windows. A blaze
shutdown helped the situation.
    
While at it, also do a blaze shutdown in the general build matrix,
can't harm.


mithro (approved):
LGTM.

PR #1069, opened 2021-10-29
Only clang-tidy files whose content hash changed

hzeller:
clang-tidy takes almost 30min in the CI, so let's remember the
content hashes of all files we processed so that we only have to
look at changed files.


hzeller:
Here is an example where the 30min clang-tidy now only takes 15 seconds as there were no changed files:
https://github.com/chipsalliance/verible/runs/4048584990?check_suite_focus=true#step:7:16

mithro (approved):
LGTM

PR #1070, opened 2021-10-29
Put the smoke-test results in the toplevel output.

hzeller:
That way, it is not necessary to unfold the part that
might contain interesting fixed issues.


mithro:
@hzeller Do you have some screenshots or links to it working?

hzeller:
Here is a link to the summary section in the smoke test
https://github.com/chipsalliance/verible/runs/4047767825?check_suite_focus=true#step:7:695

mithro (approved):
LGTM.

PR #1071, opened 2021-10-30
Fix sed in lint_tool_test.sh on macOS

mglb:
macOS doesn't understand `;`:
```
sed: 1: ":join;N;$!b join; s/ /� ...": unused label 'join;N;$!b join; s/ /·/g;s/	/├───/g;s/\r/⁋/g;s/\n/¶\n/g'
```

### DIFF
--- a/verilog/tools/lint/lint_tool_test.sh
+++ b/verilog/tools/lint/lint_tool_test.sh
@@ -348,6 +348,20 @@ EOF
 "$lint_tool" --ruleset=none --rules_config="${RULES_CONFIG_FILE}" --autofix=inplace \
     "${TEST_FILE}" > /dev/null 2>&1
 
+visualize_invisible_characters() {
+  local -r TAB=$'\t'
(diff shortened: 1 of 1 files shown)


verilog/tools/lint/lint_tool_test.sh
```diff
@@ -359,8 +371,8 @@ check_diff() {
 
   # Show invisible characters
   local invisible_chars_expr_=':join;N;$!b join; s/ /·/g;s/'$'\t''/├───/g;s/'$'\r''/⁋/g;s/\n/¶\n/g'
```
hzeller, line 373: Do you still need this `invisible_chars_expr_` now ?

mglb: Line removed

mglb:
I did a few more changes to the sed call to make it fully posix-compliant and added error detection.

hzeller (approved):

PR #1077, opened 2021-11-01
LanguageServer: Create symbol tree outline in editor

hzeller:
We adapt the verible internals with the language server by
walking the parsed structure and emit some select DocumentSymbols.
For now, these are just higher level structural components, but more
can be added easily.

This is the first baseline implementation and is planned to be re-structured to call callbacks that then allow easy testing and building of the output structure.


common/lsp/lsp-protocol-enums.h
```diff
@@ -0,0 +1,57 @@
...
+// Enums are currently not handled yet by jcxxgen, so enumeration are
+// done separately here.
+namespace verible {
+namespace lsp {
+// These are the SymbolKinds defined by the LSP specifcation
+// https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#symbolKind
+// Interesting is, that not all of them are actually supported by all
+// editors (playing around with Kate (https://kate-editor.org/)). So maybe
+// these editors need to be made understanding.
+enum SymbolKind {
```
hzeller, line 27: Deliberately no `enum class` as this enum is used as set of integer constants in the context of the Language Server Protocol, so `static_cast<int>`-ing them would be needed every time

fangism: Ok, add this comment too.

common/lsp/lsp-protocol-enums.h
```diff
@@ -0,0 +1,57 @@
...
+  Field = 8,
+  Constructor = 9,
+  Enum = 10,  // SV enum
+  Interface = 11,
+  Function = 12,  // SV function
+  Variable = 13,
+  Constant = 14,
+  String = 15,
+  Number = 16,
+  Boolean = 17,
```
fangism, line 44: How is Boolean not already a Variable? (same general question about the classifications of Variable, Constant, Number)

common/lsp/lsp-protocol-enums.h
```diff
@@ -0,0 +1,57 @@
...
+  File = 1,
+  Module = 2,  // SV module. Kate does not seem to support that ?
+  Namespace = 3,
+  Package = 4,  // SV package
+  Class = 5,    // SV class
+  Method = 6,   // SV class -> method
+  Property = 7,
+  Field = 8,
+  Constructor = 9,
+  Enum = 10,  // SV enum
```
fangism, line 37: Enum type? (do distinguish from EnumMember, below)

verilog/tools/ls/document-symbol-filler.cc
```diff
@@ -0,0 +1,157 @@
...
+      }
+      break;
+    }
+
+    default:
+      is_visible_node = false;
+      break;
+  }
+
+  // Independent of visible or not, we always descent to our children.
```
fangism, line 119: descend

verilog/tools/ls/document-symbol-filler.h
```diff
@@ -0,0 +1,48 @@
...
+
+ protected:
+  void Visit(const verible::SyntaxTreeLeaf &leaf) final;
+  void Visit(const verible::SyntaxTreeNode &node) final;
+
+ private:
+  verible::lsp::Range RangeFromLeaf(const verible::SyntaxTreeLeaf &leaf) const;
+  verible::lsp::Range RangeFromToken(const verible::TokenInfo &token) const;
+
+  const int kModuleSymbolKind;  // Might be different if kate-workaround.
```
fangism, line 41: static? or constexpr?

fangism (approved):

verilog/tools/ls/document-symbol-filler.h
hzeller: They are constants within the lifetime of the object , but chosen and set in the constructor depending on `kate_workaround`.

verilog/tools/ls/document-symbol-filler.cc
hzeller: Done.

common/lsp/lsp-protocol-enums.h
hzeller: Added comment (but note: currently they are actually not emitted yet)

common/lsp/lsp-protocol-enums.h
hzeller: They are all upper-case, so the C++ compiler does not have a name-clash. Or do you mean the general classification ? This is from the Language Server Protocol, so they slapped in all sorts of generic classifications that might be useful in tagging syntax tree nodes.

common/lsp/lsp-protocol-enums.h
hzeller: done.

PR #1078, opened 2021-11-01
Use clang-tidy-11 as this checks for google-runtime-references.

hzeller:
The mandate to not use Non-const references but pointers
have been removed from the google style-guide. However, for
readability this is still the preferred style in this project.
clang-tidy-11 still checks for that.


hzeller:
In #1076 the configuration is taken into account, so the change of the variable will flush the cache.

fangism (approved):

PR #1079, opened 2021-11-02
Add convenience function: TextStructureView::GetRangeForToken().

hzeller:
There are various places in which we need the line column range
for a token. Put that in one place.

### DIFF
--- a/common/text/text_structure.h
+++ b/common/text/text_structure.h
@@ -95,9 +95,13 @@ class TextStructureView {
 
   const LineColumnMap& GetLineColumnMap() const { return line_column_map_; }
 
+  // Given a byte offset, return the line/column
   LineColumn GetLineColAtOffset(int bytes_offset) const {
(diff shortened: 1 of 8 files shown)


common/text/text_structure.h
```diff
@@ -95,10 +95,14 @@ class TextStructureView {
 
   const LineColumnMap& GetLineColumnMap() const { return line_column_map_; }
 
+  // Given a byte offset, return the line/column
   LineColumn GetLineColAtOffset(int bytes_offset) const {
     return line_column_map_.GetLineColAtOffset(contents_, bytes_offset);
   }
 
+  // Convenience function: Given the token, return the range it covers.
+  LineColumnRange GetRangeForToken(const TokenInfo& token) const;
```
fangism, line 104: small request for a few unit-tests on this function

fangism (approved):

common/text/text_structure.h
hzeller: Done.

PR #1080, opened 2021-11-02
move ViolationHandler to common

wsipak:
This is a refactoring step before introducing Reviewdog Diagnostic Format in Verible:
https://github.com/chipsalliance/verible/pull/1018

`ViolationHandler` is the interface which is supposed to be implemented by `ViolationPrinter`, RDFormat printer, and other potential handlers.
This isn't specific to Verilog and thus we place it in `common`.


verilog/analysis/BUILD
```diff
@@ -227,14 +227,13 @@ cc_library(
         "//common/analysis:text_structure_linter",
         "//common/analysis:token_stream_lint_rule",
         "//common/analysis:token_stream_linter",
-        "//common/strings:diff",
+        "//common/analysis:violation_handler",
```
hzeller, line 230: I think the `verilog_linter_test` below also needs this dependency now.

verilog/analysis/verilog_linter_test.cc
```diff
@@ -34,6 +34,7 @@
 #include "absl/status/statusor.h"
 #include "absl/strings/match.h"
 #include "absl/strings/string_view.h"
+#include "common/analysis/violation_handler.h"
```
hzeller, line 37: ... because of this added include.

hzeller (changes requested):
LGTM!
Just one missing dependency.

hzeller:
If you add the dependency, this can be merged.

wsipak:
Now `verilog_linter_test` has `"//common/analysis:violation_handler"` as a dependency. Thank you.

hzeller:
LGTM!

PR #1081, opened 2021-11-02
Given a Line/Column, find token at that point.

hzeller:
This is needed for operations in the language server which
often start with a cursor position.

### DIFF
--- a/common/text/text_structure.cc
+++ b/common/text/text_structure.cc
@@ -147,6 +147,15 @@ TokenRange TextStructureView::TokenRangeOnLine(size_t lineno) const {
   }
 }
 
+TokenInfo TextStructureView::FindTokenAt(const LineColumn& pos) const {
+  if (pos.line < 0 || pos.column < 0) return EOFToken();
+  // Maybe do binary search here if we have a huge amount tokens per line.
+  for (const TokenInfo& token : TokenRangeOnLine(pos.line)) {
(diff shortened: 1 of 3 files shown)


common/text/text_structure.cc
```diff
@@ -147,6 +147,14 @@ TokenRange TextStructureView::TokenRangeOnLine(size_t lineno) const {
   }
 }
 
+TokenInfo TextStructureView::FindTokenAt(const LineColumn& pos) const {
+  if (pos.line < 0 || pos.column < 0) return EOFToken();
+  for (const TokenInfo& token : TokenRangeOnLine(pos.line)) {
+    if (GetRangeForToken(token).PositionInRange(pos)) return token;
+  }
```
fangism, line 154: Since positions are monotonic, can this be done with a binary search, like std::lower_bound or std::equal_range?

hzeller: I think this will be tricky as columns are counting (possibly multi-byte) characters and the underlying ranges are bytes.

(Though I didn't spend too much time thinking of it as the amount of tokens expected in a line is probably short enough to have linear search probably as good or better than binary search)

hzeller: Looking at it, yes, it is probably possible, but it will be more convoluted (as one has to compare tokens with Positions and a position comparison will be applied to the range of it), that I don't think it is worthwhile: More, harder to understand code for something that likely will not really benefit performance as we're talking about a handful tokens per line.

PR #1085, opened 2021-11-03
Add instructions how to integrate language server in vscode.

hzeller:
Thanks to @lromor

This is a bunch of files - It is unclear to me if this is the smallest possible way to add the extension. There are certainly things that can be removed to reduce it to the essence needed to integrate in VScode.

Requesting guidance from someone knowledgeable in VSCode extensions.
(till then: keep as draft)


hzeller:
This is good enough for now as there was no feedback so far.
It still looks like a lot of boilerplate files - Would be good if someone knowledgeable with VSCode can help reduce this to the essential files.

PR #1087, opened 2021-11-04
Basic implementation of language server document highlight.

hzeller:
If the cursor is on a symbol, this will highlight all the symbols
in the buffer with the same name (it does _not_ take naming
scopes into account yet).

### DIFF
--- a/verilog/tools/ls/verible-lsp-adapter.h
+++ b/verilog/tools/ls/verible-lsp-adapter.h
@@ -42,5 +42,13 @@ std::vector<verible::lsp::CodeAction> GenerateLinterCodeActions(
 nlohmann::json CreateDocumentSymbolOutline(
     const BufferTracker *tracker, const verible::lsp::DocumentSymbolParams &p,
     bool kate_compatible_tags = true);
+
+// Give a position in a document, return ranges in the buffer that should
+// be hi-lit.
+// Current implementation: if cursor is over a symbol, highlight all symbols
+// with the same name (NB: Does _not_ take scoping into account yet).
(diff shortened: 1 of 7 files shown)


hzeller:
This still needs a test, but if you'd like to have a look already...

hzeller:
(test added)

verilog/tools/ls/verible-lsp-adapter.h
```diff
@@ -42,5 +42,13 @@ std::vector<verible::lsp::CodeAction> GenerateLinterCodeActions(
 nlohmann::json CreateDocumentSymbolOutline(
     const BufferTracker *tracker, const verible::lsp::DocumentSymbolParams &p,
     bool kate_compatible_tags = true);
+
+// Give a position in a document, return ranges in the buffer that should
```
fangism, line 46: Given

verilog/tools/ls/verible-lsp-adapter.h
```diff
@@ -42,5 +42,13 @@ std::vector<verible::lsp::CodeAction> GenerateLinterCodeActions(
...
+// be hi-lit.
```
fangism, line 47: highlighted

fangism (approved):

verilog/tools/ls/verible-lsp-adapter.h
hzeller: done.

verilog/tools/ls/verible-lsp-adapter.h
hzeller: done.

PR #1097, opened 2021-11-09
Unify argument order in formatting functions

mglb:
In https://github.com/chipsalliance/verible/pull/1045#discussion_r741267915 constant `style` argument has been defined as a first argument. The PR applies the same order (i.e. constants/inputs before outputs) to similar functions that are used nearby.


hzeller (approved):
LGTM!

PR #1098, opened 2021-11-10
Support extended filelist format (+incdir+ & more types of comment) a…

ivan444:
…nd fix the handling of VirtualFiles inside VerilogProject.


hzeller:
For the formatting, you need to run `.github/bin/run-clang-format.sh` and then check in the changes.

verilog/analysis/verilog_project.cc
```diff
@@ -195,11 +197,22 @@ absl::StatusOr<VerilogSourceFile*> VerilogProject::OpenIncludedFile(
...
 
-void VerilogProject::AddVirtualFile(absl::string_view referenced_filename,
+void VerilogProject::AddVirtualFile(absl::string_view resolved_filename,
                                     absl::string_view content) {
+  std::string referenced_filename(resolved_filename);
+  // Make the virtual file relative to the include directory.
+  for (absl::string_view include_path : include_paths_) {
+    if (absl::StartsWith(resolved_filename, include_path)) {
+      referenced_filename = std::string(absl::StripPrefix(
+          resolved_filename, absl::StrCat(include_path, "/")));
```
hzeller, line 207: What happens if `include_path` already has a slash at the end ?

verilog/analysis/verilog_project.cc
```diff
@@ -195,11 +197,22 @@ absl::StatusOr<VerilogSourceFile*> VerilogProject::OpenIncludedFile(
   return inserted.first->second->Status();
 }
 
-void VerilogProject::AddVirtualFile(absl::string_view referenced_filename,
+void VerilogProject::AddVirtualFile(absl::string_view resolved_filename,
```
hzeller, line 200: Here I have some questions that might be clarified with comments and unit tests.

The implementation makes the assumption that, given a resolved name, we can get a 'shorter' referenced name because it is considered that one of the includes has the prefix, if not we key it by full resolved name. Could this be a problem ? Maybe a comment about the reasoning would be good.

It looks like we now key all files by referenced name, i.e. basename. Couldn't this result in issues if we have multiple files with the same basename but different paths ?

Let's say there are files
```
/foo/bar/baz.sv
/abc/xyz/baz.sv
```

and include paths `.`, `/foo/bar` and `/abc/xyz`. This `AddVirtualFile()` seems to add these under the key `bar.sv` (and possibly overwriting two `bar.sv` with last one winning ?).

Depending on the context, `baz.sv` might also need to be resolved differently. Say we have a file `/foo/bar/hello.sv` and `/abc/xyz/world.sv` which each include `baz.sv`.

Given the include search path (that has as first element `.`), the first one should open `/foo/bar/baz.sv`, the second one `/abc/xyz/baz.sv`. Will this work ?

I suspect that instead of the reference filename we need to key the files with the full filename ('resolved_filename') and whenever we get a reference file to open resolve it with the current context, which includes the dirname of the file we resolve it from (in which `.` has a different meaning).

hzeller (changes requested):
Mostly I have some questions which probably can be answered with comments and extensive tests of corner cases.

verilog/analysis/verilog_project.cc
ivan444: I agree that would be a better approach. But the includes are opened while processing AST. We don't know the current directory of the file that lists the include. If `.` is an implicit include dir as in C++, then we need to rework the AST visitor to propagate the current filename.

Let me change the PR back to resolving at the _opening_ time instead the _file adding_ time. The last time I missed to change `OpenTranslationUnit`. Let's see how the whole pipeline works if I change both of them.

verilog/analysis/verilog_project.cc
ivan444: This is obsolete.

verilog/analysis/verilog_project.cc
ivan444: Pushed a commit to resolve includes at the opening time instead the file adding time. This doesn't include the current directory, as that requires more plumbing.

verilog/analysis/verilog_project.cc
```diff
@@ -240,23 +292,41 @@ const VerilogSourceFile* VerilogProject::LookupFileOrigin(
...
+  constexpr absl::string_view kIncludeDirPrefix = "+incdir+";
+  FileList file_list_out;
+  file_list_out.file_list_path = std::string(file_list_path);
+  file_list_out.include_dirs.push_back(".");
+  std::string file_path;
+  std::istringstream stream(file_list_content);
+  while (std::getline(stream, file_path)) {
+    absl::RemoveExtraAsciiWhitespace(&file_path);
+    // Ignore blank lines
+    if (file_path.empty()) continue;
```
hzeller, line 306: Are there test-cases for each of these to-be-ignored line types ?

With 
```
MODE=coverage .github/bin/build-and-test.sh
```

... you can run a coverage and look at the resulting HTML (requires `lcov` pacakge to be installed)

(We do run coverage in the CI, but for whatever reason, codecov did not seem to have added it to this PR https://app.codecov.io/gh/chipsalliance/verible )

hzeller (approved):
As discussed offline, it would be good to also have a test that documents that current directory `.` includes will not work currently, as we don't have the path of the file that includes the referenced file available. Good starting point to fix that later.

Other than that, LGTM. Can be merged after the added test (and possibly coverage improvements if needed, see other comment).

verilog/analysis/verilog_project.cc
ivan444: Added a test.

ivan444:
> it would be good to also have a test that documents that current directory . includes will not work currently

Tried to add it. But the API doesn't support it at all. So it can't be expressed in a test. Instead, I created a Github issue that explains what's missing: https://github.com/chipsalliance/verible/issues/1117

PR #1125, opened 2021-12-21
Have GetSubtreeAsLeaf() return a pointer instead of a reference.

hzeller:
This way, if the subtree is nullptr, we don't check-fail right there,
but give the caller a chance to recover.

Use this to fix the kythe facts extractor that crashed before on
encountering .* module instance ports.

The DIE_IF_NULL is not eliminated all the way, in some cases it was pushed one level up to avoid having this change much larger (to be addressed in follow-up changes).

### DIFF
--- a/common/text/tree_utils_test.cc
+++ b/common/text/tree_utils_test.cc
@@ -1436,7 +1436,7 @@ TEST(GetSubtreeAsNodeTest, ValidatedFoundNodeEnumChildMismatches) {
 TEST(GetSubtreeAsLeafTest, ValidatedFoundLeaf) {
   auto root = TNode(FakeEnum::kZero, Leaf(7, "foo"));
   const auto& leaf = GetSubtreeAsLeaf(*root, FakeEnum::kZero, 0);
-  EXPECT_EQ(leaf.get().token_enum(), 7);
+  EXPECT_EQ(leaf->get().token_enum(), 7);
 }
 
 TEST(GetSubtreeAsLeafTest, GotNodeInsteadOfLeaf) {
--- a/common/text/tree_utils_test.cc
+++ b/common/text/tree_utils_test.cc
@@ -1435,7 +1435,7 @@ TEST(GetSubtreeAsNodeTest, ValidatedFoundNodeEnumChildMismatches) {
 
 TEST(GetSubtreeAsLeafTest, ValidatedFoundLeaf) {
   auto root = TNode(FakeEnum::kZero, Leaf(7, "foo"));
-  const auto& leaf = GetSubtreeAsLeaf(*root, FakeEnum::kZero, 0);
+  const SyntaxTreeLeaf* leaf = GetSubtreeAsLeaf(*root, FakeEnum::kZero, 0);
   EXPECT_EQ(leaf->get().token_enum(), 7);
 }
 
--- a/verilog/analysis/checkers/always_comb_blocking_rule.cc
+++ b/verilog/analysis/checkers/always_comb_blocking_rule.cc
@@ -72,7 +72,7 @@ void AlwaysCombBlockingRule::HandleSymbol(const verible::Symbol& symbol,
 
       if (node == nullptr) continue;
 
-      const auto& leaf = verible::GetSubtreeAsLeaf(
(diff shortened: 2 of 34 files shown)


hzeller:
CC @ivan444

hzeller:
This should fix the [`GetSubTreeAsSymbol()` check failures](https://chipsalliance.github.io/sv-tests-results/#VeribleExtractor|ibex|ibex) we see in the Verible Kythe extractor.

hzeller:
(Also adding Rafal, in case David is already on holiday vacation)

common/text/tree_utils_test.cc
```diff
@@ -1436,7 +1436,7 @@ TEST(GetSubtreeAsNodeTest, ValidatedFoundNodeEnumChildMismatches) {
 TEST(GetSubtreeAsLeafTest, ValidatedFoundLeaf) {
   auto root = TNode(FakeEnum::kZero, Leaf(7, "foo"));
   const auto& leaf = GetSubtreeAsLeaf(*root, FakeEnum::kZero, 0);
```
ivan444, line 1438: s/&/*/

verilog/analysis/checkers/always_comb_blocking_rule.cc
```diff
@@ -75,8 +75,8 @@ void AlwaysCombBlockingRule::HandleSymbol(const verible::Symbol& symbol,
       const auto& leaf = verible::GetSubtreeAsLeaf(
```
ivan444, line 75: s/&/*/

verilog/analysis/checkers/module_filename_rule.cc
```diff
@@ -114,9 +114,11 @@ void ModuleFilenameRule::Lint(const TextStructureView& text_structure,
   }
 
   // Only report a violation on the last module declaration.
-  const auto& last_module_id = GetModuleName(*module_cleaned.back().match);
-  violations_.insert(verible::LintViolation(
-      last_module_id.get(), absl::StrCat(kMessage, "\"", unitname, "\"")));
+  const auto* last_module_id = GetModuleName(*module_cleaned.back().match);
+  if (last_module_id) {
```
ivan444, line 118: Add log-error or vlog at least if `last_module_id` is null? It looks like a bug.

verilog/analysis/checkers/one_module_per_file_rule.cc
```diff
@@ -78,9 +78,12 @@ void OneModulePerFileRule::Lint(const TextStructureView& text_structure,
 
   if (module_cleaned.size() > 1) {
     // Report second module declaration
-    const auto& second_module_id = GetModuleName(*module_cleaned[1].match);
-    violations_.insert(verible::LintViolation(
-        second_module_id.get(), absl::StrCat(kMessage, module_cleaned.size())));
+    const auto* second_module_id = GetModuleName(*module_cleaned[1].match);
+    if (second_module_id) {
```
ivan444, line 82: Add log-error or vlog if null?

ivan444 (approved):
Might be useful to run tests in ASAN mode. This is a switch from const references to const pointers. It's difficult to say if we relied on const reference lifetime extension at any point. Asan should surface it.

Also, might be good to add more logging at the endpoints (as bubbling up nullptr implies a but at either caller or parser side).

common/text/tree_utils_test.cc
hzeller: Done (and use fully qualified type instead of auto)

verilog/analysis/checkers/always_comb_blocking_rule.cc
hzeller: Done.

verilog/analysis/checkers/module_filename_rule.cc
hzeller: Done.

verilog/analysis/checkers/one_module_per_file_rule.cc
hzeller: Done.

hzeller:
Yes, ran tests in asan mode (The CI also has it as one of its runs)

PR #1127, opened 2021-12-22
Add explicit typed structure literals

### LINKED ISSUE #717: Explicitly typed structure literals are marked as syntax error
**Describe the bug**

Verible marks explicit types on structure literals as a syntax error.

**To Reproduce**

```systemverilog
module foobar;
  always_comb begin
    foo = t_foo'{
      bar: '1,
      boo: '0
    };
  end
endmodule
```

Include any options used.

**Actual behavior:**

```
test.sv:3:16: syntax error, rejected "'{" (syntax-error).
```
It works fine without the explicit `t_foo` type.

**Expected behavior**

Section 5.10 of the LRM specifies that structure literals may have their data type defined, sometimes it is needed to remove ambiguity.

zegervdv:
A naive attempt to fix  #717.

I have introduced a new node `kAssignmentPatternExpression` to handle the struct literals (inspired by the parser from [tree-sitter-verilog](https://github.com/tree-sitter/tree-sitter-verilog)).

I have zero experience with yacc so I've probably overlooked things. 
Any comments or feedback would be welcome.

### DIFF
--- a/.github/bin/smoke-test.sh
+++ b/.github/bin/smoke-test.sh
@@ -96,6 +96,7 @@ declare -A KnownIssue
 
 #--- Opentitan
 KnownIssue[formatter:$BASE_TEST_DIR/opentitan/hw/ip/aes/dv/aes_model_dpi/aes_model_dpi_pkg.sv]=1006
+KnownIssue[formatter:$BASE_TEST_DIR/opentitan/hw/dv/sv/cip_lib/seq_lib/cip_base_vseq.sv]=1130
 # There is also bug 1008 which only shows up if compiled with asan
 
 #--- ivtest
--- a/verilog/CST/verilog_nonterminals.h
+++ b/verilog/CST/verilog_nonterminals.h
@@ -276,6 +276,7 @@ enum class NodeEnum {
   kMacroFormalArg,
   kAssignmentPattern,
   kPatternExpression,
+  kAssignmentPatternExpression,
   kMinTypMaxList,
   kConstRef,
   kInterfacePortHeader,
--- a/verilog/parser/verilog.y
+++ b/verilog/parser/verilog.y
@@ -999,7 +999,12 @@ assignment_pattern
   | TK_LP '}'
     { $$ = MakeTaggedNode(N::kAssignmentPattern, $1, $2); }
   ;
-
+assignment_pattern_expression
+  : assignment_pattern
+    { $$ = move($1); }
+  | data_type_base assignment_pattern
+    { $$ = MakeTaggedNode(N::kAssignmentPatternExpression, $1, $2); }
+  ;
 structure_or_array_pattern_expression_list
   : structure_or_array_pattern_expression_list ',' structure_or_array_pattern_expression
     { $$ = ExtendNode($1, $2, $3); }
@@ -4662,7 +4667,7 @@ expr_primary
    { $$ = move($1); }
   | expr_primary_braces
     { $$ = move($1); }
-  | assignment_pattern
+  | assignment_pattern_expression
     { $$ = move($1); }
   ;
 expr_primary_parens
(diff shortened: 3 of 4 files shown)


hzeller:
The smoke test found that this triggered a missed assumption about number of children in 
```
F20211222 19:52:48.046043 11449 layout_optimizer.cc:505] Check failed: node.Children().size() == 2 (3 vs. 2) 
```
https://github.com/chipsalliance/verible/runs/4611023027?check_suite_focus=true#step:7:140

Maybe @mglb can help resolve what needs to be done here.

zegervdv:
The failing file has a struct literal in it, so before my change it was failing on a syntax error (and therefore skipped, if I read the smoke test script correctly).

I've verified with the latest master version that the same error occurs when I comment out the specific [line](https://github.com/lowRISC/opentitan/blob/6f74005afea94f1fc0b73c92b494318d23275dde/hw/dv/sv/cip_lib/seq_lib/cip_base_vseq.sv#L833) with the literal.

mglb:
> The smoke test found that this triggered a missed assumption about number of children in
> 
> ```
> F20211222 19:52:48.046043 11449 layout_optimizer.cc:505] Check failed: node.Children().size() == 2 (3 vs. 2) 
> ```
> 
> https://github.com/chipsalliance/verible/runs/4611023027?check_suite_focus=true#step:7:140
> 
> Maybe @mglb can help resolve what needs to be done here.

Function formatting in layout optimizer supports only single rigid partition tree layout:
https://github.com/chipsalliance/verible/blob/c926ededf7953febb5f577b081474987d0f0209d/common/formatting/layout_optimizer.cc#L504-L508

I guess you have to add a case for `kAssignmentPatternExpression` in https://github.com/chipsalliance/verible/blob/c926ededf7953febb5f577b081474987d0f0209d/verilog/formatting/tree_unwrapper.cc#L601 with `VisitIndentedSection()` and probably prevent it from being hoisted later. You should also check reshaping code in the same file for function calls and other node types that wrap your node (see syntax tree of .sv file, find your node, and see its ancestors).

You can see partitions passed to layout optimizer (look for `OptimizeTokenPartitionTree, before`) with:
```
GLOG_logtostderr=1 GLOG_vmodule='layout_optimizer=9' file.sv
```

To print syntax tree (useful for modifications in tree_unwrapper) use:
```
verible-verilog-syntax --printtree code.sv
```

zegervdv:
With the extra logging I figured the error is coming from this line:
https://github.com/lowRISC/opentitan/blob/6f74005afea94f1fc0b73c92b494318d23275dde/hw/dv/sv/cip_lib/seq_lib/cip_base_vseq.sv#L280
After commenting it out, the formatting succeeds without error.

Which does not seem related to the changes I did.

Should I add the `kAssignmentPatternExpression` still to that function?

mglb:
> With the extra logging I figured the error is coming from this line: https://github.com/lowRISC/opentitan/blob/6f74005afea94f1fc0b73c92b494318d23275dde/hw/dv/sv/cip_lib/seq_lib/cip_base_vseq.sv#L280 After commenting it out, the formatting succeeds without error.
> Which does not seem related to the changes I did.

Can confirm.

> Should I add the `kAssignmentPatternExpression` still to that function?

No need to.

hzeller:
@mglb are you then taking care of the issue surfaced with the layout_optimizer ?

hzeller:
Ok, I've now recorded this as todo to fix in #1130 so that @mglb can take a look.

To make this test pass for now: @zegervdv In this pull request, can you also modify [`.github/bin/smoke-test.sh`](https://github.com/chipsalliance/verible/blob/c926ededf7953febb5f577b081474987d0f0209d/.github/bin/smoke-test.sh#L98)  around line 98 and add the following line so that the smoke test treats it as a known issue ?

```
KnownIssue[formatter:$BASE_TEST_DIR/opentitan/hw/dv/sv/cip_lib/seq_lib/cip_base_vseq.sv]=1130
```

zegervdv:
done

hzeller (approved):
Thanks!

hzeller:
Thanks, merged (sorry for the delay, winter holidays and such).

Issue #1128, opened 2021-12-22
[kythe] ReferenceExtendsInheritance() sometimes runs into nullptr scopes.


hzeller:
## Input test case

```
git clone https://github.com/accellera/uvm.git
echo uvm/distrib/src/uvm_pkg.sv > filelist.txt
verible-verilog-kythe-extractor --include_dir_paths=uvm/distrib/src --file_list_path filelist.txt > /dev/null
```

This results in a nullptr scope (which is now prevented from creating a crash but still needs investigation)

hzeller:
After #1129, log error output is

```
E20211222 12:02:51.243124 923014 kythe_facts_extractor.cc:1068] ReferenceExtendsInheritance: NULL scope for vname {
  "signature": "uvm/distrib/src/base/uvm_objection.svh#m_uvm_test_done_objection_base#",
  "path": "uvm/distrib/src/base/uvm_objection.svh",
  "language": "verilog",
  "root": "",
  "corpus": ""
}
E20211222 12:02:51.340545 923014 kythe_facts_extractor.cc:1068] ReferenceExtendsInheritance: NULL scope for vname {
  "signature": "uvm/distrib/src/base/uvm_port_base.svh#uvm_port_base#IF#",
  "path": "uvm/distrib/src/base/uvm_port_base.svh",
  "language": "verilog",
  "root": "",
  "corpus": ""
}
E20211222 12:02:51.806788 923014 kythe_facts_extractor.cc:1068] ReferenceExtendsInheritance: NULL scope for vname {
  "signature": "uvm/distrib/src/reg/uvm_reg_sequence.svh#uvm_reg_sequence#BASE#",
  "path": "uvm/distrib/src/reg/uvm_reg_sequence.svh",
  "language": "verilog",
  "root": "",
  "corpus": ""
}
...
```

Suspicion is that we hit the root scope. Needs further investigation.

PR #1129, opened 2021-12-22
Handle null scopes gracefully in ReferenceExtendsInheritance().

hzeller:
This is a stop-gap as these null scopes were encountered in the
field, further analysis is needed.

Issues #1128

### DIFF
--- a/verilog/tools/kythe/kythe_facts_extractor.cc
+++ b/verilog/tools/kythe/kythe_facts_extractor.cc
@@ -1058,6 +1058,18 @@ void KytheFactsExtractor::ReferenceExtendsInheritance(
     return;
   }
 
+  // TODO(hzeller): should this have been detected before ?
+  // NULL vname is not encountered, but NULL scope. Issue #1128
+  if (!definitions.back().first) {
+    LOG(ERROR) << "ReferenceExtendsInheritance: NULL vname";
+    return;
+  }
+  if (!definitions.back().second) {
+    LOG(ERROR) << "ReferenceExtendsInheritance: NULL scope for vname "
+               << *definitions.back().first;
+    return;
+  }
+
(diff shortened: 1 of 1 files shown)


hzeller:
These were found looking at https://chipsalliance.github.io/sv-tests-results/ comparing where the extractor fails compared to regular verible.

verilog/tools/kythe/kythe_facts_extractor.cc
```diff
@@ -1058,6 +1058,18 @@ void KytheFactsExtractor::ReferenceExtendsInheritance(
...
   }
 
+  // TODO(hzeller): should this have been detected before ?
+  // NULL vname is not encountered, but NULL scope. Issue #1128
+  if (!definitions.back().first) {
+    LOG(ERROR) << "ReferenceExtendsInheritance: NULL vname";
+    return;
+  }
+  if (!definitions.back().second) {
+    LOG(ERROR) << "ReferenceExtendsInheritance: NULL scope for vname "
```
ivan444, line 1068: Check the lines 292 and 316. IIUC, null scope is the global scope. But this whole behavior looks buggy. Including `ScopeResolver::SearchForDefinitions` (verilog/tools/kythe/scope_resolver.cc) -- it doesn't check if the `current_scope` is null  before adding it to the `declarations` vector. But, most of the callers perfom the null-check.

IMO, this PR is OK as a workaround. But we need to revisit the handling of scopes (and clarify the semantic for the global scope).

ivan444 (approved):

verilog/tools/kythe/kythe_facts_extractor.cc
hzeller: Yes, this is why I opened #1128 for further investigation.

PR #1131, opened 2021-12-26
Add nvim-lspconfig to language server documentation

zegervdv:
Add documentation on how to enable the verible language server with neovim using the built-in client.


hzeller (approved):
Thanks!

PR #1132, opened 2021-12-27
Use host m4/flex/bison on linux

### LINKED ISSUE #957: Support building with glibc 2.34
glibc 2.34, which is currently being rolled out in first distributions, has broken some vendored dependencies of Verible. Let's use this as a tracking issue to update them.

* [ ] m4 doesn't build. Dependency update not yet available. Filed https://github.com/jmillikin/rules_m4/issues/9
* [x] absl doesn't build. Dependency update: https://github.com/chipsalliance/verible/pull/956

(There might be more, but that's currently where I'm stuck.)

### LINKED ISSUE #1122: Build error on Fedora 35
I am getting the following error while trying to build verible on fedora 35:

```
➜  verible git:(master) bazel build -c opt //...
Starting local Bazel server and connecting to it...
INFO: Analyzed 1864 targets (111 packages loaded, 5362 targets configured).
INFO: Found 1864 targets...
bazel: Entering directory `/home/jordi/.cache/bazel/_bazel_jordi/13967d7c87a057fae689ac53148986f9/execroot/com_google_verible/'
ERROR: /home/jordi/.cache/bazel/_bazel_jordi/13967d7c87a057fae689ac53148986f9/external/m4_v1.4.18/gnulib/BUILD.bazel:192:11: Compiling gnulib/lib/c-stack.c [for host] failed: (Exit 1): gcc failed: error executing command /usr/bin/gcc -U_FORTIFY_SOURCE -fstack-protector -Wall -Wunused-but-set-parameter -Wno-free-nonheap-object -fno-omit-frame-pointer -g0 -O2 '-D_FORTIFY_SOURCE=1' -DNDEBUG -ffunction-sections ... (remaining 25 argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox
In file included from /usr/include/signal.h:328,
                 from external/m4_v1.4.18/gnulib/lib/c-stack.c:49:
external/m4_v1.4.18/gnulib/lib/c-stack.c:55:26: error: missing binary operator before token "("
   55 | #elif HAVE_LIBSIGSEGV && SIGSTKSZ < 16384
      |                          ^~~~~~~~
external/m4_v1.4.18/gnulib/lib/c-stack.c:139:8: error: variably modified 'buffer' at file scope
  139 |   char buffer[SIGSTKSZ];
      |        ^~~~~~
bazel: Leaving directory `/home/jordi/.cache/bazel/_bazel_jordi/13967d7c87a057fae689ac53148986f9/execroot/com_google_verible/'
INFO: Elapsed time: 10.128s, Critical Path: 3.96s
INFO: 42 processes: 13 internal, 29 linux-sandbox.
FAILED: Build did NOT complete successfully
```
My toolset version:
```
➜  ~ g++ --version
g++ (GCC) 11.2.1 20210728 (Red Hat 11.2.1-1)
Copyright (C) 2021 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

➜  ~ bazel --version
bazel 4.2.1
```

Any idea why this happens?
I was able to build in RHEL8 but not in Fedora35, which is strange.

tgorochowik:
This should allow us to skip the problematic update of the m4 dependency. Other operating systems can keep using the outdated version, Linux should be allowed to use host tools.

Closes #1122 
Closes #957

### DIFF
--- a/.github/workflows/verible-ci.yml
+++ b/.github/workflows/verible-ci.yml
@@ -132,7 +132,7 @@ jobs:
       run: |
         set -x
         source ./.github/settings.sh
-        sudo apt -qq -y install clang-10
+        sudo apt -qq -y install clang-10 flex bison
         ./.github/bin/set-compiler.sh 9
         ./.github/bin/install-bazel.sh
         ./.github/bin/install-python.sh
--- a/bazel/bison.bzl
+++ b/bazel/bison.bzl
@@ -32,10 +32,12 @@ def genyacc(
         outs = [header_out, source_out] + extra_outs,
         cmd = select({
             "@platforms//os:windows": "$(BISON) --defines=$(location " + header_out + ") --output-file=$(location " + source_out + ") " + " ".join(extra_options) + " $<",
+            "@platforms//os:linux": "bison --defines=$(location " + header_out + ") --output-file=$(location " + source_out + ") " + " ".join(extra_options) + " $<",
             "//conditions:default": "M4=$(M4) $(BISON) --defines=$(location " + header_out + ") --output-file=$(location " + source_out + ") " + " ".join(extra_options) + " $<",
         }),
         toolchains = select({
             "@platforms//os:windows": ["@rules_bison//bison:current_bison_toolchain"],
+            "@platforms//os:linux": [], # use host m4/bison
             "//conditions:default": [
                 "@rules_bison//bison:current_bison_toolchain",
                 "@rules_m4//m4:current_m4_toolchain",
--- a/README.md
+++ b/README.md
@@ -45,6 +45,8 @@ Verible's code base is written in C++.
 To build, you need the [bazel] build system and a C++17 compatible compiler
 (e.g. >= g++-9), as well as python3.
 
+Linux systems also require the following tools: m4, flex, bison.
+
 Use your package manager to install the dependencies; on a system with
 the nix package manager simply run `nix-shell` to get a build environment.
 
--- a/bazel/flex.bzl
+++ b/bazel/flex.bzl
@@ -26,10 +26,12 @@ def genlex(name, src, out):
         outs = [out],
         cmd = select({
             "@platforms//os:windows": "$(FLEX) --outfile=$@ $<",
+            "@platforms//os:linux": "flex --outfile=$@ $<",
             "//conditions:default": "M4=$(M4) $(FLEX) --outfile=$@ $<",
         }),
         toolchains = select({
             "@platforms//os:windows": ["@rules_flex//flex:current_flex_toolchain"],
+            "@platforms//os:linux": [], # use host m4/flex tools
             "//conditions:default": [
                 "@rules_flex//flex:current_flex_toolchain",
                 "@rules_m4//m4:current_m4_toolchain",
--- a/verilog/parser/BUILD
+++ b/verilog/parser/BUILD
@@ -128,6 +128,7 @@ cc_library(
         "//conditions:default": [
             "-Wno-implicit-fallthrough",
             "-Wno-type-limits",
+            "-Wno-unreachable-code",
         ],
     }),
     deps = [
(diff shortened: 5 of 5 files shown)


bazel/bison.bzl
```diff
@@ -32,10 +32,12 @@ def genyacc(
         outs = [header_out, source_out] + extra_outs,
         cmd = select({
             "@platforms//os:windows": "$(BISON) --defines=$(location " + header_out + ") --output-file=$(location " + source_out + ") " + " ".join(extra_options) + " $<",
+            "@platforms//os:linux": "bison --defines=$(location " + header_out + ") --output-file=$(location " + source_out + ") " + " ".join(extra_options) + " $<",
             "//conditions:default": "M4=$(M4) $(BISON) --defines=$(location " + header_out + ") --output-file=$(location " + source_out + ") " + " ".join(extra_options) + " $<",
         }),
         toolchains = select({
             "@platforms//os:windows": ["@rules_bison//bison:current_bison_toolchain"],
+            "@platforms//os:linux": [], # use host m4/flex
```
hzeller, line 40: use host `m4/bison` toolchain (not flex)

hzeller:
This requires to have flex and bison installed, but wouldn't it be sufficient if we just use m4 as backend for these ?
I suspect it would be more changes, but maybe it would be worthwhile for lower dependency count (m4 is installed on essentially every Linux machine because it is used for all kinds of ancient config file processing, while flex, bison need to be installed). Anyway, can be a separate change.

For completelness, I'd mention the flex/bison requirement in [the readme](https://github.com/chipsalliance/verible#build) and possibly in the [CI dependencies](https://github.com/chipsalliance/verible/blob/master/.github/workflows/verible-ci.yml#L135) which work 'by accident' now (I suppose because the CI machines are set-up as development machines and have these installed by default, but it would be a good for completeness sake (and documentation) to include that).

corco:
Should we use installed flex/bison on all platform instead of only Linux? I could make that change in a future PR. Right now, rules_flex/bison/m4 is only used on MacOS, which could have flex/bison installed locally as well.

Another alternative would be to make using local flex/bison an option, and keep "official" platforms on rules_flex/bison. I would guess Ubuntu 22.04 will use glibc 2.34, in that case it would be pointless IMHO.

hzeller:
I tend to think that ideally we would like to have flex/bison be built in the hermetic build of verible to minimize issues if it is hard to install a particular version on the given platform (I think MacOS might at most come with lex/yacc by default (as Apple doesn't like the GPL) and would require some additional brew install for flex/bison).

Using the platform flex/bison is mostly a concession right now because we don't have good enough bazel rules that make it all work, but ideally we should be able to come back to that state at some point.

corco:
In that case, I would opt to leave the current behavior as is and add an optional flag to use local flex/bison (similar to [this](https://github.com/bazelbuild/examples/tree/HEAD/rules/starlark_configurations/select_on_build_setting)) for platforms where the bazel rules fail.

bazel/bison.bzl
hzeller: This comment is still old.

.github/workflows/verible-ci.yml
```diff
@@ -132,6 +132,7 @@ jobs:
       run: |
         set -x
         source ./.github/settings.sh
+        # dependencies provided by the container: gcc, m4, flex, bison
```
hzeller, line 135: Should we explicitly install flex and bison ? Even if they are by default on the CI system, this would document it.

hzeller (approved):
LGTM, thanks!

hzeller:
Oh, a thing we missed: this now broke all Linux release builds as the releases don't install the packages. Will have a look.

hzeller:
WIP #1135

hzeller:
So even when installing flex and bison ( #1135), we run into problems with older versions of these tools not working well with verible. In particular Ubuntu xenial, Ubuntu groovy and centos7 fail to compile.
https://github.com/chipsalliance/verible/runs/4794009064?check_suite_focus=true

So I am going to roll back this #1132 and we would need version where we can switch at compile time if we use m4 locally or from bazel (as @corco suggested)

PR #1133, opened 2022-01-10
WIP: Create a debian package in bazel

hzeller:
Bottom line: looks like it might be easier to create this
with an external script as we do today with the tar-balls.


hzeller:
(mostly used for discussion right now. Looking how duplication is needed using bazel, it would probably be better to use the scripts we use to create tarballs right now and base debian packaging on that)

PR #1134, opened 2022-01-11
WIP: Fix abort when an EOL comment precedes nested `kSystemTFCall`

### LINKED ISSUE #1130: Check failure in layout_optimizer.cc
**Test case**

When processing opentitan `hw/dv/sv/cip_lib/seq_lib/cip_base_vseq.sv`, the layout optimizer fails with:

layout_optimizer.cc:505] Check failed: node.Children().size() == 2 (3 vs. 2) 

This was first discovered in #1127 that allowed to parse a file that previously didn't parse (see details there)

mglb:
Fixes #1130 

Test case:
```systemverilog
// verilog_syntax: parse-as-class-body
task t;
  `FOO(
      // foobar
      $bar("baz"))
endtask
```

---

**TODO**:

* Fix source of the problem in order to avoid fallback formatting in this known case.
* Tests.

### DIFF
--- a/common/formatting/layout_optimizer.cc
+++ b/common/formatting/layout_optimizer.cc
@@ -502,29 +502,44 @@ LayoutFunction TokenPartitionsLayoutOptimizer::CalculateOptimalLayout(
 
     case PartitionPolicyEnum::kOptimalFunctionCallLayout: {
       // Support only function/macro/system calls for now
-      CHECK_EQ(node.Children().size(), 2);
+      if (node.Children().size() == 2) {
+        const auto& function_header = node.Children()[0];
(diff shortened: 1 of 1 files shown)


hzeller:
While at it, you can also remove the line at https://github.com/chipsalliance/verible/blob/1360e497e858379d7057f691446f5b39ff39a871/.github/bin/smoke-test.sh#L99 (the smoke test now correctly determined that this issue is _not_ happening anymore)

hzeller:
Any update here ?

mglb:
I've fixed this problem in #1172.
Closing.

PR #1136, opened 2022-01-12
Revert "Use host m4/flex/bison on linux"

hzeller:
Reverts chipsalliance/verible#1132

This needs more work as releases stop working with it as the releases need a local bison/flex installed ; a first attempt to do this in #1135 showed that this then creates issues with older versions of bison/flex.

So we probably need an approach in which we _optionally_ can choose to use to use the local m4; something along the lines of [this comment](https://github.com/chipsalliance/verible/pull/1132#issuecomment-1006614936)

But for now, revert this to unbreak.

### DIFF
--- a/.github/workflows/verible-ci.yml
+++ b/.github/workflows/verible-ci.yml
@@ -132,7 +132,7 @@ jobs:
       run: |
         set -x
         source ./.github/settings.sh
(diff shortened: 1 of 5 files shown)


corco (approved):
LGTM, I will look into making the local flex/bison a command line option

PR #1144, opened 2022-01-15
Added Ubuntu 22.04 (Jammy) to releases

corco:
It's coming out in 3 months, might as well get a release ready for Ubuntu Jammy.

It's using the locally installed flex/bison since Jammy is on glibc 2.34

### DIFF
--- a/releasing/docker-run.sh
+++ b/releasing/docker-run.sh
@@ -60,6 +60,13 @@ case "$TARGET_OS" in
     # Compiler
     [ "$TARGET_VERSION" = xenial ] || [ "$TARGET_VERSION" = bionic ] && _version="$TARGET_VERSION" || _version="common"
     cat ${TARGET_OS}/${_version}/compiler.dockerstage >> ${OUT_DIR}/Dockerfile
+    
+    # Use local flex/bison on Jammy
+    if [ "$TARGET_VERSION" = jammy ]; then
+      BAZEL_OPTS="${BAZEL_OPTS} --//bazel:use_local_flex_bison"
+      cat ${TARGET_OS}/${TARGET_VERSION}/flex_bison.dockerstage >> ${OUT_DIR}/Dockerfile
+    fi
+
     # Bazel
     cat ${TARGET_OS}/common/bazel.dockerstage >> ${OUT_DIR}/Dockerfile
   ;;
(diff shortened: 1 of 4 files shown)


hzeller:
Can you temporarily uncomment the release building in https://github.com/chipsalliance/verible/blob/master/.github/workflows/verible-ci.yml#L215 so that we can see that all releases build properly (because it is disabled by default in the regular CI).

Once it works, we can rever that back in verible-ci.

corco:
I enabled the release.

Would you be interested to enable release build through [workflow input](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow) (in addition to current behavior), I could look into that.

releasing/docker-run.sh
```diff
@@ -60,6 +60,13 @@ case "$TARGET_OS" in
     # Compiler
     [ "$TARGET_VERSION" = xenial ] || [ "$TARGET_VERSION" = bionic ] && _version="$TARGET_VERSION" || _version="common"
     cat ${TARGET_OS}/${_version}/compiler.dockerstage >> ${OUT_DIR}/Dockerfile
+    
+    # Use local flex/bison on Jammy
+    if [ "$TARGET_VERSION" = jammy ]; then
+      BAZEL_OPTS="${BAZEL_OPTS} --//bazel:use_local_flex_bison"
+      cat ${TARGET_OS}/${TARGET_VERSION}/flex_bison.dockerstage >> ${OUT_DIR}/Dockerfile
```
umarcor, line 67: I would suggest hardcoding `jammy` in this line, instead of using the TARGET_VERSION envvar.

umarcor: Moreover, since it's a one-liner, you might avoid the `.dockerstage` file:

```
echo 'RUN apt-get install -y flex bison' >> ${OUT_DIR}/Dockerfile
```

umarcor:
> I enabled the release.
> 
> Would you be interested to enable release build through [workflow input](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow) (in addition to current behavior), I could look into that.

@corco, the workflow_dispatch event is added already. Hence, you'd only need to add the `inputs` fields below. See, for instance, https://github.com/hdl/containers/blob/main/.github/workflows/dispatch.yml. Those inputs are available through `github.event.inputs`. Find more info in [pyTooling/Actions: Context](https://github.com/pyTooling/Actions/#context) and github/feedback#9092.

hzeller:
So this looks like almost ready to merge; @corco , did you consider @umarcor 's suggestions ?

corco:
Yes, I applied @umarcor's suggestions and the build with releases is passing. I re-disabled building release on every push, it should be ready to merge.

hzeller (approved):
LGTM, thanks!

PR #1148, opened 2022-01-18
Prevent juxtaposition of "MustWrap" lines.

mglb:
This change will prevent juxtaposition of code fragments that should never be joined into one line, e.g. `//`-comment and regular code.
Instead of aborting, an attempt of such juxtaposition emits non-fatal error and calls `stack` combinator (which is safe in all cases I can think of).

### DIFF
--- a/common/formatting/layout_optimizer_test.cc
+++ b/common/formatting/layout_optimizer_test.cc
@@ -667,86 +667,101 @@ class LayoutFunctionFactoryTest : public ::testing::Test,
             "Exactly at 30 limit (30 cols).\n"
             "One over 30 limit (31 columns).\n"
 
-            "10 columns\r"
-
-            // Same lines, but without "MustWrap" set:
-
-            "This line is short.\r"
-            "This line is so long that it exceeds column limit.\r"
-            "        Indented  line  with  many  spaces .\r"
-
-            "One under 40 column limit (39 columns).\r"
-            "Exactly at 40 column limit (40 columns).\r"
-            "One over 40 column limit (41 characters).\r"
-
-            "One under 30 limit (29 cols).\r"
-            "Exactly at 30 limit (30 cols).\r"
-            "One over 30 limit (31 columns).\r"
-
             "10 columns"),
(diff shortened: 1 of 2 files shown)


common/formatting/layout_optimizer_test.cc
```diff
@@ -667,9 +667,25 @@ class LayoutFunctionFactoryTest : public ::testing::Test,
             "Exactly at 30 limit (30 cols).\n"
             "One over 30 limit (31 columns).\n"
 
+            "10 columns\r"
```
hzeller, line 670: mmh, instead of repeating the same thing (which is harder to manually verify or extend), can we formulate that in a way that it comes from a single source (take the \n-separated version and create the \r version from it). Also that will allow us to determine `kWrappable` programmatically.

common/formatting/layout_optimizer_test.cc
```diff
@@ -672,59 +672,96 @@ class LayoutFunctionFactoryTest : public ::testing::Test,
             absl::StrSplit(sample_, absl::ByAnyChar(" \n"), absl::SkipEmpty())),
         style_(CreateStyle()),
         factory_(LayoutFunctionFactory(style_)) {
-    for (const auto token : tokens_) {
-      ftokens_.emplace_back(TokenInfo{1, token});
-    }
+    // Create two sets of tokens
+    const ssize_t number_of_tokens_in_set = tokens_.size();
```
hzeller, line 676: Windows compile seems to [have a problem](https://github.com/chipsalliance/verible/runs/4898984612?check_suite_focus=true#step:7:684) with the signed `ssize_t` but here we only really need `size_t`.

hzeller (approved):
LGTM (after the one-line change to fix the Windows build)

common/formatting/layout_optimizer_test.cc
mglb: Fixed. Also changed int to size_t in the `for` below.

tgorochowik (approved):

PR #1157, opened 2022-01-27
Reduce CHECK()-fails: Make tree utils return pointers to be able to return invalid situations

hzeller:
Make it simpler to deal with unexpected syntax tree shape: return nullptr instead of CHECK()-failing.

Background: improve the resilience of Kythe indexing.

### DIFF
--- a/common/text/tree_utils.h
+++ b/common/text/tree_utils.h
@@ -96,7 +96,7 @@ const SyntaxTreeLeaf& CheckLeafEnum(const SyntaxTreeLeaf& leaf,
 
 namespace internal {
 template <typename S>
-void MustBeCSTSymbolOrNode(S&) {
+void StaticAssertMustBeCSTSymbolOrNode(S&) {
   typedef typename std::remove_const<S>::type base_type;
   static_assert(std::is_same<base_type, Symbol>::value ||
                     std::is_same<base_type, SyntaxTreeNode>::value,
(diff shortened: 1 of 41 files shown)


hzeller:
This substantially reduces the number of CHECK()s that we'll encounter while kythe indexing and in general encourages defensive programming in other places.

It looks like a large diff, but are mostly fairly mechanical changes.

fangism (approved):

hzeller:
Some data point
```
find common/text verilog/CST/ -type f | grep -v _test.cc | xargs egrep "(CHECK|DIE_IF_NULL)" | wc -l
```

Was **`116`** before in head and is **`82`** after this PR.

common/text/tree_utils.h
```diff
@@ -67,36 +67,49 @@ const SyntaxTreeLeaf& SymbolCastToLeaf(const Symbol&);
...
-const SyntaxTreeNode& CheckNodeEnum(const SyntaxTreeNode& node,
-                                    E expected_node_enum) {
+const SyntaxTreeNode* MatchNodeEnumOrNull(const SyntaxTreeNode& node,
+                                          E expected_node_enum) {
   // Uses operator<<(std::ostream&, E) for diagnostics.
-  CHECK_EQ(E(node.Tag().tag), expected_node_enum);
-  return node;
+  const bool enum_matches = (E(node.Tag().tag) == expected_node_enum);
+  if (!enum_matches) {
+    LOG(ERROR) << "Node: Programming error: expected " << expected_node_enum
```
ivan444, line 79: Would it be possible to add more information to VLOG? The whole tree etc. We already got tags from check-failures and they weren't useful.

ivan444 (approved):

PR #1158, opened 2022-01-27
Refresh protobuf.

hzeller:


hzeller:
Will need some work to compile with kythe, and the actual problem I'd like to see fixed (`-Wstringop-overflow` warning with gcc-11) is even in latest protobuf. Will work with protobuf team first to get that resolved.

PR #1159, opened 2022-01-27
Add  -Wctad-maybe-unsupported in CI (for clang) (also prepare for gcc-11 new warning)

hzeller:

### DIFF
--- a/common/formatting/layout_optimizer_test.cc
+++ b/common/formatting/layout_optimizer_test.cc
@@ -680,10 +680,11 @@ class LayoutFunctionFactoryTest : public ::testing::Test,
 
     CreateTokenInfosExternalStringBuffer(ftokens_);
 
-    auto must_wrap_pre_format_tokens = verible::iterator_range(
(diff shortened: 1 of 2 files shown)


common/formatting/layout_optimizer_test.cc
```diff
@@ -680,10 +680,11 @@ class LayoutFunctionFactoryTest : public ::testing::Test,
 
     CreateTokenInfosExternalStringBuffer(ftokens_);
 
-    auto must_wrap_pre_format_tokens = verible::iterator_range(
+    using ItType = std::vector<PreFormatToken>::iterator;
+    auto must_wrap_pre_format_tokens = verible::iterator_range<ItType>(
```
fangism, line 684: recommend using `verible::make_range`, because functions are allowed to deduce template arguments.

fangism (approved):

common/formatting/layout_optimizer_test.cc
hzeller: Good idea, done.

PR #1164, opened 2022-01-31
Don't store 'virtual file' content to the analyzer buffer or the stri…

ivan444:
…ng_view_map. The virtual files are already in memory and don't benefit from the additional memory mapping. What's worse, this breaks semantics (as the virtual files aren't opened the same way as the regular files) -- buffer_to_analyzer_map_ was storing random data (and eventually breaking on the conflict)


hzeller (approved):
LGTM

PR #1167, opened 2022-02-01
Removed CentOS 8 release as it is EOL

### LINKED ISSUE #1165: CentOS8 docker build starting to fail
https://github.com/chipsalliance/verible/runs/5013790643

Looks like it can't download from a URL for a repo. This might either be a temporary glitch on that server or a changed URL in general that requires to update the docker to have the URLs it uses point to the right one.

corco:
CentOS 8 is EOL and fails to build as a consequence

Closes #1165

### DIFF
--- a/releasing/docker-run.sh
+++ b/releasing/docker-run.sh
@@ -72,8 +72,7 @@ case "$TARGET_OS" in
   ;;
   centos)
     # Compiler
-    [ "$TARGET_VERSION" = 8 ] && _version="$TARGET_VERSION" || _version="common"
-    cat ${TARGET_OS}/${_version}/compiler.dockerstage >> ${OUT_DIR}/Dockerfile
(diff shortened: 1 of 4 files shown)


releasing/docker-run.sh
```diff
@@ -72,8 +72,7 @@ case "$TARGET_OS" in
   ;;
   centos)
```
umarcor, line 73: Maybe it's worth adding a comment here explaining that CentOS 8 is EOL but CentOS 7 is "alive" until 2024. That will avoid us (others) wondering in the future why is an "older" version tested but not the "newer".

umarcor (approved):

releasing/docker-run.sh
corco: I updated the releasing/README with that information

umarcor (approved):

PR #1175, opened 2022-02-03
Use flat_hash_map instead of std::map in ScopeResolver.

hzeller:
This sped up indexing of earlgray about 3x (3:30min -> 1:10min)


fangism (approved):
Nice one!

PR #1181, opened 2022-02-04
Remove unused return value.

hzeller:

### DIFF
--- a/verilog/tools/ls/verilog_ls.cc
+++ b/verilog/tools/ls/verilog_ls.cc
@@ -85,7 +85,7 @@ static void SendDiagnostics(const std::string &uri,
 }
 
 int main(int argc, char *argv[]) {
-  const auto file_args = verible::InitCommandLine(argv[0], &argc, &argv);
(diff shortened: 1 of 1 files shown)


verilog/tools/ls/verilog_ls.cc
```diff
@@ -85,7 +85,7 @@ static void SendDiagnostics(const std::string &uri,
 }
 
 int main(int argc, char *argv[]) {
-  const auto file_args = verible::InitCommandLine(argv[0], &argc, &argv);
+  verible::InitCommandLine(argv[0], &argc, &argv);
```
fangism, line 88: Looks ok, or alternatively, we could issue a diagnostic saying that this tool expects no positional arguments.

fangism (approved):

verilog/tools/ls/verilog_ls.cc
hzeller: Nobody will ever see the output - the language server is called from some other program.

PR #1184, opened 2022-02-07
Replace linear search in Scope::SearchForDefinition() with map lookup.

hzeller:
In a particular sample file, that reduced the processing time
from 48min to 8 seconds.

### DIFF
--- a/verilog/tools/kythe/kythe_facts.cc
+++ b/verilog/tools/kythe/kythe_facts.cc
@@ -32,10 +32,6 @@ bool Signature::operator<(const Signature& other) const {
                                       other.names_.begin(), other.names_.end());
 }
 
-bool Signature::IsNameEqual(absl::string_view name) const {
-  return names_.back() == name;
-}
-
 std::string Signature::ToString() const {
   std::string signature;
   for (absl::string_view name : names_) {
--- a/verilog/tools/kythe/kythe_facts.h
+++ b/verilog/tools/kythe/kythe_facts.h
@@ -52,10 +52,6 @@ class Signature {
   // Returns the signature concatenated as a string in base 64.
   std::string ToBase64() const;
 
-  // Checks whether this signature represents the same given variable in its
-  // scope.
-  bool IsNameEqual(absl::string_view) const;
-
   const std::vector<std::string>& Names() const { return names_; }
 
  private:
--- a/verilog/tools/kythe/kythe_facts_extractor.cc
+++ b/verilog/tools/kythe/kythe_facts_extractor.cc
@@ -1197,7 +1197,7 @@ void KytheFactsExtractor::ReferenceIncludeFile(
 
   // Create child of edge between the parent and the member of the included
   // file.
-  for (const ScopeMemberItem& member : included_file_scope->Members()) {
+  for (const auto& [_, member] : included_file_scope->Members()) {
     CreateEdge(member.vname, kEdgeChildOf, vnames_context_.top());
   }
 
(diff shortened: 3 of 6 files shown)


hzeller:
Keeping as Draft until it is more clear if the peculiar reverse alphabetical search in the original implementation had a good reason or if it was just a copy/paste fluke.

hzeller:
Here is a difference of the previous implementation and this one
```
git clone https://github.com/Nic30/hdlConvertor.git /tmp/hdlConvertor
```

We're looking at the file `/tmp/hdlConvertor/tests/sv_test/std2017/p600.sv`
```SystemVerilog
// source code
// $typename would return
typedef bit node;
// "bit"
node [2:0] X;
// "bit [2:0]"
int signed Y;
// "int"
package A;
enum {A,B,C=99} X;
// "enum{A=32'sd0,B=32'sd1,C=32'sd99}A::e$1"
typedef bit [9:1'b1] word; // "A::bit[9:1]"
endpackage : A
import A::*;
module top;
typedef struct {node A,B;} AB_t;
AB_t AB[10];
// "struct{bit A;bit B;}top.AB_t$[0:9]"

endmodule
```

Now running our extractor:
```
bazel run verilog/tools/kythe:verible-verilog-kythe-extractor -- --print_kythe_facts json_debug --file_list_root "/tmp/hdlConvertor" --file_list_path <(echo ./tests/sv_test/std2017/p600.sv)
```

There is a difference for this PR vs. head for token a at `261:262`
Looking at the area of the code, this is what this is referring to (with a bit more area around, this refers to the `A` after the import):

```
cut -z -c241-282 /tmp/hdlConvertor/tests/sv_test/std2017/p600.sv
ndpackage : A
import A::*;
module top;
```

This is the json snippet for the _old_ implementation. Note the reference `/tmp/hdlConvertor/tests/sv_test/std2017/p600.sv#A#` _and_ the reference `/tmp/hdlConvertor/tests/sv_test/std2017/p600.sv#A#A#`.

```json
{
  "source": {
    "signature": "@261:262#",
    "path": "/tmp/hdlConvertor/tests/sv_test/std2017/p600.sv",
    "language": "verilog",
    "root": "",
    "corpus": ""
  },
  "edge_kind": "/kythe/edge/ref/imports",
  "target": {
    "signature": "/tmp/hdlConvertor/tests/sv_test/std2017/p600.sv#A#",
    "path": "/tmp/hdlConvertor/tests/sv_test/std2017/p600.sv",
    "language": "verilog",
    "root": "",
    "corpus": ""
  },
  "fact_name": "/"
},
{
  "source": {
    "signature": "@261:262#",
    "path": "/tmp/hdlConvertor/tests/sv_test/std2017/p600.sv",
    "language": "verilog",
    "root": "",
    "corpus": ""
  },
  "edge_kind": "/kythe/edge/ref/imports",
  "target": {
    "signature": "/tmp/hdlConvertor/tests/sv_test/std2017/p600.sv#A#A#",
    "path": "/tmp/hdlConvertor/tests/sv_test/std2017/p600.sv",
    "language": "verilog",
    "root": "",
    "corpus": ""
  },
  "fact_name": "/"
},
```

The new version only has one (the first), but not the double-A one (because we explicitly match the last part of the name `A`, while the previous implementation looked through a reverse sorted list and found the `#A#A` kind ... I suppose).

To me, the old implementation actually looks buggy (there is no `A` nested in another `A`), but I'd like to leave it to @ivan444 and possibly  @MinaToma to help me understand what I am looking at.

fangism:
> 

The `import` statement is what brings the `enum A` (`A::A`) into scope.  You also have `package A` itself.  `node A` could still be legal after it.  Need to consult a language lawyer to determine whether these _shadow_ declarations are valid, as they may hide previous declarations.  I don't know if the metatypes place these symbols into different namespaces.

ivan444:
This change flattens the vnames and breaks scoping. We first need to add a kythe verification test which does scope shadowing to verify this.

ivan444 (approved):
Let's go with this a temp fix. After fixing the performance issue we can revisit this code (add proper scoping).

hzeller:
Filed issue #1195 to track this.

PR #1192, opened 2022-02-15
Switch from idiomatic to performant way to read a file.

hzeller:
Reading large generated files showed that GetContent() was super slow.
With this, a 245MiB file reading from a local filesystem is
sped up from 11 seconds -> 0.045 seconds.


fangism (approved):
I wonder if we should start benchmarking certain operations, maybe more interesting ones than file operations?

hzeller:
Yes operations that allocate the tree structures are all individually deleted, spending a considerable amount in destructors in the end; these could be possibly arena allocated.

PR #1193, opened 2022-02-15
Move more Kythe indexing artifacts into flat_hash_set.

hzeller:

### DIFF
--- a/verilog/tools/kythe/BUILD
+++ b/verilog/tools/kythe/BUILD
@@ -93,6 +93,7 @@ cc_library(
         "//common/util:logging",
         "//verilog/analysis:verilog_project",
         "@com_google_absl//absl/container:btree",
+        "@com_google_absl//absl/container:flat_hash_set",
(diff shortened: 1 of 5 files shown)


hzeller:
WIP. TODO

 * Quantify performance improvement
 * Remove now unnecessary code for various `operator<()`

hzeller:
A particular large 240MiB input file was parsed about 9% faster (425s instead of 467s) and about 10% less memory (52GiB instead of 58GiB).
(also with #1184 applied beforehand)

hzeller:
The savings don't seem super-significant quite yet.

hzeller:
Tested in different environment also using tcmalloc (TODO: add to Verible by default), it is actually 43% faster. So looks worthwhile

hzeller:
#1194 moves the struct into the *.cc file, so if that is submitted first, this will need to have a manual merge.

ivan444 (approved):

PR #1194, opened 2022-02-17
Always stream the final Kythe result.

hzeller:
Instead of collecting the final result in an ever-growing set
before printing, write them out as they are received to
minimize memory use.

This _could_ result in the same facts be emitted multiple times
if there are duplicates (can there ?), but this is benign.

Make StreamKytheFactsEntries() the main toplevel function.
Remove the ExtractKytheFacts() that accumulates everything in
sets first.

### DIFF
--- a/verilog/tools/kythe/kythe_facts_extractor.cc
+++ b/verilog/tools/kythe/kythe_facts_extractor.cc
@@ -72,12 +72,14 @@ std::vector<absl::string_view> ConcatenateReferences(
 
 }  // namespace
 
-void StreamKytheFactsEntries(KytheOutput* kythe_output,
-                             const IndexingFactNode& file_list_facts_tree,
-                             const VerilogProject& project) {
-  const auto indexing_data = ExtractKytheFacts(file_list_facts_tree, project);
-  kythe_output->Emit(indexing_data);
-}
+// Extracted Kythe indexing facts and edges.
+struct KytheIndexingData {
+  // Extracted Kythe indexing facts.
+  std::set<Fact> facts;
+
+  // Extracted Kythe edges.
+  std::set<Edge> edges;
+};
 
 // KytheFactsExtractor processes indexing facts for a single file.
 // Responsible for traversing IndexingFactsTree and processing its different
@@ -278,8 +280,9 @@ class KytheFactsExtractor {
   KytheIndexingData kythe_data_;
 };
 
-KytheIndexingData ExtractKytheFacts(const IndexingFactNode& file_list,
-                                    const VerilogProject& project) {
+void StreamKytheFactsEntries(KytheOutput* kythe_output,
+                             const IndexingFactNode& file_list,
+                             const VerilogProject& project) {
   VLOG(1) << __FUNCTION__;
   // Create a new ScopeResolver and give the ownership to the scope_resolvers
   // vector so that it can outlive KytheFactsExtractor.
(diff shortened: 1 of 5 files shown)


verilog/tools/kythe/kythe_facts_extractor.cc
```diff
@@ -1298,40 +1300,44 @@ void KytheFactsExtractor::CreateEdge(const VName& source_node,
 
 std::ostream& KytheFactsPrinter::PrintJsonStream(std::ostream& stream) const {
   // TODO(fangism): Print function should not be doing extraction work.
-  const auto indexing_data =
-      ExtractKytheFacts(file_list_facts_tree_, *project_);
+  class Printer : public KytheOutput {
+   public:
+    Printer(std::ostream& stream) : stream_(stream) {}
```
fangism, line 1305: `explicit`

verilog/tools/kythe/kythe_facts_extractor.cc
```diff
@@ -1298,40 +1300,44 @@ void KytheFactsExtractor::CreateEdge(const VName& source_node,
...
-      stream << "," << std::endl;
+    void Emit(const Edge& edge) final {
+      if (add_comma) stream_ << "," << std::endl;
+      edge.FormatJSON(stream_, /*debug=*/true) << std::endl;
+      add_comma = true;
     }
-    edge.FormatJSON(stream, /*debug=*/true);
-    should_separate_entries = true;
-  }
+    std::ostream& stream_;
```
fangism, line 1335: `private:` and `add_comma_`

verilog/tools/kythe/kythe_facts_extractor.cc
```diff
@@ -1298,40 +1300,44 @@ void KytheFactsExtractor::CreateEdge(const VName& source_node,
...
-      ExtractKytheFacts(file_list_facts_tree_, *project_);
-
-  stream << "[";
-  bool should_separate_entries = false;
-  for (const Fact& fact : indexing_data.facts) {
-    if (should_separate_entries) {
-      stream << "," << std::endl;
+  class Printer final : public KytheOutput {
+   public:
+    Printer(std::ostream& stream) : stream_(stream) {}
```
fangism, line 1324: `explicit`

verilog/tools/kythe/kythe_facts_extractor.h
```diff
@@ -57,13 +57,13 @@ class KytheFactsPrinter {
...
 
-// Extracted Kythe indexing facts and edges.
-struct KytheIndexingData {
-  // Extracted Kythe indexing facts.
-  std::set<Fact> facts;
-
-  // Extracted Kythe edges.
-  std::set<Edge> edges;
+// Output sink interface for producing the Kythe output.
+class KytheOutput {
```
fangism, line 61: Alternatively: you could formulate this as a pair functions, a FactHandler and an EdgeHandler, no need for any virtual function dispatch, perhaps simpler?  

Up to you, it's a matter of taste.

fangism (approved):

verilog/tools/kythe/kythe_facts_extractor.h
hzeller: I agree (and in fact my first implementation was like that), but I wanted to keep the change small in the first step (as there was already a streaming interface that needed extension).
More refactoring will happen in this file :)

verilog/tools/kythe/kythe_facts_extractor.cc
hzeller: done.

verilog/tools/kythe/kythe_facts_extractor.cc
hzeller: Done.

verilog/tools/kythe/kythe_facts_extractor.cc
hzeller: Done.

PR #1199, opened 2022-02-18
Store string_view in Signature() array, not std::string.

hzeller:
All strings are backed by the text structure ultimately.
There is one string that is generated ad-hoc, which is the
assembled location string "@123:456"; this is now kept in
a stable backing store.

In a test with a particular large file, this resulted in
some significant improvement:
311s -> 232s (25% improvement).

TODO: storing these vectors of strings seems like a waste
of space; couldn't we just have a string-view and a pointer to
the parent signature ?


fangism (approved):
Finally!

PR #1209, opened 2022-02-22
Unroll OR expression on initializer list in MatchesTagAnyOf()

hzeller:
In a particularly large file, this sped up the formatting by
a factor of > 2x (34sec -> 16sec).

### DIFF
--- a/common/text/concrete_syntax_tree.h
+++ b/common/text/concrete_syntax_tree.h
@@ -156,16 +156,16 @@ class SyntaxTreeNode : public Symbol {
     switch (enums.size()) {
       case 4:
         if (*it++ == EnumType(tag_)) return true;
-        [[fallthrough]];
+        ABSL_FALLTHROUGH_INTENDED;
       case 3:
         if (*it++ == EnumType(tag_)) return true;
-        [[fallthrough]];
+        ABSL_FALLTHROUGH_INTENDED;
       case 2:
         if (*it++ == EnumType(tag_)) return true;
-        [[fallthrough]];
+        ABSL_FALLTHROUGH_INTENDED;
       case 1:
         if (*it++ == EnumType(tag_)) return true;
-        [[fallthrough]];
+        ABSL_FALLTHROUGH_INTENDED;
       case 0:
         return false;
       default:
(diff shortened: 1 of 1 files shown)


common/text/concrete_syntax_tree.h
```diff
@@ -149,12 +149,30 @@ class SyntaxTreeNode : public Symbol {
...
   bool MatchesTagAnyOf(std::initializer_list<EnumType> enums) const {
-    // Linear search.
-    auto iter = std::find(enums.begin(), enums.end(), EnumType(tag_));
-    return iter != enums.end();
+    auto it = enums.begin();
+    // Unroll OR expression. Right now, we never have more than 4.
+    switch (enums.size()) {
+      case 4:
+        if (*it++ == EnumType(tag_)) return true;
+        [[fallthrough]];
```
fangism, line 159: There is probably a macro like `ABSL_FALLTHROUGH` for this.

fangism (approved):
A more scalable way to express this may be with a variadic function, instead of an `initializer_list`.  That might make it easier for most compilers to effectively unroll loops like this.

common/text/concrete_syntax_tree.h
hzeller: Done.

hzeller:
I like the idea of a variadic template as it will force the compiler to actually do this at compile time, while now, we only hope that the `initializer_list` is ending up as no overhead abstraction.
However it is used in a context which gets two of these sets in [`IsInsideFirst()`](https://github.com/chipsalliance/verible/blob/348c1f8b083f04bfbe5fd45399d910d1d52075ab/common/text/syntax_tree_context.h#L80-L88), so that would need to be reformulated. Ideas ?

mglb:
> However it is used in a context which gets two of these sets in [`IsInsideFirst()`](https://github.com/chipsalliance/verible/blob/348c1f8b083f04bfbe5fd45399d910d1d52075ab/common/text/syntax_tree_context.h#L80-L88), so that would need to be reformulated. Ideas ?

I think you can use array reference instead of initializer list. This will give you array size as a template parameter while still allowing use of `{...}` in arguments list.
Two example implementations: https://godbolt.org/z/Y95dGYTvE

PR #1219, opened 2022-02-24
Fix possible correctness bug in VectorTree.

hzeller:
After erase(), the iterator to use it is invalid. Use
the new location it returns for the subsequent insert().


fangism (approved):
This might be difficult to unit-test as it is exercising implementation-defined behavior.

PR #1220, opened 2022-02-24
Prepare users of VectorTree to be more agnostic about the internal co…

hzeller:
…ntainer.

This will make it simpler to change the container later. In particular,
not all containers might have reserve(), so pull that operation into a
method that can be implemented according to the containers' needs.

Also, make the `subnodes_type` typedef in VectorTree public for users
to properly define their iterator types.

This is a preparation for other changes to be simpler, so
no functional or container type change in this PR.

### DIFF
--- a/common/util/vector_tree.h
+++ b/common/util/vector_tree.h
@@ -232,13 +232,11 @@ class _VectorTreeImpl {
 template <typename T>
 class VectorTree : private _VectorTreeImpl {
   typedef VectorTree<T> this_type;
-
-  // Self-recursive type that represents children in an expanded view.
-  typedef std::vector<this_type> subnodes_type;
-
   typedef _VectorTreeImpl impl_type;
 
  public:
+  // Self-recursive type that represents children in an expanded view.
+  typedef std::vector<this_type> subnodes_type;
   typedef T value_type;
 
   VectorTree() = default;
@@ -316,6 +314,9 @@ class VectorTree : private _VectorTreeImpl {
     return *this;
   }
 
+  // Hint the implementation how many children are to be expected.
+  void SetExpectedChildrenUpperBound(size_t s) { children_.reserve(s); }
+
   // Builders
 
   // Appends a new child node to the tree at this level.
@@ -359,7 +360,7 @@ class VectorTree : private _VectorTreeImpl {
   // There need not be any relationship between this node and the other.
   void AdoptSubtreesFrom(this_type* other) {
     auto& src_children = other->Children();
-    children_.reserve(children_.size() + src_children.size());
+    SetExpectedChildrenUpperBound(children_.size() + src_children.size());
     AdoptSubtreesFromUnreserved(&src_children);
   }
 
@@ -636,8 +637,7 @@ class VectorTree : private _VectorTreeImpl {
   VectorTree<S> Transform(
       const std::function<S(const this_type& node)>& f) const {
     VectorTree<S> return_tree(f(*this));  // sets Value()
-    auto& return_children = return_tree.Children();
-    return_children.reserve(Children().size());
+    return_tree.SetExpectedChildrenUpperBound(Children().size());
     // Construct children subtree.
     for (const auto& child : Children()) {
       return_tree.AdoptSubtree(child.template Transform<S>(f));
@@ -699,7 +699,7 @@ class VectorTree : private _VectorTreeImpl {
       for (const auto& child : temp) {
         num_grandchildren += child.Children().size();
       }
-      children_.reserve(num_grandchildren);
+      SetExpectedChildrenUpperBound(num_grandchildren);
     }
 
     // Concatenate all grandchildren.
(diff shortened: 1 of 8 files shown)


common/util/vector_tree.h
```diff
@@ -316,6 +314,9 @@ class VectorTree : private _VectorTreeImpl {
     return *this;
   }
 
+  // Hint the implementation how many children are to be expected.
+  void SetExpectedChildrenUpperBound(size_t s) { children_.reserve(s); }
```
fangism, line 318: could just name it something like `ReserveChildren()`?

fangism (approved):

common/util/vector_tree.h
hzeller: I wanted to make the intention clear: that we provide an upper bound.
`Reserve` sounds too close to the actual implementation of the idea.

mglb:
@hzeller
What are your plans related to VectorTree? I'm refactoring VectorTree (and as a side effect also MapTree) too as I'm working on TokenPartitionTree interface.

My concept (roughly):

* Extract all extra methods from VectorTree as template functions. Leave only `Parent()`, `Children()`, `Value()`.
  * The extracted functions operate on all types that fulfil **TreeNode** "concept":
    * has `Parent()` method returning pointer to TreeNode
    * has `Children()` method returning reference to a TreeNodes range with methods: `push_back` (for adding children) and `assign` (for adopting subtrees)
    * has `Value()` method returning some value.
  * A lot of duplicated code/declarations is removed:
    * The functions replace many methods from MapTree (which is in large part a copy of VectorTree)
    * Single template function is able to handle both const and mutable nodes.
  * It looks like some methods (`Flatten*()`, `HoistOnlyChild()`, etc) are here just for use in TokenPartitionTree - they can be moved to TokenPartitionTree implementation itself.
  * It is now really simple to create a custom tree type - just implement `Parent()`, `Children()`, and optionally `Value()` and the tree functions will work.

* Create `ContainerProxyBase` class, intended to be subclassed using CRTP which would then be returned by a TreeNode's `Children()` method.
  * The subclass would need to implement following functions:
    * `SomeSTLContainer& UnderlyingContainer()` - returns "real" container
    * `void ItemsAdded/ItemsBeingRemoved/ItemsBeingReplaced(Iterator first, Iterator last, ...)` - handle added/removed/replaced items
  * The class' public interface would mimic standard STL container and would just forward method calls to a container returned by `UnderlyingContainer()`. However, it would also call `ItemsAdded`/`ItemsBeingRemoved`/`ItemsBeingReplaced` where suitable. For convenience, it could have extra methods like `reserve()` that would just be noop (or SFINAE-removed) when underlying container does not support them.
  * Example subclass implementation (inside stripped VectorTree class):
    ```c++
    template <typename T>
    class VectorTree {
      using node_type = VectorTree<T>;
      using child_iterator = typename std::vector<node_type>::iterator;
    
      class ContainerProxy; // forward declaration
    
    public:
      node_type* Parent() { return parent_; }
      ContainerProxy Children() { return ContainerProxy(this); }
    
    private:
      class ContainerProxy: private ContainerProxyBase<ContainerProxy, node_type> {
        using base_type = ContainerProxyBase<ContainerProxy, node_type>;
        friend base_type;
        
      public:
        ContainerProxy(node_type* node): node_(node) {}
    
        using base_type::push_back;
        using base_type::emplace_back;
        using base_type::back;
        // ...
    
      private:
        std::vector<node_type>& UnderlyingContainer() { return node_->children_; }
        void ItemsAdded(child_iterator first, child_iterator last) {
          // Set inserted nodes' parent
          for (auto it = first; it != last; ++it) {
            it->parent_ = node_;
          }
        }
        // ItemsBeingRemoved not needed here;
        // ItemsReplaced skipped - it could just call ItemsAdded
    
        node_type* node_;
      };
      friend ContainerProxy;
    
      node_type* parent_ = nullptr;
      std::vector<node_type> children_;
    };
    ```
    In fact, this is almost complete VectorTree implementation (there's no `Value()`). Example use:
    ```c++
    VectorTree<int> tree;
    tree.Children().push_back(VectorTree<int>());
    printf("tree:         %p\n", &tree);
    printf("child:        %p\n", &tree.Children().back());
    printf("child parent: %p\n", tree.Children().back().Parent());

    // Prints:
    // tree:         0x7ffd8f3c45c0
    // child:        0x10a3eb0
    // child parent: 0x7ffd8f3c45c0
    ```
    Still works after `s/vector/list/g`.

hzeller:
Oh very nice, that sounds like excellent improvement ideas @mglb !

The motivation of my change was that I think the fact that we have to `reserve()` to keep stability of pointers _very_ brittle, which results in error-prone use and hard-to-debug latent issues.
This PR's change makes it easier to switch to another container that at least allows random access `operator[]` (e.g. `std::deque`) to be used (there are actually still some test failures if one would actually do that right now, would need some investigation what is going on there).

Anyway, I don't have any large change pending for this class and I am looking forward to your implementation!

PR #1222, opened 2022-02-25
Use final instead of override if possible.

hzeller:
Mark all methods at the end of the inheritance chain
as final. This serves as documentation for the human reader
and also provides opportunities for the compiler to consider
devirtualization.

### DIFF
--- a/common/analysis/command_file_lexer_test.cc
+++ b/common/analysis/command_file_lexer_test.cc
@@ -40,7 +40,7 @@ class FilteredCommandFileLexer : public CommandFileLexer {
     }
   }
 
-  const verible::TokenInfo& DoNextToken() override {
+  const verible::TokenInfo& DoNextToken() final {
     do {
       CommandFileLexer::DoNextToken();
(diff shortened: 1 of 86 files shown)


verilog/tools/kythe/indexing_facts_tree_extractor_test.cc
```diff
@@ -71,9 +71,9 @@ struct TestFileEntry {
       absl::string_view code_text, absl::string_view temp_dir,
       const std::function<const VerilogSourceFile*(absl::string_view)>&
           file_opener,
-      absl::string_view override_basename = "")
+      absl::string_view final_basename = "")
```
fangism, line 74: The original name probably makes more sense here.

fangism (approved):

verilog/tools/kythe/indexing_facts_tree_extractor_test.cc
hzeller: Oops, that happens if `sed` is used for mass-editing :)
Fixed.

PR #1226, opened 2022-02-28
Improve KeywordIdentifier detection in symbol table using

### LINKED ISSUE #917: verible-verilog-project crashes on this sample file
Need to drill down what the problem is, but here some random file from ibex `examples/simple_system/rtl/ibex_simple_system.sv` make the project tool crash:

To reproduce
```
bazel build verilog/tools/project:verible-verilog-project

IBEX_DIR=/tmp/test/ibex
git clone https://github.com/lowrisc/ibex ${IBEX_DIR}

echo "${IBEX_DIR}/examples/simple_system/rtl/ibex_simple_system.sv" > /tmp/simple.list
bazel-bin/verilog/tools/project/verible-verilog-project symbol-table-defs --file_list_root=/ --file_list_path=/tmp/simple.list
```

Output
```
F20210907 16:05:04.105469 1043670 logging.h:39] Check failed: t != nullptr reference_branch_point_
    @     0x556d1ab1fd58  google::LogMessage::Fail()
    @     0x556d1ab1fca3  google::LogMessage::SendToLog()
    @     0x556d1ab1f5ca  google::LogMessage::Flush()
    @     0x556d1ab22dae  google::LogMessageFatal::~LogMessageFatal()
    @     0x556d1aa4d5e7  verible::DieIfNull<>()
    @     0x556d1aa4884e  verilog::SymbolTable::Builder::HandleIdentifier()
    @     0x556d1aa48c62  verilog::SymbolTable::Builder::Visit()
    @     0x556d1ab14250  verible::SyntaxTreeLeaf::Accept()
    @     0x556d1aaf5f81  verible::TreeContextVisitor::Visit()
    @     0x556d1aa46b7f  verilog::SymbolTable::Builder::Descend()
    @     0x556d1aa4699b  verilog::SymbolTable::Builder::Visit()
    ...
```

hzeller:
IsIdentifierLike()

The overlapping keywords and identifiers are the result of
mis-classification in the SymbolTable by simply looking
at SymbolIdentifier; instead, use the IsIdentifierLike()
predicate function.

Also, while at it, reduce the number of special cases by
only keep the KeywordIdentifier tokens that are also used
in other places of the grammar.

Long term, we need to make sure to properly handle
KeywordIdentifiers in the whole stack; it seems like there
are other subtle bugs lurking.

One approach would be to change the token tag to
SymbolIdentifier if we happen to encounter it in
KeywordIdentifiers, but currently there are hidden assumptions
that this is not happening.

Fixes #917


hzeller:
Ok, looks like using `IsIdentifierLike()` as an indicator to treat a leaf as identifier results in more issues surfaced by the smoke test. So for now, just reduce the number of unnecessary keywords.

verilog/analysis/symbol_table_test.cc
```diff
@@ -1469,6 +1469,23 @@ TEST(BuildSymbolTableTest, ModuleInstanceNamedParameterAssignment) {
   }
 }
 
+TEST(BuildSymbolTableTest, KeywordAsIdentifierRegressionIssue917) {
```
fangism, line 1472: Maybe edit test name to describe the issue being tested?

fangism (approved):
Just be warned that you might see failures on ivltests (which might exercise Verilog-AMS).  

Might also be worth checking with any open issues able dialect selection.

hzeller:
all the removed keywords were _only_ used in the KeywordIdentifier, not anywhere else in the grammar. In which way could we see failures here ?

verilog/analysis/symbol_table_test.cc
hzeller: Done.

hzeller:
Files with syntax errors detected in the [ivtest](https://github.com/steveicarus/ivtest) (200 out of 2789) is the same for
  * source head CI https://github.com/chipsalliance/verible/runs/5366850371?check_suite_focus=true#step:7:969
  * This pull request https://github.com/chipsalliance/verible/runs/5366870155?check_suite_focus=true#step:7:907

PR #1237, opened 2022-03-01
Don't treat built-in functions such as cos() special.

hzeller:
We are not treating built-in functions in a special way but
it makes code operating on the syntax tree more brittle as
it potentially misses out on a class of identifiers.


verilog/CST/net_test.cc
```diff
@@ -149,15 +149,6 @@ TEST(GetIdentifiersFromNetDeclarationTest, EmptySource) {
   EXPECT_TRUE(net_declarations.empty());
 }
 
-TEST(GetIdentifiersFromNetDeclarationTest, NoNetIdentifier) {
```
hzeller, line 152: Now that `sin` is not a special token anymore, this will not work as expected.

(Also, I have honestly no idea what this test is testing for)

hzeller:
The downside I could see would be that `verible-verilog-obfuscate` would now translate all these build-in functions to random identifier strings. Not sure how the obfuscator is used typically, so don't know if this would be an issue in practice.

fangism (approved):
Ok, but can you file some issues for these, while we've thought of them:

* protect some of these built-in functions from the obfuscator, because this would alter semantics (we lose this with this PR)
* check number of arguments for built-in math functions (we don't already do this, so low priority)
* handle built-in functions in the symbol table (not implemented yet)

hzeller:
I've modified the obfuscator to keep bulilt-in functions as-is (at the expense that they will make it to the saved mapping file, but that can't harm)

fangism (approved):

PR #1240, opened 2022-03-02
Don't rely on implicit const char* -> bool conversion.

hzeller:
While here only used in a test, this is an error-prone pattern.

Also fix slightly changed dependencies needed with the recent move
of using random identifier generation closer to the specific
verilog tool.

### DIFF
--- a/verilog/transform/obfuscate_test.cc
+++ b/verilog/transform/obfuscate_test.cc
@@ -26,7 +26,7 @@ namespace {
 using verible::IdentifierObfuscator;
 
 static std::string ExpectNeverToBeCalled(absl::string_view) {
-  EXPECT_FALSE("This identifier generator should've never been called");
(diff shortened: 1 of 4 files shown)


verilog/transform/obfuscate_test.cc
```diff
@@ -26,7 +26,7 @@ namespace {
 using verible::IdentifierObfuscator;
 
 static std::string ExpectNeverToBeCalled(absl::string_view) {
-  EXPECT_FALSE("This identifier generator should've never been called");
+  EXPECT_TRUE(false) << "This identifier generator should've never been called";
```
mglb, line 29: Alternative: `ADD_FAILURE() << "...";` - does the same thing, but without "expected: true, actual: false"

hzeller: Good idea. Done.

PR #1244, opened 2022-03-03
Add ContainerProxy

mglb:
TL;DR: Utility class needed for https://github.com/chipsalliance/verible/pull/1252, potentially useful in some other cases.

---

CRTP base class for creating modification tracking STL container proxies.

The class allows safe exposition of a STL container to an user in cases where
changes to the container must be pre-/post-processed. This is done through
implementation of several event handling methods in a derived class.

Example use case: List of child nodes in a tree which sets parent when
a child is inserted.

Note that the proxy only tracks container changes. It does not track
individual elements.

Derived class interface
-----------------------

- Mandatory:
  - `ContainerType& underlying_container()`,
    `const ContainerType& underlying_container() const`:
    Returns reference to the wrapped container.

- Optional:
  - `void ElementsInserted(iterator first, iterator last)`:
    Called when new elements were inserted. `first` and `last` (exclusive)
    refer to inserted elements.
  - `void ElementsBeingRemoved(iterator first, iterator last)`:
    Called just before removing elements from `first` to `last` (exclusive).
  - `void ElementsBeingReplaced()`:
    Called when all elements are going to be replaced (due to assignment).
  - `void ElementsWereReplaced()`:
    Called just after all elements were replaced.

Usage
-----

1. Subclass:

       using ContainerType = std::vector<int>
       class MyProxy: private ContainerProxyBase<MyProxy, ContainerType> {
         using Base = ContainerProxyBase<MyProxy, ContainerType>;
         friend Base;
         // ...
       };

2. Implement mandatory interface:

       private:
         ContainerType& underlying_container() {
           return container_;
         }
         const ContainerType& underlying_container() const {
           return container_;
         }
       // ...
       ContainerType container_; // not a part of the interface

3. Inherit public interface (all or only some members):

       public:
         using typename Base::container_type;
         using typename Base::value_type;
         using typename Base::reference;
         using typename Base::const_reference;
         // ...
         using Base::begin;
         using Base::end;
         // ...

   Note: inheriting (or implementing) `void swap(MyProxy&)` will also provide
   standalone function `void swap(MyProxy&, MyProxy&)`.

4. Implement optional interface:

       private:
         void ElementsInserted(iterator first, iterator last) {
           /* ... */
         }
         void ElementsBeingRemoved(iterator first, iterator last) {
           /* ... */
         }
         void ElementsBeingReplaced() { /* ... */ }
         void ElementsWereReplaced() { /* ... */ }

5. Optional: implement operator=(const MyProxy&) and operator=(MyProxy&&)
   (base class' `operator=` works only with `ContainerType`):

       public:
         MyProxy& operator=(const MyProxy& other) {
           *this = other.underlying_container();
           return *this;
         }
         MyProxy& operator=(MyProxy&& other) {
           // You probably want to notify `other` about invalidation
           // of its **whole container** (not just elements).
           *this = std::move(other.underlying_container());
           return *this;
         }

Limitations
-----------

- Tested only with std::vector and std::list, but it should work with other
  SequenceContainers.

### DIFF
--- a/common/util/container_proxy.h
+++ b/common/util/container_proxy.h
@@ -1,0 +1,458 @@
+// Copyright 2017-2022 The Verible Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef VERIBLE_COMMON_UTIL_CONTAINER_PROXY_H_
+#define VERIBLE_COMMON_UTIL_CONTAINER_PROXY_H_
+
+#include <initializer_list>
+#include <iterator>
+#include <type_traits>
+#include <utility>
+
+namespace verible {
+
+// CRTP base class for creating modification tracking STL container proxies.
+//
+// The class allows safe exposition of a STL container to an user in cases where
+// changes to the container must be pre-/post-processed. This is done through
+// implementation of several optional event handling methods in a derived class.
+// Event handling methods that are not implemented do not introduce overhead.
+//
+// Example use case: List of child nodes in a tree which sets parent when
+// a child is inserted.
+//
+// Note that the proxy only tracks container changes. It does not track
+// individual elements.
+//
+// Derived class interface
+// -----------------------
+//
+// - Mandatory:
+//   - `ContainerType& underlying_container()`,
+//     `const ContainerType& underlying_container() const`:
+//     Returns reference to the wrapped container.
+//
+// - Optional:
+//   - `void ElementsInserted(iterator first, iterator last)`:
+//     Called when new elements were inserted. `first` and `last` (exclusive)
+//     refer to inserted elements.
+//   - `void ElementsBeingRemoved(iterator first, iterator last)`:
+//     Called just before removing elements from `first` to `last` (exclusive).
+//   - `void ElementsBeingReplaced()`:
+//     Called when all elements are going to be replaced (due to assignment).
+//   - `void ElementsWereReplaced()`:
+//     Called just after all elements were replaced.
+//
+// Usage
+// -----
+//
+// 1. Subclass:
+//
+//        using ContainerType = std::vector<int>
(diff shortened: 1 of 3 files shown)


common/util/container_proxy.h
```diff
@@ -0,0 +1,454 @@
...
+  }
+
+  reference at(size_type index) { return container().at(index); }
+  const_reference at(size_type index) const { return container().at(index); }
+
+  // Modifiers (inserting)
+
+  // TODO(mglb): provide methods taking `iterator` instead of `const_iterator`.
+  // In most cases `iterator` is convertible to `const_iterator` without any
+  // cost, but that,'s not guaranteed.
```
hzeller, line 188: `s/that,'s/that is/`

common/util/container_proxy.h
```diff
@@ -0,0 +1,454 @@
...
+//          }
+//          MyProxy& operator=(MyProxy&& other) {
+//            // You probably want to notify `other` about invalidation
+//            // of its **whole container** (not just elements).
+//            *this = std::move(other.underlying_container());
+//            return *this;
+//          }
+//
+template <class DerivedType, class ContainerType>
+class ContainerProxyBase {
```
hzeller, line 123: The name is not sufficiently conveying what this class does.
I think it should be possible to see that this provides way to observe change, so maybe something along the lines of `ObservableContainerDelegatorBase` ... even though it is a mouth-full.
Maybe you have better ideas.

We can also defer the decision to somewhere in the near future once we have a good name, so that we don't stall the current effort.

hzeller (approved):
LGTM
Some smallish remarks, but essentially ready to merge.

common/util/container_proxy.h
```diff
@@ -0,0 +1,416 @@
...
+#include <type_traits>
+#include <utility>
+
+namespace verible {
+
+// CRTP base of STL container wrapper that notifies about container changes.
+//
+// The class allows safe exposition of a STL container to an user in cases where
+// changes to the container must be pre-/post-processed. This is done through
+// implementation of several event handling methods in a derived class.
```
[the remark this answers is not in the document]
mglb: Done

common/util/container_proxy.h
mglb: Done

common/util/container_proxy.h
mglb: I prefer to defer this decision. I've added a TODO comment in the code.

Ideas/notes for future reference:
- Something like "ContainerObserverBase". The derived class is the one that observes a container. Delegation is an option. I assume that since the base class is a CRTP base/template it shouldn't be viewed as a separate entity that delegates something.
- It would be useful to convey that the class not only observes, but also provides container-like interface.

PR #1246, opened 2022-03-03
Flatten associative binary operations.

### LINKED ISSUE #1236: Long chains of associative operator expressions result in tree whose processing typically crashes the stack
Not so much in hand-written code, but in machine generated code, we often see huge chains of expressions with the same associative operator. These end up in a degenerated parse tree essentially being a linked list but are typically processed recursively and make several of the verible tools very unhappy.

let's generate such a long expression:

`gen-long_expr.sh`
```bash
#!/bin/bash
LEN=1000000
echo -n "module long_expr(input logic in[$LEN]); wire combine = 0 "
for (( i=0; i < $LEN; ++i)); do
   echo "  ^in[$i]"
done
echo ";endmodule"
```

```
./gen-long_expr.sh > long_expr.sv
```

Processing these, the tools suffer from walking the stack or simply segfault due to recursion limits they run into

```
verible-verilog-syntax long_expr.sv
verible-verilog-project symbol-table-defs --file_list_root=. --file_list_path <(echo "long_expr.sv")
verible-verilog-kythe-extractor --file_list_root=. --file_list_path <(echo "long_expr.sv")
verible-verilog-lint long_expr.sv
verible-verilog-format long_expr.sv
```

It might be worthwhile to internally replace the super-deep nested tree with an expression node for the particular operator with a large amount of sibling children that can be processed iteratively.

fangism:
This reduces stack depth for long chained expressions.

Fixes #1236

### DIFF
--- a/verilog/CST/BUILD
+++ b/verilog/CST/BUILD
@@ -296,6 +296,7 @@ cc_library(
         "//common/text:tree_utils",
         "//common/util:casts",
         "//common/util:logging",
+        "//verilog/parser:verilog_token_classifications",
         "//verilog/parser:verilog_token_enum",
         "@com_google_absl//absl/strings",
     ],
--- a/verilog/CST/expression.cc
+++ b/verilog/CST/expression.cc
@@ -98,6 +98,11 @@ const verible::Symbol* GetConditionExpressionFalseCase(
   return GetSubtreeAsSymbol(condition_expr, NodeEnum::kConditionExpression, 4);
 }
 
+std::vector<verible::TreeSearchMatch> FindAllBinaryOperations(
+    const verible::Symbol& root) {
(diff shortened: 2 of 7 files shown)


hzeller:
Can you have a look at the [smoke-tests](https://github.com/chipsalliance/verible/runs/5414627731?check_suite_focus=true) that became less happy (and apparently found things not covered by any tests)

hzeller:
`verible-verilog-format` still crashes on the long expression generated with the script in #1236, but that seems to be due to some nested allocation in `common/formatting/state_node.h` in a very deeply nested stack; so there probably something is still allocated recursively.
But that should probably be fixed in a separate PR.

hzeller (approved):
LGTM!

PR #1249, opened 2022-03-05
Show an example implementation of using project policies.

hzeller:
This just provides an empty GetBuiltinProjectPolicies()
in project_policies.cc file that shows how built-in policies
can be added to verible.

Better document the ProjectPolicy struct.

Use the these built-in policies in verble-verilog-lint and
verible-verilog-ls.

### DIFF
--- a/verilog/analysis/BUILD
+++ b/verilog/analysis/BUILD
@@ -332,6 +332,14 @@ cc_test(
 )
 
 cc_library(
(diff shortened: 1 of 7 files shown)


hzeller:
(Remark by David: ideally, we actually remove the code instead of suggesting it would be a usable idea)

PR #1252, opened 2022-03-06
Use ContainerProxy for VectorTree's children list

mglb:
This PR implements `VectorTreeChildrenList` class, which extends `ContainerProxyBase` to create a wrapper for `std::vector` containing VectorTree nodes. The class automatically sets correct `parent_` pointer in each inserted node, where the "correct" pointer is passed in a constructor.
The class is used in VectorTree as a child nodes storage (`children_` member), and as a public interface for children list access and manipulation (retured as a reference from `Children()` method).
As a result VectorTree's `parent_` member is only set/modified either in VectorTree's constructor or assignment, or inside VectorTreeChildrenList.

Only VectorTree methods were modified to take advantage of this new class. Other code still uses methods like `NewChild()` for inserting nodes. I'll update the codebase to use the new interface in near future, after finishing one more VectorTree refactoring.

### DIFF
--- a/common/util/vector_tree.h
+++ b/common/util/vector_tree.h
@@ -200,166 +200,6 @@ class _VectorTreeImpl {
   }
 };
 
-namespace vector_tree_internal {
-
-// A wrapper of a sequence container for storing VectorTree nodes that sets
-// correct parent pointer in each inserted node. The reference to the "correct"
-// parent is passed to a constructor.
-//
-// The sole purpose of this class is to function as a children list in
-// VectorTree, both as a storage and as a public interface for tree manipulation
-// (through reference).
-//
-// This class handles parent pointer assignment for all cases where the children
-// list itself is modified. However, it does not clear or otherwise change
-// parent pointer in removed nodes.
-template <class Container>
-class VectorTreeChildrenList
-    : ContainerProxyBase<VectorTreeChildrenList<Container>, Container> {
-  using Base = ContainerProxyBase<VectorTreeChildrenList<Container>, Container>;
-  friend Base;
-
-  using ThisType = VectorTreeChildrenList<Container>;
-
-  using VectorTreeType = typename Container::value_type;
-  friend VectorTreeType;
-
- public:
-  using typename Base::container_type;
-
-  // Sequence Container interface
-
-  using typename Base::value_type;
-
-  using typename Base::const_reference;
-  using typename Base::reference;
-
-  using typename Base::const_iterator;
-  using typename Base::iterator;
-
-  using typename Base::difference_type;
-  using typename Base::size_type;
-
-  using typename Base::const_reverse_iterator;
-  using typename Base::reverse_iterator;
-
-  using Base::begin;
-  using Base::cbegin;
-  using Base::cend;
-  using Base::end;
-
-  using Base::crbegin;
-  using Base::crend;
-  using Base::rbegin;
-  using Base::rend;
-
-  using Base::back;
-  using Base::front;
-  using Base::operator[];
-  using Base::at;
-
-  using Base::empty;
-  using Base::max_size;
-  using Base::size;
-
-  using Base::emplace_back;
-  using Base::push_back;
-
-  using Base::emplace_front;
-  using Base::push_front;
-
-  using Base::emplace;
-  using Base::insert;
-
-  using Base::clear;
-  using Base::erase;
-  using Base::pop_back;
-  using Base::pop_front;
-
-  using Base::assign;
-  using Base::operator=;
-  using Base::swap;
-
-  using Base::capacity;
-  using Base::reserve;
-  using Base::resize;
-
- protected:
-  // ContainerProxy interface
-
-  container_type& underlying_container() { return container_; }
-  const container_type& underlying_container() const { return container_; }
-
-  void ElementsInserted(iterator first, iterator last) {
-    LinkChildrenToParent(iterator_range(first, last));
-  }
-
-  // Unused:
-  // void ElementsBeingRemoved(iterator first, iterator last)
-
-  // Unused:
-  // void ElementsBeingReplaced()
-
-  void ElementsWereReplaced() { LinkChildrenToParent(container_); }
-
- private:
-  // Sets parent pointer of nodes from `children` range to address of `node_`.
-  template <class Range>
-  void LinkChildrenToParent(Range&& children) {
-    for (auto& child : children) {
-      child.parent_ = &node_;
-    }
-  }
-
-  // Hide constructors and assignments from the world. This object is created
-  // and assigned-to only in VectorTree.
-
-  explicit VectorTreeChildrenList(VectorTreeType& node) : node_(node) {}
-
-  // Construction requires parent node reference.
-  VectorTreeChildrenList(const VectorTreeChildrenList&) = delete;
-
-  VectorTreeChildrenList(VectorTreeType& node,
-                         const VectorTreeChildrenList& other)
-      : node_(node), container_(other.container_) {
-    LinkChildrenToParent(container_);
-  }
-
-  VectorTreeChildrenList& operator=(const VectorTreeChildrenList& other) {
-    container_ = other.container_;
-    LinkChildrenToParent(container_);
-    return *this;
-  }
-
-  // Construction requires parent node reference.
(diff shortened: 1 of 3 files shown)


common/util/vector_tree.h
```diff
@@ -239,22 +407,18 @@ class VectorTree : private _VectorTreeImpl {
   typedef std::vector<this_type> subnodes_type;
```
hzeller, line 407: Shouldn't this be typedef'ed to`VectorTreeChildrenList` now ?

common/util/vector_tree.h
```diff
@@ -374,9 +523,9 @@ class VectorTree : private _VectorTreeImpl {
 
   const this_type* Parent() const { return parent_; }
 
-  subnodes_type& Children() { return children_; }
+  auto& Children() { return children_; }
```
hzeller, line 526: Since we know the return type, why `auto` ?

hzeller (commented):
First quick round.

common/util/vector_tree.h
mglb: Not really. It is used for:
- Getting iterator types - `VectorTreeChildrenList` would be OK for this.
- Creating some temporary container which is compatible with node's children list ([align_test.cc](https://github.com/chipsalliance/verible/blob/master/common/formatting/align_test.cc#L1492), [vector_tree.h](https://github.com/chipsalliance/verible/pull/1252/files#diff-46de95dbc32f853fc3fbec5c4e9e6e01c7906fe79b5e49006d447c13a5a23335R843-R844)) - `VectorTreeChildrenList` is either impossible to use (it can't be constructed outside of VectorTree) or adds overhead of parent pointer assigning when this is not needed.

I think in the next refactoring (conversion of VectorTree methods into functions operating on a template type fulfilling "Tree" concept) this typedef will be probably removed and replaced with use of `tree_traits` class.

common/util/vector_tree.h
mglb: Replaced with `VectorTreeChildrenList`.

common/util/vector_tree.h
```diff
@@ -198,6 +200,172 @@ class _VectorTreeImpl {
...
+
+  explicit VectorTreeChildrenList(VectorTreeType& node) : node_(node) {}
+
+  // Construction requires parent node reference.
+  VectorTreeChildrenList(const VectorTreeChildrenList&) = delete;
+
+  VectorTreeChildrenList(VectorTreeType& node,
+                         const VectorTreeChildrenList& other)
+      : node_(node), container_(other.container_) {
+    for (auto& child : container_) {
```
fangism, line 323: Same function as `ElementsWereReplaced()`?

common/util/vector_tree.h
```diff
@@ -198,6 +200,172 @@ class _VectorTreeImpl {
...
+  using Base::reserve;
+  using Base::resize;
+
+ protected:
+  // ContainerProxy interface
+
+  container_type& underlying_container() { return container_; }
+  const container_type& underlying_container() const { return container_; }
+
+  void ElementsInserted(iterator first, iterator last) {
```
fangism, line 293: As far as naming goes, this function doesn't seem to be aware (or need to be aware) that elements were inserted, it is merely re-linking children back to the parent.  Consider renaming this something that suggests `RelinkChildrenRange()`?

common/util/vector_tree.h
```diff
@@ -198,6 +200,172 @@ class _VectorTreeImpl {
...
+    }
+  }
+
+  // Unused:
+  // void ElementsBeingRemoved(iterator first, iterator last)
+
+  // Unused:
+  // void ElementsBeingReplaced()
+
+  void ElementsWereReplaced() {
```
fangism, line 305: Similar naming suggestion, something like `RelinkContainerChildren()` would be more self-descriptive.

common/util/vector_tree.h
```diff
@@ -198,6 +200,172 @@ class _VectorTreeImpl {
...
+                         const VectorTreeChildrenList& other)
+      : node_(node), container_(other.container_) {
+    for (auto& child : container_) {
+      child.parent_ = &node;
+    }
+  }
+
+  VectorTreeChildrenList& operator=(const VectorTreeChildrenList& other) {
+    container_ = other.container_;
+    for (auto& child : container_) {
```
fangism, line 330: Call one of the above helper functions.
(Same elsewhere below.)

common/util/vector_tree.h
```diff
@@ -266,14 +434,15 @@ class VectorTree : private _VectorTreeImpl {
...
+      : node_value_(v), children_(*this) {
     AdoptSubtree(std::forward<Args>(args)...);
   }
 
   template <typename... Args>
   explicit VectorTree(value_type&& v, Args&&... args)
-      : node_value_(std::move(v)), children_() {
-    AdoptSubtree(std::forward<Args>(args)...);
+      : node_value_(std::move(v)), children_(*this) {
+    children_.reserve(sizeof...(args));
```
fangism, line 444: Nice!

common/util/vector_tree.h
```diff
@@ -198,6 +200,172 @@ class _VectorTreeImpl {
...
+    container_ = std::move(other.container_);
+    for (auto& child : container_) {
+      child.parent_ = &node_;
+    }
+    return *this;
+  }
+
+  // Reference to a VectorTree node in which this object represents a list of
+  // children.
+  VectorTreeType& node_;
```
fangism, line 361: One minor concern is that relative to the previous implementation, this member increases the size of a node by an extra pointer.

common/util/vector_tree.h
```diff
@@ -198,6 +200,172 @@ class _VectorTreeImpl {
...
+template <class Container>
+class VectorTreeChildrenList
+    : ContainerProxyBase<VectorTreeChildrenList<Container>, Container> {
+  using Base = ContainerProxyBase<VectorTreeChildrenList<Container>, Container>;
+  friend Base;
+
+  using ThisType = VectorTreeChildrenList<Container>;
+
+  using VectorTreeType = typename Container::value_type;
+  friend VectorTreeType;
```
fangism, line 225: Alternative to `friend` relationship, might this helper class be suitable as a private nested class of `VectorTree`?  (Then private access becomes easier.)

common/util/vector_tree.h
mglb: I've created [`LinkChildrenToParent`](https://github.com/chipsalliance/verible/pull/1252/commits/48ff8a3d4d4933600a4de710aaf8394df2f342d1) method containing the parent assigning for loop. This method is called here and in all places where the assigning code has been used before.

common/util/vector_tree.h
mglb: This name is a part of [generic base class interface](https://github.com/chipsalliance/verible/blob/master/common/util/container_proxy.h#L38-L55) (i.e. an event handler).

common/util/vector_tree.h
mglb: The name is a part of generic base class interface (like `ElementsInserted`).

common/util/vector_tree.h
mglb: Calling `LinkChildrenToParent` now.

common/util/vector_tree.h
mglb: Yes, unfortunately. Alternatives I've tried:
- `VectorTreeChildrenList` as a reference-like class returned from `Children()`. Problems: requires two classes (for const and non-const references) and use of `auto` without `&` everywhere.
- `offsetof` used to calculate pointer to node from pointer to `children_` (i.e. from `this` inside `VectorTreeChildrenList`). Problem: `offsetof` requires stanadard layout class, while both value type and container can have non-standard layout (e.g. with `GLIBCXX_DEBUG` std::vector is not standard-layout).

common/util/vector_tree.h
mglb: Good idea, thanks. Implemented & changed the class name to just `ChildrenList`.

common/util/vector_tree.h
hzeller: Maybe for now, we should add a TODO here for later reducing the footprint.

hzeller (approved):
Merging now so that follow-up changes can proceed, but please add the potential memory footprint TODO at the `node_` instance variable.

Issue #1255, opened 2022-03-07
symbol-table: binding with parameters crashes (in DeclareInstance())


hzeller:
Example:

```SystemVerilog
bind a b c (.d);
```

(e.g. [real world example](https://github.com/lowRISC/ibex/blob/master/vendor/lowrisc_ip/ip/prim/fpv/tb/prim_secded_inv_hamming_76_68_bind_fpv.sv))

Crashes with
```
F20220307 14:30:01.151845 2584532 logging.h:40] Check failed: (t != nullptr) verilog/analysis/symbol_table.cc:939: declaration_type_info_
*** Check failure stack trace: ***
    @     0x557a213934a2  google::LogMessage::Fail()
    @     0x557a213a4b75  google::LogMessage::SendToLog()
    @     0x557a21392b06  google::LogMessage::Flush()
    @     0x557a213944fd  google::LogMessageFatal::~LogMessageFatal()
    @     0x557a2119548f  verible::DieIfNull<>()
    @     0x557a211b28c6  verilog::SymbolTable::Builder::EmplaceTypedElementInCurrentScope()
    @     0x557a211bae70  verilog::SymbolTable::Builder::DeclareInstance()
    @     0x557a211c079d  verilog::SymbolTable::Builder::Visit()
    @     0x557a21375253  verible::SyntaxTreeNode::Accept()
[...]
```

This looks related/similar to #1241 where a `bind` attempt results in a `nullptr` declaration_type_info_ in `DeclareRegister()`

hzeller:
A simple work-around would be to
```patch
--- a/verilog/analysis/symbol_table.cc
+++ b/verilog/analysis/symbol_table.cc
@@ -1146,6 +1146,7 @@ class SymbolTable::Builder : public TreeContextVisitor {
     const verible::TokenInfo* instance_name_token =
         GetModuleInstanceNameTokenInfoFromGateInstance(instance);
     if (!instance_name_token) return;
+    if (!declaration_type_info_) return;  // #1255
     const absl::string_view instance_name(instance_name_token->text());
     const SymbolTableNode& new_instance(EmplaceTypedElementInCurrentScope(
         instance, instance_name, SymbolMetaType::kDataNetVariableInstance));
```

... but I suspect there is a deeper reason why it fails here.

hzeller:
(worked around for now by skipping `bind` in https://github.com/hzeller/verible/commit/4d65a5c517fbd8d6cc2ac5d8de1c2b790fe15aa8 )

Issue #1241, opened 2022-03-02
symbol-table: crash


hzeller:
The following (fuzzer-generated) code will crash the buliding of the symbol table in the project-tool

/tmp/testfile.sv
```SystemVerilog
bind a b c;
```

Looks like it happens somwhere in `DeclareRegister()` when calling `EmplaceTypedElementInCurrentScope()`
```
bazel build -c dbg verilog/tools/project:verible-verilog-project
bazel-bin/verilog/tools/project/verible-verilog-project symbol-table-defs --file_list_root=/ --file_list_path <(echo /tmp/testfile.sv)
F20220302 13:58:07.653908 1449685 logging.h:40] Check failed: (t != nullptr) verilog/analysis/symbol_table.cc:896: declaration_type_info_
*** Check failure stack trace: ***
    @     0x558643e25702  google::LogMessage::Fail()
    @     0x558643e2564c  google::LogMessage::SendToLog()
    @     0x558643e24f63  google::LogMessage::Flush()
    @     0x558643e288c8  google::LogMessageFatal::~LogMessageFatal()
    @     0x558643d53e76  verible::DieIfNull<>()
    @     0x558643d4f8e3  verilog::SymbolTable::Builder::EmplaceTypedElementInCurrentScope()
    @     0x558643d5085c  verilog::SymbolTable::Builder::DeclareRegister()
[...]
```

hzeller:
(worked around for now by skipping `bind` in https://github.com/hzeller/verible/commit/4d65a5c517fbd8d6cc2ac5d8de1c2b790fe15aa8 )

PR #1257, opened 2022-03-07
symbol-table: Skip bind statement for now to avoid crashes.

hzeller:
Issues #1241 #1255


fangism (approved):
Good strategy to avoid crashing on the unhandled cases.

PR #1265, opened 2022-03-15
Extract VectorTree methods

mglb:
uThis PR introduces TreeNode concept and converts VectorTree methods to standalone generic functions that operate on types fulfilling the concept.
There's also `TreeNodeTraits` struct which allows obtaining information (supported features, types, etc) about various types fullfiling TreeNode concept.

## TreeNode concept

Class `T` must contain at least the following member in order to fulfill the
TreeNode concept:

- `Container<T>& Children()`:
  Returns reference to a STL-like container or a range containing the node's
  children.
  The Container must at least support `begin()` and `end()` methods.
  Pointer and iterator stability is not required.

To check whether class `T` fulfills the concept one can use the
`TreeNodeTraits<T>` traits struct. It is defined only for classes fulfilling
the TreeNode concept, what makes it suitable for SFINAE tests. For
readability and consistency with Value and Parent checks (described below)
there is a member `TreeNodeTraits<T>::available` which is always true.

Optional members:

- `T* Parent()`:
  Returns pointer to a parent node or nullptr when the node is a tree root.
  Types with this member can be detected by checking whether
  `TreeNodeTraits<T>::Parent::available` is true.

- `ValueType& Value()`:
  Returns reference to a value stored in the node.
  Types with this member can be detected by checking whether
  `TreeNodeTraits<T>::Parent::available` is true.

- `subnodes_type` (typename):
  Container type which should be used for storing arrays of detached
  child nodes. It must be move-assignable to, and move-constructible from,
  a value returned by `Children()` method.
  When the type is absent, it is assumed to be the type returned by
  `Children()` method with reference and other modifiers removed.
  It is available through `TreeNodeTraits<T>::Children::container_type`.

## To be done at some point in the future

* Rename extracted functions to something less generic. E.g. `Path(node, container)` could be named `NodePath` (and merged with the Path printer that uses this name now), `Root(node)` could become `FindTreeRoot`.
* Replace (at least some) `enable_if` template parameters with `static_assert`s.
* Use extracted functions for operations on `MapTree` and remove substituted MapTree methods.
* Maybe: Replace `Value()` with `get()` and provide tuple-like interface to allow [strutured bindings](https://en.cppreference.com/w/cpp/language/structured_binding#Case_2:_binding_a_tuple-like_type)


common/util/trees.h
```diff
@@ -0,0 +1,856 @@
...
+  static inline constexpr bool available = true;
+};
+
+// TreeNodeTraits implementation details:
+
+// Alias to Node::subnodes_type if it exists.
+// Intended for use with detected_or_t.
+template <typename Node,  //
+          typename Type_ = typename Node::subnodes_type>
+using TreeNodeSubnodesType = Type_;
```
fangism, line 93: should this type alias be hidden an a private or internal namespace?

common/util/trees.h
```diff
@@ -0,0 +1,856 @@
...
+template <typename Node,  //
+          typename Parent_ = decltype(*std::declval<Node>().Parent()),
+          typename = std::void_t<TreeNodeChildrenTraits<Parent_>>>
+struct TreeNodeParentTraits : FeatureTraits {};
+
+// BirthRank implementation details:
+
+// BirthRank implementation supportin any container
+template <class T>
+inline static size_t BirthRank(const T& node, std::input_iterator_tag) {
```
fangism, line 133: I'd remove `inline` and `static` because these are already templates in a public header.

common/util/trees.h
```diff
@@ -0,0 +1,856 @@
...
+// Defined when `Node` contains `Parent()` method which dereferences to a type
+// fulfilling NodeTraits concept.
+template <typename Node,  //
+          typename Parent_ = decltype(*std::declval<Node>().Parent()),
+          typename = std::void_t<TreeNodeChildrenTraits<Parent_>>>
+struct TreeNodeParentTraits : FeatureTraits {};
+
+// BirthRank implementation details:
+
+// BirthRank implementation supportin any container
```
fangism, line 131: supporting

common/util/trees.h
fangism, line 133: I would add a comment that this overload is a linear-time operation.

common/util/trees.h
```diff
@@ -0,0 +1,856 @@
...
+  return 0;
+}
+
+// BirthRank optimized for random access containers
+template <class T>
+inline static size_t BirthRank(const T& node, std::random_access_iterator_tag) {
+  if (node.Parent() != nullptr) {
+    return std::distance(&(node.Parent()->Children().front()), &node);
+  }
+  return 0;
```
fangism, line 150: // a root node has no parent, but is considered rank 0

common/util/trees.h
```diff
@@ -0,0 +1,856 @@
...
+                              trees_internal::TreeNodeValueTraits, Node>;
+  using Children = Children_;
+};
+
+// Functions operating on tree nodes:
+
+// Returns true when `node` is a leaf, false otherwise.
+template <class T,  //
+          std::enable_if_t<TreeNodeTraits<T>::available>* = nullptr>
+bool is_leaf(T& node) {
```
fangism, line 175: `const T&`

common/util/trees.h
```diff
@@ -0,0 +1,856 @@
...
+  return node.Children().empty();
+}
+
+// Descend through children using indices specified by iterator range.
+// Iterator type must dereference to an integral value.
+// This works on any internal node, not just the root.
+template <class T, class Iterator,  //
```
fangism, line 182: // This implementation fatally exits if any indices are out-of-range.
// The const-ness of the returned reference matches the const-ness of the node argumnent.

[The const-ness comment is good to repeat where applicable.]

common/util/trees.h
```diff
@@ -0,0 +1,856 @@
...
+    class DstTree, class SrcTree, class SrcNodeToDstValueFunc,
+    class DstValue_ =
+        std::invoke_result_t<SrcNodeToDstValueFunc, const SrcTree&>,  //
+    std::enable_if_t<TreeNodeTraits<DstTree>::Value::available &&
+                     TreeNodeTraits<SrcTree>::Value::available &&
+                     std::is_constructible_v<DstTree, DstValue_>>* = nullptr>
+DstTree Transform(const SrcTree& src_node, const SrcNodeToDstValueFunc& f) {
+  // Using invoke() to allow passing SrcTree's method pointers as `f`
+  DstTree dst_node(std::invoke(f, src_node));
+  dst_node.Children().reserve(src_node.Children().size());
```
fangism, line 552: Worth noting that this implementation requires the children container to support `reserve`.  This was done to guarantee node address-stability.

(Probably worth commenting in other functions that call `reserve`)

common/util/trees.h
```diff
@@ -0,0 +1,856 @@
+// Copyright 2017-2022 The Verible Authors.
```
fangism, line 1: Maybe name this file `tree_operations`?

fangism (commented):
Only did quick partial review for now.

I generally like the direction this is going, separating implementation from interface.  Thanks for moving the tree operations to their own file with this.

common/util/trees.h
```diff
@@ -0,0 +1,856 @@
...
+//   Returns pointer to a parent node or nullptr when the node is a tree root.
+//   Types with this member can be detected by checking whether
+//   `TreeNodeTraits<T>::Parent::available` is true.
+//
+// - `ValueType& Value()`:
+//   Returns reference to a value stored in the node.
+//   Types with this member can be detected by checking whether
+//   `TreeNodeTraits<T>::Parent::available` is true.
+//
+// - `subnodes_type` (typename):
```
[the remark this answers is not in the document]
mglb: I've added that it has to be a container **or a range**, support `begin()` and `end()` methods, and lack of pointer/iterator stability guarantee.

I've also added explicit check for `back()` in functions that use it.
`reserve()` is now called conditionally (only if supported by a container).
Functions that use other methods (`push_back()`, `insert()`, etc) mention this fact in a comment.
`size()` method has been replaced with `std::size()`.

common/util/trees.h
mglb: It is already hidden in `namespace types_internal`.

common/util/trees.h
mglb: Fixed.

common/util/trees.h
mglb: Fixed.

common/util/trees.h
mglb: Mentioned. Also added following comment to public `BirthRank()`:
// This function is aware of children container type: it uses `std::distance()`
// to calculate node index in random access containers (which usually performs
// in O(1)), and O(n) loop for other containers.

common/util/trees.h
mglb: Added in the public function (cited one is an implementation detail).

common/util/trees.h
mglb: Fixed.

common/util/trees.h
mglb: Added.
Mentioned in all (I hope so) relevant functions.

common/util/trees.h
mglb: I've changed it (and other similar functions) to call `reserve()` only if supported by the container.

common/util/trees.h
mglb: Fixed

common/util/trees.h
mglb: Yes, this sound better. Renamed.

common/util/tree_operations.h
```diff
@@ -0,0 +1,926 @@
...
+
+#include "common/util/logging.h"
+#include "common/util/spacer.h"
+#include "common/util/type_traits.h"
+
+namespace verible {
+
+namespace tree_operations_internal {
+
+// TODO(mglb): move (Unavailable)FeatureTraits to a more generic header, or
```
fangism, line 77: So far there is
https://cs.opensource.google/verible/verible/+/master:common/util/type_traits.h

common/util/tree_operations.h
```diff
@@ -0,0 +1,926 @@
...
+template <typename Node,  //
+          typename ChildrenType_ = decltype(std::declval<Node>().Children()),
+          typename =
+              std::void_t<decltype(*std::declval<Node>().Children().begin()),
+                          decltype(std::declval<Node>().Children().end())>>
+struct TreeNodeChildrenTraits : FeatureTraits {
+  // Container type which should be used for storing arrays of detached
+  // child nodes. It must be move-assignable to, and move-constructible from,
+  // a value returned by `Children()` method.
+  using container_type = verible::remove_cvref_t<
```
fangism, line 110: Can you use https://www.cplusplus.com/reference/type_traits/remove_cv/ ?

common/util/tree_operations.h
```diff
@@ -0,0 +1,926 @@
...
+// Defined when `Node` contains `Parent()` method which dereferences to a type
+// fulfilling NodeTraits concept.
+template <typename Node,  //
+          typename Parent_ = decltype(*std::declval<Node>().Parent()),
+          typename = std::void_t<TreeNodeChildrenTraits<Parent_>>>
+struct TreeNodeParentTraits : FeatureTraits {};
+
+// BirthRank implementation details:
+
+// BirthRank implementation supporting any container.
```
fangism, line 135: // Requires that the tree node type provides 'Parent()'.

common/util/tree_operations.h
```diff
@@ -0,0 +1,926 @@
...
+}
+
+// std::function type representing a printer function in PrintTree.
+template <class Node>
+using PrintTreePrinterFunction = std::function<std::ostream&(
+    std::ostream&, typename TreeNodeTraits<Node>::Value::const_reference)>;
+
+// Pretty-print in tree-form.  Value() is enclosed in parens, and the whole
+// node is enclosed in braces.
+// This variant supports a custom value 'printer'.
```
fangism, line 255: // Maybe worth noting where many of these functions do not require 'Parent()'.

common/util/tree_operations.h
```diff
@@ -0,0 +1,926 @@
...
+  std::set<T*> ancestors_a, ancestors_b;
+  // In alternation, insert a/b into its respective set of ancestors,
+  // and check for membership in the other ancestor set.
+  // Return as soon as one is found in the other's set of ancestors.
+  while (a != nullptr || b != nullptr) {
+    if (a != nullptr) {
+      if (ancestors_b.find(a) != ancestors_b.end()) {
+        return a;
+      }
+      ancestors_a.insert(a);
```
fangism, line 427: doesn't have to be in this PR, but we should avoid the double lookup of find-and-insert by using the return value of `insert`.

common/util/tree_operations.h
```diff
@@ -0,0 +1,926 @@
...
+//   Types with this member can be detected by checking whether
+//   `TreeNodeTraits<T>::Parent::available` is true.
+//
+// - `ValueType& Value()`:
+//   Returns reference to a value stored in the node.
+//   Types with this member can be detected by checking whether
+//   `TreeNodeTraits<T>::Parent::available` is true.
+//
+// - `subnodes_type` (typename):
+//   Container type which should be used for storing arrays of detached
```
fangism, line 49: // Contiguous storage (array) is required (?) or only bidirectional iterability?

common/util/tree_operations.h
```diff
@@ -0,0 +1,926 @@
...
+  node = std::move(only);
+  return true;
+}
+
+// Combines the Nth and (N+1) sibling using a custom function 'joiner' on the
+// nodes' values, and the Nth sibling will adopt N+1's children.
+// The 'joiner' function does: *left = f(*left, *right);
+// The (N+1) sibling will be erased in the process, and every sibling
+// thereafter will be shifted back one position (same inefficiency as shifting
+// vector contents).  This invalidates all iterators after position N,
```
fangism, line 658: This applies for vector-like containers, with contiguous storage.

common/util/type_traits.h
```diff
@@ -45,6 +45,39 @@ using match_const =
...
+// Alias to `Op<Args...>` if that type is valid; otherwise alias to `Default`.
+template <class Default, template <class...> class Op, class... Args>
+using detected_or_t =
+    typename type_traits_internal::detected_impl<Default, void, Op,
+                                                 Args...>::type;
+
+// Alias to type T with removed reference, const, and volatile qualifiers.
+// Backport from C++20.
+template <class T>
+using remove_cvref_t = std::remove_cv_t<std::remove_reference_t<T>>;
```
fangism, line 79: Is https://www.cplusplus.com/reference/type_traits/remove_cv/ enough? (C++11)

fangism (approved):

hzeller:
@mglb ready to merge after addressing the smallish comments by David.

common/util/tree_operations.h
```diff
@@ -0,0 +1,926 @@
...
+template <class Container>
+auto /* void */ TryReserve(Container& container,
+                           typename Container::size_type new_cap)
+    -> std::void_t<decltype(container.reserve(new_cap))> {
+  container.reserve(new_cap);
+}
+
+// No-op candidate used when Container doesn't provide `reserve()` method.
+template <class Container>
+void TryReserve(Container&, ...) {}
```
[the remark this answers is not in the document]
mglb: Changed to `ReserveIfSupported`.

common/util/tree_operations.h
```diff
@@ -0,0 +1,926 @@
...
+bool is_leaf(const T& node) {
+  return node.Children().empty();
+}
+
+// Descend through children using indices specified by iterator range.
+// Iterator type must dereference to an integral value.
+// This works on any internal node, not just the root.
+// This implementation fatally exits if any indices are out-of-range.
+// The const-ness of the returned reference matches the const-ness of the node
+// argumnent.
```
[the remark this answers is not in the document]
mglb: Fixed

common/util/tree_operations.h
```diff
@@ -0,0 +1,926 @@
...
+// there is a member `TreeNodeTraits<T>::available` which is always true.
+//
+// Optional members:
+//
+// - `T* Parent()`:
+//   Returns pointer to a parent node or nullptr when the node is a tree root.
+//   Types with this member can be detected by checking whether
+//   `TreeNodeTraits<T>::Parent::available` is true.
+//
+// - `ValueType& Value()`:
```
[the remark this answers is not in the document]
mglb: Most operations do not need `Value()` method (and if they do, there's `TreeNodeTraits<T>::Value::available` to check for that), so I think it shouldn't be mandatory.
Use cases:
- Basic node types could provide data members directly in the node.
- The node class could be a polymorphic base with data members available only in derived classes.

common/util/tree_operations.h
```diff
@@ -0,0 +1,926 @@
...
+//   Returns reference to a value stored in the node.
+//   Types with this member can be detected by checking whether
+//   `TreeNodeTraits<T>::Parent::available` is true.
+//
+// - `subnodes_type` (typename):
```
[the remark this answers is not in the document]
mglb: It shouldn't be required - `TreeNodeTraits::Children::container_type` just uses type returned by `Children()` method when `subnodes_type` alias is not available (see rest of the comment below).

common/util/tree_operations.h
mglb: This code also needs to strip reference, so it would turn into a bit longer `std::remove_cv_t<std::remove_reference_t<...>>`.

common/util/tree_operations.h
mglb: I've added the comment to the public function (the cited one is internal). Should it be here too?

common/util/tree_operations.h
mglb: I did it the other way around - I've added comments to functions that DO require `Parent()` or `Value()`.

common/util/tree_operations.h
mglb: The find and insert are called on different sets (ancestors_`b`.find(a) and ancestors_`a`.insert(a)).

common/util/tree_operations.h
mglb: "Array" is a misleading word here - replaced with "collections". I've also added more explanation. New comment:
```
// - `subnodes_type` (typename):
//   Container type which should be used for storing collections of detached
//   child nodes.
//   It must be move-assignable to, and move-constructible from a value returned
//   by `Children()` method. It must provide at least the same STL container's
//   interface and iterator/pointer stability guarantees as a return type of
//   `Children()` method.
//   The use case where this alias is useful is when `Children()` returns a
//   container wrapped by some internal wrapper type that is either not
//   constructible outside of the node class or has some extra logic that is
//   not useful when used outside of a node. The `subnodes_type` in such case
//   should be an alias to the container's real type.
//   When the type is absent, it is assumed to be the type returned by
//   `Children()` method with reference and other modifiers removed.
//    It is available through `TreeNodeTraits<T>::Children::container_type`.
```

common/util/tree_operations.h
mglb: New comment:
```
// The (N+1) sibling will be erased in the process, and every sibling
// thereafter will be shifted back one position. Depending on a children
// container type this can invalidate all interators after position N and
// interators to Nth node children. This applies e.g. for vector-like
// containers, with contiguous storage.
```

common/util/type_traits.h
mglb: It is not - `remove_cv` doesn't strip reference.
I've decided to backport this from C++20 because stripping of reference and const is a bit tricky in C++ and can be easily made in wrong order. With a type `T = const U&`:
- `std::remove_cv_t<std::remove_reference_t<T>>` is `U`: correct.
- `std::remove_reference_t<std::remove_cv_t<T>>` is `const U`: incorrect. `remove_cv_t` operates on a reference type itself (doing nothing), not on `const U`.

common/util/tree_operations.h
mglb: Moved it there.

PR #1266, opened 2022-03-16
LSP Range: allow empty range overlap predicate.

hzeller:
codeAction requests often just include the cursor position in the
request which is an empty range. If the diagnostic in question is
a one-character range, make sure we recognize that as overlap.

### DIFF
--- a/common/lsp/lsp-protocol-operators.h
+++ b/common/lsp/lsp-protocol-operators.h
@@ -31,8 +31,16 @@ inline constexpr bool operator<(const Position &a, const Position &b) {
 inline constexpr bool operator>=(const Position &a, const Position &b) {
   return !(a < b);
 }
+inline constexpr bool operator==(const Position &a, const Position &b) {
+  return a.line == b.line && a.character == b.character;
+}
+
+// Ranges overlap if some part of one is inside the other range.
+// Also empty ranges are considered overlapping if their start point is within
+// the other range.
+// rangerOverlap() is commutative.
 inline constexpr bool rangeOverlap(const Range &a, const Range &b) {
-  return !(a.start >= b.end || b.start >= a.end);
+  return !(a.start >= b.end || b.start >= a.end) || (a.start == b.start);
 }
 }  // namespace lsp
 }  // namespace verible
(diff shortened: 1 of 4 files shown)


common/lsp/lsp-protocol-operators.h
```diff
@@ -31,8 +31,12 @@ inline constexpr bool operator<(const Position &a, const Position &b) {
 inline constexpr bool operator>=(const Position &a, const Position &b) {
   return !(a < b);
 }
+inline constexpr bool operator==(const Position &a, const Position &b) {
+  return a.line == b.line && a.character == b.character;
+}
+
 inline constexpr bool rangeOverlap(const Range &a, const Range &b) {
-  return !(a.start >= b.end || b.start >= a.end);
+  return !(a.start >= b.end || b.start >= a.end) || (a.start == b.start);
```
fangism, line 39: How about: `!(a.end <= b.start || b.end <= a.start)`
Overlap check should only require two comparisons at most.

common/lsp/lsp-protocol-operators.h
```diff
@@ -31,8 +31,12 @@ inline constexpr bool operator<(const Position &a, const Position &b) {
 inline constexpr bool operator>=(const Position &a, const Position &b) {
   return !(a < b);
 }
+inline constexpr bool operator==(const Position &a, const Position &b) {
+  return a.line == b.line && a.character == b.character;
+}
+
 inline constexpr bool rangeOverlap(const Range &a, const Range &b) {
```
fangism, line 38: comment whether abutment counts as overlap

fangism (approved):

common/lsp/lsp-protocol-operators.h
hzeller: That expression would be equivalent to the original one (`a.end <= b.start` equiv to `b.start >= a.end` as well as `b.end <= a.start` equiv to `a.start >= b.end`) and will not catch the empty range.
It probably can be formulated in two expressions, but haven't had coffee yet. Do you see a way ?

common/lsp/lsp-protocol-operators.h
hzeller: Updated comment (abutment does not count as overlap).

common/lsp/lsp-protocol-operators.h
fangism: If there is a "special case", then you probably can't avoid the extra comparison.

fangism (approved):

PR #1267, opened 2022-03-16
Make `IndexingNodeData` move-only

mglb:


mglb:
Abandoned due to invalidity. Details: https://github.com/chipsalliance/verible/issues/1275

PR #1270, opened 2022-03-17
Parsing/Preprocessing: add the concept of 'warning'.

hzeller:
So far, all messages in the lex/preprocessing/parse phase have been
of 'error severity'. However, there are issues whose severity can
be lower e.g. warning about re-definition of a macro.

In this change
  * Make RejectedToken contain a severity (name of that object
    is still the same, might need some re-name later)
  * Change printing of rejected tokens to include the severity
    and don't talk about a 'rejected token', but just the token
    a hand; that does sound more user-friendly (the 'rejected'
    sounds more like an artifact of parsing).
    CAVE: this might trip up scripts of users that grep for
    messages.
  * Surface preprocessing re-definition as warning message.
  * Do _not_ include the preprocessing warning yet in the output
    of the parser; merely LOG(INFO) it at this point.
    (Reason: might be too noisy as the preprocessor has limited
    understanding of `ifdef branches).
    At least, this provides a properly formatted filename + line-number
    info in the log.

Added a few TODOs for further cleanups.

### DIFF
--- a/common/analysis/file_analyzer.cc
+++ b/common/analysis/file_analyzer.cc
@@ -35,7 +35,7 @@
 namespace verible {
 
 // Translates phase enum into string for diagnostic messages.
-static const char* AnalysisPhaseName(const AnalysisPhase& phase) {
+const char* AnalysisPhaseName(const AnalysisPhase& phase) {
   switch (phase) {
     case AnalysisPhase::kLexPhase:
       return "lexical";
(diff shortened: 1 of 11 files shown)


hzeller:
Example of the LOG(INFO) output of the preprocessor issues it sees (redifinitions due to incomplete `ifdef` handling)

```
I20220317 13:46:25.170456 770932 verilog_analyzer.cc:258] /tmp/picorv32.v:30:31: preprocessing warning at token "" : Re-defining macro 'debug'
I20220317 13:46:25.170563 770932 verilog_analyzer.cc:258] /tmp/picorv32.v:38:25: preprocessing warning at token "(* keep *)" : Re-defining macro 'FORMAL_KEEP'
I20220317 13:46:25.170578 770932 verilog_analyzer.cc:258] /tmp/picorv32.v:40:24: preprocessing warning at token "" : Re-defining macro 'FORMAL_KEEP'
I20220317 13:46:25.170589 770932 verilog_analyzer.cc:258] /tmp/picorv32.v:42:31: preprocessing warning at token "empty_statement" : Re-defining macro 'assert'
```

fangism (approved):

PR #1271, opened 2022-03-17
Format syntax errors always with column range.

hzeller:
If the token mentioned in a syntax error covers more than one character,
print syntax error with the full column range.

### DIFF
--- a/common/analysis/file_analyzer_test.cc
+++ b/common/analysis/file_analyzer_test.cc
@@ -60,14 +60,14 @@ TEST(FileAnalyzerTest, TokenErrorMessageSameLine) {
   const TokenInfo error_token(1, analyzer.Data().Contents().substr(17, 5));
   {
     const auto message = analyzer.TokenErrorMessage(error_token);
-    EXPECT_EQ(message, "token: \"w0rld\" at 2:5-9");
+    EXPECT_EQ(message, "token: \"w0rld\" at 2:5:2:9:");
   }
   {
     constexpr bool with_diagnostic_context = false;
     const auto message = analyzer.LinterTokenErrorMessage(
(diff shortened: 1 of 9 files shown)


common/analysis/file_analyzer_test.cc
```diff
@@ -60,14 +60,14 @@ TEST(FileAnalyzerTest, TokenErrorMessageSameLine) {
   const TokenInfo error_token(1, analyzer.Data().Contents().substr(17, 5));
   {
     const auto message = analyzer.TokenErrorMessage(error_token);
-    EXPECT_EQ(message, "token: \"w0rld\" at 2:5-9");
+    EXPECT_EQ(message, "token: \"w0rld\" at 2:5:2:9:");
```
fangism, line 63: `2:5:2:9` reads a little weird.  Is there a different middle delimiter we could use?  `2:5..2:9`? or `2:5--2:9`?

hzeller: So it seems like `2:5-9:` is something that e.g. emacs parses naturally. Changed to that format.
(it does not work with multiple lines, so for covering them, I kept `42:5:45:2:` - needs to be figured out in the future what would be best)

fangism: I might be nice if diagnostic rendering logic were separated from the analyzer, or accepted a range-formatter function.  
It might come in handy as this gets integrated more places.  

I won't hold up this change.

fangism (approved):

common/analysis/file_analyzer_test.cc
hzeller: Yes, if the logic gets more complicated or we need to change stuff for other IDEs, then this might be a good idea.

PR #1276, opened 2022-03-18
Don't limit the diagnostic messages for textDocument/diagnostics

hzeller:
While we limit the number of diagnostic messages we send in the
unsolicted notification to the editor, don't do that if explicitly
requested in the textDocument/diagnostics RPC request.

### DIFF
--- a/verilog/tools/ls/verible-lsp-adapter.cc
+++ b/verilog/tools/ls/verible-lsp-adapter.cc
@@ -48,24 +48,35 @@ static verible::lsp::Diagnostic ViolationToDiagnostic(
 }
 
 std::vector<verible::lsp::Diagnostic> CreateDiagnostics(
-    const BufferTracker &tracker) {
+    const BufferTracker &tracker, int message_limit) {
   // Diagnostics should come from the latest state, including all the
   // syntax errors.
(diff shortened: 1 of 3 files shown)


verilog/tools/ls/verible-lsp-adapter.cc
```diff
@@ -48,22 +48,32 @@ static verible::lsp::Diagnostic ViolationToDiagnostic(
...
+  //
+  // We might consider emitting them around the last known
+  // edit point in the document as this is what the user sees (if we get
+  // individual edits, not full files pushed).
+  //
+  // TODO(hzeller): to limit repetition, maybe limit the number of messages
+  // coming from the _same_ source if we have a "message_limit". So for
+  // instance, don't complain on every single line not to use tabs as
+  // indentation.
+  if (message_limit > 0 && remaining > message_limit) {
```
fangism, line 73: Below you pass -1 to signal no-limit, so should this be `>= 0`?

fangism (approved):

verilog/tools/ls/verible-lsp-adapter.cc
hzeller: Done.

PR #1277, opened 2022-03-18
Start documenting more behavior in preprocessor unit test.

hzeller:
* Simplify testing: move repetetive parse testing into
        separate macro.
   * Also, simplify the warning for macro definition - directly
        fill in warning.
   * Document that `define stay in the filtered token stream.

### DIFF
--- a/verilog/preprocessor/verilog_preprocess_test.cc
+++ b/verilog/preprocessor/verilog_preprocess_test.cc
@@ -14,6 +14,7 @@
 
 #include "verilog/preprocessor/verilog_preprocess.h"
 
+#include <algorithm>
 #include <map>
 #include <vector>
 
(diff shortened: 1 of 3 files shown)


verilog/preprocessor/verilog_preprocess_test.cc
```diff
@@ -99,6 +101,13 @@ TEST(VerilogPreprocessTest, InvalidPreprocessorInputs) {
   }
 }
 
+#define EXPECT_PARSE_OK()                                                \
+  do {                                                                   \
```
fangism, line 105: maybe try `SCOPED_TRACE` from googletest, which I believe shows test stacktraces useful for test helper functions.

fangism (approved):

verilog/preprocessor/verilog_preprocess_test.cc
hzeller: It looks like it will require one more line to add to each place I call this, so the advantage of only having one thing to call is diminishing..

PR #1284, opened 2022-03-23
Implement `undef

hzeller:

### DIFF
--- a/verilog/preprocessor/verilog_preprocess_test.cc
+++ b/verilog/preprocessor/verilog_preprocess_test.cc
@@ -234,6 +234,25 @@ TEST(VerilogPreprocessTest, TwoMacroDefinitions) {
   }
 }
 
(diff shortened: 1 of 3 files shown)


verilog/preprocessor/verilog_preprocess_test.cc
```diff
@@ -234,6 +234,16 @@ TEST(VerilogPreprocessTest, TwoMacroDefinitions) {
   }
 }
 
+TEST(VerilogPreprocessTest, UndefMacro) {
```
fangism, line 237: also suggest a test case that undef's a nonexistent macro.

fangism (approved):

verilog/preprocessor/verilog_preprocess_test.cc
hzeller: Done.

PR #1289, opened 2022-03-24
Always use branching preprocessing with the syntax checker.

hzeller:
I was considering making this configurable via flags, but this is
a strong case of 'this should just do the right thing'.

We _might_ need to start allowing `+define+` command line options,
but that is on my short-list of things to implement.

### DIFF
--- a/verilog/tools/syntax/verilog_syntax.cc
+++ b/verilog/tools/syntax/verilog_syntax.cc
@@ -298,8 +298,11 @@ int main(int argc, char** argv) {
       continue;
     }
 
-    // TODO(hzeller): make preprocessing configurable with flags.
-    const verilog::VerilogPreprocess::Config preprocess_config;
+    // TODO(hzeller): is there ever a situation in which we do not want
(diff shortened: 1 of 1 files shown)


verilog/tools/syntax/verilog_syntax.cc
```diff
@@ -298,8 +298,11 @@ int main(int argc, char** argv) {
       continue;
     }
 
-    // TODO(hzeller): make preprocessing configurable with flags.
-    const verilog::VerilogPreprocess::Config preprocess_config;
+    // TODO(hzeller): is there ever a situation in which we do not want
```
fangism, line 301: If you are fed an input which is known to be preprocessed, you might want to disable this (flag).
C compilers have an option like this to skip preprocessing.

fangism (approved):

verilog/tools/syntax/verilog_syntax.cc
hzeller: But even then, there would be no difference.

PR #1290, opened 2022-03-24
Enable Preprocess-branch filtering for syntax-table applications.

hzeller:
Tools that build a syntax table such as the project tool or the
kythe indexer are only interested in the relevant syntax tree.
Not filtering will increase the chance of syntax errors in case we
have branches that are not complete verilog statements.


fangism (approved):

hzeller:
@ivan444 @MinaToma 
This will turn on proper branch preprocessing, which will fix probably most of the issues we see now (running into syntax errors with files containing preprocessing branches).

It looks like there is a Kythe test [`verilog/tools/kythe/testdata/two_definitions.sv`](https://github.com/chipsalliance/verible/blob/master/verilog/tools/kythe/testdata/two_definitions.sv) that tests for the old behavior (the new behavior only should see one of the branches). I tried to ['fix'](https://github.com/chipsalliance/verible/pull/1290/files#diff-c24dd65ced8e9ba8c3d06a9686e61d97f6fb0599fee56a710c42aef057a50e68) it, but as you can see that didn't work. How do I fix the test-case ?

hzeller:
So if Kythe can actually produce meaningful multiple facts for the same symbol in different branches (and consumers of such data can make use of it as in 'defined in one of these two places'), then it would probably make sense to do parsing with fallback as in the linter (#1293) and language server (#1294).

ivan444:
`two_definitions.sv` test is now broken or working fine?
If it's broken, which kythe facts does it produce? IIUC, before we saw two definitions (breaks semantics), but now only one branch should be executed, so we should se a single definition. Maybe you picked the wrong one? (which branch is kept?)

hzeller:
currently, it is broken, but maybe I picked the wrong branch ? But it would require that `USE_PARAMETER` would be set by the kythe testing environment, which I doubt. Anyway, changed the `NOTDEFINED` to `DEFINED` and let the CI do its thing.

I am slighlty tending towards (in this order):

  * Do the fallback evaluation
  * remove this test

PR #1293, opened 2022-03-25
Linter: if needed, make two parse attempts with and w/o pp-branches.

hzeller:
Parsing without filtering the preprocessing branches will work
most of the time and provide more 'material' for the linter to look
over. But if parsing fails in that case (because it can result
in invalid syntactic constructs), retry with preprocessing
branches enabled.


fangism (approved):
Recommend one integration test, so that we know if this feature has regressed.

hzeller:
> Recommend one integration test, so that we know if this feature has regressed.

Done. Though not integration test, but using `VerilogAnalyzer::AnalyzeAutomaticPreprocessFallback()`, which is tested separately.

PR #1295, opened 2022-03-25
Fixing syntax error on wire delay value 

karimtera:
This PR introduces a fix for declaring a net with delay, which is allowed by LRM (A.2.1.3) as:
`net_declaration ::= net_type [...] data_type_or_implicit [delay3] list_of_net_decl_assignments ;`

Added test: 
```verilog
module m; 
wire #(100) x = y,z=k,foo; 
endmodule
```
Proposed modification to 'verilog.y' grammer:

- 'net_declaration' rule now accepts:
` net_type delay3 net_variable_or_decl_assigns ';'`

Explanation:
'net_declaration' rule accepts  ` net_type data_type_or_implicit net_variable_or_decl_assigns ';'`
Which pares this line correctly
```verilog
wire  [12:0] #(100) x = y; 
```
Because 'data_type_or_implicit' accepts `decl_dimensions delay3_or_drive_opt`
Notice 'decl_dimensions' is not optional here, so removing it causes a problem.
An easy workaround would be to add a new rule for 'net_declaration' that works for this line only.

Issues fixed: [271](https://github.com/chipsalliance/verible/issues/271)

### DIFF
--- a/verilog/parser/verilog.y
+++ b/verilog/parser/verilog.y
@@ -5453,7 +5453,7 @@ net_declaration
                           MakePackedDimensionsNode($3),
                           $4, $5, $6); }
   | net_type delay3 net_variable_or_decl_assigns ';'
-  { $$ = MakeTaggedNode(N::kNetDeclaration, $1, $2, $3, $4); }
+  { $$ = MakeTaggedNode(N::kNetDeclaration, $1, nullptr, nullptr, $2, $3, $4); }
   /* TODO(fangism): net_type_identifer [ delay_control ] list_of_net_decl_assignments */
   /* TODO(fangism): TK_interconnect ... */
   ;
--- a/verilog/parser/verilog.y
+++ b/verilog/parser/verilog.y
@@ -5452,6 +5452,8 @@ net_declaration
     { $$ = MakeTaggedNode(N::kNetDeclaration, $1, $2,
                           MakePackedDimensionsNode($3),
                           $4, $5, $6); }
+  | net_type delay3 net_variable_or_decl_assigns ';'
+  { $$ = MakeTaggedNode(N::kNetDeclaration, $1, $2, $3, $4); }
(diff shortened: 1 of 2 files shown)


verilog/parser/verilog.y
```diff
@@ -5452,6 +5452,8 @@ net_declaration
     { $$ = MakeTaggedNode(N::kNetDeclaration, $1, $2,
                           MakePackedDimensionsNode($3),
                           $4, $5, $6); }
+  | net_type delay3 net_variable_or_decl_assigns ';'
+  { $$ = MakeTaggedNode(N::kNetDeclaration, $1, $2, $3, $4); }
```
fangism, line 5456: Let's try to make the fields of `kNetDeclarations` consistent with each other as much as possible.  For example, the one right above this has 6 arguments.  You may insert `nullptr`s to make similar fields line up.  

The consistency will make it easier to access common fields with functions in `CST/*`.

This consistency isn't strictly enforced today because we lack a proper abstract syntax tree, but some day we might get one.

verilog/parser/verilog_parser_unittest.cc
```diff
@@ -1786,6 +1786,7 @@ static const ParserTestCaseArray kModuleTests = {
     "swimming bool;  // bool is an Icarus Verilog extension\n"
     "endmodule",
     // net declarations
+    "module m; wire #(100) x = y,z=k,foo; endmodule\n",
```
fangism, line 1789: Let's also include a minimalist case that exercise the new syntax rule, e.g.

```
module m; wire #(100) x = y; endmodule;
```

fangism (changes requested):
Thank you for this fix!

fangism:
@hzeller  Internally, this happens to be b/161138189.

fangism (approved):

PR #1299, opened 2022-03-28
Document for each project the expected number of tool failures.

hzeller:
Tools fail on some of the files. So far, we only were concerned about
crashes in tool in the smoke tests, but recently improvements have
been sen in also reducing the number of files we can't process.

This change documents how many of these failures are to be expected
to make sure we don't add regressions (if failures increase, the
smoke test will fail), but also detect when we see less than expected
failures.

### DIFF
--- a/.github/bin/smoke-test.sh
+++ b/.github/bin/smoke-test.sh
@@ -132,16 +132,16 @@ ExpectedFailCount[syntax:ibex]=13
 ExpectedFailCount[lint:ibex]=13
 ExpectedFailCount[project:ibex]=175
 
-ExpectedFailCount[syntax:opentitan]=29
-ExpectedFailCount[lint:opentitan]=29
+ExpectedFailCount[syntax:opentitan]=27
+ExpectedFailCount[lint:opentitan]=27
 ExpectedFailCount[formatter:opentitan]=1
-ExpectedFailCount[project:opentitan]=648
+ExpectedFailCount[project:opentitan]=646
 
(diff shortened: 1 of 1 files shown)


.github/bin/smoke-test.sh
```diff
@@ -119,6 +119,88 @@ KnownIssue[formatter:$BASE_TEST_DIR/ivtest/ivltests/packed_dims_invalid_class.v]
...
+  echo "  -> Non zero exit code ${OBSERVED_NONZERO_COUNT}/${NUM_FILES}"
+
+  expected_count_key="${TOOL_SHORT_NAME}:${PROJECT_NAME}"
+  if [[ -v ExpectedFailCount[${expected_count_key}] ]]; then
+    expected_count=${ExpectedFailCount[${expected_count_key}]}
+    if [ ${OBSERVED_NONZERO_COUNT} -gt ${expected_count} ] ; then
+      echo "::error:: 😱 More failures than expected ${expected_count}"
+      return 1
+    elif [ ${OBSERVED_NONZERO_COUNT} -lt ${expected_count} ] ; then
+      echo "::notice:: 🎉 Yay, reduced non-zero exit count than expected ${expected_count}"
```
fangism, line 192: Is there an action to recommend here, like "update the improved expected count?"

fangism (approved):

.github/bin/smoke-test.sh
hzeller: Yes - printed out now what to change in the scrip twhen this happens.

Issue #1308, opened 2022-04-01
Smoke test update expected failures


karimtera:
Hello, 

I am facing the same issue as [#1306](https://github.com/chipsalliance/verible/issues/1306).

Needed fixes:
- `project:ibex`: 173 instead of 172
- `syntax:opentitan`: 28 instead of 27
- `lint:opentitan`: 28 instead of 27
- `project:opentitan`: 650 instead of 648

hzeller:
Fixed with #1311 that also adds a little bit of padding to the expected failure numbers.

PR #1309, opened 2022-04-01
smoke test was expecting less failures

karimtera:
Description:
Updates smoke test expected failures.

Changes:
- `project:ibex`: 173 instead of 172
- `syntax:opentitan`: 28 instead of 27
- `lint:opentitan`: 28 instead of 27
- `project:opentitan`: 650 instead of 648

Fixes: [#1308](https://github.com/chipsalliance/verible/issues/1308)


hzeller:
I've now addressed this also with a little bit of 'allowance' around the particular problem #1311

PR #1312, opened 2022-04-04
Get bash from /usr/bin/env instead of hard-coded path /bin/bash

hzeller:
In Posix-compliant systems, bash is not guaranteed to be installed
in /bin/bash; but if it is installed, it can be queried with
/usr/bin/env


mithro (approved):
LGTM.

PR #1315, opened 2022-04-06
SymbolTable: provide more detailed information about re-defined symbols.

hzeller:
Provide location of new and previous definition of a symbol that
is found out to be re-defined.

### DIFF
--- a/common/text/text_structure.cc
+++ b/common/text/text_structure.cc
@@ -132,8 +132,21 @@ LineColumnRange TextStructureView::GetRangeForToken(
     const LineColumn eofPos = GetLineColAtOffset(Contents().length());
     return {eofPos, eofPos};
   }
+  // TODO(hzeller): This should simply be GetRangeForText(token.text()),
+  // but the more thorough error checking in GetRangeForText()
+  // exposes a token overrun in verilog_analyzer_test.cc
+  // Defer to fix in separate change.
   return {GetLineColAtOffset(token.left(Contents())),
           GetLineColAtOffset(token.right(Contents()))};
+}
+
+LineColumnRange TextStructureView::GetRangeForText(
+    absl::string_view text) const {
+  const int from = std::distance(Contents().begin(), text.begin());
+  const int to = std::distance(Contents().begin(), text.end());
+  CHECK_GE(from, 0) << '"' << text << '"';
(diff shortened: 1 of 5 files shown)


common/text/text_structure.cc
```diff
@@ -132,10 +132,23 @@ LineColumnRange TextStructureView::GetRangeForToken(
...
   return {GetLineColAtOffset(token.left(Contents())),
           GetLineColAtOffset(token.right(Contents()))};
 }
 
+LineColumnRange TextStructureView::GetRangeForText(
+    absl::string_view text) const {
+  const int from = std::distance(Contents().begin(), text.begin());
+  const int to = std::distance(Contents().begin(), text.end());
+  CHECK_GE(from, 0) << '"' << text << '"';
+  CHECK_LT(to, Contents().length()) << '"' << text << '"';
```
mglb, line 148: `CHECK_LE`? There's a possibility that `text.end() == Contents().end()`.

common/text/text_structure_test.cc
```diff
@@ -216,6 +216,25 @@ TEST_F(TokenRangeTest, GetRangeOfTokenVerifyAllRangesExclusive) {
...
+  const TokenInfo& token = data_.FindTokenAt({0, 7});
+  EXPECT_EQ(token.text(), "world");
+  const LineColumnRange token_range = data_.GetRangeForToken(token);
+  EXPECT_EQ(token_range.start.line, 0);
+  EXPECT_EQ(token_range.start.column, 7);
+
+  const LineColumnRange text_range = data_.GetRangeForText(token.text());
+  EXPECT_EQ(text_range.start.line, 0);
+  EXPECT_EQ(text_range.start.column, 7);
+}
```
mglb, line 236: Let's add test case that fails due to CHECK_LE problem mentioned in the other comment:
```
data_.GetRangeForText(data_.Contents());
```

mglb (changes requested):

common/text/text_structure.cc
hzeller: Oops, that was a leftover from some experiment. Fixed and changed to `CHECK_LE()`

common/text/text_structure_test.cc
hzeller: Done.

mglb (approved):

PR #1318, opened 2022-04-07
[WIP] implementing ExpandMacro method in the preprocessor

karimtera:
**Description:**
This PR introduces a way to handle macro calls, instead of substituting the macro identifier with the definition text (`PP_define_body` token), the preprocessor re-lexes the definition text, and feed the new lexed tokens to the next stage (parser by default).

**Example:**
```verilog
`define DELAY #100
module m;
     wire `DELAY x=y;
endmodule
````
Old behavior:
`DELAY` is substituted with a ``PP_define_body`` token, then parser matches it to a  `GenericIdentifier`.

New behavior:
`DELAY` is substituted with 2 token describing `#` and `100`

**Changes:**

- `MacroDefinition`:
  - Added a field to save the lexed macro definitions as a vector of tokens
  - Added two methods to set/ get this field.
- `VerilogPreprocess`:
  - Added `HandleMacroIdentifier` method which searches for the macro definition, and if found, it will add the lexed body to the preprocessor stream view.
  - Added `ExpandMacro` method which lexes the macro body, and then sets the `MacroDefinition`'s `lexed_macro_body`.
  - `HandleTokenIterator`:
    - if a token is `MacroIdentifier` or `MacroIdItem` then call `HandleMacroIdentifier`.
  - `HandleDefine`:
    - if the macro is callable (contains parameters) then `ExpandMacro` is not ready for it.

**To-Do list:**

- [x] Apply clang-format.
- [x] Add new testcases (I am currently testing by building, then using the lint tool)
- [x] These changes introduced many errors in the test suite, still investigating these errors.
- [x] In case there is an \`undef to an expanded macro, it is deleted too from the preprocessor stream view, because it is just pointing to its address as saved in `MacroDefinition`.
- [x] For now, only use this new approach when a macro is expanded directly into the SV main code, not inside another macro, or preprocessor directive (I think these are most of the cases that fail in the test suite).
- [x] I think it is not totally correct to expand the macro once it's defined, but it should be expanded (and lexed) on calls, e.g.
```verilog
`define FIRST 1
`define SECOND `FIRST
`define FIRST 2
````
Calling `SECOND` should be expanded to `2`, not `1`.

**Future Work:**
- [x] Handle the callable macros.
- [x] Handle the nested macros.

**Fixes:** 
Issue [#1208](https://github.com/chipsalliance/verible/issues/1208)

### DIFF
--- a/common/text/macro_definition.h
+++ b/common/text/macro_definition.h
@@ -105,6 +105,8 @@ class MacroDefinition {
   // Create a text substitution map to be used for macro expansion.
   absl::Status PopulateSubstitutionMap(const std::vector<TokenInfo>&,
                                        substitution_map_type*) const;
+  absl::Status PopulateSubstitutionMap(const std::vector<DefaultTokenInfo>&,
+                                       substitution_map_type*) const;
 
   // Replace formal parameter references with actuals.
   static const TokenInfo& SubstituteText(const substitution_map_type&,
--- a/common/text/macro_definition.h
+++ b/common/text/macro_definition.h
@@ -86,9 +86,13 @@ class MacroDefinition {
   void SetDefinitionText(const TokenInfo& t) { definition_text_ = t; }
 
   // Lexes the definition_text_ and store it in lexed_macro_body
-  void SetLexedDefinitionText(const std::vector<TokenInfo> l) {lexed_macro_body = l ;} 
+  void SetLexedDefinitionTokens(const std::vector<TokenInfo>& l) {
+    lexed_definition_tokens_ = l;
+  }
 
-  const std::vector<TokenInfo>& GetLexedDefinitionText() const { return lexed_macro_body; }
+  const std::vector<TokenInfo>& GetLexedDefinitionText() const {
+    return lexed_definition_tokens_;
+  }
 
   // Macro definitions with empty () should call this.
   void SetCallable() { is_callable_ = true; }
@@ -132,8 +136,7 @@ class MacroDefinition {
 
   // un-tokenized text
   DefaultTokenInfo definition_text_;
-  std::vector<TokenInfo> lexed_macro_body;
-
+  std::vector<TokenInfo> lexed_definition_tokens_;
 };
 
 }  // namespace verible
--- a/common/text/macro_definition.h
+++ b/common/text/macro_definition.h
@@ -85,6 +85,11 @@ class MacroDefinition {
 
   void SetDefinitionText(const TokenInfo& t) { definition_text_ = t; }
 
+  // Lexes the definition_text_ and store it in lexed_macro_body
+  void SetLexedDefinitionText(const std::vector<TokenInfo> l) {lexed_macro_body = l ;} 
+
+  const std::vector<TokenInfo>& GetLexedDefinitionText() const { return lexed_macro_body; }
+
   // Macro definitions with empty () should call this.
   void SetCallable() { is_callable_ = true; }
 
@@ -127,6 +132,8 @@ class MacroDefinition {
 
   // un-tokenized text
   DefaultTokenInfo definition_text_;
+  std::vector<TokenInfo> lexed_macro_body;
+
 };
 
 }  // namespace verible
--- a/common/text/macro_definition.h
+++ b/common/text/macro_definition.h
@@ -85,15 +85,6 @@ class MacroDefinition {
 
   void SetDefinitionText(const TokenInfo& t) { definition_text_ = t; }
 
-  // Lexes the definition_text_ and store it in lexed_macro_body
-  void SetLexedDefinitionTokens(const std::vector<TokenInfo>& l) {
-    lexed_definition_tokens_ = l;
-  }
-
-  const std::vector<TokenInfo>& GetLexedDefinitionText() const {
-    return lexed_definition_tokens_;
-  }
-
   // Macro definitions with empty () should call this.
   void SetCallable() { is_callable_ = true; }
 
@@ -136,7 +127,6 @@ class MacroDefinition {
 
   // un-tokenized text
   DefaultTokenInfo definition_text_;
-  std::vector<TokenInfo> lexed_definition_tokens_;
 };
 
 }  // namespace verible
--- a/verilog/analysis/verilog_analyzer.cc
+++ b/verilog/analysis/verilog_analyzer.cc
@@ -209,11 +209,17 @@ std::unique_ptr<VerilogAnalyzer>
 VerilogAnalyzer::AnalyzeAutomaticPreprocessFallback(absl::string_view text,
                                                     absl::string_view name) {
   std::unique_ptr<verilog::VerilogAnalyzer> parser;
-  for (bool preprocess_filter_branches : {false, true}) {
-    parser = verilog::VerilogAnalyzer::AnalyzeAutomaticMode(
-        text, name, {.filter_branches = preprocess_filter_branches});
-    if (parser && parser->LexStatus().ok() && parser->ParseStatus().ok()) break;
-    VLOG(1) << "Retry parsing with filter branches enabled";
+  for(bool preprocess_expand_macros : {true, false}){
+    bool expand_macro_status=0;
+    for (bool preprocess_filter_branches : {false, true}) {
+      parser = verilog::VerilogAnalyzer::AnalyzeAutomaticMode(
+          text, name, {.filter_branches = preprocess_filter_branches,.expand_macros=preprocess_expand_macros});
+      if (parser && parser->LexStatus().ok() && parser->ParseStatus().ok()) {expand_macro_status=1; break;}
+      VLOG(1) << "Retry parsing with filter branches enabled";
+    }
+    if(expand_macro_status) break;
+    VLOG(1) << "Retry parsing with macro expanding disabled";
+    
   }
   return parser;
 }
--- a/verilog/analysis/verilog_analyzer.cc
+++ b/verilog/analysis/verilog_analyzer.cc
@@ -209,17 +209,21 @@ std::unique_ptr<VerilogAnalyzer>
 VerilogAnalyzer::AnalyzeAutomaticPreprocessFallback(absl::string_view text,
                                                     absl::string_view name) {
   std::unique_ptr<verilog::VerilogAnalyzer> parser;
-  for(bool preprocess_expand_macros : {true, false}){
-    bool expand_macro_status=0;
+  for (bool preprocess_expand_macros : {true, false}) {
+    bool expand_macro_status = 0;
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

**Comment by karimtera** [comment]
P.S. I know these are not ready by any means to be merged, I am opening this PR to get any guidance if there are better ideas, or in case I am totally going in the wrong direction :)

**Comment by hzeller** [comment]
The 'expand needed when called' in particular you'll see once you work with callable macros as it has to expand with the parameters in mind.

One additional thing: SystemVerilog macros can be recursively defined, so a macro can have macro definitions which then need to be expanded. So prepare for a situation in which you're inside an expansion and hit macro definitions and expansions.

Running clang-format is simple, there is already a script: Just invoke `.github/bin/run-clang-format.sh` before submitting.

File: common/text/macro_definition.h [hunk 1]
```diff
@@ -85,6 +85,11 @@ class MacroDefinition {
 
   void SetDefinitionText(const TokenInfo& t) { definition_text_ = t; }
 
+  // Lexes the definition_text_ and store it in lexed_macro_body
+  void SetLexedDefinitionText(const std::vector<TokenInfo> l) {lexed_macro_body = l ;}
```
hzeller, line 89: Don't pass vectors by value.

I'd probably call it `SetLexedDefinitionTokens()` to better reflect what we do. Also `lexed_macro_body` should be `lexed_definition_tokens_` then so that we don't have many names for the same thing.

common/text/macro_definition.h
```diff
@@ -127,6 +132,8 @@ class MacroDefinition {
 
   // un-tokenized text
   DefaultTokenInfo definition_text_;
+  std::vector<TokenInfo> lexed_macro_body;
```
hzeller, line 135: Note the style to use underscores at the end of instance variables of classes (exception: `struct`s).

verilog/preprocessor/verilog_preprocess.cc
```diff
@@ -212,6 +232,19 @@ void VerilogPreprocess::RegisterMacroDefinition(
   // TODO(hzeller): multiline warning with 'previously defined here' location
 }
 
+absl::Status VerilogPreprocess::ExpandMacro(MacroDefinition& definition){
+  VerilogLexer lexer(definition.DefinitionText().text());
+  verible::TokenSequence myseq;
+  while (true) {
```
hzeller, line 238: this `while(true)` ... `DoNextToken()` ... `break` loop sounds like it can be formulated as `for`-loop.

verilog/preprocessor/verilog_preprocess.cc
```diff
@@ -229,6 +262,12 @@ absl::Status VerilogPreprocess::HandleDefine(
   verible::MacroDefinition macro_definition(*define_tokens[0], *macro_name);
   const auto parse_error_ptr =
       ParseMacroDefinition(define_tokens, &macro_definition);
+
+  if(!macro_definition.IsCallable()){
+    auto expanding_status=ExpandMacro(macro_definition);
```
hzeller, line 267: We often typically use this pattern for these kind of things.
```
if (auto status = ExpandMacro(...); !status.ok())
   return status;
```

verilog/preprocessor/verilog_preprocess.cc
```diff
@@ -212,6 +232,19 @@ void VerilogPreprocess::RegisterMacroDefinition(
   // TODO(hzeller): multiline warning with 'previously defined here' location
 }
 
+absl::Status VerilogPreprocess::ExpandMacro(MacroDefinition& definition){
```
hzeller, line 235: Don't pass parameters to be modified by reference, but rather as pointer. This is the style used throughout so that it is easy to see at the call-site for the reader that something is modified.

verilog/preprocessor/verilog_preprocess.cc
```diff
@@ -199,6 +201,24 @@ std::unique_ptr<VerilogPreprocessError> VerilogPreprocess::ParseMacroDefinition(
   return nullptr;
 }
 
+// Responds to `define directives.  Macro definitions are parsed and saved
+// for use within the same file.
+absl::Status VerilogPreprocess::HandleMacroIdentifier(
+    const TokenStreamView::const_iterator iter  // points to `MACROIDENTIFIER token
+    ) {
+    const absl::string_view sv = (*iter)->text();
```
hzeller, line 209: At some point, you also want to make this dependent on the config (that now only has the `filter_branches` field). There might be situations in which we do not want to expand macros, so this will help with that (and i suspect a bunch of tests fail now because they assume that macro expansion does not take place, so there you then can set the values you need for the particular test; so maybe you need to support such configuration option rather earlier than later)

hzeller (commented):
A first pass, mostly stylistic comments. Once we have some tests, we can dig deeper into the implementation.

karimtera:
> The 'expand needed when called' in particular you'll see once you work with callable macros as it has to expand with the parameters in mind.
> 
That i noticed, there may be different expansions for one macro, depending on the arguments, or in case it is redefined.
The tokens of each expansion needs to be saved somewhere, is `MacroDefinition` still the right place for this?

> One additional thing: SystemVerilog macros can be recursively defined, so a macro can have macro definitions which then need to be expanded. So prepare for a situation in which you're inside an expansion and hit macro definitions and expansions.
> 

Yeah, so the macro definition body itself needs to be "preprocessed".

> Running clang-format is simple, there is already a script: Just invoke `.github/bin/run-clang-format.sh` before submitting.

Yup, I added that as a self-reminder.

verilog/analysis/verilog_analyzer.cc
```diff
@@ -209,11 +209,17 @@ std::unique_ptr<VerilogAnalyzer>
...
+    bool expand_macro_status=0;
+    for (bool preprocess_filter_branches : {false, true}) {
+      parser = verilog::VerilogAnalyzer::AnalyzeAutomaticMode(
+          text, name, {.filter_branches = preprocess_filter_branches,.expand_macros=preprocess_expand_macros});
+      if (parser && parser->LexStatus().ok() && parser->ParseStatus().ok()) {expand_macro_status=1; break;}
+      VLOG(1) << "Retry parsing with filter branches enabled";
+    }
+    if(expand_macro_status) break;
+    VLOG(1) << "Retry parsing with macro expanding disabled";
+
```
karimtera, line 222: `expand_macros` is a new configuration option for the preprocessor.
It is disabled by default, since it is not fully implemented, and also might not be used by formatter.

karimtera:
`expand_macros` is a new configuration option for the preprocessor.
It is disabled by default, since it is not fully implemented, and also might not be used by formatter.

hzeller (commented):
Alright, this looks like a good first step, now we need various tests that test various aspects.

 * Defininig and then later expanding.
   * Expanding a second time to make sure that this still works as expected
 * Defining, expanding somewhere, then un-defing. (Now, the sequence is removed, so our memory of storing it in the definition does not work anymore, so this might need to change how we store the expanded tokens)
 * Thinking about expanding with macro parameters. In a way, the expansion is always the same, only where the placeholders are, we inject the parameters from the original place.
 * Essentially writing tests to find ways to break the current implementation until there is nothing anymore to fix :)

w.r.t. the undef, maybe we just need a expanded tokens token sequence stored with the preprocessor, which is just a large std:;vector<> of the expanded tokens and each MacroDefinition just remembers the range from where to where its 'own' tokens are (essentially our own arena allocation).
Or we have a vector of `TokenSequence`s in the preprocessor and use that as backing store, while the MacroDefinition just remembers the index into that. Lots of possibilities, so when you play with it, you'll find some that is a good choice.

karimtera:
> Alright, this looks like a good first step, now we need various tests that test various aspects.
> 
> * Defininig and then later expanding.
>   
>   * Expanding a second time to make sure that this still works as expected
> * Defining, expanding somewhere, then un-defing. (Now, the sequence is removed, so our memory of storing it in the definition does not work anymore, so this might need to change how we store the expanded tokens)
> * Thinking about expanding with macro parameters. In a way, the expansion is always the same, only where the placeholders are, we inject the parameters from the original place.
> * Essentially writing tests to find ways to break the current implementation until there is nothing anymore to fix :)
> 

Where can i add the tests? in `lint_tool_test.sh`?
I need to view the preprocessed tokens to validate the tests output, not sure how to do this for now, but i am looking into it.

> w.r.t. the undef, maybe we just need a expanded tokens token sequence stored with the preprocessor, which is just a large std:;vector<> of the expanded tokens and each MacroDefinition just remembers the range from where to where its 'own' tokens are (essentially our own arena allocation). Or we have a vector of `TokenSequence`s in the preprocessor and use that as backing store, while the MacroDefinition just remembers the index into that. Lots of possibilities, so when you play with it, you'll find some that is a good choice.

I chose your second suggestion with a little modification, I think there is no need for the `MacroDefinition` to remember the index into the vector of `TokenSequence`s.
Because, anyway I think it's needed to lex the macro body on each call, as there may be any macros inside it that are redefined.
For example:
```verilog
`define FIRST 0
`define SECOND `FIRST
assign x = `SECOND

`define FIRST 1
assign x = `SECOND
```

karimtera:
Current limitation:
- ExpandMacro can't deal with any preprocessor directive other than macro calls.

To implement this I would need to add something like the `HandleTokenIterator` method, and similarly other handling methods, the only difference will be that new methods would not append to the preprocessor `TokenStreamView`, but only get interpreted, e.g.

```verilog
`define CHILD1 x
`define CHILD2 y
`define PARENT \
    `ifndef CHILD3 \
        `define CHILD3 z \
    `endif \
    `CHILD1,`CHILD2, `CHILD3
```

Calling PARENT shall define CHILD3 and register it to the macro definition map normally, then PARENT expands as `x, y, z`

hzeller:
w.r.t. tests: there is a corresponding [verilog_preprocess_test.cc](https://github.com/chipsalliance/verible/blob/master/verilog/preprocessor/verilog_preprocess_test.cc) into which you can write code to verify that the tokens you get are the tokens you expect.
You can then test with `bazel test verilog/preprocessor/...`

The shell tests such as lint_tool_test.sh are more a final end-2-end test that are much harder to pinpoint a particular problem - these are good to modify later in the project; if the unit-test works, we should be confident that the preprocessing does the right thing.

hzeller:
LGTM, I think we're ready for a merge and iterate from here.

hzeller:
Can you re-base the pull request to current head ? That should make the CI happy.

So roughly the following

```bash
git pull # Get latest from your fork
git fetch upstream  # might first need git remote add upstream https://github.com/chipsalliance/verible.git
git checkout master
git merge upstream/master
git push  # push back to your fork

git checkout clocking_macro
git rebase master
git push -f
```

hzeller (approved):
Thanks!
(sorry, just came back from vacation)

PR #1325, opened 2022-04-29
ContainerProxy: Add macros for inheriting common sets of members.

mglb:
Use of a single macro instead of a long list of methods is more readable and makes it easier to inherit correct methods.


hzeller (approved):
Neat!
Usually I'd suggest to limit the use of macros, but this looks like a good readability enhancer.

Issue #1321, opened 2022-04-26
Forbid implicit begin/end 


suzizecat:
**Summary**

It would be nice to have a style rule checking the presence of a pair of begin/end keywords for each construction that support it.
the obvious example would be with if/else if/else.

It might be interesting to split the case structure from the if one, as it might be convenient and less error-prone to have one-line case statements.

The name could be something like `explicit-begin` ou `forbid-implicit-begin` (a bit too long to my taste)

**Test cases and examples**

```systemverilog
// This example should be diagnosed.
if my_condition // <-- issue
    signal <= 1;
else // <-- issue
    signal <= 0;

other_signal <= 1;
```

```systemverilog
// This variant should NOT be diagnosed (negative test case)
// Considered correct: 
if my_condition begin
    signal <= 1;
end else begin
    signal <= 0;
end

other_signal <= 1;
```

This should be done regardless of the actual position of begin/end:
```systemverilog
// This variant should NOT be diagnosed (negative test case)
// Considered correct: 
if my_condition 
begin
    signal <= 1;
end 
else 
begin
    signal <= 0;
end

other_signal <= 1;
```
The case situation might be a little bit trickier:
```systemverilog
// Maybe valid ? Configuration would be nice...
case (stuff)
    0 : foo = 1; // <-- Arguably OK
    1 : foo = 0; // <-- Same
    3 :          // <-- Arguably not OK
          foo = bar;
    default:
        begin // <-- Always OK
              foo = 0;
       end
endcase
```
**Proposal**

N/A

**Additional context**

N/A

Best regards,
Julien FAUCHER

sconwayaus:
Also mentioned here: #2040

PR #1336, opened 2022-05-16
[LINT] Add rule explicit-begin

suzizecat:
Add rule to check if the begin keyword always follows a if/else/for.
Related to #1321

### DIFF
--- a/verilog/analysis/checkers/BUILD
+++ b/verilog/analysis/checkers/BUILD
@@ -23,8 +23,8 @@ cc_library(
         ":create_object_name_match_rule",
         ":disable_statement_rule",
         ":endif_comment_rule",
-        ":explicit_begin_rule",
         ":enum_name_style_rule",
+        ":explicit_begin_rule",
         ":explicit_function_lifetime_rule",
         ":explicit_function_task_parameter_type_rule",
         ":explicit_parameter_storage_type_rule",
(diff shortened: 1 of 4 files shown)


suzizecat:
This should be OK (clang-format was ran and tests were performed on a testcase).
I have not run the integrated tests because I didn't found a way to only run my test (I'm a noob with bazel).

hzeller:
You can run your test with 

```
bazel test //verilog/analysis/checkers:explicit_begin_rule_test
```

suzizecat:
Thanks, I understood the "reject" pattern thing.
This should be ok. however I need to add tests for "for" constructions

hzeller:
What is the status of this PR @suzizecat ? Looks like we were almost there, just needed to make all the tests work.
Can you re-base and make the tests work, so that we're ready for review ?

suzizecat:
Hi @hzeller , I don't have access to the computer used for developing this before next week, but I'll have a look again at this time. 
IIRC I have some issues with cases I hadn't accounted for at first glance.

I'll let you know if I run in any issues.

sconwayaus:
Hi @suzizecat. I was interested in this feature as a user and came across this PR. Are you wanting to finish this one off? Otherwise I'm happy to pick this up.

FWIW: I copied your code into a branch here: https://github.com/sconwayaus/verible/tree/begin_end_rule. Wasn't a lot of work to make this build at the head, (couple of BUILD files was about it), and I added some for loop tests too.

suzizecat:
Hi @sconwayaus, I had issues with building (probably the BUILD files you fixed) but had no time to work on this. 
If I remember correctly, there was also quite a few edge cases that weren't taken into account by this PR, but if you are willing to pick this up, please feel free to do so.

PR #1345, opened 2022-06-06
Build and upload macOS binaries for releases

### LINKED ISSUE #1344: [macOS] CI not building release binaries for macOS
Hi all, it looks like the GitHub Actions process doesn't prepare any release artifacts for macOS, and as a result there aren't any macOS binaries on the Releases page: https://github.com/chipsalliance/verible/blob/dd5e91a4c632b47c8721784e23bf3acc6f46afbc/.github/workflows/verible-ci.yml#L244-L266

I'm not sure if there's an automatic connection between the two, but [the homebrew repo for this project](https://github.com/chipsalliance/homebrew-verible) also hasn't seen a new release since October of last year.

I can try putting together a PR for releasing macOS Verible binaries, but since I'm not that familiar with GH Actions, I'm mostly going to be pattern-matching based on what I see for Windows/Linux binaries.

tymcauley:
I'm a GitHub-Actions novice, so I'm open to suggestions if this isn't the right way to solve this problem.

Closes #1344.

Status: Pending resolution to #1404.

### DIFF
--- a/.github/workflows/verible-ci.yml
+++ b/.github/workflows/verible-ci.yml
@@ -360,6 +360,20 @@ jobs:
     - name: Tests
       run: bazel test --test_output=errors --cxxopt=-Wno-range-loop-analysis //...
 
+    - name: Build Verible Binaries
+      run: bazel build -c opt :install-binaries
+
+    - name: Prepare release
+      run: |
+        VERSION=$(git describe --match=v*)
+        bazel run -c opt :install -- "/private/var/tmp/verible-${VERSION}-macos"
+        tar -zcvf "verible-${VERSION}-macos.tar.gz" "/private/var/tmp/verible-${VERSION}-macos"
+
+    - name: 📤 Upload artifact
+      uses: actions/upload-artifact@v2
+      with:
+        path: verible-*.tar.gz
+
   PrepareVSPlugin:
     container: debian:bullseye
     runs-on: [self-hosted, Linux, X64]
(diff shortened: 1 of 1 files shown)


hzeller:
Thanks!
We also already have set up a homebrew repo, but it needs to have some triggering enabled so that we can automatically build a new release there:
https://github.com/chipsalliance/homebrew-verible

... however, we need some owner who can help automating things; would you like to have a look @tymcauley ?

hzeller:
the CI now ran, but it had some trouble in running the MacOS build, can you have a look ?

hzeller:
The MacOS build is still failing BTW, there might be more refining needed.

tymcauley:
Sorry for the delay on this, this still needs some work to get it functional, hopefully I'll have time to work on it in the near future.

tymcauley:
Looks like the macOS build is failing now at the install step, most likely due to the issue raised here: https://github.com/chipsalliance/verible/issues/1404#issuecomment-1400726752

HendrikMennen:
Any news on this? Would be awesome to have macos binaries

tymcauley:
> Any news on this? Would be awesome to have macos binaries

I think this won't work until #1404 is resolved.

hzeller:
Until someone with access to a Mac has solved #1404 and sent a pull request, note also that there is a brew package available:
https://github.com/chipsalliance/homebrew-verible

hzeller:
If the `:install` is not working yet due to #1404, this could probably be circumvented by manually installing the binaries.

That is also how the homebrew formula avoids that problem:

https://github.com/chipsalliance/homebrew-verible/blob/46664b28ed8f0dfe2d709faf45691cf42d9628b6/Formula/verible.rb#L34-L43

hzeller:
I've now manually installed the binaries and packaged up things.

Issue #1358, opened 2022-07-03
[WIP][Discussion] Standalone preprocessor tool design


karimtera:
# Standalone Preprocessor Tool Design

## Introduction

The goal of this thread is to discuss and put the outlines of the standalone preprocessor tool.

The preprocessor tool should fully support the directives described in 2018 SV-LRM, and provide APIs that can be easily used by other Verible tools.

This proposal is a continuation on the already built pseudo-preprocessing tool.

#### Description
This replaces comments with equal number of whitespaces, to preserve the same byte offsets.

(_IN-PROGRESS_)

tgorochowik:
These parameters look good to me. @hzeller @fangism please comment if you have some additional ideas.

One thing that we should consider early on is how to handle conditionals in the code, here are some notes from my discussion with @mglb 

Consider following code:

```
module mod_name (input bus_a, output bus_b
`ifdef USE_PG_PIN
    , input pg_pin
`endif
    );
endmodule
```

To parse this into a tree, Verible could preprocess the code and generate two sources: one for condition being true, one for condition being false. Both sources would be parsed into alternate trees. The trees would be then merged to share common branches. Alternate branches would be wrapped in a dedicated node.
The preprocessor lines (and inactive in a given variant conditional code blocks) could be treated like comments for parsing and formatting purposes.

What is needed

- Separate preprocessor parser. Can be derived from Verilog parser. If possible, it should share common parts with Verilog parser. It should output location of all preprocessor instructions and macros, together with their semantic information (i.e. what they are, what is their expansion, possibly alternate expansions for macros defined in ifdef blocks, etc).
- Using the preprocessor parser to generate all possible source variants for Verilog parsing purposes. Or just single variant when defines are given.
- Implementation of token list and CST structures for storing alternate parts of preprocessed sources.
- Support for processing (formatting, linting) of multiple source variants. Preprocessor instructions (ifdef, define, …) could be ignored just like comments, same with code blocks that are inactive in a given variant.
- TBD: Macro calls which expand to partial structures (e.g. those with “begin” but without “end”)

Note that the integration with linter and formatter can be done at a later step, but we have to keep in mind that this is the ultimate goal so we have to think about how to make that as easy as possible with the implementation we're doing

tgorochowik:
And in general, the standalone tool should take raw systemverilog sources as input and ouput preprocessed code (only one branch of the conditional code based on the defines, macros should be resolved etc).

[...]

PR #1372, opened 2022-08-04
[VeriblePreProcessor][1]: multiple-cu and generate-variants modes

karimtera:

**Description:**
`--generate_variants` mode generates all the possible variants for conditionals directives.

In case of `--multiple-cu` is true, and multiple files are passed to the tool, each one of them will be compiled in a separate compilation unit, thus declarations scopes will end by the end of each file (this is the default behavior).


verilog/analysis/flow_tree.cc
```diff
@@ -0,0 +1,147 @@
...
+    variants_.push_back(current_sequence_);
+  }
+  if (current_token.token_enum() != PP_Identifier &&
+      current_token.token_enum() != PP_ifndef &&
+      current_token.token_enum() != PP_ifdef &&
+      current_token.token_enum() != PP_define &&
+      current_token.token_enum() != PP_define_body &&
+      current_token.token_enum() != PP_elsif &&
+      current_token.token_enum() != PP_else &&
+      current_token.token_enum() != PP_endif)
```
hzeller, line 141: For readability reasons, if your `if` is more than one line long, use `{ ... }` blocks

```
  if (this_thing ||
      the_other_thing) {
   do_stuff();
 }
```

(also in other places in this file)

verilog/analysis/flow_tree.cc
```diff
@@ -0,0 +1,147 @@
...
+    current_sequence_
+        .pop_back();  // remove tokens to back track into other variants.
```
hzeller, line 143: if the code breaks funny with a too long comment behind, put the comment in a separate line before.

verilog/tools/preprocessor/verilog_preprocessor.cc
```diff
@@ -49,53 +50,169 @@ static absl::Status StripComments(const SubcommandArgsRange& args,
...
 
-  const std::string usage = absl::StrCat("usage: ", argv[0],
-                                         " command args...\n"
-                                         "available commands:\n",
-                                         commands.ListCommands());
+  // Control flow tree constructing.
+  verilog::FlowTree control_flow_tree(lexed_sequence);
+  auto status =
+      control_flow_tree
+          .GenerateControlFlowTree();  // construct the control flow tree.
```
hzeller, line 130: It looks like the funny break in this line just happend because the comment is behind the line.
If you put it in the line before, it should be fine.
(also, this comment should probably not be there: it is just stating the obvious. Use comments for things that need description, not re-iterate what the code just says).

verilog/analysis/flow_tree.cc
```diff
@@ -0,0 +1,147 @@
...
+            token_index);  // edges from a token to the one coming after it
+                           // directly.
+    } else {
+      if (token_index > 0)
+        edges_[token_index - 1].push_back(
+            edges_[token_index]
+                .back());  // edges from the last token in `ifdef/`ifndef body
+                           // to `endif from the same if block.
+    }
+    token_index++;  // increment the token index.
```
hzeller, line 98: Is the comment adding anything to the understanding what the code says ?

verilog/analysis/flow_tree.cc
```diff
@@ -0,0 +1,147 @@
...
+        }
+        ifs_.pop_back();  // the if block edges were added, ready to pop it.
+      }
+    }
+    token_index++;  // increment the token index.
+  }
+
+  // Adding edges for non-if blocks.
+  token_index = 0;
+  for (auto current_token : source_sequence_) {
```
hzeller, line 84: Do you modify the `current_token` ? If not, make it `auto const`.

verilog/tools/preprocessor/verilog_preprocessor.cc
```diff
@@ -49,53 +50,169 @@ static absl::Status StripComments(const SubcommandArgsRange& args,
...
   }
+  if (args.size() ==
+      1) {  // No Files are passed (files are passed as positional arguments).
+    std::cerr << "ERROR: No System-Verilog files were passed.\n\n"
+              << absl::ProgramUsageMessage() << std::endl;
+    return 1;
+  }
+
+  // Select the operation mode and execute it.
+  if (strip_comments_flag) {
```
hzeller, line 189: this nested if/else if with for loop and another if inside is tedious to read. Suggest having separate small functions that are called on the various flags.

hzeller (commented):
First round of comments and suggestions.

Stylistically, I think you like to put comments at the end of a code line, but then the formatting makes that unusally awkward as it breaks it into multiple lines. In many cases it makes sense to put the comment in front for eaiser readabiilyt.

Also a suggestion for the API of flow tree which might be nicer with a callback funciton.

hzeller:
Also question that needs to be considered and tested.

Consider the following example:

```SystemVerilog
`ifdef A
   // foo
`else
   // bar 
`endif

`ifdef A
   // baz
`else
   // quux 
`endif

`ifndef A
   // foobar
`else
   // bazfoo
`endif
```

will that create (2^3=) 8 variants of (number of different conditions for `A`=) 2 ? It should be two as there is only one condition in different places; so it is important to keep track of the different conditions not necessarily be branches.

(In the real world, you see files with many ifdefs but they only look at a handful of different conditions. So it  will be a practical difference if we get 16 or 2^3122 variants out of it)

karimtera:
Hello @hzeller,

Thanks a lot for your review, and sorry for my silly mistakes! :)

Regarding the question you discussed in the previous comment, I think this would require that we know (while traversing the tree) which macros are assumed to be defined or not, maybe in the form of a bit mask.

So, if we are generating a variant in which `A` is defined, we shouldn't freely try all possible next nodes from an `ifdef A` condition, because we already know `A` should be defined in this variant.

Suggestion:
1. Have a vector `condional_macros` to save all macros used in conditionals by the order of appearing.
2. Add a bit mask argument to `DepthFirstSearch`, such that each bit refers to a macro index in `conditonal_macros` vector, initially the bit mask should be 0, and a bit is set when the referred macro is assumed to be defined in the currently generating variant.

Would this idea be a good one? a bit mask can only be 32 bit long, or 64 in case of long int, would that be enough in practical cases?

Thanks,
Karim

hzeller:
I like the idea to represent the different conditions as a bit in an int, because then we also have a way to uniquely identify a variant by its number: 0x42 represents the variant in which all conditions are considered 'false' except the second (0x02) and the 7th (0x40). And numbering can happen from the top of the file. Also we can exactly query how many different variants we actually have.
(it is probably ok to have the conditions a bit fuzzy a big and have `ifdef A`, `ifdef B` and `ifdev A || B` just simply three distinct conditions).

We do need to be aware that there might be many more than 64 conditions, so I'd be wary to store it in a unsigned int, but rather some sort of bitset.

verilog/analysis/flow_tree.cc
```diff
@@ -0,0 +1,150 @@
...
+               current_token_enum == PP_elsif ||
+               current_token_enum == PP_endif) {
+      // Add iterator of `elsif/`else/`endif to else_block of the last recent if
+      // block.
+      if_blocks_.back().else_block.push_back(iter);
+
+      // If the current token is an `endif, then we are ready to create edges
+      // for this if block.
+      if (current_token_enum == PP_endif) {
+        auto& current_if_block = if_blocks_.back().else_block;
```
hzeller, line 63: can this be const `const auto& current_if_block = ....`

In general, whenever you do an assignment, make it const by default unless it is not possible (I also sometimes leave out the `const` if it would force an awkward line-break which wouldn't there without).

That helps everyone reading the code to keep less things in their mind as they don't have to worry if there might be an update to a variable.

verilog/analysis/flow_tree.h
```diff
@@ -0,0 +1,73 @@
...
+using TokenSequenceConstIterator = verible::TokenSequence::const_iterator;
+
+struct ConditionalBlock {
+  // Start of a ifdef/ifndef block
+  TokenSequenceConstIterator if_block_begin;
+
+  // Subsequent else/elsif blocks
+  std::vector<TokenSequenceConstIterator> else_block;
+};
+
```
hzeller, line 37: The `FlowTree` wants a class documentation comment.

verilog/analysis/flow_tree.h
```diff
@@ -0,0 +1,73 @@
...
+class FlowTree {
+ public:
+  explicit FlowTree(verible::TokenSequence source_sequence)
+      : source_sequence_(std::move(source_sequence)){};
+
+  // Constructs the control flow tree by adding the tree edges in edges_.
+  absl::Status GenerateControlFlowTree();
+
+  // Generates all possible variants.
+  absl::Status GenerateVariants();
```
hzeller, line 47: The output via a callback has not been addressed yet (maybe because they showed up as `furtther comments` hidden in the github review ?).

verilog/analysis/flow_tree.h
```diff
@@ -0,0 +1,73 @@
...
+
+  // A memory for all variants generated.
+  std::vector<verible::TokenSequence> variants_;
```
hzeller, line 50: With the callback reporting, we don't need this output variable.

hzeller:
Please also add a `flow_tree_test.cc` to test every aspect of the functionality.

verilog/analysis/flow_tree.h
karimtera: No, I saw the comments but I am still figuring out how to use it from the user's or the tool's side.

karimtera:
> (it is probably ok to have the conditions a bit fuzzy a big and have ifdef A, ifdef B and ifdev A || B just simply three distinct conditions).

I checked the LRM, and here is what I found:

Section: 22.6
> ifdef_directive ::=
ifdef text_macro_identifier ifdef_group_of_lines
{ elsif text_macro_identifier elsif_group_of_lines }
[ else else_group_of_lines ]
endif

Section: 22.5.1
> text_macro_identifier ::=
identifier

Thus, I think it shouldn't be supported, do you agree?

hzeller:
ah, so you mean the only 'expression' in an ifdef really can be the existence of a single identifier. That is good, that makes things simple.
(of cousre, there could be other ifdefs that then define/redefine identifiers, but let's not worry about that in the first iteration)

karimtera:
I implemented the `std::bitset` solution, for example:
```verilog
`ifdef A
  A_TRUE
`else
  A_FALSE
`endif

`ifdef A
  A_TRUE
`else
  A_FALSE
`endif

`ifndef A
  A_FALSE
`else
  A_TRUE
`endif
```

Result:
```
Variant number 1:
(#293: "A_TRUE")
(#293: "A_TRUE")
(#293: "A_TRUE")

Variant number 2:
(#293: "A_FALSE")
(#293: "A_FALSE")
(#293: "A_FALSE")
```

I assumed the maximum number of macros to be 128 just for now, It seams to be working.

verilog/analysis/flow_tree.h
```diff
@@ -0,0 +1,114 @@
...
+  std::vector<ConditionalBlock> if_blocks_;
+
+  // The original source code lexed token seqouence.
+  const verible::TokenSequence source_sequence_;
+
+  // The variant's token sequence currrently being built by DepthFirstSearch.
+  verible::TokenSequence current_sequence_;
+
+  // Macros assumed in the variant that is currently begin built.
+  // TODO(karimtera): Is it better to use dymanic_bitset?
```
hzeller, line 98: or a vector in which we store `uints_64` and do some bid-fiddeling.
For now, this is fine, but make sure in a unit-test that if you generate more than 128 different macros, that the code reacts gracefully, e.g. by returning a Status in which it returns a failure.

Anyway, since we use the same type in other places as well (e.g. in the parameter to DepthFirst search), I suggest to make it a using declaration/typedef. The common place to define it would be right after `'private:`

```c++
private:
  using BitSet = std::bitset<128>;
  // ...
  absl::Status DepthFirstSearch(..., BitSet,...);
  //...
  BitSet current_macros_;
```

verilog/analysis/flow_tree.h
```diff
@@ -0,0 +1,114 @@
...
+#include "absl/status/status.h"
+#include "common/lexer/token_stream_adapter.h"
+#include "verilog/parser/verilog_token_enum.h"
+
+namespace verilog {
+using TokenSequenceConstIterator = verible::TokenSequence::const_iterator;
+
+// Receive a complete token sequence of one variant.
+// 1st argument: generated variant.
+// 2nd argument: number of generated variants (to decide if more is needed).
```
hzeller, line 33: I'd remove this parameter. You already can deal with 2^128 different variants, but you only provide a number that can hold 2^31 positive values.

I'd leave it out for now. If we ever want to tell the callback which exact variant we're talking about, we might need to pass the bitset (as reference). But I guess for now let's not worry about it.

verilog/analysis/flow_tree.cc
```diff
@@ -0,0 +1,348 @@
...
+
+  return absl::OkStatus();
+}
+
+// Traveses the control flow tree in a depth first manner, appending the visited
+// tokens to current_sequence_, then provide the completed variant to the user
+// using a callback function (VariantReceiver).
+absl::Status FlowTree::DepthFirstSearch(const VariantReceiver &receiver,
+                                        TokenSequenceConstIterator current_node,
+                                        std::bitset<128> assumed) {
```
hzeller, line 256: passing the bitset by value sounds expensive.

verilog/tools/preprocessor/verilog_preprocessor.cc
```diff
@@ -16,29 +16,44 @@
...
+bool PrintGeneratedVariant(const verible::TokenSequence& recieved_variant,
+                           const int variants_counter,
+                           const bool is_variant_complete) {
+  if (!is_variant_complete) {
+    // Only accepts 2 variants at most. (to test the functionality).
+    // TODO(karimtera): How many variants are practical?
+    return variants_counter < 20;
   }
-  const char* source_file = args[0];
+  std::cout << "Variant number " << variants_counter + 1 << ":\n";
```
hzeller, line 45: Since we discussed above to probably not include the variant counter in the callback function, if you really want to have a counter, just count it separately, if you keep track of how often times you've been called

```c++
int counter = 0;
auto status = control_flow_tree.GenerateVariants([&counter](const verible::TokenSequence& variant) {
   ++counter;
   // Do whatever needs to be done to print
  return counter < kNumberOfVariantLimit;  // probably should come from a flag
});
```

hzeller (commented):
Alright, starting to look good!

Now very important is to write a unit test for flow_tree. It is actually always a good idea to start the test right at the time when you begin implement something, because it makes life so much simpler - you always know what is working, and it is much easier to change things. And since the unit test is the first 'user' of the API, you also get a feel if it provides everything one would need, long before it is wired up in some other library or binary somewhere.
Ah well, we're a bit late now, but better late than never :)

So start a `flow_tree_test.cc`, make all the boilerplate to set up a particular input stream and test that you get all the expected responses.

verilog/analysis/flow_tree.h
```diff
@@ -0,0 +1,114 @@
...
+// 3rd argument: checks if the variant is ready to be sent.
```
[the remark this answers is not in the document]
karimtera: Not every variant that the VariantReceiver receives is ready to be used.
Because I use the receiver callback in 2 modes:
1. To check if the user wants more variants or not (main functionality).
2. Also to terminate other DFS calls, that was started before the user says "no more" (this is where the 3rd argument will be false).

At the beginning of `FlowTree:DepthFirstSearch` I do:
```c++
  if (!receiver(current_sequence_, variants_counter_, false)) {
    return absl::OkStatus();
  }
```

verilog/analysis/flow_tree.h
```diff
@@ -0,0 +1,114 @@
...
+// source code. Furthermore, enabling doing the following queries on the graph:
+// 1) Generating all the possible variants (provided via a callback function).
+// 2) TODO(karimtera): uniquely identify a variant with a bitset.
+class FlowTree {
+ public:
+  explicit FlowTree(verible::TokenSequence source_sequence)
+      : source_sequence_(std::move(source_sequence)){};
+
+  // Constructs the control flow tree by adding the tree edges in edges_.
+  absl::Status GenerateControlFlowTree();
```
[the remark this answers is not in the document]
karimtera: It only populates `edges_`, would it be better if I added this in the beginning of `GenerateVariants`?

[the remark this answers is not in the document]
hzeller: For all user-facing API calls, think from the point of view of the user: is this something they expect to do when they work with this, or is this some implementation detail that requires the user to know what is going up internally to be able to call the functions in the right sequence.

It sounds like an implementation detail to me, so if `GenerateVariants()` sees that edges have not been populated, it should do that.

verilog/analysis/flow_tree.h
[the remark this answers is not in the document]
hzeller: so if it is used in two modes, this needs to be clearly explained (the current third parameter description is not super helpful), but it also sounds like some unnecessary leaking of an implementation detail.

My expectation as a use of such an API is: I get called back with fully usable variants and once I return 'false', the callback will not be called anymore.

verilog/analysis/flow_tree.h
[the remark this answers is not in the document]
karimtera: Yes, Thanks for clarification.
I agree that when the user calls `GenerateVariants()` He doesn't expect that he should use `GenerateControlFlowTree` before it, `GenerateVariants()` should be complete in that regard.

verilog/analysis/flow_tree.cc
karimtera: I can add it as a class member just like `current_macros_` bitset.
Actually, I may create a struct containing everything about the variant being generated:
1. Its `TokenSequence`.
2. bitset of visited/assumed macros so far (1 means visited or assumed, 0 means don't care or not visited).
3. bitset of assumed macros (1 means assumed to defined, 0 vice versa).

hzeller: Yes, a struct sounds good as it will help document what is needed.

verilog/tools/preprocessor/verilog_preprocessor.cc
karimtera: Sorry, I don't quite understand this syntax.
I'll just count variants inside `VariantReceiver` using a static local variable for now.

hzeller: Static counter sounds like a good first step, we can refine that later.

The syntax is a c++ closure, so the parameters and {} with return are the same as a function. As with closures in other language, it can also capture values from the surrounding scope; in this case, we capture a reference to the counter using the `[&counter]` part.

verilog/analysis/flow_tree.h
```diff
@@ -0,0 +1,123 @@
...
+  verible::TokenSequence sequence;
+
+  // The i-th bit in "macros_mask" is 1 when the macro (with ID = i) is assumed
+  // to be defined, o.w. it is assumed to be undefined.
+  BitSet macros_mask;
+
+  // The i-th bit in "assumed" is 1 when the macro (with ID = i) is visited or
+  // assumed (either defined or not), o.w. it is not visited (its value doesn't
+  // affect this variant).
+  BitSet assumed;
```
hzeller, line 52: I don't really understand the description of these fields.
What is the difference of `macro_mask` and `assumed` ?

verilog/analysis/flow_tree.h
```diff
@@ -0,0 +1,123 @@
...
+#include <string>
+#include <vector>
+
+#include "absl/status/status.h"
+#include "common/lexer/token_stream_adapter.h"
+#include "verilog/parser/verilog_token_enum.h"
+
+namespace verilog {
+
+using BitSet = std::bitset<128>;
```
hzeller, line 30: Having `BitSet`, `ConditionalBlock`,`Variant`, and `VariantReceiver` here pollutes the `verilog::` namespace but it is really only needed in the context of the FlowTree.

Define these types inside the FlowTree.

verilog/analysis/flow_tree.h
karimtera: `assumed` is used to identify if a macro's value doesn't matter a variant, e.g.

```verilog
`ifdef A
  `ifdef B
   ...
   `endif
`endif
```

In this case we have 2 macros, then each variant should be identified by a `bitset` (which is `macro_mask`) of length 2.
For example the `bitset macro_mask` of value `01`  shows the variant in which `A` is defined and `B` is not.
Similarly the `bitset macro_mask` of value `00` refers to the variant in which both `A` and `B` is undefined, but in fact when `A` is undefined, then `B` doesn't affect the variant and using `0` in `macro_mask` is not accurate enough because `10` will still refers to the same variant.

That's the use of `assumed` it is like another layer of information, if a bit in assumed is not set, then it means that this macro doesn't affect the variant in any way.

hzeller: It might be useful to add this description in the header file

hzeller:
FYI, the preprocessor changed slightly in the meantime, the pull request needs to be rebase.

verilog/analysis/flow_tree_test.cc
```diff
@@ -0,0 +1,185 @@
...
+    `endif
+    )",
+      R"(
+    `endif
+    )"};
+
+  for (auto test : test_cases) {
+    FlowTree tree_test(LexToSequence(test));
+    ResetGlobalTestingVariables();
+    auto status = tree_test.GenerateVariants(GetVariant);
```
hzeller, line 126: ... and in the cases where you are not really interested in the variants, the closure can be quite small

``` c++
auto status = tree_test.GenerateVariants([&variants](const FlowTree::Variant&) { return true; });
```

verilog/tools/preprocessor/verilog_preprocessor.cc
```diff
@@ -16,29 +16,41 @@
...
-        "Missing file argument.  Use '-' for stdin.");
-  }
-  const char* source_file = args[0];
+bool PrintGeneratedVariant(const verilog::FlowTree::Variant& variant) {
+  static int variants_counter = 1;
+  std::cout << "Variant number " << variants_counter << ":\n";
+  for (auto token : variant.sequence) std::cout << token << '\n';
+  puts("");
+  variants_counter++;
+  // Only accepts 20 variants at most. (to test the functionality).
```
hzeller, line 43: This should be a flag, for instance something like

```c++
ABSL_FLAG(int, limit_variants, 20,
                      "Maximum number of variants printed"
````

File: verilog/tools/preprocessor/verilog_preprocessor.cc [hunk 21]
```diff
@@ -49,53 +61,163 @@ static absl::Status StripComments(const SubcommandArgsRange& args,
...
+  std::string source_contents;
+  if (auto status = verible::file::GetContents(source_file, &source_contents);
+      !status.ok()) {
+    std::cerr << "ERROR: passed file can't be open\n.";
+    return status;  // check if the the file is not readable or doesn't exist.
+  }
 
-int main(int argc, char* argv[]) {
-  // Create a registry of subcommands (locally, rather than as a static global).
-  verible::SubcommandRegistry commands;
```
hzeller, line 68: You have removed the whole subcommand registry stuff. That should be kept.

hzeller (commented):
Looks like the subcommand stuff had been removed, I suspect that was in the early implementation stages to get things going ? That should of course be kept.

For the test: looks pretty good, some small suggestions there.

verilog/analysis/flow_tree_test.cc
```diff
@@ -0,0 +1,185 @@
...
+    `else
+      A_TRUE
+    `endif)";
+
+  FlowTree tree_test(LexToSequence(test_case));
+  ResetGlobalTestingVariables();
+  auto status = tree_test.GenerateVariants(GetVariant);
+  EXPECT_TRUE(status.ok());
+  EXPECT_EQ(g_variants.size(), 2);
+  for (const auto& variant : g_variants[0].sequence) {
```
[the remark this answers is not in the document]
karimtera: Couldn't use `ElementsAre` with `TokenInfo`, as comparing tokens fails due to comparing the bounds of the `string_view` of the text.
Used this instead:
```c++
EXPECT_THAT(variants[0].sequence[0].text(), "A_TRUE_1");
```

verilog/tools/preprocessor/verilog_preprocessor.cc
karimtera: Yes sorry, but why would we need it?
I think using 'FileList' from 'verilog_project.h' for positional arguments, and `absl/flags` for other flags is enough.
I don't understand the difference between treating 'strip_comments' as a subcommand, or as a boolean flag.

hzeller: the subcommand is something where we can nicely implement features, such as listing them with their own description, have its own flags, can do its own tests (e.g. make sure that only one is there at a time), create separate manpages etc. (consider it similar to the idea of the git subcomments. like 'git push', 'git commit' etc are all subcommands.

You can get an inferior subset of the same functionality by using absl-flags, but since we already have Subcommands, why should we downgrade ?

karimtera: Ah okay, I get it.

I will have a deeper look into the subcommand class, and use it instead of the absl flags. 🙏🏻

verilog/analysis/flow_tree.h
```diff
@@ -0,0 +1,133 @@
...
+
+namespace verilog {
+
+// FlowTree class builds the control flow graph of a tokenized System-Verilog
+// source code. Furthermore, enabling doing the following queries on the graph:
+// 1) Generating all the possible variants (provided via a callback function).
+// 2) TODO(karimtera): uniquely identify a variant with a bitset.
+class FlowTree {
+ public:
+  using BitSet = std::bitset<128>;
```
hzeller, line 35: Maybe you can add a
```
static constexpr int kMaxDistinctMacros = 128;
```
and then use it in the `std::bitset<kMaxDistinctMacros>`. Then it reads nicely and makes aware of the limitations.

verilog/analysis/flow_tree.h
```diff
@@ -0,0 +1,133 @@
...
+  using TokenSequenceConstIterator = verible::TokenSequence::const_iterator;
+
+  struct ConditionalBlock {
```
hzeller, line 38: The name `ConditionalBlock` and then seeing the fields it has does not necessarily connect, mostly because of the naming of the fields inside (suggestion below). But also, like any class or struct, this of course wants a description.

The naming of the fields is odd. Typically we call something an `iterator` if it is something you'd want to use to advance through some sequence of things. So while these are an iterator type, here, we're mostly using them to remember a position; maybe call them '`pos` or `postition`, `location` or something of that sort ?
e.g. `else_location`.

There needs to be a description of where the iterators point to when they are not pointing anywhere; possibly the `end()` of some token sequence, but it is not entirely clear what that would be. Maybe a short description that this `ConditionalBlock` struct is always used in conjunction with some corresponding `TokenSequence` and that all of the fields in here are pointing inside there.

Also, it seems ifdef and ifndef could be exclusive, maybe that requires some documentation or a different code to make it impossible to mis-use. Maybe only one `if_location` that either points to either an `ifdef` if `ifndef`  and a boolean indicating if this is a positive condition ?

verilog/analysis/flow_tree.h
```diff
@@ -0,0 +1,133 @@
...
+    TokenSequenceConstIterator else_iterator;
+    TokenSequenceConstIterator endif_iterator;
+  };
+
+  struct Variant {
+    // Contains the token sequence of the variant.
+    verible::TokenSequence sequence;
+
+    // The i-th bit in "macros_mask" is 1 when the macro (with ID = i) is
+    // assumed to be defined, o.w. it is assumed to be undefined.
```
hzeller, line 51: o.w. stands for otherwise ? This is an unusual abbreviation; either use `o/w` or write it out `otherwise`.

verilog/analysis/flow_tree.h
```diff
@@ -0,0 +1,133 @@
...
+    BitSet macros_mask;
+
+    // The i-th bit in "assumed" is 1 when the macro (with ID = i) is visited or
```
hzeller, line 54: Still don't fully understand what assumed means here, but maybe the meaning is `visited` ?

verilog/analysis/flow_tree.h
```diff
@@ -0,0 +1,133 @@
...
+  // Receive a complete token sequence of one variant.
+  // variant_sequence: the generated variant token sequence.
+  using VariantReceiver = std::function<bool(const Variant &variant)>;
+
+  explicit FlowTree(verible::TokenSequence source_sequence)
+      : source_sequence_(std::move(source_sequence)){};
+
+  // Generates all possible variants.
+  absl::Status GenerateVariants(const VariantReceiver &receiver);
+
```
hzeller, line 80: As an user of the API, I'd probably want to know what the meaning of each of the numbered macros is. So we should maybe have a
```
  const std::vector<TokenSequenceConstIterator> &GetUsedMacros();
```

That way, a user can associate bit N they get back in Variant in the BitSet with a meaningful name. Also they get an idea if the number of possible variants.

verilog/analysis/flow_tree_test.cc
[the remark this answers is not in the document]
hzeller: Did you git push your changes yet ?

verilog/analysis/flow_tree_test.cc
```diff
@@ -0,0 +1,185 @@
...
+    ResetGlobalTestingVariables();
+    auto status = tree_test.GenerateVariants(GetVariant);
+    EXPECT_TRUE(status.ok());
+    EXPECT_EQ(g_variants.size(), 3);
+    for (const auto& variant : g_variants) {
+      EXPECT_EQ(variant.sequence.size(), 1);
+    }
+  }
+}
+
```
hzeller, line 183: Since `macro_mask` and `assumed` are part of the public API, you also want tests that verify that they contain the expected bitset.

verilog/tools/preprocessor/verilog_preprocessor.cc
hzeller: Don't see this here yet, but I guess you have not git pushed that version yet.

verilog/analysis/flow_tree_test.cc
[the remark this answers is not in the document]
karimtera: Not yet, I am working on the subcommand part.

verilog/analysis/flow_tree.h
karimtera: Yes, It means that during the generation of a variant some paths may not be visited, and their macro value won't make a difference in the variant.

I will change its name to `visited`.

hzeller:
Did you re-base your branch so that we don't have the conflicts ?

verilog/tools/preprocessor/verilog_preprocessor.cc
```diff
@@ -42,25 +51,177 @@ static absl::Status StripComments(const SubcommandArgsRange& args,
...
+  // for debugging.
+  for (auto& u : preprocessed_data.errors) outs << u.error_message << '\n';
+
+  //  parsing just as a trial
+  std::string post_preproc;
+  for (auto u : preprocessed_stream) post_preproc += std::string{u->text()};
+  std::string source_view{post_preproc};
+  verilog::VerilogAnalyzer analyzer(source_view, "file1", config);
+  auto analyze_status = analyzer.Analyze();
+  /* const auto& mydata = analyzer.Data().Contents(); */
```
hzeller, line 113: Don't leave commented-out debug code in the final PR.

verilog/tools/preprocessor/verilog_preprocessor.cc
```diff
@@ -42,25 +51,177 @@ static absl::Status StripComments(const SubcommandArgsRange& args,
...
+  verilog::StripVerilogComments(source_contents, &outs, replace_char);
 
   return absl::OkStatus();
 }
 
-static const std::pair<absl::string_view, SubcommandEntry> kCommands[] = {
-    {"strip-comments",  //
-     {&StripComments,   //
-      R"(strip-comments file
+static absl::Status PreprocessSingleFile(const char* source_file, std::istream&,
```
hzeller, line 75: Could the `source_file` be of type `std::string_view` ?

verilog/tools/preprocessor/verilog_preprocessor.cc
```diff
@@ -42,25 +51,177 @@ static absl::Status StripComments(const SubcommandArgsRange& args,
...
+                                         std::ostream& outs, std::ostream&) {
+  std::string source_contents;
+  if (auto status = verible::file::GetContents(source_file, &source_contents);
+      !status.ok()) {
+    std::cerr << "ERROR: passed file can't be open\n.";
```
hzeller, line 80: We have more information like the filename and the status which actually contains the exact problem, so we can print that directly:

```
  message_stream << filename << ": " << status;
```

verilog/tools/preprocessor/verilog_preprocessor.cc
hzeller, line 75: What is the `std::istream&` intended to be used for ? Looks like it can be removed.

Also the second `std::ostream&` ? It looks like the second ostream is the `message_stream`; we can use that instead of `std::cerr` below.

Parameters that we don't need: let's not include them, as it just gets confusing.

verilog/tools/preprocessor/verilog_preprocessor.cc
```diff
@@ -42,25 +51,177 @@ static absl::Status StripComments(const SubcommandArgsRange& args,
...
+  }
+  verible::TokenStreamView lexed_streamview;
+  // Initializing the lexed token stream view.
+  InitTokenStreamView(lexed_sequence, &lexed_streamview);
+  verilog::VerilogPreprocessData preprocessed_data =
+      preprocessor.ScanStream(lexed_streamview);
+  auto& preprocessed_stream = preprocessed_data.preprocessed_token_stream;
+  for (auto u : preprocessed_stream) outs << *u << '\n';
+  // for debugging.
+  for (auto& u : preprocessed_data.errors) outs << u.error_message << '\n';
```
hzeller, line 105: Wouldn't we expect the error messages go to the `message_stream` ?

verilog/tools/preprocessor/verilog_preprocessor.cc
```diff
@@ -42,25 +51,177 @@ static absl::Status StripComments(const SubcommandArgsRange& args,
...
+
+  //  parsing just as a trial
```
hzeller, line 107: What are the following lines for ? Looks like an expensive string concatenation of everything, then parsing it wihtout using the result. I suspect this is some leftover debugging code ?

verilog/tools/preprocessor/verilog_preprocessor.cc
```diff
@@ -42,25 +51,177 @@ static absl::Status StripComments(const SubcommandArgsRange& args,
...
+  }
+
+  // Control flow tree constructing.
+  verilog::FlowTree control_flow_tree(lexed_sequence);
+  int counter = 0;
+  auto status = control_flow_tree.GenerateVariants(
+      [limit_variants, &counter](const verilog::FlowTree::Variant& variant) {
+        if (counter == limit_variants) return false;
+        counter++;
+        std::cout << "Variant number " << counter << ":\n";
```
hzeller, line 177: This might be good for the message stream, so that command redirect works and we only get the actual file in there.

verilog/tools/preprocessor/verilog_preprocessor.cc
```diff
@@ -42,25 +51,177 @@ static absl::Status StripComments(const SubcommandArgsRange& args,
...
+        for (auto token : variant.sequence) std::cout << token << '\n';
```
hzeller, line 178: is this meant to be to `std::cout` ? Looks like `outs` is the stream we should be targeting.
This looks like it prints the raw tokens, separated with newline which might not be as useful. Ideally, we get a re-creation of a processable SystemVerilog file, but with the conditional blocks removed.

verilog/tools/preprocessor/verilog_preprocessor.cc
```diff
@@ -42,25 +51,177 @@ static absl::Status StripComments(const SubcommandArgsRange& args,
...
+        puts("");
```
hzeller, line 179: What do we need the `puts("")` for ?

verilog/tools/preprocessor/verilog_preprocessor.cc
```diff
@@ -42,25 +51,177 @@ static absl::Status StripComments(const SubcommandArgsRange& args,
...
+    auto status = PreprocessSingleFile(source_file, std::cin, outs, std::cerr);
+    if (!status.ok()) return status;
+    outs << '\n';
+  }
+  return absl::OkStatus();
+}
+
+static absl::Status GenerateVariants(const SubcommandArgsRange& args,
+                                     std::istream&, std::ostream& outs,
+                                     std::ostream&) {
```
hzeller, line 143: This third stream is not named, this is probably our `message_stream` ?

verilog/tools/preprocessor/verilog_preprocessor.cc
```diff
@@ -42,25 +51,177 @@ static absl::Status StripComments(const SubcommandArgsRange& args,
...
+  const int limit_variants = absl::GetFlag(FLAGS_limit_variants);
+  if (args.size() > 1) {
+    return absl::InvalidArgumentError(
+        "ERROR: generate-variants only works on one file.");
+  }
+  const char* source_file = args[0];
+  std::string source_contents;
+  if (auto status = verible::file::GetContents(source_file, &source_contents);
+      !status.ok()) {
+    std::cerr << "ERROR: passed file can't be open\n.";
```
hzeller, line 153: Similar here as in the preprocess function; this should probably go to the message stream and include filename and status message.

verilog/tools/preprocessor/verilog_preprocessor.cc
```diff
@@ -42,25 +51,177 @@ static absl::Status StripComments(const SubcommandArgsRange& args,
...
 Output: (stdout)
   Contents of original file with // and /**/ comments removed.
 )"}},
+    {"multiple-cu",
+     {&MultipleCU,
+      R"(multiple-cu file [more_files]
+Inputs:
+  'file' is a Verilog or SystemVerilog source file.
+   There can be multiple SystemVerilog source files.
+   Each one of them will be compiled in a separate compilation unit.
```
hzeller, line 211: i'd probably write out compilation unit (`multiple-compilation-unit` ?).

'compiled in a separate compilation unit` is a bit ambiguous: it sounds like each file will get its own output, but since we have only stdout as output, we generate only one output ? 

Also, we're not really 'compiling' here, we are preprocessing to generate a compilation unit.
Can you wordsmith the command description a bit so that it is easier to understand ?

verilog/tools/preprocessor/verilog_preprocessor.cc
```diff
@@ -42,25 +51,177 @@ static absl::Status StripComments(const SubcommandArgsRange& args,
...
+  return absl::OkStatus();
+}
+
+static absl::Status MultipleCU(const SubcommandArgsRange& args, std::istream&,
+                               std::ostream& outs, std::ostream&) {
+  if (args.empty()) {
+    return absl::InvalidArgumentError("Missing file arguments.");
+  }
+  for (const char* source_file : args) {
+    outs << source_file << ":\n";
```
hzeller, line 133: this should probably go to the `message_stream`

verilog/tools/preprocessor/verilog_preprocessor.cc
```diff
@@ -42,25 +51,177 @@ static absl::Status StripComments(const SubcommandArgsRange& args,
...
+Inputs:
+  'file' is a Verilog or SystemVerilog source file.
+   There can be multiple SystemVerilog source files.
+   Each one of them will be compiled in a separate compilation unit.
+Output: (stdout)
+  The preprocessed files content (same contents with directives interpreted).
+)"}},
+    {"generate-variants",
+     {&GenerateVariants,
+      R"(genera-variants file [-limit_variants number]
```
hzeller, line 217: (probably in a follow-up change: if we already have a file-list with some `+define+` set, we can 'fix' these define and thus can generate an output with less degrees of freedom, reducing the variants number.)

verilog/tools/preprocessor/verilog_preprocessor.cc
```diff
@@ -42,25 +51,177 @@ static absl::Status StripComments(const SubcommandArgsRange& args,
...
+Inputs:
+  'file' is a Verilog or SystemVerilog source file.
+  '-limit_variants' flag limits variants to 'number' (20 by default).
```
hzeller, line 220: Also for future: we might consider creating an output file per variant. Either with a number or maybe with all assumed names of the macros in the filename:

```
  foo+HELLO-BAR-BAZ.sv  # HELLO defined, but BAR, BAZ not
  foo-HELLO+BAR-BAZ.sv  # HELLO undefined, BAR defined, BAZ undefined.
```
Would be the variant in which `HELLO` is defined, but

verilog/tools/preprocessor/verilog_preprocessor.cc
```diff
@@ -42,25 +51,177 @@ static absl::Status StripComments(const SubcommandArgsRange& args,
...
+Output: (stdout)
+   Generates every possible variant considering the conditional directives.
+)"}},
```
hzeller, line 223: We have an opportunity for another sub-command: `list-macros`, which would be the output of the `GetUsedMacros()` function from the FlowTree (can be a separate change)

verilog/tools/preprocessor/verilog_preprocessor.cc
```diff
@@ -42,25 +51,177 @@ static absl::Status StripComments(const SubcommandArgsRange& args,
...
+      });
+  if (!status.ok()) {
+    return status;
+  }
 
-Input:
+  return absl::OkStatus();
+}
+
+static const std::pair<absl::string_view, SubcommandEntry> kCommands[] = {
```
hzeller, line 189: We need a basic test for the new commands for the compilation unit and variants.

hzeller (commented):
Nice. A few comments w.r.t. which output stream to use so that it is easier for users to redirect the contents to the right output.

And the usual: we need a test for ach new subcommand

karimtera:
> Did you re-base your branch so that we don't have the conflicts ?

Would re-basing using a similar approach to the one you showed me before in PR #1318 solve the conflict? or I should solve the conflict manually here, then pull the changes?

hzeller:
Yes, just like in https://github.com/chipsalliance/verible/pull/1318#issuecomment-1169001516
You'd first get your master fresh, then go into your feature branch (`pp_generate_variants_mode`) and run `git rebase master`. Then there might be conflicts you have to fix, and git will guide you through the process (`git add` the files you have fixed, how to continue the rebase).

(what I do whenver I do some large thing like that, I create a copy of my full git-tree just in case :) ).

verilog/tools/preprocessor/verilog_preprocessor.cc
karimtera: I think it couldn't be, because `source_file` comes from `GetSubcommandEntry` which returns `std::vector<char*>`.
I mean it can be changed of course, but I will have to change `SubcommandFunction` to be a template, and maybe change something in `common/util/init_command_line.h` too. 
https://github.com/chipsalliance/verible/blob/4faba50b9fa6c30cf18b59f1254c15fc214d72c3/common/util/subcommand.h#L31-L42

verilog/tools/preprocessor/verilog_preprocessor.cc
karimtera: Yes, I removed it.

verilog/tools/preprocessor/verilog_preprocessor.cc
karimtera: Yes, this should be `outs`.
I added the support for white-spaces in PR #1376 so that I can recreate the SystemVerilog file after preprocessing.
Can we keep the output for now as tokens until we get there?

verilog/tools/preprocessor/verilog_preprocessor.cc
karimtera: Better visual isolation between variants.

verilog/tools/preprocessor/verilog_preprocessor.cc
karimtera: That would be fun, i will add it as a to-do comment.

verilog/tools/preprocessor/verilog_preprocessor.cc
karimtera: Check my previous comment https://github.com/chipsalliance/verible/pull/1372#discussion_r960428463

hzeller: `const char*` converts naturally into a `std::string_view`, so in the loop in `MultipleCU` it will just work

hzeller: PreprocessSingleFile() is never called from the subcommand. You only register `MultipleCU()` and `GenerateVarints()` with the subocmmands so that you have to adhere to that function signature.
But only from within these functions, you then call `PreprocessSingleFile()`, but you don't need the extra parameters here.

hzeller (changes requested):
I think the superfluous function parameters are not the ones that we need due to subcommands (as these functions are actually not hooked as subcommands. Or am I missing somehting ?)

Other than that, this looks like we're nearing completion of this change!
To see if you have covered all the relevant code parts, take a look how to run coverage on your machine https://github.com/chipsalliance/verible#test (you might need to install `lcov` if not already on your machine)

To run the CI and get ready to merge, please also do the rebase.

karimtera:
I have done the rebase, I just have a problem using `bazel build`:
```log
ERROR: /home/ktera/.cache/bazel/_bazel_ktera/3011e1bac261c5d8b2ad99395406c1ad/external/rules_pkg/pkg/mappings.bzl:473:22: name 'json' is not defined
ERROR: /home/ktera/git/VERIBLE/verible/third_party/proto/kythe/BUILD:23:14: error loading package '@com_google_protobuf//': in /home/ktera/.cache/bazel/_bazel_ktera/3011e1bac261c5d8b2ad99395406c1ad/external/rules_pkg/mappings.bzl: Extension 'pkg/mappings.bzl' has errors and referenced by '//third_party/proto/kythe:analysis_proto'
ERROR: /home/ktera/git/VERIBLE/verible/third_party/proto/kythe/BUILD:23:14: error loading package '@com_google_protobuf//': in /home/ktera/.cache/bazel/_bazel_ktera/3011e1bac261c5d8b2ad99395406c1ad/external/rules_pkg/mappings.bzl: Extension 'pkg/mappings.bzl' has errors and referenced by '//third_party/proto/kythe:analysis_proto'
ERROR: /home/ktera/git/VERIBLE/verible/third_party/proto/kythe/BUILD:23:14: every rule of type proto_library implicitly depends upon the target '@com_google_protobuf//:protoc', but this target could not be found because of: error loading package '@com_google_protobuf//': in /home/ktera/.cache/bazel/_bazel_ktera/3011e1bac261c5d8b2ad99395406c1ad/external/rules_pkg/mappings.bzl: Extension 'pkg/mappings.bzl' has errors
ERROR: Analysis of target '//third_party/proto/kythe:analysis_proto' failed; build aborted: Analysis failed
```

I am trying to figure out the problem, I tried to build the master branch and it gives me the same error.

Note: I use local flex and bison `--//bazel:use_local_flex_bison`.

hzeller:
You need bazel version 4 now

hzeller:
did it work ? Can you `push -f` your rebased branch ?

karimtera:
Yes, it worked.
I found a mistake in some tests, so I am rebasing again.
Sorry for the delay!

karimtera:
It is okay, everything looks fine, but there is some minor issue in `tools/preprocessor/verilog_preprocessor_test.sh`.

Just a new line missing in an expected result of a test, should I commit it to fix this issue?

hzeller:
Yes, and don't forget to push, so that we can run the CI also.

hzeller (commented):
This looks very good; only missing before merge
  * a few places that are not exercised by tests yet in the flow tree
  * PreprocessSingleFile() the superfluous parameters (which are not needed there as it is not directly registered as subcommand)

Everything else, we then can see in the subsequent reviews.

Sorry that it took so many rounds, but that is just happening the first time when contributing to a code-base as there are many assumptions already how things work. Good work!

karimtera:
> This looks very good; only missing before merge
>     * a few places that are not exercised by tests yet in the flow tree

Yes, I added tests for those, but I am curious to know how did you point them out? :) by looking into the coverage report?
I am installed `lcov`, and will try to run it.

> Sorry that it took so many rounds, but that is just happening the first time when contributing to a code-base as there are many assumptions already how things work. Good work!

Thanks for your effort, Henner!
That was a lot of fun actually :)

hzeller:
Yes, I ran

```
MODE=coverage .github/bin/build-and-test.sh //verilog/analysis:flow_tree_test
```

Which then generates some HTML output that we can look at.
Line 65, looks like, is not covered yet.
(and a bunch of error conditions. If they are easy to trigger, it is worthwhile looking at them as well)

verilog/analysis/flow_tree.h
```diff
@@ -0,0 +1,158 @@
...
+  // Receive a complete token sequence of one variant.
+  // variant_sequence: the generated variant token sequence.
+  using VariantReceiver = std::function<bool(const Variant &variant)>;
+
+  explicit FlowTree(verible::TokenSequence source_sequence)
+      : source_sequence_(std::move(source_sequence)){};
+
+  // Generates all possible variants.
+  absl::Status GenerateVariants(const VariantReceiver &receiver);
```
hzeller, line 82: Not for now, but one TODO for future work would be to make this a const function
```
absl::Status GenerateVariants(const VariantReceiver &receiver) const;
```

For that to work, we would have to call `GenerateControlFlowTree()` outside as it modifies the state. You previously had that, but that was at a state when it was not quite clear yet what the final API looks like, so we can revisit.
We discussed before that it might be slightly confusing in the API to have such call, when it is called `GenerateControlFlowTree()` as it might not be quite clear to a reader of the API if it is required.

So we can consider these options: 
   * Do it in the constructor. But it would somewhat violate the [don't do work in the constructor](https://google.github.io/styleguide/cppguide.html#Doing_Work_in_Constructors) guidance, in particular since there can be error conditions that we can't return in the constructor.
   * Require the user to call `GenerateControlFlowTree()`, but since that is unnecessarily obscure to the potential usr of the API, we could just call it `Initialize()` ? That way, for someone reading the API, they know what to do.
   * If the user has not called `Initialize()` have `GenerateVariants()` return a status error, maybe `absl::FailedPreconditionError()`. If Initialize() was called, but had some non-ok status, we can return that. So in general, we can have an `absl::Status initialize_status_` in the `FlowTree` class, that we initialize in the constructor with the failed precondition status, and then again when calling `Initialize()`. Then the only thing `GenerateVariants()` has to do is to `if (!initialize_status_.ok()) return initialize_status_;`

Also, once the initialization question is figured out, we need to have it carry its own state, like `current_variant_` and `wants_more_` so that it does ont modify anything.

But all of that, let's wait until we have the other pull requests done; we always can go back and make things 'prettier'.

verilog/analysis/flow_tree.cc
```diff
@@ -0,0 +1,358 @@
...
+  for (TokenSequenceConstIterator iter = source_sequence_.begin();
+       iter != source_sequence_.end(); iter++) {
+    current_token_enum = iter->token_enum();
+
+    if (IsConditional(iter)) {
+      switch (current_token_enum) {
+        case PP_ifdef: {
+          if_blocks_.push_back(empty_block);
+          if_blocks_.back().if_location = iter;
+          if_blocks_.back().positive_condition = 1;
```
hzeller, line 195: If we have boolean values it is better to use `true` and `false` to initialize.

hzeller (commented):
Just little comments with idea for future improvements.

karimtera:
I found a bug in `FlowTree::AddBlockEdges()` function, the coverage tool is so useful, I will be using it. (y)

hzeller:
Yay. 
Yes, it is very helpful to find all the corner-cases one hasn't thought of.

A word of caution though: It is good to _first_ start out _not_ using the coverage tool but try to think of all possible ways the code should work. And thinking of ways how to break it - is there a corner cases we're unsure of ? Make an attempt to write a test to break things. If yes, it is good and we can improve the code; if not it is also good, because, well not a bug, but also because we made sure that it won't happen in the future.

Then, once we think we have covered everything, the coverage tool will show what is left to do and getting things as close to 100% as possible.

Starting with the coverage tool is not as effective: because then one focuses on just making sure that everything is somehow brushed past, but thinking about the code first and how to break it is helping to get a good understanding of potential issues here and for future code - and improve it accordingly.

The coverage tool mostly provides insight into path coverage while thinking about corner cases is better in providing condition coverage (path coverage and condition coverage are the technical terms for that).

karimtera:
Interesting, because I thought it would be enough to use the tool, but now that you explained the difference between path and condition coverage, I get why we still need the condition coverage, because we may hit a line of code, and still not break it.

hzeller:
Ready to mege, or are there things you'd like to add to this PR ? (the remaining comments I made can be addressed in a separate change)

karimtera:
Ready to merge, we will need to revisit some parts while reviewing the next PRs anyway.

hzeller:
Thanks, merged!
Now next PRs I guess will start out with a rebase ...

hzeller:

PR #1373, opened 2022-08-04
[VeriblePreProcessor][2]: Positional arguments to pass defines and inclusion directories.

karimtera:

**Description:**
This PR implements a class that handles positional arguments for the preprocessor tool.
Any of the following positional argument can be passed one or more times.

| Argument | Description 
| --- | --- |
| <file.sv> | Specifies a SV file to be preprocessed. 
| +define+\<foo>[=\<text>] | Defines one or more macros same as ``` `define foo text``` 
| +incdir+\<dir> | Specifies the directory to search for files included with ``` `include "file_name"```


common/util/cmd_positional_arguments.h
```diff
@@ -0,0 +1,67 @@
...
+  absl::Status SetIncludeDir(absl::string_view include_dir);
+
+  // Sets a define arguments.
+  absl::Status SetDefine(
+      std::pair<absl::string_view, absl::string_view> define);
+
+  // Sets a file argument.
+  absl::Status SetFile(absl::string_view file);
+
+  std::vector<char *>
```
hzeller, line 54: Similar comments like the previous PR: if you put the comment above, then the variable declaration itself can be in one line.

verilog/tools/preprocessor/verilog_preprocessor.cc
```diff
@@ -49,53 +51,172 @@ static absl::Status StripComments(const SubcommandArgsRange& args,
...
-  const SubcommandArgsRange command_args(args.cbegin() + 2, args.cend());
-
-  const auto& sub = commands.GetSubcommandEntry(args[1]);
-  // Run the subcommand.
-  const auto status = sub.main(command_args, std::cin, std::cout, std::cerr);
-  if (!status.ok()) {
-    std::cerr << status.message() << std::endl;
+
+  // Get flags.
+  bool strip_comments_flag = absl::GetFlag(FLAGS_strip_comments);
```
hzeller, line 165: make the values assigned to `const`.

Also in other places. By default, make every variable you assign to and that is not modified later `const`; this will help readers of the code as they have to keep track of less things that are potentially mutable.

common/util/cmd_positional_arguments.cc
```diff
@@ -0,0 +1,128 @@
...
+      continue;
+    }  // all_args_[0] is the tool's name, which can be skipped.
+    if (argument[0] != '+') {  // the argument is a SV file name.
+      if (auto status = SetFile(argument); !status.ok())
+        return status;  // add it to the file arguments.
+    } else {            // it should be either a define or incdir.
+      std::vector<absl::string_view> argument_plus_splitted =
+          absl::StrSplit(argument, absl::ByChar('+'), absl::SkipEmpty());
+      if (argument_plus_splitted.size() < 2) {
+        // Unknown argument.
```
hzeller, line 83: We say the same thing in the next line, so not sure if this comment is helping or distracting.

common/util/cmd_positional_arguments.cc
```diff
@@ -0,0 +1,128 @@
...
+}
+
+// Main function that parses arguments and add them to the correct data memeber.
+absl::Status CmdPositionalArguments::ParseArgs() {
+  // Positional arguments types:
+  // 1) <file>
+  // 2) +define+<foo>[=<value>]
+  // 3) +incdir+<dir>
+
+  int argument_index = 0;
```
hzeller, line 70: What is this argument index used for ?
It looks like you're only using it to skip the first argument, which you can probably do by for-looping over the range with the first excluded:
`verible::make_range(args.begin() + 1, args.end()`

common/util/cmd_positional_arguments.cc
```diff
@@ -0,0 +1,128 @@
...
+    std::pair<absl::string_view, absl::string_view> define) {
+  defines_.push_back(define);
+  return absl::OkStatus();
+}
+
+// Sets a include directory argument.
+absl::Status CmdPositionalArguments::SetIncludeDir(
+    absl::string_view include_dir) {
+  include_dirs_.push_back(include_dir);
+  return absl::OkStatus();
```
hzeller, line 43: It looks like all these 'SetDefine()`, `SetIncludeDir()` and `SetFile()` always return `absl::OkStatus()`.
So we could just leave out these methods and directly call `defines_.push_back(...)` below in the parser. It improves readability as we don't have to do error handling of an error that doesn't happen.

If the argument is "yes, but in the future we might need to return an error", then, yes, let's define these methods when that future has come. Paramout is to keep the code as simple as possible at all times without adding complications that might never be needed.

common/util/cmd_positional_arguments.cc
```diff
@@ -0,0 +1,128 @@
...
+
+// Gets include directories arguments.
+std::vector<absl::string_view> CmdPositionalArguments::GetIncludeDirs() {
+  return include_dirs_;
+}
+
+std::vector<std::pair<absl::string_view, absl::string_view>>
+
+// Gets macro defines arguments.
+CmdPositionalArguments::GetDefines() {
```
hzeller, line 54: The return value and the function header got separate here, which looks very odd. I guess this might've happened due to clang-formatter got confused. If you put the comment before the return value, things should be fine.

Having said this, this simple function is probably best implemented in the header.

common/util/cmd_positional_arguments.h
```diff
@@ -0,0 +1,67 @@
...
+class CmdPositionalArguments {
+ public:
+  explicit CmdPositionalArguments(const std::vector<char *> &args)
+      : all_args_(args){};
+
+  // Main function that parses arguments.
+  absl::Status ParseArgs();
+
+  // Gets include directories arguments.
+  std::vector<absl::string_view> GetIncludeDirs();
```
hzeller, line 35: Given that the GetIncludeDirs(), GetDefines() and GetFiles() only return a member, implementing them directly in the header is probably improving readability. No need to have a multi-line implementation in the *.cc file, when we can write directly here

```c++
const std::vector<absl::string_view>& GetIncludeDirs() const { return defines_; }
```

Also
  * The return value should be a const reference, as we don't want to copy the full vector.
  * The method itself should be const qualified (the `const` at the end), as this does not change the state of the object.

(but also see comment above: in favor of parsing everyhing into a `FileList` struct)

common/util/cmd_positional_arguments.h
```diff
@@ -0,0 +1,67 @@
...
+#include <vector>
+
+#include "absl/status/status.h"
+#include "absl/strings/string_view.h"
+
+namespace verible {
+
+class CmdPositionalArguments {
+ public:
+  explicit CmdPositionalArguments(const std::vector<char *> &args)
```
hzeller, line 28: We need to state the assumptions about the lifetime of the `char*` here. In the parsing stage, we're putting these in string-views, so we need to remind the user that the `char*` arguments they put in here need to outlive the use-time of this `CmdPositionialArguments` object.

(but also see comment above: the `FileList` struct actually copies verything into a string, so possibly not needed)

common/util/cmd_positional_arguments.h
```diff
@@ -0,0 +1,67 @@
...
+
+#include <string>
+#include <vector>
+
+#include "absl/status/status.h"
+#include "absl/strings/string_view.h"
+
+namespace verible {
+
+class CmdPositionalArguments {
```
hzeller, line 26: We already have some functionality that extracts these information from a file, check out this:

https://github.com/chipsalliance/verible/blob/6b2724257d54225491ee48214fa08f4ddfb4d29c/verilog/analysis/verilog_project.h#L322-L340

The struct currently only contains files and incdirs, but we can easily add macro defines.

Maybe we can just move your functionality there and have another toplevel function (not even a class like this); we can call it

```c++
// Parse positional parameters from command line and extract files, +incdir+ and +define+
FileList ParseSourceFileListFromCommandline(std::vector<char *> cmdline);
```

That then will help us to merge the functionality.

hzeller (commented):
I think if we use your code to parse things into a `struct FileList` (and add a container in there for the defines), we have a good starting point to use it for all tools.

common/util/cmd_positional_arguments.h
```diff
@@ -0,0 +1,67 @@
...
+  // Sets a file argument.
+  absl::Status SetFile(absl::string_view file);
+
+  std::vector<char *>
+      all_args_;  // contains all arguments (tool's name is included).
+  std::vector<absl::string_view>
+      include_dirs_;  // contains all arugments that follows +incdir+<dir>.
+  std::vector<absl::string_view>
+      files_;  // contains all SV files passed to the tool.
+  std::vector<std::pair<absl::string_view, absl::string_view>>
```
hzeller, line 60: instead of a `std::pair<>`. make this a struct with

```c++
struct TextMacroDefinition {
   absl::string_view  name;
   absl::string_view value;
}
```
(and this will be a member `std::vector<TextMacroDefinition> macros` in the `struct FileList` then).

Also, given that the other members in FileList actually store `std::string`, we should do the same in `TextMacroDefinition`, thus having the concern about the std::vector<char*>` having to outlive the class eliminated.

Note, we also already have a [`struct MacroDefinition`](https://github.com/chipsalliance/verible/blob/6b2724257d54225491ee48214fa08f4ddfb4d29c/common/text/macro_definition.h#L76) which internally uses Tokens to refer to values.These tokens come from parsed files so it is more complicated. So for now it would be ok to just come up with our own struct and then later at some point consolidate. So when you declare the proposed `struct TextMacroDefinition`, add a comment pointing to `MacroDefinition` as future work.

common/util/cmd_positional_arguments.cc
karimtera: Okay, I get it, I think in this case there won't be any error that will be returned.

verilog/analysis/verilog_project.cc
```diff
@@ -341,4 +341,50 @@ absl::StatusOr<FileList> ParseSourceFileListFromFile(
   return ParseSourceFileList(file_list_file, content);
 }
 
+absl::StatusOr<FileList> ParseSourceFileListFromCommandline(
```
hzeller, line 344: Please also add a test in `verilog_project_test.cc`

verilog/analysis/verilog_project.cc
```diff
@@ -341,4 +341,50 @@ absl::StatusOr<FileList> ParseSourceFileListFromFile(
...
+      // Then "argument" is a SV file name.
+      file_list_out.file_paths.push_back(argument);
+    } else {
+      // it should be either a define or incdir.
+      std::vector<std::string> argument_plus_splitted =
+          absl::StrSplit(argument, absl::ByChar('+'), absl::SkipEmpty());
+      if (argument_plus_splitted.size() < 2) {
+        return absl::InvalidArgumentError("Unkown argument.");
+      }
+      std::string plus_argument_type = argument_plus_splitted[0];
```
hzeller, line 360: std::string_view

verilog/analysis/verilog_project.cc
```diff
@@ -341,4 +341,50 @@ absl::StatusOr<FileList> ParseSourceFileListFromFile(
...
+      if (absl::StrContains(plus_argument_type, "define")) {
```
hzeller, line 361: why 'contains' and not equal ? Otherwise we'd match `+foodefinebar+` which is not really what we want

hzeller (changes requested):
Nice. Make sure to add a unit test. Also, use absl::string_view where possible, avoid std::string unless you actually need a copy of the data.

hzeller:
Here the first start would of course to rebase to current head.

But since there are a lot of changes that you essentially have to integrate, and this one is a somewhat independent (just the commandline changes), it might be easier if you start a fresh pull request: you just copy over the `verilog_project.h`, `verilog_project.cc` and `verilog_project_test.cc` as well as relevant BUILD rules over from here into a fresh fork from head and create a fresh pull request from that (and abandon this)

Up to you.

karimtera:
That would be easier, I am on it.

PR #1375, opened 2022-08-04
[VeriblePreProcessor][4]: Handling file inclusion in "VerilogPreprocess" class.

karimtera:

**Description:**
- Added a feature to `VerilogPreprocess` which allows to store paths,
  That can be used later to find the SV file to include.
- The preprocessor tool takes these paths from the user,
  as a `+incdir+<path>[+<another_path>]` and set them in `VerilogPreprocess`.
- The included files macros and conditionals can be expanded and evaluated.
- Some limitations exists and were written as TODOs in place,
  need to open issues for these.

### DIFF
--- a/verilog/preprocessor/verilog_preprocess.h
+++ b/verilog/preprocessor/verilog_preprocess.h
@@ -74,6 +74,10 @@ struct VerilogPreprocessData {
   verible::TokenStreamView preprocessed_token_stream;
   std::vector<TokenSequence> lexed_macros_backup;
 
+  // A backup memory of pairs sequence, which owns the content text unlike a
+  // TokenInfo or a TokenSequence.
+  std::vector<std::vector<std::pair<int, std::string>>> child_included_content;
+
   // Map of defined macros.
   MacroDefinitionRegistry macro_definitions;
 
@@ -103,6 +107,9 @@ class VerilogPreprocess {
     // want to emit all tokens.
     bool filter_branches = false;
 
+    // Inlude files with `include.
+    bool include_files = false;
+
     // Expand macro definition bodies, this will relexes the macro body.
     bool expand_macros = false;
     // TODO(hzeller): Provide a map of command-line provided +define+'s
@@ -126,6 +133,9 @@ class VerilogPreprocess {
   // Add defines passed to the tool with +define+<foo>[=<value>].
   absl::Status AddDefineFromCmdLine(
       std::pair<absl::string_view, absl::string_view> define);
+
+  // Add search directory paths passed to the tool with +incdir+<path>.
+  absl::Status AddIncludeDirFromCmdLine(absl::string_view include_dir_path);
 
  private:
   using StreamIteratorGenerator =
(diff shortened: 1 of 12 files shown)


hzeller:
Let's await until we're done with the other PRs, to not have too many review comments going on in parallel that might be harder to merge later.
With regard to include files it might be worthwhile looking around what happens in the context of the verilog project supprt `verilog/analysis/verilog_project.h` which is used by the Kythe indexer and the symbol table.

Not everything might apply, but it is good to have a look to have an informed choice knowing what already exists and could be re-used.

verilog/preprocessor/verilog_preprocess.h
```diff
@@ -123,6 +130,13 @@ class VerilogPreprocess {
   // TODO(fangism): ExpandMacro, ExpandMacroCall
   // TODO(b/111544845): ExpandEvalStringLiteral
 
+  // Add defines passed to the tool with +define+<foo>[=<value>].
+  absl::Status AddDefineFromCmdLine(
+      std::pair<absl::string_view, absl::string_view> define);
+
+  // Add search directory paths passed to the tool with +incdir+<path>.
+  absl::Status AddIncludeDirFromCmdLine(absl::string_view include_dir_path);
```
karimtera, line 138: I will start by replacing both of these functions with `SetFileList()`.

verilog/preprocessor/verilog_preprocess.h
```diff
@@ -219,6 +235,9 @@ class VerilogPreprocess {
     bool current_branch_condition_met_;
   };
 
+  // Paths to search into during handling `include
+  std::vector<std::string> search_paths_;
```
karimtera, line 239: This should be the `FileList` containing defines, incdirs and SystemVerilog files.

A question: The SystemVerilog files will not be needed here, is there a way to make use of it? or remove it?

hzeller: We can change everything as needed.
Maybe we have a sub-type in `FileList` that contain preprocessing information.

```c++
struct FileList {
   struct PreprocessingInfo {
         std::vector<std::string> include_dirs;
         // ... macro defintions
   };
   std::vector<std::string> file_paths;
   PreprocessingInfo preprocessing;
};
```

Then use that type with only the relevant parts:
```c++
   FileList::PreprocessingInfo
```

Might need to change all the places the current `FileList` is used, but it is worthwhile to improve clarity.

hzeller:
Do you want to rebase this one or create a fresh new one as it might be easier ?

karimtera:
I will create a fresh branch for this.

verilog/preprocessor/verilog_preprocess.h
karimtera: There is a cycle in dependency problem whenever I try to include `verilog_project.h` in `verilog_preprocess.h`, because the former includes `verilog_analyzer.h`, which also includes `verilog_preprocess.h`. 

Would it be a good idea if i isolates the definition of the `FileList` in a separte file?

hzeller: Sounds good. Put `FileList` in a separate file.

karimtera:
Closed by #1482

Issue #1139, opened 2022-01-13
Reduce boilerplate of VSCode language server integration to essentials


hzeller:
The initial integration of the language server with VSCode was partially auto-generated with tools, and created a lot of boilerplate:
  https://github.com/chipsalliance/verible/tree/master/verilog/tools/ls/vscode
(while all other editors essentially only require 2-3 lines of configuration to start using the Verible Verilog language server)

It would be good to find out what is actually necessary and reduce the boilerplate to the essence needed.

PR #1377, opened 2022-08-04
Canonicalize package.json file

hzeller:


verilog/tools/ls/vscode/package.json
```diff
@@ -65,9 +72,9 @@
         "mocha": "^9.1.1",
         "ts-loader": "^9.2.5",
         "typescript": "^4.4.3",
+        "vsce": "^1.100.1",
         "webpack": "^5.52.1",
-        "webpack-cli": "^4.8.0",
-        "vsce": "^1.100.1"
+        "webpack-cli": "^4.10.0"
```
tgorochowik, line 77: Is the version bump intentional?

tgorochowik (approved):

verilog/tools/ls/vscode/package.json
hzeller: not directly, it was the consequence of running it with current versions (that then also modified this package.json file. So I am not entirely sure if this file is even needed if it is anyway, which is essentially the question in #1139)

PR #1387, opened 2022-08-12
Don't be unnecessarily strict on the bazel version requirements.

hzeller:
It looks like this used to be important, but let's see if the
latest version of Ubuntu used in the CI can deal with it directly.


hzeller:
Ok, looks like bazel is still not a standard installable pacakge in the CI and we need to to it manually.

Issue #1386, opened 2022-08-12
Crash in hierarchy invariant check in token partition tree


hzeller:
**Test case**

Crash with [this input](https://github.com/chipsalliance/verible/blob/13938b495abd6958280c70c3adf4d0a7a9a29597/verilog/formatting/formatter_test.cc#L18375)

This is generated with a fuzzer, so the input itself looks funny, but it shows that it is possible to hit a real bug this way.

Documented in #1385

Result:
```
F20220812 00:05:11.462432    12 token_partition_tree.cc:63] Check failed: parent_begin == children_begin (2 vs. 3) node:
{ (>>>>>>>>[<auto>], policy: append-fitting-sub-partitions) @{0,1,1,0}, (origin: "`c()")
  { (>>>>>>>>[(], policy: fit-else-expand, (origin: "(")) }
  { (>>>>>>>>[)], policy: fit-else-expand, (origin: "")) }
}
```

PR #1399, opened 2022-08-22
Flatten whole subtree in argumentless kMacroCalls.

mglb:
Partial fix of #1386.

The tests that remain commented-out fail with status message:
```
Formatted output is lexically different from the input.    Please file a bug.  Details:
Mismatched token enums.  got: (MacroCallCloseToEndLine) (#711: ")") vs. (')') (#41: ")")
First mismatched token [2]: (MacroCallCloseToEndLine) (#711: ")") vs. (')') (#41: ")")
First mismatched token [2]: (MacroArg) (#710: "`c()\0;") vs. (MacroArg) (#710: "`c();")
```
The first mismatch is caused by existence of two `)` tokens (real `)` and `MacroCallCloseToEndLine`), which are detected differently due to changed formatting.
The second mismatch is caused by removal of `\0` (`\n` in the other test case).

### DIFF
--- a/verilog/formatting/formatter_test.cc
+++ b/verilog/formatting/formatter_test.cc
@@ -18380,19 +18380,24 @@ TEST(FormatterEndToEndTest, FuzzingRegression_UseAfterFree_1384) {
 }
 #endif
 
-#if 0
 // https://github.com/chipsalliance/verible/issues/1386
 TEST(FormatterEndToEndTest, FuzzingRegressionHierachyInvariant) {
-  TestForNonCrash(string_view_from_literal("`c(`c()\0;);"));
-}
-#endif
-
 #if 0
+  // Original sample input from fuzzer
+  TestForNonCrash(string_view_from_literal("`c(`c()\0;);"));
+#endif
+#if 0
+  // Lexically correct variant (`\0` -> `\n`)
+  TestForNonCrash(string_view_from_literal("`c(`c()\n;);"));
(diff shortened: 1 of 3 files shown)


verilog/formatting/formatter_test.cc
```diff
@@ -18346,14 +18347,23 @@ foobar, input    bit [4] foobaz,
   }
 }
 
+// Creates a string_view spanning a whole string literal.
+// Works correctly with strings containing null bytes.
+template <std::size_t N>
+constexpr absl::string_view string_view_from_literal(const char (&s)[N]) {
```
hzeller, line 18353: Nice!

hzeller:
The equivalence check can probably be relaxed in 

https://github.com/chipsalliance/verible/blob/1adb872b2179df99afeab67a3bb0d7cdac9352e4/verilog/analysis/verilog_equivalence.cc#L248

```c++
 return l.text() == r.text();
```

I think it is benign, as we really only want to make sure that the tokens coming out of it represent the same text as coming in. Needs a test though.

(Encountered that recently as well so probably worthwhile addressing. Though can be separate change)

hzeller (approved):
LGTM

hzeller:
A failure due to `\0` not being emitted in the output is probably ok for this fuzzing situation.

We only want to make sure that we do not crash in this situation, but if there are other issues. Sending #1402

PR #1402, opened 2022-08-22
Fuzz tests: only worry about crashes.

### LINKED ISSUE #1386: Crash in hierarchy invariant check in token partition tree

hzeller:
It is ok if our formatter fails for other reasons (that should be
addressed in different tests and functional implementations). Our
main criteria for the fuzzer passing is that we do not crash.

While at it: add another test-case for 1381.

Finalizes and fixes #1386


mglb (approved):
LGTM

PR #1405, opened 2022-08-26
Use newer rules for m4, fixing the glibc 2.34 compile issues.

hzeller:
So now this should compile cleanly there as well.

Fixing #957


hzeller:
Only fail is the Kythe test that started to fail recently, so all good from the perspective of this change.

tgorochowik (approved):

PR #1407, opened 2022-08-26
Add tool for producing Kythe kzips based on System Verilog filelists

ivan444:
Definition of Kythe kzips can be found at https://kythe.io/docs/kythe-kzip.html

This is required for Kythe indexing.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,7 +1,6 @@
 workspace(name = "com_google_verible")
 
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
-load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 
 # Bazel platform rules, needed as dependency to absl.
 http_archive(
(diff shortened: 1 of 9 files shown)


hzeller:
Introducing Python is of course a bit icky because it is hard to maintain.
Some of the functionality (like filelist parsing) we already have in the Verible codebase, but I guess the main motivation here was to have a simple way to zip-up everything ?

hzeller:
(if you haven't already, also make sure to rebase to current head, as I updated a few WORKSPACE entries this morning)

ivan444:
Already rebased.

> Introducing Python is of course a bit icky because it is hard to maintain.

Not the first python code in Verible repo. IMO, of all ecosystems, Python is probably the easiest to maintain.

> Some of the functionality (like filelist parsing) we already have in the Verible codebase, but I guess the main motivation here was to have a simple way to zip-up everything ?

Yep. It was easier to rewrite everything than to fight with adding libzip dependency. We might reconsider if libzip becomes compileable with the setup that we use.

hzeller:
Can you file an issue to include libzip and change this code-base to C++

hzeller (approved):
LGTM
Let's await the CI results.

PR #1408, opened 2022-08-26
Update Kythe version used.

hzeller:

### DIFF
--- a/.github/settings.sh
+++ b/.github/settings.sh
@@ -34,7 +34,4 @@ export BAZEL_OPTS="-c opt --show_progress_rate_limit=10.0"
 # Used to fetch the BAZEL version where needed.
 export BAZEL_VERSION=4.0.0
 
-# TODO(b/171989992): revert to using release version after upgrading system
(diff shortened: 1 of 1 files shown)


hzeller:
Looks like this pulls in a [dependency to rust](https://github.com/chipsalliance/verible/runs/8043721927?check_suite_focus=true#step:4:182) now. More work needed.

hzeller:
@ivan444 do you see how we can reduce the dependencies that we don't actually need in our context ?

ivan444:
This is something for discussion with Kythe team. I'm going to open an issue with them.

I know how to reduce the dependency, but that would imply a surgery of their release. Pretty fragile.

ivan444:
Opened https://github.com/kythe/kythe/issues/5371

BTW. the workaround is not as complicated as I feared. Just fork `extractors.bazelrc`, drop the languages that we don't care about, and edit `--bazelrc` at `.github/bin/run-kythe.sh`. Tested locally.

PR #1413, opened 2022-08-29
Only extract from kythe repo what we need. Re-enable kythe CI.

hzeller:


hzeller:
(not working yet. Converting to draft)

hzeller:
@ivan444 you might be able to see more easily what is wrong here.

PR #1419, opened 2022-09-02
Make InitCommandLine() return vector of string_view instead of char*

hzeller:
For all intendt and purposes, we're only interested in string_view
for the positional parameters and in particular do not want to encourage
modifying them (in particular since all kind of split operations can
cheaply and easily be done in the string_view domain).

The `char*` was just a leaked implementation detail of the
underlying main() entry point which happens to return a writable
buffer.


hzeller:
@karimtera FYI - this will make dealing with the positional parameters more easily in #1373 (or its successor)

PR #1421, opened 2022-09-02
Adding support of +define+ in FileList.

karimtera:
This PR implements: `ParseSourceFileListFromCommandline(const std::vector<absl::string_view>& cmdline_args)`
Which parses the command line arguments into a `FileList` that now contains a vector holding the definitions passed through `+define+`.

Note: `ParseSourceFileListFromCommandline` is not yet used in the preprocessor tool.

### DIFF
--- a/verilog/analysis/verilog_project.cc
+++ b/verilog/analysis/verilog_project.cc
@@ -328,6 +328,7 @@ FileList ParseSourceFileList(absl::string_view file_list_path,
 
     if (absl::StartsWith(file_path, kIncludeDirPrefix)) {
       // Handle includes
+      // TODO(karimtera): split directories by comma, to allow multiple dirs.
       file_list_out.include_dirs.emplace_back(
           absl::StripPrefix(file_path, kIncludeDirPrefix));
     } else {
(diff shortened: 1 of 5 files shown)


verilog/analysis/verilog_project.cc
```diff
@@ -345,4 +346,51 @@ absl::StatusOr<FileList> ParseSourceFileListFromFile(
...
+    if (argument.empty()) continue;
+    if (argument[0] != '+') {
+      // Then "argument" is a SV file name.
+      file_list_out.file_paths.push_back(std::string(argument));
+    } else {
+      // it should be either a define or incdir.
+      std::vector<absl::string_view> argument_plus_splitted =
+          absl::StrSplit(argument, absl::ByChar('+'), absl::SkipEmpty());
+      if (argument_plus_splitted.size() < 2) {
+        return absl::InvalidArgumentError("Unkown argument.");
```
hzeller, line 364: Maybe we can make the message so that the user, who just typed a wrong command line, has some actionable information

```c++
  return absl::InvalidArgumentError(absl::StrCat("expected either '+define+' or '+incdir+', but got '", argument, "'"));
```

verilog/analysis/verilog_project.cc
```diff
@@ -345,4 +346,51 @@ absl::StatusOr<FileList> ParseSourceFileListFromFile(
...
+      } else if (plus_argument_type == "incdir") {
+        // incdir argument.
+        for (const absl::string_view incdir_argument :
+             verible::make_range(argument_plus_splitted.begin() + 1,
+                                 argument_plus_splitted.end())) {
+          // argument_plus_splitted[0] is 'incdir' so it is safe to skip it.
+          file_list_out.include_dirs.push_back(std::string(incdir_argument));
+        }
+      } else {
+        return absl::InvalidArgumentError("Unkown argument.");
```
hzeller, line 389: Similar here: we have all the information to emit a more actionable error message.

verilog/analysis/verilog_project.cc
```diff
@@ -345,4 +346,51 @@ absl::StatusOr<FileList> ParseSourceFileListFromFile(
   return ParseSourceFileList(file_list_file, content);
 }
 
+absl::StatusOr<FileList> ParseSourceFileListFromCommandline(
+    const std::vector<absl::string_view>& cmdline) {
+  FileList file_list_out;
```
hzeller, line 351: I'd probably just call this `result` instead of `file_list_out`. Not only is it shorter, but since it is somewhat common to use, it makes things slightly easier to read.

verilog/analysis/verilog_project_test.cc
```diff
@@ -592,5 +592,37 @@ TEST(VerilogProjectTest, ParseSourceFileList) {
...
+  std::vector<absl::string_view> cmdline;
+  // The first argument will be discarded.
+  cmdline.push_back("the_tool_name");
+  cmdline.push_back("+define+macro1=text1+macro2+macro3=text3");
+  cmdline.push_back("file1");
+  cmdline.push_back("+define+macro4");
+  cmdline.push_back("file2");
+  cmdline.push_back("+incdir+~/path/to/file1+path/to/file2");
+  cmdline.push_back("+incdir+./path/to/file3");
+  cmdline.push_back("+define+macro5");
```
hzeller, line 605: Maybe also a test with an equals sign but nothing following `+define+macro6=`

hzeller (changes requested):
Nice!

verilog/analysis/verilog_project_test.cc
```diff
@@ -592,5 +592,37 @@ TEST(VerilogProjectTest, ParseSourceFileList) {
...
+  macros.push_back(TextMacroDefinition{"macro1", "text1"});
+  macros.push_back(TextMacroDefinition{"macro2", ""});
+  macros.push_back(TextMacroDefinition{"macro3", "text3"});
+  macros.push_back(TextMacroDefinition{"macro4", ""});
+  macros.push_back(TextMacroDefinition{"macro5", ""});
+  EXPECT_THAT(
+      parsed_file_list->defines,
+      ElementsAre(macros[0], macros[1], macros[2], macros[3], macros[4]));
+}
+
```
hzeller, line 626: We should also have a test that triggers the error conditions
  * What happens if there is `+define+` with nothing following
  * Something not `+define+` or `+incdir+`

verilog/analysis/verilog_project.cc
hzeller: Can you double-check how we treat thesse situation in the `FileList` parser from a file ? A defensive strategy might as well be to just treat `argument` as a `file_path` (in case someone has a crazy filename `+foobar/baz.sv`), and then just have further stages that attempt to open the file complain about its non-existence.

Anyway, just make sure that we are consistent with what would happen in the file list parser parsing and command line parsing.

verilog/analysis/verilog_project.h
```diff
@@ -319,6 +319,15 @@ class VerilogProject {
       buffer_to_analyzer_map_;
 };
 
+// TODO(karimtera): Using "MacroDefiniton" struct might be better.
+struct TextMacroDefinition {
+  std::string name;
+  std::string value;
+  bool operator==(const TextMacroDefinition& macro_definition) const {
```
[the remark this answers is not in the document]
karimtera: I had to add it for the tests, Before I added this I got this error:

```log
external/com_google_googletest/googletest/include/gtest/gtest-matchers.h:184:14: error: no match for 'operator==' (operand types are 'const verilog::TextMacroDefinition' and 'const verilog::TextMacroDefinition')
```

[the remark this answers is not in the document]
hzeller: Interesting, well, then leave it in :)

verilog/analysis/verilog_project.cc
karimtera: They are both on the same page, as both don't check if the SystemVerilog file or `incdir` exists, it just adds it.
I would only add defines parsing in `ParseSourceFileList()`.

karimtera: Also judging by the way it looks `ParseSourceFileList()` doesn't support `+incdir+dir1+dir2`, it only removes the prefix.

hzeller: alright, so some TODO to improve the ParseSourceFileList() at some point.

verilog/analysis/verilog_project_test.cc
karimtera: That should be accepted?

hzeller: You can decide if it should be accepted or not at this stage. But if we have a test, then we document that.

karimtera: I added as an invalid syntax for now.

karimtera:
There was an issue reported by @mglb in the original first PR, if the argument is `+define+a=b=c`, the macro value should be `b=c`, I added the fix in here, because I was afraid It will be lost in the first PR.

hzeller:
did you also add a test to confirm that `+define+a=b=c` works ?

karimtera:
Yup, I added a test for that.

verilog/analysis/verilog_project.cc
```diff
@@ -345,4 +347,59 @@ absl::StatusOr<FileList> ParseSourceFileListFromFile(
...
+        absl::StrSplit(argument, absl::ByChar('+'), absl::SkipEmpty());
+    if (argument_plus_splitted.size() < 2) {
+      return absl::InvalidArgumentError(
+          absl::StrCat("ERROR: Expected either '+define+' or '+incdir+' "
+                       "followed by the parameter but got '",
+                       argument, "'"));
+    }
+    absl::string_view plus_argument_type = argument_plus_splitted[0];
+    if (plus_argument_type == "define") {
+      // define argument.
```
hzeller, line 371: To reduce visual noise, I'd probably leave out this comment as it more or less repeats what the code line says.
(similar below for incdir).

hzeller (approved):
LGTM.
Do you want to add anything here, or want the TODO 'send defines to preprocessor'  in follow-up changes ?

karimtera:
> LGTM. Do you want to add anything here, or want the TODO 'send defines to preprocessor' in follow-up changes ?

I've already done the 'send defines to preprocessor' in the the remaining PRs.

hzeller:
Alright, so ready to merge ?

karimtera:
Yes, ready to merge.

hzeller:
Alright, it's in!

karimtera:
I will close PR #1373 then.

hzeller:

PR #1422, opened 2022-09-02
Fix missing dependency, drop header not needed

hzeller:
* Since we're using the flag header, need to also link the
   corresponding library
 * We never use token_stream_adapter, so let's not add
   that dependency.
 * Remove some unnecessary trailing whitespace.


hzeller:
@karimtera I did some tiny changes to tell bazel exactly what is needed.

tgorochowik (approved):

karimtera:
LGTM, Thanks Henner!

PR #1423, opened 2022-09-02
Narrow dependencies to needed headers/libs.

hzeller:


hzeller:
I found some more opportunities to trim headers and libraries to what is minimum needed.
cc: @karimtera

karimtera:
Thanks Henner, LGTM

PR #1424, opened 2022-09-04
Adds the defines passed through +define+ into the preprocessor.

karimtera:
The goal is to register the defines passed to the processor tool through `+define+<macro>[=<value>]` into the preprocessor defines map.

Added the method `VerilogPreprocess::AddDefineFromCmdLine` wihch internally registers defines using the method `VerilogPreprocess::RegisterMacroDefinition` to ensure that everything is aligned, and these passed defines are treated the same way as macros defined by `` `define <macro> <body>``.

Suggestion:
- It might be the time to use `MacroDefinition` instead of `TextMacroDefinition` to save defines in the `FileList` struct.
But, I am not sure about who will own the data itself as `MacroDefinition` only holds Tokens (string-views).

### DIFF
--- a/verilog/preprocessor/verilog_preprocess.cc
+++ b/verilog/preprocessor/verilog_preprocess.cc
@@ -586,6 +586,20 @@ absl::Status VerilogPreprocess::HandleTokenIterator(
   return absl::OkStatus();
 }
 
+// Adds a define passed to the tool with +define+<foo>[=<value>].
+void VerilogPreprocess::AddDefineFromCmdLine(absl::string_view define_name,
+                                             absl::string_view define_body) {
+  // manually create the tokens to save them into a MacroDefinition.
+  verible::TokenInfo macro_directive(PP_define, "`define");
+  verible::TokenInfo macro_name(PP_Identifier, define_name);
+  verible::TokenInfo macro_body(PP_define_body, define_body);
+  verible::MacroDefinition macro_definition(macro_directive, macro_name);
+  macro_definition.SetDefinitionText(macro_body);
+
+  // Registers the macro definition to memeory.
+  RegisterMacroDefinition(macro_definition);
+}
+
 VerilogPreprocessData VerilogPreprocess::ScanStream(
     const TokenStreamView& token_stream) {
   preprocess_data_.preprocessed_token_stream.reserve(token_stream.size());
--- a/verilog/preprocessor/verilog_preprocess.cc
+++ b/verilog/preprocessor/verilog_preprocess.cc
@@ -586,9 +586,9 @@ absl::Status VerilogPreprocess::HandleTokenIterator(
   return absl::OkStatus();
 }
 
-// Adds a define passed to the tool with +define+<foo>[=<value>].
-void VerilogPreprocess::AddDefineFromCmdLine(absl::string_view define_name,
-                                             absl::string_view define_body) {
+// Sets a define passed to the tool with +define+<foo>[=<value>].
+void VerilogPreprocess::SetExternalDefine(absl::string_view define_name,
+                                          absl::string_view define_body) {
   // manually create the tokens to save them into a MacroDefinition.
   verible::TokenInfo macro_directive(PP_define, "`define");
   verible::TokenInfo macro_name(PP_Identifier, define_name);
--- a/verilog/preprocessor/verilog_preprocess.h
+++ b/verilog/preprocessor/verilog_preprocess.h
@@ -125,8 +125,9 @@ class VerilogPreprocess {
 
   // Add defines passed to the tool with +define+<foo>[=<value>].
   // NOTE: Since both of the arguments are string_views, it doesn't own the
-  // referenced memory. The user should ensure that owning strings outlive these
-  // string_views.
+  // referenced memory.
+  // The user should ensure that owning strings outlive the VerilogPreprocess
+  // object.
   void SetExternalDefine(absl::string_view define_name,
                          absl::string_view define_body);
   // TODO(karimtera): It would be better to pass the "FileList" all at once,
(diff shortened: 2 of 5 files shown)


verilog/preprocessor/verilog_preprocess.cc
```diff
@@ -586,6 +586,20 @@ absl::Status VerilogPreprocess::HandleTokenIterator(
   return absl::OkStatus();
 }
 
+// Adds a define passed to the tool with +define+<foo>[=<value>].
+void VerilogPreprocess::AddDefineFromCmdLine(absl::string_view define_name,
```
hzeller, line 590: I'd probably call it 'SetExternalDefine()`

verilog/preprocessor/verilog_preprocess.cc
```diff
@@ -586,6 +586,20 @@ absl::Status VerilogPreprocess::HandleTokenIterator(
   return absl::OkStatus();
 }
 
+// Adds a define passed to the tool with +define+<foo>[=<value>].
```
hzeller, line 589: We need to specify lifetime expectations here in the comment: the string-views need to point to valid memory for the lifetime of this `VerilogPreprocess` object (as we retain just the string_views, so they need to be available).

Later, we for error messages, we might want to be able to point where that was defined, but that can be a TODO for later (ideally, I'd like to have FileList also just contain string_views, because then it can be exactly pin-pointed from where things where coming from)

hzeller (changes requested):
Nice.
I think you'd need to describe the lifetime requirements a bit in the verilog_preprocess.h (it will work in our case but describing it is good).

It might be worthwhile actually considering passing the preprocessing directives directly as the `FileList` object, as it also contains incdirs, which we'll need later. But can be a TODO for later.

One thing missing: a test in `verilog_preprocess_test.cc` for the new function. Even if tested later in the verilog_preprocessor, we always want a test as close as possible to the implementation as this is the place where we look first when making changges.

verilog/preprocessor/verilog_preprocess.h
```diff
@@ -123,6 +123,10 @@ class VerilogPreprocess {
   // TODO(fangism): ExpandMacro, ExpandMacroCall
   // TODO(b/111544845): ExpandEvalStringLiteral
 
+  // Add defines passed to the tool with +define+<foo>[=<value>].
```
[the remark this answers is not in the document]
hzeller: sorry for the repeated comment. I originally wrote it in the CC file, but then moved it to the header file and forgot to remove in cc.

karimtera:
> It might be worthwhile actually considering passing the preprocessing directives directly as the `FileList` object, as it also contains incdirs, which we'll need later. But can be a TODO for later.

The next PR is about the incdirs, I will change it then (y).

verilog/preprocessor/verilog_preprocess_test.cc
```diff
@@ -791,5 +791,25 @@ endmodule
...
+  VerilogPreprocess preprocessor(test_config);
+
+  preprocessor.SetExternalDefine("MACRO1", "VALUE1");
+  preprocessor.SetExternalDefine("MACRO2", "VALUE2");
+
+  const auto& pp_data = preprocessor.ScanStream(test_case_stream_view);
+
+  EXPECT_THAT(pp_data.preprocessed_token_stream[0]->text(), "VALUE1");
+  EXPECT_THAT(pp_data.preprocessed_token_stream[1]->text(), "VALUE2");
+}
```
karimtera, line 812: I couldn't use the testing class in here, because it doesn't use `SetExternalDefine()`, I also didn't want to directly use the lexer, as that would need some extra includes, and would be the same way the preprocessor tool is used.

This test also uses expanding macros feature.

hzeller: So ideally weProbably in one of the next steps in this journey, we need to tell `VerilogAnalyzer` about `FileList`, so that we can inject the necessary information that way.
Make sure to add a TODO here, so that we don't forget to later run everything through `PreprocessorTester`

verilog/tools/preprocessor/verilog_preprocessor_test.sh
```diff
@@ -491,5 +491,74 @@ diff --strip-trailing-cr -u "$MY_EXPECT_FILE" "$MY_OUTPUT_FILE" || {
...
+
+[[ $status == 0 ]] || {
+  "Expected exit code 0, but got $status"
+  exit 0
+}
+
+diff --strip-trailing-cr -u "$MY_EXPECT_FILE" "$MY_OUTPUT_FILE" || {
+  exit 1
+}
+
```
hzeller, line 562: Would we also already get expanded macros at this point ? So if we have an input file

```
`A
```

and preprocess that with `+define+A=foobar`, would we get an output file
```
foobar
```
?

If so: it should be part of the test.
If not yet: it needs to be a TODO.

hzeller (commented):
LGTM, very nice. I think only a few TODO comments we might need.

verilog/tools/preprocessor/verilog_preprocessor_test.sh
karimtera: Yes, it can be expanded, but I currently have the preprocessor tool configured with expand_macros=0.

verilog/preprocessor/verilog_preprocess_test.cc
hzeller: Once we also have ``undef implemented, we can also see if undef properly undefines the value set by SetExternalDefine.

verilog/tools/preprocessor/verilog_preprocessor_test.sh
hzeller: sounds like a command-line flag that wants to be added :)

hzeller:
Alright, looks like ready to merge unless there is more you want to change.
Let's await the CI to finish.

karimtera:
I am okay with it, ready to merge.

hzeller:

Issue #1400, opened 2022-08-22
`std::iterator` deprecation warning with newer compilers


hzeller:
Running with newer compilers reveals a deprecation of  `std::iterator` as base class; context https://gcc.gnu.org/gcc-12/porting_to.html

```
./common/util/expandable_tree_view.h:214:19: warning: 'iterator<std::forward_iterator_tag, const verible::UnwrappedLine, long, const verible::UnwrappedLine *, const verible::UnwrappedLine &>' is deprecated [clang-diagnostic-deprecated-declarations]
    : public std::iterator<std::forward_iterator_tag, const value_type> {
                  ^
verilog/formatting/formatter.cc:574:36: note: in instantiation of member class 'verible::ExpandableTreeView<verible::VectorTree<verible::UnwrappedLine>>::iterator' requested here
      format_tokens_partition_view.begin(), format_tokens_partition_view.end());
                                   ^
/usr/lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/bits/stl_iterator_base_types.h:127:12: note: 'iterator<std::forward_iterator_tag, const verible::UnwrappedLine, long, const verible::UnwrappedLine *, const verible::UnwrappedLine &>' has been explicitly marked deprecated here
    struct _GLIBCXX17_DEPRECATED iterator
           ^
/usr/lib/gcc/x86_64-linux-gnu/12/../../../../include/x86_64-linux-gnu/c++/12/bits/c++config.h:119:34: note: expanded from macro '_GLIBCXX17_DEPRECATED'
# define _GLIBCXX17_DEPRECATED [[__deprecated__]]
```

PR #1427, opened 2022-09-06
Bump glog

sschaetz:
This bumps the glog dependency to the latest version.

It solved the following problem on macOS with strict warning settings:

```bash
ERROR: /private/var/tmp/_bazel_sschaetz/58178f5cbadb135b3179e421ef008716/external/com_github_google_glog/BUILD.bazel:5:13: Compiling src/raw_logging.cc failed: (Aborted): wrapped_clang_pp failed: error executing command external/local_config_cc/wrapped_clang_pp '-D_FORTIFY_SOURCE=1' -fstack-protector -fcolor-diagnostics -Wall -Wthread-safety -Wself-assign -fno-omit-frame-pointer -g0 -O2 -DNDEBUG ... (remaining 61 arguments skipped)

Use --sandbox_debug to see verbose messages from the sandbox
external/com_github_google_glog/src/raw_logging.cc:139:3: error: 'syscall' is deprecated: first deprecated in macOS 10.12 - syscall(2) is unsupported; please switch to a supported interface. For SYS_kdebug_trace use kdebug_signpost(). [-Werror,-Wdeprecated-declarations]
  safe_write(STDERR_FILENO, buffer, strlen(buffer));
  ^
external/com_github_google_glog/src/raw_logging.cc:63:34: note: expanded from macro 'safe_write'
# define safe_write(fd, s, len)  syscall(SYS_write, fd, s, len)
                                 ^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.3.sdk/usr/include/unistd.h:746:6: note: 'syscall' has been explicitly marked deprecated here
int      syscall(int, ...);
         ^
1 error generated.
Error in child process '/usr/bin/xcrun'. 1
```

PR #1428, opened 2022-09-08
Enable C++20 Support

sschaetz:
This change-set enables compilation of all of verible targets with C++20 support. Integration of verible into a C++20 enabled repo motivated this change.

There are 2 main deprecations that this change-set remedies:

1. The deprecation of `std::iterator` (a good summary of what is going on can be found [here](https://www.fluentcpp.com/2018/05/08/std-iterator-deprecated/)) - this prompted the change in `common/util/expandable_tree_view.h`
2. The deprecation of support for aggregate initialization for types with user-declared constructors (prompted by [P1008](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1008r1.pdf)); a good explanation of the situation can be found in [this stackoverflow question](https://stackoverflow.com/questions/57271400/why-does-aggregate-initialization-not-work-anymore-since-c20-if-a-constructor).
  This change-set works around this change by adding constructors to the types in question. I don't think this is the only solution; I believe one could also try to remove the user-declared constructors but I'm not 100% sure this would work for all types.

I'm happy to address any comments including trying to solve (2.) in a different manner if what is proposed here is not a good path forward.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -12,12 +12,12 @@ common:linux --cxxopt=-std=c++17 --host_cxxopt=-std=c++17 --client_env=BAZEL_CXX
 
 # https://github.com/abseil/abseil-cpp/issues/848
 # https://github.com/bazelbuild/bazel/issues/4341#issuecomment-758361769
-common:macos --features=-supports_dynamic_linker --cxxopt=-std=c++17 --host_cxxopt=-std=c++17 --client_env=BAZEL_CXXOPTS=-std=c++17
+common:macos --features=-supports_dynamic_linker --cxxopt=-std=c++20 --host_cxxopt=-std=c++20 --client_env=BAZEL_CXXOPTS=-std=c++20
 
 # Force the use clang-cl on Windows instead of MSVC. MSVC has some issues with
 # the codebase, so we focus the effort for now is to have a Windows Verible
 # compiled with clang-cl before fixing the issues unique to MSVC.
-common:windows --compiler=clang-cl --cxxopt=/std:c++17 --host_cxxopt=/std:c++17 --client_env=BAZEL_CXXOPTS=/std:c++17
+common:windows --compiler=clang-cl --cxxopt=/std:c++20 --host_cxxopt=/std:c++20 --client_env=BAZEL_CXXOPTS=/std:c++20
 
 # Address sanitizer settings.
 build:asan --strip=never
--- a/common/formatting/layout_optimizer_internal.h
+++ b/common/formatting/layout_optimizer_internal.h
@@ -201,6 +201,20 @@ using LayoutTree = VectorTree<LayoutItem>;
 // Single segment of LayoutFunction
 // Maps starting column to a linear cost function and its optimal layout.
 struct LayoutFunctionSegment {
+  LayoutFunctionSegment(
+    int column,
+    const LayoutTree& layout,
+    int span,
+    float intercept,
+    int gradient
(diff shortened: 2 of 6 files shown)


sschaetz:
Mhh, I'm not exactly sure why ClangTidy complains about things I did not change. Could this be the first time ClangTidy runs on these files?

karimtera:
> Mhh, I'm not exactly sure why ClangTidy complains about things I did not change. Could this be the first time ClangTidy runs on these files?

No, it is not the first time.
A lot of errors are in glog, maybe related to #1427, not sure about the other issues.

I am also getting similar issues in vim, that some glog headers are missing.

.bazelrc
```diff
@@ -12,12 +12,12 @@ common:linux --cxxopt=-std=c++17 --host_cxxopt=-std=c++17 --client_env=BAZEL_CXX
 
 # https://github.com/abseil/abseil-cpp/issues/848
 # https://github.com/bazelbuild/bazel/issues/4341#issuecomment-758361769
-common:macos --features=-supports_dynamic_linker --cxxopt=-std=c++17 --host_cxxopt=-std=c++17 --client_env=BAZEL_CXXOPTS=-std=c++17
```
hzeller, line 15: I would _not_ choose c++20 right now here, as we still want to have the code compatible with c++17 for people on older platforms with older compilers.

common/formatting/layout_optimizer_internal.h
```diff
@@ -201,6 +201,14 @@ using LayoutTree = VectorTree<LayoutItem>;
 // Single segment of LayoutFunction
 // Maps starting column to a linear cost function and its optimal layout.
 struct LayoutFunctionSegment {
+  LayoutFunctionSegment(int column, const LayoutTree& layout, int span,
```
hzeller, line 204: If we just remove he constructors below, we should not need this workaround (the constructors below are just default copy and move)

verilog/analysis/symbol_table.h
```diff
@@ -131,6 +131,15 @@ struct ReferenceComponent {
   const SymbolTableNode* resolved_symbol = nullptr;
 
  public:
+  ReferenceComponent(
```
hzeller, line 134: If we just remove he constructors below, we should not need this workaround (the constructors below are just default copy and move)

hzeller (changes requested):
I think instead of adding a constructor, suggest to remove the other ones (copy/move).

That way, we can keep the improved readability with the aggregate construction.

hzeller:
Let's not worry about the clang-tidy problem for now, it indeed seems to be related to glog that has been updated recently.

This PR fixes #1400

hzeller:
I've disabled `clang-tidy` run for now, so if you re-base, this will not be in the way anymore.

hzeller:
Is there anything else you need to proceed @sschaetz ?

hzeller:
(I'll start putting the necessary changes in separate PRs)

sschaetz:
@hzeller apologies, I'm not abandoning this, just got busy with other stuff. Great if you can take it over though!

hzeller:
I started a pull request with the constructor issue #1451, if you want to have a quick review.

I'll then leave the iterator issue to you (good to fix, but we also need to make sure it keeps compiling with C++17)

hzeller:
Since this PR is implemented with two separate PRs #1451 and #1463, I'll close this now.

PR #1430, opened 2022-09-08
Use recommended gflags for glog

hzeller:


hzeller:
ah, that has an undeclared implicit fallthrough. There was probably a reason why I was using the unlabeled newer version before :)

PR #1431, opened 2022-09-08
Initialize positive condition always. Simplify ifdef/ifndef block init

hzeller:
Provide a constructor in ConditionalBlock that enforces setting the whole struct correctly at instruction time. Use that constructor when emplacing_back it into our `if_blocks`.

Since the positive/negative conditional can be expressed as a boolean, use that as an opportunity to simplify the initialization and eliminate duplicate code.


hzeller:
for you to review @karimtera , but will only submit after you've added a test that can verify if this is doing the correct thing (#1432)

karimtera:
LGTM, ready to merge.

Issue #1412, opened 2022-08-29
[kythe] Replace python kzip writer with C++ implementation


ivan444:
Either introduce libzip (if it manage to compile with compilers used by this project) or https://github.com/hzeller/ziplain

PR #1436, opened 2022-09-08
Use protobuf from head as it fixes a warning.

hzeller:


hzeller:
The internal bazel rules don't work in this intermediate version.

hzeller:
It looks like the new protobuf librarry is not compatible with  [`py_proto_library`](https://github.com/chipsalliance/verible/blob/5d59ab431786aa4a0def52edbf183fd20b306838/third_party/proto/kythe/BUILD#L1) anymore.

But this will be a non-issue once #1412 is implemented.

hzeller:
BTW, the warning to eliminate would be the following, showing up in newer compilers with current 21.5 of protobuf

```
In function 'void* memcpy(void*, const void*, size_t)',
    inlined from 'uint8_t* google::protobuf::io::EpsCopyOutputStream::WriteRaw(const void*, int, uint8_t*)' at external/com_google_protobuf/src/google/protobuf/io/coded_stream.h:684:16,
    inlined from 'virtual uint8_t* google::protobuf::internal::ImplicitWeakMessage::_InternalSerialize(uint8_t*, google::protobuf::io::EpsCopyOutputStream*) const' at external/com_google_protobuf/src/google/protobuf/implicit_weak_message.h:103:28,
    inlined from 'bool google::protobuf::MessageLite::SerializePartialToZeroCopyStream(google::protobuf::io::ZeroCopyOutputStream*) const' at external/com_google_protobuf/src/google/protobuf/message_lite.cc:411:30:
/path-to-glibc-2.35-163-dev/include/bits/string_fortified.h:29:33: warning: 'void* __builtin___memcpy_chk(void*, const void*, long unsigned int, long unsigned int)' specified size between 18446744071562067968 and 18446744073709551615 exceeds maximum object size 9223372036854775807 [-Wstringop-overflow=]
```

PR #1440, opened 2022-09-09
Passing external defines and incdirs to the preprocessor.

karimtera:
"VerilogPreprocess":
- Adds "PreprocessingInfo" to storing external defines and incdirs.
- Adds a public function "SetPreprocessingInfo()" to set these info.

The preprocessor tool:
- Modifies "PreprocessSingleFile()" arguments to accept a "PreprocessingInfo", rather than just defines.

Implementation details:
- Needed to move `FileList` to a separate file outside `verilog_project.h` to avoid dependency cycle.


karimtera:
I don't understand why there was a conflict on `preprocessor/BUILD`.

verilog/preprocessor/verilog_preprocess.cc
```diff
@@ -552,6 +553,134 @@ absl::Status VerilogPreprocess::HandleEndif(
   return absl::OkStatus();
 }
 
+// Handle `include directives.
+absl::Status VerilogPreprocess::HandleInclude(
```
hzeller, line 557: Did you have a look how the kythe extractor and/or project_tool (I forgot where ti is implemented) does the include handling ?
Ideally we keep the ownership of the full file in one place (and then the string ownership where the tokens are pointing to is clear).

hzeller (commented):
I think we should move the actual implementation of the include feature to its own pull request (see comment, I think we should see if we can re-use some functionality).

So in this change, just breaking out the FileList into its own file and the corresponding changes. Then we can concentrate on the actual include implementation.

verilog/preprocessor/verilog_preprocess.cc
karimtera: No, I will try to find it.

hzeller:
w.r.t the BUILD conflict, I was running buildifier, so that moved some lines around. 

essentially
```
buildifier $(find . -name BUILD)
```

This was the change https://github.com/chipsalliance/verible/commit/591054191ac660fd1aefa4611db3492d53c80088

hzeller:
LGTM! Thanks, merged.

PR #1443, opened 2022-09-12
Add c++ tool for converting filelist+source to Kythe kzips

ivan444:
Related to #1412 (will close after the cleanup)

It was simpler to implement the CC version than to fight with the side product of `bazel --build_python_zip`...

NOTE: this is not included in the install target yet, as `boringssl` can't be compiled with `-c opt` -- fails with `-Werror=array-bounds`. It's a false warning, and already fixed in the main branch (but not ported to the Bazel branch yet). WIll address that in a followup.


hzeller:
The directory `third_party/` might be hard to import in certain environments, so for that we have `external_libs/`

For the formatting to be ignored in third-party-ish files, I sent #1445
The `kRoot` and stuff constants don't seem to be needed anymore according to the [asan-clang ci](https://github.com/chipsalliance/verible/actions/runs/3037314001/jobs/4889599093#step:7:510), so these can be safely removed.

For the [protobuffer warning w.r.t. unused inline functions](https://github.com/chipsalliance/verible/actions/runs/3037314001/jobs/4889599093#step:7:396), I wonder if attempting to apply #1436 will work (using protobuffer from head), maybe that is fixed upstream ? I originally abandoned that PR, as it wasn't working with the Python protobuffer anymore, but since we now don't need it anymore it might just work.

hzeller:
Alright, #1445 is merged, so if you rebase, that verify formatting issue will go away.

third_party/ziplain/ziplain.cc
```diff
@@ -0,0 +1,247 @@
...
+#include <endian.h>
+#include <string.h>
+#include <zlib.h>
+
+#include <cstdio>
+#include <memory>
+#include <vector>
+
+namespace ziplain {
+ByteSource MemoryByteSource(std::string_view input) {
```
hzeller, line 12: maybe as part of our import efforts to this project, we should `sed 's/std::string_view/absl::string_view/'` ?

hzeller:
To apply the protobuffer patch locally, easiest is to use the patch that github can create conveniently: https://patch-diff.githubusercontent.com/raw/chipsalliance/verible/pull/1436.diff

hzeller:
I think I'd just copy ziplain.{h,cc} into `common/util/` and not leave it as external library (we can probably link to the external lib and say that the author expressed that it is ok to integrate yadayada).

Since I have written that (one of the Verible authors), it can be just integrated with the standard Apache copyright header with `The Verible Authors` with not too much fuss with third_party/ or external_libs/.

(I did the same thing in common/lsp, where I prototyped things in https://github.com/hzeller/bare-lsp and then copied to this project and adapted).

Well and then, we _do_ need to fix the formatting and use absl::string_view :)

WORKSPACE
```diff
@@ -205,3 +207,30 @@ http_archive(
     strip_prefix = "abseil-py-1.2.0",
```
hzeller, line 207: Do we still need absl.py ?

WORKSPACE
```diff
@@ -205,3 +207,30 @@ http_archive(
...
+
+git_repository(
+    name = "boringssl",
+    # Commits must come from the main-with-bazel branch.
+    commit = "b9232f9e27e5668bc0414879dcdedb2a59ea75f2",
+    # Use the GitHub mirror because the official source at
+    # https://boringssl.googlesource.com/boringssl does not allow
+    # unauthenticated git clone and the archives suffer from
+    # https://github.com/google/gitiles/issues/84 preventing the use of
+    # sha256sum on archives.
```
hzeller, line 233: Is this really an issue here ? The generated github zip-archives are guaranteed to generate stable files, so we can rely on them being sha256-able. With that in mind, I prefer if we do it the same way like the other external references in this `WORKSPACE` file.

Also we tend to name the libraries with a more globally unique namespaced form, so instead of `"boringssl"` maybe better `"com_google_boringssl"` ?
```bzl
# For sha256
http_archive(
    name = "com_google_boringssl",
    sha256 = "7128f303ded3a647c4ff08d9396ac86a03f3f2f8ab3080fe8183f9092860736d",
    strip_prefix = "boringssl-b9232f9e27e5668bc0414879dcdedb2a59ea75f2",
    urls = ["https://github.com/google/boringssl/archive/b9232f9e27e5668bc0414879dcdedb2a59ea75f2.zip"],
)
```

third_party/ziplain/ziplain.cc
ivan444: Done

WORKSPACE
ivan444: Done. Thanks! I didn't know Github provides that.

ivan444:
> For the [protobuffer warning w.r.t. unused inline functions](https://github.com/chipsalliance/verible/actions/runs/3037314001/jobs/4889599093#step:7:396), I wonder if attempting to apply #1436 will work (using protobuffer from head), maybe that is fixed upstream ? I originally abandoned that PR, as it wasn't working with the Python protobuffer anymore, but since we now don't need it anymore it might just work.

Let me try that in a followup PR when I remove the python version

common/util/ziplain.h
```diff
@@ -0,0 +1,78 @@
...
+ByteSource MemoryByteSource(absl::string_view input);
+
+// Utility function that reads the content of a file and provides a ByteSource.
+// May return an empty function if file could not be opened
+// (no other errors are reported. If you need error handling, write your own).
+ByteSource FileByteSource(const char *filename);
+
+// Encode a zip file. Call AddFile() 0..n times, then finalize with Finish()
+// No more files can be added after Finish().
+class Encoder {
```
hzeller, line 57: Maybe we should rename it `SimpleZip` or `SimpleZipEncoder` as the original name only really made sense in the namespace it was in `ziplain::Encoder`.
Or nest the namespace: `verible::ziplain::Encoder`. Up to you.

While at it, we probably then also should rename the file to a more boring `simple-zip.{h,cc}` or `ziplain-encoder.{h,cc}` depending on the decision above.

ivan444:
MacOs and win builds are broken. I have a fix, but it's going to wait for tomorrow morning.

common/util/ziplain.cc
```diff
@@ -0,0 +1,265 @@
...
+struct Encoder::Impl {
+  static constexpr int16_t kPkZipVersion = 20;  // 2.0, pretty basic.
+  struct CompressResult {
+    uint32_t input_crc;
+    size_t input_size;
+    size_t output_size;
+  };
+
+  Impl(int compression_level, ByteSink out)
+      : compression_level_(compression_level < 0   ? 0
```
hzeller, line 103: If you add (empty) comments at th end of the line, hopefully the formatter will latch onto the right formatting
```c++
  compression_level_(compression_level < 0 ? 0  //
                      : compression_level > 9  ? 9 //
                                               : compression_level)
```

You can run the formatter locally with `.github/bin/run-clang-format.sh`

.github/bin/run-clang-format.sh
```diff
@@ -16,7 +16,7 @@
 FORMAT_OUT=${TMPDIR:-/tmp}/clang-format-diff.out
 
 # Run on all files.
-find . -name "*.h" -o -name "*.cc" | grep -v third_party/|external_libs/ \
+find . -name "*.h" -o -name "*.cc" | grep -v 'third_party/|external_libs/' \
```
hzeller, line 19: LOL, looks like I didn't know what I was doing.

hzeller (commented):
Ah, the endianness, probably needs `OSSwapHostToLittleInt32()` or something. Sorry, didn't test ziplain on other platforms.

common/util/ziplain.cc
ivan444: Didn't help. I tried many things. But it puzzles me why CI and my local run of the script produce different results.

I'll try again later today.

common/util/ziplain.h
ivan444: Renamed and placed under verible::zip namespace

hzeller (approved):
Nice.

common/util/simple_zip.cc
```diff
@@ -0,0 +1,276 @@
...
+
+ private:
+  const char *begin_;
+  char *pos_;
+};
+
+// Returns the value clamped to the [low, high] interval. If the value is in the
+// interval, returns the value. Otherwise returns low or high depending which
+// one is closer.
+// TODO(ikr): Repace with std::clamp once we move to c++17
```
hzeller, line 99: We do have the c++17 minimum requirement already, so can be done even now.

PR #1447, opened 2022-09-12
Use ABSL_DIE_IF_NULL instead of CHECK_NOTNULL.

hzeller:
The absl form is the newer call.


fangism (approved):

hzeller:
ah, uhm, `ABSL_DIE_IF_NULL` is defined with `ABSL_MUST_USE_RESULT`. Nevermind.

PR #1455, opened 2022-09-15
Updated the Style Guide topic for the forbid-defparam rule

snsokolov:


hzeller:
I think we should work on publishing the style guide where all these topics make sense...

snsokolov:
> I think we should work on publishing the style guide where all these topics make sense...

I agree, but I think we may have a bigger issue here, filed https://github.com/chipsalliance/verible/issues/1462

PR #1459, opened 2022-09-15
Make clang-tidy happy.

hzeller:
Copying the ByteSource and ByteSink std::function objects is less preferred than using const references.

Don't pass member variable as parameter when calling a member function (for Copy.., Compress..).

While at it: use suggested std::clamp as we already require C++17.


hzeller:
Note, clang-tidy is disabled here, but will be ready to be used again when glog is removed ( #1448 )

ivan444 (approved):

PR #1460, opened 2022-09-16
Experiment how the MSVC STL incompatibility with string view manifests.

hzeller:


hzeller:
long story short, the MSVC STL library rivals with the prick in social media comments as in pointing out that 

"well _actually_ you can't expect a string view iterator to cast to `const char*`".

Changing that will be some work and reduce readability in a few places. Maybe something for some cold winter afternoon to work on.

PR #1461, opened 2022-09-16
Add a unit test for simple_zip

hzeller:


ivan444:
The test is OK, but maybe a better approach would be shell-based test with unzip. What do you think?

hzeller:
then we have the ordeal of compiling unzip with baze just to run the test ... something we tried to avoid in the first place.

ivan444:
Shell script test with unzip linux tool. Something safe to assume that
exists/easy to install (except on windows).

hzeller:
Yes, but we also want to distribute it on Windows.
We want to be hermetic, so we don't want to assume anything existing on the target machine in question.

Maybe as manual test ? That should stop bazel from attempting to run it.

hzeller:
In any case, I'd put that in a separate PR, as it sounds like wrangling with potential system specific things.

ivan444:
> Yes, but we also want to distribute it on Windows. We want to be hermetic, so we don't want to assume anything existing on the target machine in question.
> 
> Maybe as manual test ? That should stop bazel from attempting to run it.

We already have a similar thing for kythe verification tests. I'd guess it's easy to exclude the run from windows build.

In any case, this PR LGTM.

ivan444 (approved):

PR #1463, opened 2022-09-16
Remove std::iterator

sschaetz:
This is another part of making the code compatible with C++20 (started #1428 and continued in earnest in #1451). [`std::iterator`](https://en.cppreference.com/w/cpp/iterator/iterator) is deprecated starting from C++20.


hzeller:
Fixes #1400

hzeller:
Nice.
Thanks, merged!

PR #1464, opened 2022-09-16
Provide a FileList::Append() method to concatenate.

hzeller:
This is a preparation to use them in other tools that previously assembled files just from a file list in a file, but now might merge them from that file list and the command line.

### DIFF
--- a/verilog/analysis/verilog_filelist.h
+++ b/verilog/analysis/verilog_filelist.h
@@ -32,20 +32,39 @@ struct TextMacroDefinition {
 };
 
 // File list for compiling a System Verilog project.
+// TODO: ideally, all the strings would be string_views, but we need to make
+//       sure to have the relevant backing store live.
+// TODO: are there cases in which files and incdirs are interleaved so that
+//       the first files don't see all the incdirs yet, but after more incdirs
+//       are added, all of them are relevant ? If so, this would require some
+//       restructering.
+// TODO: document if files are relative to tool invocation or relative to
+//       file_list_path (the latter makes more sense, but I think currently that
+//       is underspecified).
+// TODO: Alongside previous: also introduce file_list_root field ?
 struct FileList {
   // A struct holding information relevant to "VerilogPreprocess" preprocessor.
   struct PreprocessingInfo {
     // Directories where to search for the included files.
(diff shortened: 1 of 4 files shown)


hzeller:
cc: @karimtera to review.

verilog/analysis/verilog_filelist.h
```diff
@@ -32,20 +32,39 @@ struct TextMacroDefinition {
...
+// TODO: ideally, all the strings would be string_views, but we need to make
+//       sure to have the relevant backing store live.
+// TODO: are there cases in which files and incdirs are interleaved so that
+//       the first files don't see all the incdirs yet, but after more incdirs
+//       are added, all of them are relevant ? If so, this would require some
+//       restructering.
+// TODO: document if files are relative to tool invocation or relative to
+//       file_list_path (the latter makes more sense, but I think currently that
+//       is underspecified).
+// TODO: Alongside previous: also introduce file_list_root field ?
```
karimtera, line 44: Would that mean that paths in `file_paths` should exist under `file_list_root`?

karimtera: Also, I don't get the first TODO.
Isn't `file_list_path` a file that contains the arguments (files, defines and incdirs)? why should files passed through the command line be relative to it?

karimtera:
LGTM, I didn't know we needed this, thanks for adding it!

verilog/analysis/verilog_filelist.h
hzeller: That TODO is mostly for me to figure out as I don't really know what happens if file_list_root is not specified.

The list of files in the file-list either are absolute `/path/to/file/...` or relative `relative/to/somedir`, and I suspect file_list_root is what provides the prefix `${file_list_root}/relative/to/some/dir`.
However, in absence of `file_list_root` specifically defined, is that just `$(dirname file_list_path)` ? Probably yes, but since I don't know yet I wrote it as TODO for me to figure out.

From the command line, the situation is more clear: any files are provided relative to the current working directory.

hzeller:
Maybe I change the Append() at some point to the various functions creating FileList just getting that as `FileList *`  argument and append directly. So this might change, but is a good starting point.

PR #1466, opened 2022-09-19
Don't implicitly capture [this] with [=]

hzeller:
In c++20, this will be deprecated.


common/formatting/layout_optimizer.cc
```diff
@@ -520,27 +520,27 @@ LayoutFunction TokenPartitionsLayoutOptimizer::CalculateOptimalLayout(
...
-                         lf = factory_.Indent(this->CalculateOptimalLayout(n),
-                                              relative_indentation);
-                       else
-                         lf = this->CalculateOptimalLayout(n);
-
-                       return lf;
-                     });
+      std::transform(
+          node.Children().begin(), node.Children().end(), layouts.begin(),
+          [this, node, indentation](const TokenPartitionTree& n) {
```
mglb, line 525: Looks like the original `[=]` is a bug. We definitely don't want to copy `TokenPartitionTree` here.
```suggestion
          [this, &node, indentation](const TokenPartitionTree& n) {
```

mglb (changes requested):

PR #1471, opened 2022-09-20
Use bazel query to create a list of targets to index in compilation db.

hzeller:
We can't use bazel query from within bazel, so this file is created in a script separately. The file is checked-in, so whenever set of files change, we could use that in the CI to test if it has been kept up-to-date.


hzeller:
I have no idea why this fails on the CI, I can't reproduce it locally, maybe bazel-export @QuantamHD can help.

All this is doing providing a wider range of files that the compilation database contains for use in `clang-tidy` or `clangd` for instance. But now some tests fail to compile with some bazel failure which looks like some internal disagreement of how a particular rule is built...

hzeller:
Rebased this old PR and played around with it a bit more, but still the same issue. And only on Mac and Windows, so I can't test it locally

  * Mac: https://github.com/chipsalliance/verible/actions/runs/4206710045/jobs/7300556984#step:5:13
  * Windows: https://github.com/chipsalliance/verible/actions/runs/4206710045/jobs/7300558908#step:7:11

It sounds like it there are two internal bazel actions that are conflicting ?

More context: what happens in this change is a new [file with all targets](https://github.com/chipsalliance/verible/blob/f77eaf3ed04b00ee7d7359b8412fdc41666a2047/bazel/all_compdb_rules.bzl) imported in the [toplevel BUILD-file](https://github.com/chipsalliance/verible/pull/1471/files#diff-3053677e4333a96b20ed76c662bca07fb2e1ce059479170139209356787da09c) (the new file is generated with a [bazel query](https://github.com/chipsalliance/verible/blob/f77eaf3ed04b00ee7d7359b8412fdc41666a2047/.github/bin/make-compilation-db.sh#L36-L41)).

All the rules running on Linux seem to work fine (this is all using `bazel 4` to be compatible with the oldest active LTS bazel). Do I need to update some skylark/rules_cc/... update ?

hzeller:
Issue narrowed down and documented in https://github.com/hzeller/bazel-filegroup-issue

hzeller:
Underlying issue: 
https://github.com/bazelbuild/bazel/issues/14294

This is broken since bazel 5.0.0 and no fix yet in sight.

PR #1472, opened 2022-09-20
Prepare VLOG macros to exist at some point.

hzeller:
... only define our own VLOG-ing #ifndef VLOG


fangism (approved):
Hopefully a smooth future transition.

PR #1476, opened 2022-09-21
Move file-list related functions to verilog_filelist.{h,cc}

hzeller:
Since we now have the separate library, let's move also the parsing functions into there. Makes it easier to navigate the project.

No code-changes, just moving things around.


hzeller:
Looks like something you can LGTM @karimtera

karimtera:
LGTM, thanks for the work!

PR #1478, opened 2022-09-21
Change functions parsing file list from return to append.

hzeller:
We want to merge parameters from the command line and files in file-lists and accumulate them into an effective filelist (possibly with multiple `-f` parameters, interspersed with commandline parameters ?).

Change the parse functions to append to an existing file list, to make it easy programmatically to do the accumulation. Remove the FileList::Append() method.


verilog/analysis/verilog_filelist.cc
```diff
@@ -27,24 +27,13 @@
...
+absl::Status AppendFileListFromContent(absl::string_view file_list_path,
+                                       const std::string& file_list_content,
+                                       FileList* append_to) {
   // TODO(hzeller): parse +define+ and stash into preprocessing configuration.
   constexpr absl::string_view kIncludeDirPrefix = "+incdir+";
-  FileList file_list_out;
-  file_list_out.file_list_path = std::string(file_list_path);
-  file_list_out.preprocessing.include_dirs.push_back(".");
+  append_to->file_list_path = std::string(file_list_path);
+  append_to->preprocessing.include_dirs.push_back(".");  // Should we do that?
```
fangism, line 36: Not entirely sure if the standard has an opinion about this, or if it is implementation defined.

hzeller:
Also CC @karimtera FYI

verilog/analysis/verilog_filelist_test.cc
```diff
@@ -89,18 +69,18 @@ TEST(FileListTest, ParseSourceFileListFromCommandline) {
       "file3",
       "+define+macro6=a=b",
       "+incdir+../path/to/file4+./path/to/file5"};
-  auto parsed_file_list = ParseSourceFileListFromCommandline(cmdline);
-  ASSERT_TRUE(parsed_file_list.ok());
+  FileList result;
+  auto status = AppendFileListFromCommandline(cmdline, &result);
+  ASSERT_TRUE(status.ok());
```
fangism, line 74: Maybe print the actual status received when it fails, something like
`ASSERT_TRUE(status.ok()) << "Got status: " << status.message();`

I can't remember of `ASSERT_OK` (and relatives) already do this.

verilog/tools/preprocessor/verilog_preprocessor.cc
```diff
@@ -128,13 +129,14 @@ static absl::Status MultipleCU(const SubcommandArgsRange& args, std::istream&,
                                std::ostream& message_stream) {
   // Parse the arguments into a FileList.
   std::vector<absl::string_view> cmdline_args(args.begin(), args.end());
-  auto parsed_file_list =
-      verilog::ParseSourceFileListFromCommandline(cmdline_args);
-  if (!parsed_file_list.ok()) {
-    return parsed_file_list.status();
+  verilog::FileList file_list;
+  if (auto status =
```
fangism, line 133: I think there's a `RETURN_IF_ERROR` macro for this pattern, if you prefer.

fangism (approved):

verilog/analysis/verilog_filelist.cc
hzeller: At least for the accumulation use-case, it would then probably be something that would be done once at the toplevel.

verilog/analysis/verilog_filelist_test.cc
hzeller: We don't have `ASSERT_OK()` unfortunately in the public gtest. Changing the asserts to print the status.

verilog/tools/preprocessor/verilog_preprocessor.cc
hzeller: Neat, yes, that makes things more readable.

PR #1482, opened 2022-09-23
Handling includes in the preprocessor

karimtera:
### Introduction:
The goal is to handle `` `include "<file>"`` in `VerilogPreprocess`.

### What's done here:
The included files contents is saved in  "TextStructure"s to own the memory, these `TextStructure`s are held inside `VerilogPreprocesData` that way the preprocessor can have `string_view`s pointing to it, without fearing that the memory goes out of scope.

### Future work:
1. The `TextStructure` consists of 3 main parts:
    1. `string owned_contents_` which owns the memory that should be referenced by tokens.
    2. `TokenSequence tokens_` the lexed tokens.
    3. `TokenStreamView tokens_view_` a modified view of `tokens_`, in this case the preprocessed view.

It's designed in a way such that `tokens_view_` should only point to tokens from `tokens_`, and that is not the case if we include contents of another file (another `TextStructure`).

This is considered a limitation, we bypassed this by leaving `tokens_view_` empty, we will only use the first 2 components of a `TextStructure`.

This is fine, for the standalone preprocessor usage, but for example the `VerilogAnalyzer` fill in these `tokens_view_`s, and that causes the internal consistency checks of the `TextStructure` to fail.
This also prevented me from using `PreprocessorTester` as it uses `VerilogAnalyzer`.

2. Another point to discuss is utilizing `VerilogProject::OpenIncludedFile()` which has more advantages over the way we open included files right now, such as avoiding to open the same file multiple times, and have a more clear definition of a compilation unit.
It could be done, but here are some changes that I think need to be done first:
    - Add a member `VerilogProject project_` to `VerilogPreprocess`.
    - Add a constructor to `VerilogPreprocess` to be able to construct `project_` correctly (as a `VerilogProject` can't be assigned, copied, or moved).
    - Modify `VerilogPreprocess::ScanStream()` or replace it with `VerilogPreprocess::ScanProject()`, which should scan all `project_.files_` files.


verilog/preprocessor/verilog_preprocess.cc
```diff
@@ -552,6 +552,99 @@ absl::Status VerilogPreprocess::HandleEndif(
...
+    preprocess_data_.errors.push_back(
+        {**token_iter, "ERROR: passed file can't be open"});
+    return absl::InvalidArgumentError("ERROR: passed file can't be open.");
+  }
+
+  verilog::VerilogPreprocess child_preprocessor(config_);
+  // TODO(karimtera): share the macros defined with child pp.
+  // TODO(karimtera): limit number of nested pps.
+  absl::string_view source_contents_view = source_contents;
+  preprocess_data_.child_included_content.emplace_back(std::make_unique<verible::TextStructure>(source_contents_view));
```
karimtera, line 617: Initially, I wanted `preprocess_data_.child_included_content` to be a vector of `TextStructure`, not `unique_ptr<TextStructure>`, but I got the following error when emplacing back:
```log
error: static assertion failed: result type must be constructible from value type of input range
```
Which I don't understand because a `TextStructure` is constructible from a string_view.

karimtera: Also, this way the `TextStructure` get destructed, and the preprocessor crashes on includes.

hzeller: The reason why you can't have a vector of plain TextStructures is, that the vector internally already allocate the space for a bunch of them, even before the first push_back().
But that means, it needs a default constructor for that to work.

karimtera: This is the only constructor available for the text structure.
 
https://github.com/chipsalliance/verible/blob/882481ad4b0d29d065446f80a219b99299fc7c30/common/text/text_structure.h#L237

hzeller: Yes, and that is intentional. So for you use-case you'd do what you do now, storing raw or unique_ptrs in your vector.

verilog/preprocessor/verilog_preprocess.cc
```diff
@@ -613,13 +613,12 @@ absl::Status VerilogPreprocess::HandleInclude(
...
   // TODO(karimtera): limit number of nested pps.
-  absl::string_view source_contents_view = source_contents;
-  preprocess_data_.child_included_content.emplace_back(std::make_unique<verible::TextStructure>(source_contents_view));
+  preprocess_data_.child_included_content.emplace_back(std::make_unique<verible::TextStructure>(source_contents));
   verible::TextStructure& included_structure = *preprocess_data_.child_included_content.back();
   verible::TokenSequence& included_sequence = included_structure.MutableData().MutableTokenStream();
   verible::TokenStreamView& included_view = included_structure.MutableData().MutableTokenStreamView();
 
-  verilog::VerilogLexer lexer(source_contents);
+  verilog::VerilogLexer lexer(included_structure.Data().Contents());
```
karimtera, line 621: This was the problem, the `TextStructure` mainly contains 3 things:

1. `string owned_contents_` which owns the memory that should be referenced by tokens.
2. `TokenSequence tokens_` the lexed tokens.
3. `TokenStreamView tokens_view_` a modified view of `tokens_`, in this case the preprocessed view.

The problem was that `tokens_` pointing to `source_contents` which goes out of scope after `HandleInclude` finishes.

karimtera:
TODO:

- [ ] Add unit testing in `verilog_preprocess_test.cc`
- [ ] Add the following tests to the preprocessor tool:
    - [ ] Include a non existing file with an absolute path.
    - [ ] Add defines with `+define`, check if it's seen in the included file.
    - [ ] Add incdir, check if it's see in the included file.

verilog/preprocessor/verilog_preprocess.cc
karimtera: Update: 
`TokenStreamView tokens_view_` is left empty, because there is a restriction that it should only refer to tokens of the same text structure.

verilog/tools/preprocessor/verilog_preprocessor_test.sh
```diff
@@ -14,6 +14,18 @@
 # limitations under the License.
 
 # Find input files
+MY_RELATIVE_INCLUDED_FILE_1="myinclude_1.txt"
```
hzeller, line 17: I find it usually more readable to combine the assignment and readonly in one go; it then reads like `const` in C/C++.
```bash
readonly MY_RELATIVE_INCLUDED_FILE_1="myinclude_1.txt"
```

(also good to apply to the other variables in this file where possible).

verilog/tools/preprocessor/verilog_preprocessor_test.sh
```diff
@@ -560,5 +569,241 @@ diff --strip-trailing-cr -u "$MY_EXPECT_FILE" "$MY_OUTPUT_FILE" || {
...
+EOF
+
+cat > "$MY_ABSOLUTE_INCLUDED_FILE_1" <<EOF
+included1_content
+EOF
+
+
+cat > "$MY_EXPECT_FILE" <<EOF
+${MY_INPUT_FILE}:
+(#293: "included1_content")
```
hzeller, line 587: Is the TODO to change token output to actual content output stored in an issue somewhere, so that we don't forget to change that at some point ?

verilog/tools/preprocessor/verilog_preprocessor_test.sh
```diff
@@ -560,5 +569,241 @@ diff --strip-trailing-cr -u "$MY_EXPECT_FILE" "$MY_OUTPUT_FILE" || {
...
+
+EOF
+
+"$preprocessor" multiple-compilation-unit "$MY_INPUT_FILE" > "$MY_OUTPUT_FILE" 2>&1
+
+status="$?"
+
+[[ $status == 0 ]] || {
+  "Expected exit code 0, but got $status"
+  exit 0
```
hzeller, line 598: Wouldn't we want tto `exit 1` here as we want to indicate the failure to match the expectation ?
(also other places below)

verilog/tools/preprocessor/verilog_preprocessor_test.sh
```diff
@@ -560,5 +569,241 @@ diff --strip-trailing-cr -u "$MY_EXPECT_FILE" "$MY_OUTPUT_FILE" || {
...
+EOF
+
+cat > "$MY_ABSOLUTE_INCLUDED_FILE_1" <<EOF
+included1_content
+EOF
+
+
+cat > "$MY_EXPECT_FILE" <<EOF
+${MY_INPUT_FILE}:
+(#293: "included1_content")
```
hzeller, line 620: I think at some point we should emit

```verilog
`line "filename"
```
tokens, for further tools down the line to deal with it properly (I am not sure that verible will deal with it in itself, so that should probably be recorded in some issue).

It essentially tells tools that read that output stream that "_the following content actually comes from this filename_

(This similar to `#line` from the C preprocessor) 
[So at this point, mostly request for issue creating]

hzeller (commented):
Nice. First round of comments.

verilog/tools/preprocessor/verilog_preprocessor_test.sh
karimtera: Yes, it should be the next PR.

verilog/tools/preprocessor/verilog_preprocessor_test.sh
karimtera: Yes, I didn't want to change this, the idea here is that the `"Expected exit..."` will give a `no command found` error, because it's missing `echo`.

verilog/tools/preprocessor/verilog_preprocessor_test.sh
karimtera: I opened this issue #1484 for `` `line`` support.
Do you mean Verible should manually add `` `line`` in files? or we just need to support it if found in source files.

hzeller:
The windows test seems to have a problem https://github.com/chipsalliance/verible/actions/runs/3135473803/jobs/5091188377#step:7:86

Might be because they have funny paths with backslashes and colons in that operating system. Do we trip over any of these ?

verilog/tools/preprocessor/verilog_preprocessor_test.sh
hzeller: Nice thanks! I'll comment on that issue for details that come to mind.

(One unrelated question: how did you make the single back-tick in the `line` line ? I failed to get something like that with the github markdown...)

karimtera: It's `line included between 2 double back-ticks, the thing is there should be a space between the double back-tick and the back-tick of line.

hzeller: `` `test `` :) Ah, thank you!

karimtera:
> The windows test seems to have a problem https://github.com/chipsalliance/verible/actions/runs/3135473803/jobs/5091188377#step:7:86
> 
> Might be because they have funny paths with backslashes and colons in that operating system. Do we trip over any of these ?

Sorry, didn't see your comment earlier.
Yes, it's happening in tests with `` \`include ${file} ``.

How can I test this?

hzeller:
For the private change to work, don't use `std::make_unique<>` but `new` directly (I suspect that would still work with the emplace back).
Then add `VerilogPreprocess` as a friend.
You might need to do something like the following to have TextStructure know about correct namespace:

```c++
namespace verilog {
  class VerilogPreprocess;
} // namespace verilog
//...
namespace verible {
//...
class TextStructure { 
   friend verilog::VerilogPreprocess;
   //...
};
} // namespace verible
```

hzeller:
(if everything else fails, just make the constructors public again and I'll take care of the rest later. This is part of some refactoring effort and more for documentation purposes, so not super-important right now that they are private. Certainly not important enough to use your time)

karimtera:
I would have to do a rebase to see this change at first, is it expected that the build in the ci would have a commit that doesn't exist on my branch (yet)?

hzeller:
oh, yeah if the CI picks up this change but you don't have it in your change yet, you need to rebase.

verilog/preprocessor/verilog_preprocess.cc
```diff
@@ -552,6 +554,115 @@ absl::Status VerilogPreprocess::HandleEndif(
...
+      }
+    }
+    if (!found) {
+      preprocess_data_.errors.push_back({**token_iter, "no such file found."});
+      return absl::InvalidArgumentError("ERROR: included file can't be found.");
+    }
+  } else {
+    if (auto status = verible::file::FileExists(file_path_to_search);
+        !status.ok()) {
+      preprocess_data_.errors.push_back({**token_iter, "no such file found."});
```
hzeller, line 596: the message in the status might already contain the detailed information, so instead of the manual string `"no such file found."`, we could use `status.message()`.

verilog/preprocessor/verilog_preprocess.cc
```diff
@@ -552,6 +554,115 @@ absl::Status VerilogPreprocess::HandleEndif(
...
+  }
+
+  // Creating a new "VerilogPreprocess" object for the included file,
+  // With the same configuration and preprocessing info (defines, incdirs) as
+  // the main one.
+  verilog::VerilogPreprocess child_preprocessor(config_);
+  child_preprocessor.setPreprocessingInfo(preprocess_info_);
+
+  // TODO(karimtera): limit number of nested includes, detect cycles? maybe.
+  preprocess_data_.included_text_structure.emplace_back(
```
hzeller, line 618: Is it planned to have a map of filename -> file content at some point, so that if we encounter the same file multiple times, we only have it in memory once ?

`class VerilogProject` might already do some of this, but if this does not match right now, add a TODO, to merge that functionality (there is the `file_set_type files_` member that contains that info, also used to avoid re-loading the same file multiple times).

verilog/preprocessor/verilog_preprocess.cc
```diff
@@ -552,6 +554,115 @@ absl::Status VerilogPreprocess::HandleEndif(
...
+    return absl::InvalidArgumentError("Expected a path to a SV file.");
+  }
+  // Currently the file path looks like "path", we need to remove "".
+  const auto& token_text = file_token_iter->text();
+
+  std::filesystem::path file_path =
+      std::string(token_text.substr(1, token_text.size() - 2));
+
+  std::string file_path_to_search = file_path.string();
+
```
hzeller, line 577: Opening the include file looks like it is similar to what `VerilogProject` provides in `VerilogProject::OpenIncludedFile()`.

Might be worthwhile documenting the differences here and why we need a separate implementation (or maybe even use some of it directly if you weren't aware of it yet). Documentation then helps later potential unification efforts.

hzeller (changes requested):
Nice.
Can you document the similarities and differences why we couldn't use the `VerilogProject::OpenIncludedFile()`, so that it is easier to later merge the functionality if it can't be done right now ?

When you're adding a test for `HandleInclude()` in `verilog_preprocess_test.cc` you might need to create a few files. You can use `testing::TempDir()` inside the unit test to have a good location where to put a testing file. Also the convenience class `verible::file::testing::ScopedTestFile` is sometimes useful in that context (there are a few examples in the code base)

verilog/preprocessor/verilog_preprocess.cc
karimtera: I looked at it again, I think it can be done, But we will need to change the public API provided by `VerilogPreprocess`, because right now `ScanStream(TokenSequence)` is the only one available, I will add it as a TODO.

verilog/preprocessor/verilog_preprocess.cc
karimtera: I added my thoughts in the description of this PR.

karimtera:
Hello @hzeller, I updated PR's description.
Please let me know what you think of the first point in future work, I think it is a big problem if we can't use `VerilogAnalyzer` with files containing includes.

hzeller (commented):
Can you put the discussion about `VerilogProject::OpenIncludedFile` right into the header documentation about `VerilogProcess::HandleInclude` ? If it is in the code, it is easier to discover as it is hard to find in the description of some pull request later.

hzeller:
In general, it is probably good to consider the next steps (maybe not in this change, we can do that separate) how to integrate the preprocessing into the project, so that it doesn't have to come up with its own version of including files.

That is then also a good starting point to make it easier to have `VerilogAnalayzer` deal with includes.

Currently, there is `TextStructure` that has a view attached to it, but I intend to break this further apart to actually make it easier to have potentially multiple views (e.g. a token stream with just a file or a file with includes), but that also goes one step at a time.

So I think for now, we should document the include handling in the header comment `VerilogProcess::HandleInclude`, then submit this state and then you can work towards integrating it in VerilogProject. Once there, we can then see how we can give Analyze a view that work with a token stream that is composed of files with includes.

verilog/preprocessor/verilog_preprocess.cc
```diff
@@ -552,6 +555,116 @@ absl::Status VerilogPreprocess::HandleEndif(
   return absl::OkStatus();
 }
 
+// Handle `include directives.
+absl::Status VerilogPreprocess::HandleInclude(
+    TokenStreamView::const_iterator iter,
+    const StreamIteratorGenerator& generator) {
+  // TODO(karimtera): how to differentiate between <file> and "file"?? both are
```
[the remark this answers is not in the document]
karimtera: Yeah, right that would be the only way since they are of the same token type.

verilog/preprocessor/verilog_preprocess.cc
```diff
@@ -552,6 +555,116 @@ absl::Status VerilogPreprocess::HandleEndif(
...
+          {**token_iter,
+           absl::StrCat(file_path_to_search, ": no such file found.")});
+      return absl::InvalidArgumentError("ERROR: included file can't be found.");
+    }
+  } else {
+    if (auto status = verible::file::FileExists(file_path_to_search);
+        !status.ok()) {
+      preprocess_data_.errors.push_back(
+          {**token_iter, std::string(status.message())});
+      return absl::InvalidArgumentError("ERROR: included file can't be found.");
```
hzeller, line 601: instead of wrapping the status with a new one with a guess of what went wrong, woul'd it be better (_and_ easier) if we returned the original status that already contained all the most detailed issue ?
(similar in error cases below)

verilog/preprocessor/verilog_preprocess.cc
hzeller: It is best to add it to the declaration or implementation comment, as the PR description can't be easily be read just looking at the code, so the discussion would be lost.

hzeller (approved):
Just two little things left.

hzeller:
Can you run `.github/bin/run-clang-format.sh` to make the CI happy ?

karimtera:
> Can you run `.github/bin/run-clang-format.sh` to make the CI happy ?
Yup, my bad

PR #1483, opened 2022-09-24
Make SyntaxTree{Node,Leaf} final for better documentation.

hzeller:
Having them final makes it clear for the reader that these are the only SyntaxTree classes to worry about when working with these.

### DIFF
--- a/common/text/concrete_syntax_leaf.h
+++ b/common/text/concrete_syntax_leaf.h
@@ -28,7 +28,7 @@
 
 namespace verible {
 
-class SyntaxTreeLeaf : public Symbol {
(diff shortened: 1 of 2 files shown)


hzeller:
This doesn't make any difference code-wise (except that the compiler _might_ be able to do devirtualization in some cases).
This is mostly while thinking about how we can maybe improve the allocation/deallocation of `SyntaxTree{Node,Leaf}`s, as they  make up an substantial amount of runtime (~10% just for delete when using tcmalloc), so I try to think about some way to maybe have pool allocation of sorts for these at some point.

Then it helps to know that the amount of different classes to worry about is two.

fangism (approved):

PR #1485, opened 2022-09-28
verible ls on neovim

sfmth:


verilog/tools/ls/README.md
```diff
@@ -89,14 +89,69 @@ to get it started up on our Verilog/SystemVerilog projects.
...
+```
+Then to install it open neovim and type: ``` :PlugInstall ```
+
+After installing nvim-lspconfig, enable the verible config by appending the following script to your ~/.config/nvim/init.nvim file:
 
 ```lua
--- init.lua
-require'lspconfig'.verible.setup {}
-```
+lua << EOF
```
hzeller, line 104: Can this be trimmed down ?

To help people understand the configuration I think it is important we want to drill down as much as possible to the minimum configuration that is needed for someone to get the LSP going. Otherwise someone who is new to neovim will have a hard time to distinguish what is important ( `require'lspconfig'.verible.setup ...`) and what is noise (such as the keybindings). I suspect that the LSP configuration comes with its own stater set of keybindings anyway ? Or do we have to give the keybindings otherwise the features would not be accessible ?

We then can point them to some URL for fine-tuning keymaps etc. then it is easier to read for users here, as they only get the relevant gist. Also with the external configuration, we don't have to keep track of the neovim features that might improve and change over time.

verilog/tools/ls/README.md
```diff
@@ -89,14 +89,69 @@ to get it started up on our Verilog/SystemVerilog projects.
...
+
+local lsp_flags = {
+  -- This is the default in Nvim 0.7+
+  debounce_text_changes = 150,
+}
 
+require'lspconfig'.verible.setup {
+    on_attach = on_attach,
+    flags = lsp_flags,
+	--root_dir = root_pattern(".git"),
```
hzeller, line 148: What happens if the project does not have a `.git` directory ? Does it fall back to some reasonable default such as the current directory (good) or will it end up in the root directory `/` (bad) ?

verilog/tools/ls/README.md
```diff
@@ -89,14 +89,69 @@ to get it started up on our Verilog/SystemVerilog projects.
 
 ### Neovim
 
-Make sure to have version `0.5.0` or newer and install the [nvim-lpsconfig](https://github.com/neovim/nvim-lspconfig/) plugin.
-Enable the verible config:
+Make sure to have version `0.5.0` or newer and install the [nvim-lspconfig](https://github.com/neovim/nvim-lspconfig/) plugin.
```
hzeller, line 92: The `verible-verilog-ls` binary to call is never mentioned in this configuration. Is it because neovim has already some preconfiguration for Verible ? If so, we should mention it here. It would help understand _why_ we want version `>= 0.5.0`.

verilog/tools/ls/README.md
sfmth: that's just leftover comments that I didn't remove
lines with -- are comments.
it will use the current directory.

verilog/tools/ls/README.md
sfmth: nvim-lspconfig is hosting the verible ls in neovim, it has a default cmd attribute for verible which can be modified if need be

sfmth: I don't really know where that version comes from but it might have something to do with the nvim-lspconfig plugin

hzeller:
Thanks, merged!

sfmth:
Great!

PR #1487, opened 2022-10-04
Make Kythe extraction iterative -- output immediately instead of keep…

ivan444:
…ing the extracted facts in memory

Tested: verification tests + no diffs on a few examples.


hzeller (approved):
Very cool!

PR #1491, opened 2022-10-05
Introduce the concept of MemBlock to prepare for zero-copy file reading.

hzeller:
Right now files are read into a string, then copied into another string in TextStructure. With this abstraction of the block of memory we can allow for two things
  a) read files into the string that is then moved into the
     TextStructure with zero copy.
  b) Allow for mmap()-ing a file for platforms that support it.

This is the first change towards this goal, introducing the MemBlock concept.

### DIFF
--- a/common/strings/BUILD
+++ b/common/strings/BUILD
@@ -114,6 +114,14 @@ cc_test(
 )
 
 cc_library(
+    name = "mem_block",
+    hdrs = ["mem-block.h"],
+    deps = [
+        "@com_google_absl//absl/strings",
+    ],
+)
+
+cc_library(
     name = "naming_utils",
     srcs = ["naming_utils.cc"],
     hdrs = ["naming_utils.h"],
--- a/common/strings/mem-block.h
+++ b/common/strings/mem-block.h
@@ -1,0 +1,59 @@
+// Copyright 2022 The Verible Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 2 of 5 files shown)


common/strings/BUILD
```diff
@@ -113,6 +113,14 @@ cc_test(
     ],
 )
 
+cc_library(
+    name = "mem-block",
```
fangism, line 117: what about `mem_block` for style consistency?

common/strings/mem-block.h
```diff
@@ -0,0 +1,54 @@
...
+ public:
+  virtual ~MemBlock() {}
+  virtual absl::string_view AsStringView() const = 0;
+
+ protected:
+  MemBlock() = default;
+
+ private:
+  MemBlock(const MemBlock &) = delete;
+  MemBlock(MemBlock &&) = delete;
```
fangism, line 36: Recommend deleting the default assignment operators as well.

common/text/text_structure.h
```diff
@@ -259,9 +263,7 @@ class TextStructure {
  protected:
   // This string owns the memory referenced by all substring string_views
   // in this object.
-  // TODO(hzeller): avoid local copy and just require that the string-view
-  // we are called with outlives this object?
-  const std::string owned_contents_;
+  std::unique_ptr<MemBlock> owned_contents_;
```
fangism, line 266: If needed in the future, this could also become a `shared_ptr`.

fangism (approved):

common/strings/mem-block.h
hzeller: Sounds good, done.

PR #1493, opened 2022-10-06
Document uses of TextStructure for easier future refactoring.

hzeller:
No functional change.

Only change: make the constructors for TextStructure private and add all uses within Verible as friend classes. That way it is easy to see where refactoring efforts need to go before we can remove TextStructure.

Improvd documentation of TextStructure and TextStructureView to describe the motivation.

### DIFF
--- a/common/analysis/file_analyzer.h
+++ b/common/analysis/file_analyzer.h
@@ -88,8 +88,7 @@ std::ostream& operator<<(std::ostream&, const RejectedToken&);
 class FileAnalyzer {
  public:
   explicit FileAnalyzer(absl::string_view contents, absl::string_view filename)
-      : text_structure_(std::make_unique<TextStructure>(contents)),
-        filename_(filename) {}
+      : text_structure_(new TextStructure(contents)), filename_(filename) {}
 
   virtual ~FileAnalyzer() {}
 
--- a/common/text/text_structure.h
+++ b/common/text/text_structure.h
@@ -46,8 +46,11 @@ namespace verible {
 class TextStructure;
 
 // TextStructureView contains sequences of tokens and a tree, but all
-// string_views in this structure rely on string memory owned elsewhere,
-// for example in TextStructure.
+// string_views in this structure rely on string memory owned elsewhere.
+//
+// TODO(hzeller): This is a kitchen sink and should be split into multiple
+// aspects; tokens, concrete syntax tree or line number mapping are different
+// aspects not needed everywhere.
 class TextStructureView {
  public:
   // Deferred in-place expansion of the syntax tree.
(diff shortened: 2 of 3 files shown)


common/text/text_structure.h
```diff
@@ -230,16 +233,39 @@ class TextStructureView {
...
+// harder to actually handle memory ownership and views independently. For
+// instance the FileAnalyzer should keep track of the file content itself and
+// then choose to generate a view (or multiple) on top of that (e.g for
+// fallback parsing). Similar for VerilogSourceFile.
+// The language server already has an in-memory representation which is
+// unnecessarily copied into a TextStructure just to do the rest of the
+// analysis, etc.. Long story short: it is beneficial to separate ownership and
+// views.
+//
+// So: this class is eventually to be removed. For now, make all the
```
fangism, line 250: Acknowledged.

common/text/text_structure.h
```diff
@@ -230,16 +233,39 @@ class TextStructureView {
...
+// so a future refactoring is easier.
+// (If, in the meantime, more TextStructure use is needed in other classes,
+// not to worry: just add them here as friend class. This is merely
+// documentation of use currently).
 class TextStructure {
- public:
+ private:
+  friend class FileAnalyzer;
+  friend class TextStructureTokenized;
+  friend class TextStructureViewPublicTest_ExpandSubtreesOneLeaf_Test;
```
fangism, line 260: I've been a fan of `friend`-ing tests, but at the moment don't have another suggestion.

common/analysis/file_analyzer.h
```diff
@@ -88,8 +88,7 @@ std::ostream& operator<<(std::ostream&, const RejectedToken&);
 class FileAnalyzer {
  public:
   explicit FileAnalyzer(absl::string_view contents, absl::string_view filename)
-      : text_structure_(std::make_unique<TextStructure>(contents)),
-        filename_(filename) {}
+      : text_structure_(new TextStructure(contents)), filename_(filename) {}
```
fangism, line 91: I kind of prefer `make_unique` and `make_shared` stylistically, as a way to minimize the direct calls of `operator new` and make it clear on-the-spot what kind of ownership is involved.

common/text/text_structure_test.cc
```diff
@@ -476,7 +475,7 @@ TEST_F(TextStructureViewPublicTest, ExpandSubtreesOneLeaf) {
   const int divide = 3;
   const int new_node_tag = 7;
   std::string subtext(tokens_[0].text().data(), tokens_[0].text().length());
-  auto subanalysis = absl::make_unique<TextStructure>(subtext);
+  std::unique_ptr<TextStructure> subanalysis(new TextStructure(subtext));
```
fangism, line 478: same thought about `operator new` from earlier comment, but I don't mind the type-repetition much.

fangism (approved):

common/analysis/file_analyzer.h
hzeller: Problem is, then I have to give general `make_unique<>` friendship to the constructor; but I wanted to be specific in this case to know exactly who is making use of being friends.

common/text/text_structure_test.cc
hzeller: Same reason as above why I had to change this away from make_unique.

PR #1494, opened 2022-10-06
Make Kythe extraction iterative. File by file. Free the file after it…

ivan444:
… gets processed. The consequence is that now Kythe Anchor owns its content.

### DIFF
--- a/verilog/analysis/verilog_project.cc
+++ b/verilog/analysis/verilog_project.cc
@@ -156,6 +156,23 @@ absl::StatusOr<VerilogSourceFile*> VerilogProject::OpenFile(
   return &file;
 }
 
+bool VerilogProject::RemoveRegisteredFile(
+    absl::string_view referenced_filename) {
+  if (files_.erase(std::string(referenced_filename)) == 1) {
+    LOG(INFO) << "Removed " << referenced_filename << " from the project.";
+    return true;
+  }
+  for (const auto& include_path : include_paths_) {
+    const std::string resolved_filename =
+        verible::file::JoinPath(include_path, referenced_filename);
+    if (files_.erase(resolved_filename) == 1) {
+      LOG(INFO) << "Removed " << resolved_filename << " from the project.";
+      return true;
+    }
+  }
+  return false;
+}
+
 VerilogSourceFile* VerilogProject::LookupRegisteredFileInternal(
     absl::string_view referenced_filename) const {
   const auto opened_file = FindOpenedFile(referenced_filename);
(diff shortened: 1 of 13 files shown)


verilog/analysis/verilog_project.cc
```diff
@@ -156,6 +156,23 @@ absl::StatusOr<VerilogSourceFile*> VerilogProject::OpenFile(
   return &file;
 }
 
+bool VerilogProject::RemoveRegisteredFile(
+    absl::string_view referenced_filename) {
+  if (files_.erase(std::string(referenced_filename)) == 1) {
```
hzeller, line 161: can we change `VerilogProject::file_set_type` to some absl type ? They have heterogeneous lookup, so we can do the lookup with the string_view.

verilog/tools/kythe/indexing_facts_tree.h
```diff
@@ -30,44 +30,34 @@
...
-  // one implicit conversion at most.
-  // e.g. some_function(DisableConversion<absl::string_view>)
-  // accepts only string_views, while disabling string_view's implicit
-  // conversion from std::string (and others).
-  template <class T>
-  struct DisableConversion {
-    /* implicit */ DisableConversion(T v) : value(v) {}  // NOLINT
+// Position of the Anchor in the original text.
+struct AnchorOffset {
+  const int begin;
```
hzeller, line 35: It feels like these want to be `size_t` maybe (I guess it depends a bit what the kythe protocol buffer size of this allows). 

Given that this is representing a begin + length, wouldn't this be, naming-wise, more an `AnchorRange` or `AnchorSpan` ?

It might be worthwhile documenting why this class exists and is not simply a string_view into the larger content (because string might not exist anymore), so capturing the file offset and length is needed.

hzeller (commented):
LGTM. Just some comment in naming (and possibly types inside) `AnchorOffset`.

verilog/analysis/verilog_project.cc
ivan444: Done

verilog/tools/kythe/indexing_facts_tree.h
ivan444: All done. AnchorRange fits much better than Offset. Thanks!

hzeller:
The current compile errors don't seem to be related to the things fixed in #1495

ivan444:
I reverted the commit that's replacing `std::map` with `absl::flat_hash_map` in `VerilogProject`. We store iterators, and that doesn't play well with flat_hash_set (need to check if node_hash_set would work, it keeps the keys stable, but not sure about iterators).

PR #1504, opened 2022-10-12
Fix parsing of 'with' inside streaming operator.

hzeller:
**This is picking up where we left off in #697. Mostly the [strong suggestion](https://github.com/chipsalliance/verible/pull/697#pullrequestreview-610447868) there was to use lexer states which has to be explored here**

The stream expression has an expression followed by an optional 'with' '[' array_range_expression ']'
(LRM 11.4.14)

There are a few other cases where an expression can be trailed with 'with',

with_constraint_block: there we expect an
identifier list in parenthesis after 'with'
similar with array methods with predicate
with '(' ...)
Since the parser with one look-ahead can't see beyond the 'with', it runs into a shift/reduce conflict
as it does not know if '(' or '[' is to follow.

Disambiguate that in the lexer by creating a
separate token TK_with__followed_by_bracket which is a TK_with, where the lexer already looked ahead and knows that it will be followed by '['.

After seeing the '[', everything but the "with" is put back into the stream to be processed separately.

Changed lexer and parser as well as associated unit tests including formatter.

Fixing #693

### DIFF
--- a/verilog/formatting/formatter_test.cc
+++ b/verilog/formatting/formatter_test.cc
@@ -1066,6 +1066,10 @@ static constexpr FormatterTestCase kFormatterTestCases[] = {
         "   parameter  int  b={ << 4  {{ << 2 { a }  }} } ;",
         "parameter int b = {<<4{{<<2{a}}}};\n",
     },
+    {
+        "   parameter  int  b={ << {foo with [ 1 +: 4 ] } } ;",
+        "parameter int b = {<<{foo with [1+:4]}};\n",
+    },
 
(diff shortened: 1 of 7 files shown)


hzeller:
Hey @msierszulski maybe this would be something  interesting for you to pick up ?

I started it a long time ago in #697 but then became busy with other things so I didn't have the time to implement the reviewer-suggested changes.

Since you're currently actively looking at the yacc file in other contexts, maybe this would be something interesting for you to make a fresh pull request out of that implements the reviewer suggestions from back then ?

PR #1505, opened 2022-10-13
Add missing dependencies.

hzeller:


hzeller:
cc: @karimtera

karimtera:
I wonder how could I catch such things, I didn't see a warning while doing a bazel build, does it appear as a warning somewhere else?

hzeller:
The only way currently is to look at all the includes and add the relevant libraries to the BUILD file.

There is a tool `build_cleaner` that we use within Google, but I think it is not released anywhere :( (which is how I discovered it)

PR #1508, opened 2022-10-14
Use tcmalloc as allocator.

hzeller:


hzeller:
(starting out as draft; experimenting if there is any advantage)

hzeller:
Looks like more trouble than worth it. Abandoning.

PR #1509, opened 2022-10-14
absl::make_unique is about to be deprecated. Use std::make_unique directly

hzeller:


hzeller:
Context: cl/481156957

fangism (approved):

Issue #1502, opened 2022-10-11
Separate ownership of file content from analysis results


hzeller:
Currently, we have [`TextStructure`](https://github.com/chipsalliance/verible/blob/42782d267f11a9a754a774f6f3426bf04783bc0c/common/text/text_structure.h#L235-L268), which contains the content of the file and a `TextStructureView` (which essentially holds parse results on top of that).
Then `FileAnalyzer` [inherits from TextStructure](https://github.com/chipsalliance/verible/blob/42782d267f11a9a754a774f6f3426bf04783bc0c/common/analysis/file_analyzer.h#L87) to generate specific analysis results.

This is somewhat rigid, as it means that any ownership of file content within Verible needs to be traced back to some `FileAnalyzer` or at least `TextStructure`.

This is a bit rigid. It couples the ownership of the file content memory tightly with the analyzers but that is not necessarily desirable.

These days, we have more requirements that warrants separating opening files and keeping their content in memory from the analysis steps working on them.
  * There might be multiple analyzers corresponding to a file possibly with multiple configuration options. Right now, this requires multiple copies of the same content.
  * Sometimes, for sub-analysis, we have to copy file contents (as they are rigidly associated with analyzers), instead of just carving out string_views.
  * A preprocessed file might need to open multiple files and then create a uniform token stream on top of that. The same include file might be needed in multiple files. That means there is no 1:1 correspondence anymore between file contents and various views referring to it.
  * The Kythe indexer also possibly opens multiple files with includes, but mostly works file-by-file. Given that we want to index these sequentially (we emit facts as soon as we discover them), we also might need to retire files from memory once not needed anymore (i.e. referenced by any string view in memory). But slightly prefer files that were included to keep in memory longer as we might see them included in later files as well.
  * Other applications, such as the formatter, only need a subset of the analysis (token stream, but not necessarily full parsing), and are strictly work file-by-file and always want to purge files once done.
  * Once file content ownership is nicely stored in [`MemBlock`](https://github.com/chipsalliance/verible/blob/master/common/strings/mem-block.h#L27-L40)s, these can even come from a `mmap()` if supported by operating system and filesystem.
  * _Possibly more requirements; add them to this issue._
  
This issue is the anchor for refacoring the current situation to meet these goals. Possible path:
[...]

PR #1512, opened 2022-10-17
Provide content of file in a MemBlock living in VerilogSourceFile.

hzeller:
To be able to have VerilogSourceFile content be used in contexts other than Analysis (as currently needed with preprocessing), break out the file content to live in a MemBlock in VerilogSourceFile. That way it can be used independently.

Then use that all the way down the VerilogAnalyzer chain for the regular behavior.

Using shared_ptr<MemBlock> for now limits worry about ownership while refactoring is going on: everyone can have their own (referenced counted-) copy of the MemBlock, minimizing the changes needed otherwise. Once the final picture is clear, this might be simplified by having a true central owner of the MemBlock with guaranteed life-time; then handing out Stringviews from there is cheaper.

Issues #1502

### DIFF
--- a/common/analysis/file_analyzer.h
+++ b/common/analysis/file_analyzer.h
@@ -87,7 +87,12 @@ std::ostream& operator<<(std::ostream&, const RejectedToken&);
 // FileAnalyzer holds the results of lexing and parsing.
 class FileAnalyzer {
  public:
-  explicit FileAnalyzer(absl::string_view contents, absl::string_view filename)
+  FileAnalyzer(std::shared_ptr<MemBlock> contents, absl::string_view filename)
+      : text_structure_(new TextStructure(std::move(contents))),
+        filename_(filename) {}
+
+  // Legacy constructor.
+  FileAnalyzer(absl::string_view contents, absl::string_view filename)
       : text_structure_(new TextStructure(contents)), filename_(filename) {}
 
   virtual ~FileAnalyzer() {}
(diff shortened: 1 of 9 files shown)


common/analysis/file_analyzer.h
```diff
@@ -87,7 +87,12 @@ std::ostream& operator<<(std::ostream&, const RejectedToken&);
 // FileAnalyzer holds the results of lexing and parsing.
 class FileAnalyzer {
  public:
-  explicit FileAnalyzer(absl::string_view contents, absl::string_view filename)
+  FileAnalyzer(std::shared_ptr<MemBlock> contents, absl::string_view filename)
+      : text_structure_(new TextStructure(std::move(contents))),
```
fangism, line 91: Is the `move` necessary for `shared_ptr`?

verilog/analysis/verilog_analyzer.h
```diff
@@ -23,6 +23,7 @@
 #include "absl/status/status.h"
 #include "absl/strings/string_view.h"
 #include "common/analysis/file_analyzer.h"
+#include "common/strings/mem-block.h"
```
fangism, line 26: Not in this PR, but perhaps rename file to `mem_block.h` for consistency with the rest of the project?

fangism (approved):

common/analysis/file_analyzer.h
hzeller: clang-tidy will otherwise complain on performance grounds. Because without move, we increase the reference counter when copying, and then decrease it again once the `contents` parameter falls out of scope. With `std::move` we just take the already increased reference counter we get from the per-value parameter.

verilog/analysis/verilog_analyzer.h
hzeller: (putting on TODO list)

fangism (approved):

PR #1514, opened 2022-10-18
Provide a way to read files as MemBlock.

hzeller:
* wire it up to the first user in verilog_project.
 * Provide a memory mapping impelementation of creating a `MemBlock`
 * Rename `mem-block.h` to `mem_block.h` (for consistency)

### DIFF
--- a/common/util/file_util.cc
+++ b/common/util/file_util.cc
@@ -34,6 +34,12 @@
 #include "absl/strings/string_view.h"
 #include "common/util/logging.h"
 
+#ifndef _WIN32
+#include <fcntl.h>
+#include <sys/mman.h>
+#include <sys/stat.h>
+#endif
+
 namespace fs = std::filesystem;
 
 namespace verible {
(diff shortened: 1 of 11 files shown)


hzeller:
current CI issues are because github is phasing out an old ubuntu and wants to indicate that with a planned brown-out.

Addressed separately in #1515

common/util/file_util.cc
```diff
@@ -161,6 +167,62 @@ absl::Status GetContents(absl::string_view filename, std::string *content) {
   return absl::OkStatus();
 }
 
+absl::StatusOr<std::unique_ptr<MemBlock>> AttemptMemMapFile(
+    absl::string_view filename) {
+#ifndef _WIN32
+  class MemMapBlock final : public MemBlock {
```
ivan444, line 173: Move the class to the anonymous namespace?

common/util/file_util.cc
```diff
@@ -161,6 +167,62 @@ absl::Status GetContents(absl::string_view filename, std::string *content) {
...
+   public:
+    MemMapBlock(char *buffer, size_t size) : buffer_(buffer), size_(size) {}
+    ~MemMapBlock() final { munmap(buffer_, size_); }
```
ivan444, line 176: Check return status and log error if != 0

common/util/file_util.cc
```diff
@@ -161,6 +167,62 @@ absl::Status GetContents(absl::string_view filename, std::string *content) {
...
+#endif
+}
+
+absl::StatusOr<std::unique_ptr<MemBlock>> GetContentAsMemBlock(
+    absl::string_view filename) {
+  auto mmap_result = AttemptMemMapFile(filename);
+  if (mmap_result.status().ok()) {
+    return mmap_result;
+  }
+
```
ivan444, line 216: log info, or warning here

ivan444 (approved):

common/util/file_util.cc
hzeller: Added a `DCHECK()` now, so that it is logged.
In general, it is a ¯\_(ツ)_/¯-situation.

common/util/file_util.cc
hzeller: It already is - it is defined inside the `AttemptMemMapFile()` function itself.

common/util/file_util.h
```diff
@@ -65,6 +66,15 @@ absl::Status FileExists(const std::string& filename);
 // Read file "filename" and store its content in "content"
 absl::Status GetContents(absl::string_view filename, std::string* content);
 
+// Attempt to memory map file and return as MemBlock it it is possible.
+absl::StatusOr<std::unique_ptr<MemBlock>> AttemptMemMapFile(
+    absl::string_view filename);
```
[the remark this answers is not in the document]
hzeller: AttemptMemMapFile() was only really exported to be able to see it in the test few more error conditions there. But you're right, we should steer users to only consider `GetContentAsMemBlock()`

common/util/file_util.cc
hzeller: There are a few valid reasons tht this wouldn't work, e.g. reading from a pipe of sorts, so it is an expected failure to not be able to `mmap()` something. Logging that would be confusing.

common/util/file_util.cc
hzeller: Removed DCHECK() again. apparently our DCHECK() implementation does then _not_ call munmap.
(confirmed with `strace -emmap,munmap bazel-bin/common/util/file_util_test`)

PR #1516, opened 2022-10-19
Towards using VerilogProject in the preprocessor

karimtera:


verilog/preprocessor/verilog_preprocess.cc
```diff
@@ -587,48 +601,18 @@ absl::Status VerilogPreprocess::HandleInclude(
...
-      return status;
-    }
-  }
-
-  // Actually open the file.
-  std::string source_contents;
-  if (auto status =
-          verible::file::GetContents(file_path_to_search, &source_contents);
-      !status.ok()) {
+  const auto status_or_file = project_->OpenIncludedFile(file_path.string());
```
hzeller, line 604: Intead of using `VerilogProject` (which results in all the cyclic dependencies) maybe you can pass in a 
`std::function<absl::StatusOr<VerilogSourceFile*>(absl::string_view filename)> file_pener', which you then call here.

Whoever calls HandleInclude() then has to pass the function that can be a wrap around VerilogProject.

(Of course, longer term, we don't really want a `VerilogSourceFile` I suspect, but just the underlying file `std::shared_ptr<MemBlock>` ?)

karimtera: I tried the following:
```c++
  // verilog_preprocess.h
  using FileOpener = std::function<absl::StatusOr<VerilogSourceFile>(absl::string_view)>;
  FileOpener file_opener_ = nullptr;
```


```c++
  // verilog_preprocess.cc
  const auto status_or_file = file_opener_(file_path.string());
  /* const auto status_or_file = project_->OpenIncludedFile(file_path.string()); */
  if (!status_or_file.ok()) {
    preprocess_data_.errors.push_back(
        {**token_iter, std::string(status_or_file.status().message())});
    return status_or_file.status();
  }
  const auto source_contents = status_or_file->GetContent()->AsStringView();
```
To get the `source_contents` I would need to call `GetContent()` from `VerilogSourceFile`, that's why I should keep `#include "verilog_project.c"` in `verilog_preprocess.cc`, so the issue remains.

Am I missing something?

hzeller: what if your file_opener returns an `absl::string_view`, not a super-specific `VerilogSourceFile` ?

```
using FileOpener = std::function<absl::StatusOr<absl::string_view>(absl::string_view filename)>;
```

And an implementation can be roughly something like this:
```
 FileOpener my_fun = [project](absl::string_view filename) -> absl::StatusOr<absl::string_view {
    auto result = project->OpenTheFile(filename);
    if (!result.status().ok()) return result.status();
    return result->GetContent();
 }
```

karimtera: I think that's also won't fit in, because the implementation is needed inside `HandleInclude()` to open the files and there can't be `VerilogProject project` in there to capture it in `my_fun`, right?

karimtera: Yes, but in this case the caller is HandleInclude() itself.

karimtera: It will be the caller relative to the new VerilogPreprocess which will scan the contents of the included file.

verilog/tools/preprocessor/verilog_preprocessor.cc
```diff
@@ -91,7 +94,17 @@ static absl::Status PreprocessSingleFile(
...
-  verilog::VerilogPreprocess preprocessor(config);
+
+  verilog::VerilogProject project(".", preprocessing_info.include_dirs);
+
+  FileOpener file_opener =
+      [&project](
+          absl::string_view filename) -> absl::StatusOr<absl::string_view> {
+    auto result = project.OpenIncludedFile(filename);
+    if (!result.status().ok()) return result.status();
+    return (*result)->GetContent()->AsStringView();
```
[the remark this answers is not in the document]
karimtera: Yes, done (y)

verilog/preprocessor/verilog_preprocess_test.cc
```diff
@@ -892,7 +895,18 @@ TEST(VerilogPreprocessTest, IncludingFileWithAbsolutePath) {
   const std::string equivalent_content =
       "module included_file(); endmodule\nmodule src(); endmodule\n";
 
-  VerilogPreprocess tester(VerilogPreprocess::Config({.include_files = true}));
+  // TODO(karimtera): allow including files with absolute paths.
+  // This is a hacky solution for now.
+  verilog::VerilogProject project(".", {"/"});
+  FileOpener file_opener =
```
hzeller, line 901: The nice thing with the abstraction of the file opening is, that you can also use that to simplify unit tests:
you could have a very simple implementation of the FileOpener without need for a VerilogProject; something along the line of

```c++
file_opener = [](absl::string_view filename) -> absl::StatusOr<absl::string_view> {
   if (filename == "foo") return "content of foo file";
   if (filename == "bar") return "content of bar file";
   return NotFoundError("file does not exist");
}
```

That way you can more fine-grained test things and don't even need to create a file in the filesystem.

verilog/preprocessor/verilog_preprocess.h
```diff
@@ -53,6 +53,8 @@
 
 namespace verilog {
 
+class VerilogProject;
```
hzeller, line 56: Do you still need this forward declare ?

verilog/preprocessor/verilog_preprocess_test.cc
```diff
@@ -40,6 +41,8 @@ using verible::container::FindOrNull;
 using verible::file::CreateDir;
 using verible::file::JoinPath;
 using verible::file::testing::ScopedTestFile;
+using FileOpener =
```
hzeller, line 44: To reduce repetition (anda void having to change everything whenever the FileOpener function changes), I suggest to just refer to the type in the `VerilogProcess`

```c++
using FileOpener = VerilogPreprocess::FileOpener;
```

verilog/preprocessor/verilog_preprocess.h
```diff
@@ -242,6 +247,10 @@ class VerilogPreprocess {
 
   // Defines and incdirs Information passed externally.
   FileList::PreprocessingInfo preprocess_info_;
+
+  // A pointer to a file opener function.
+  // This is needed for opening new files while handling includes.
+  FileOpener const file_opener_ = nullptr;
```
hzeller, line 253: for consistency, I'd put the `const` to the front.

verilog/preprocessor/verilog_preprocess.cc
```diff
@@ -53,6 +54,16 @@ VerilogPreprocess::VerilogPreprocess(const Config& config) : config_(config) {
       BranchBlock(true, true, verible::TokenInfo::EOFToken()));
 }
```
hzeller, line 55: Instead of repeating the constructor init and body, could you just use a delegating constructor that calls the other one ?

```c++
VerilogPreprocess::VerilogPreprocess(const Config& config) : VerilogPreprocess(config, nullptr) {}
```

verilog/preprocessor/verilog_preprocess.cc
```diff
@@ -587,48 +598,19 @@ absl::Status VerilogPreprocess::HandleInclude(
...
-    }
-  }
-
-  // Actually open the file.
-  std::string source_contents;
-  if (auto status =
-          verible::file::GetContents(file_path_to_search, &source_contents);
-      !status.ok()) {
+  // Use the provided FileOpener to open the included file.
+  const auto status_or_file = file_opener_(file_path.string());
```
hzeller, line 602: should we check if `file_opener_` is not nullptr here ?
So early in the `HandleInclude()` 

```c++
  if (!file_opener_) return FailedPreconditionError("file_opener_ not defined");
```

hzeller (commented):
Looks pretty good!
Some suggestions for simplificdation of unit tests, less typing for using declarations and avoiding potential problem when file opener is not provided.

verilog/tools/preprocessor/verilog_preprocessor.cc
```diff
@@ -91,7 +94,17 @@ static absl::Status PreprocessSingleFile(
   config.filter_branches = true;
   config.include_files = true;
   // config.expand_macros=1;
-  verilog::VerilogPreprocess preprocessor(config);
+
+  verilog::VerilogProject project(".", preprocessing_info.include_dirs);
+
+  FileOpener file_opener =
+      [&project](
+          absl::string_view filename) -> absl::StatusOr<absl::string_view> {
```
hzeller, line 102: something to experiment with: Maybe we can even leave out the `-> absl::StatusOr<absl::string_view>` part. Ideally, the compiler will be able to deduce the return type to look at the left hand side of the expression.

(though I would also not be surprised if it can't.)

verilog/preprocessor/verilog_preprocess.cc
```diff
@@ -587,48 +598,19 @@ absl::Status VerilogPreprocess::HandleInclude(
...
+  if (!status_or_file.ok()) {
     preprocess_data_.errors.push_back(
-        {**token_iter, std::string(status.message())});
-    return status;
+        {**token_iter, std::string(status_or_file.status().message())});
+    return status_or_file.status();
   }
+  const auto source_contents = *status_or_file;
```
hzeller, line 608: Instead of `auto`, I'd probably be specific here with the type, as it might be important for the reader to know that this is a `absl::string_view`.

verilog/preprocessor/verilog_preprocess.cc
```diff
@@ -587,48 +598,19 @@ absl::Status VerilogPreprocess::HandleInclude(
...
 
   // Creating a new "VerilogPreprocess" object for the included file,
   // With the same configuration and preprocessing info (defines, incdirs) as
   // the main one.
-  verilog::VerilogPreprocess child_preprocessor(config_);
+  verilog::VerilogPreprocess child_preprocessor(config_, file_opener_);
   child_preprocessor.setPreprocessingInfo(preprocess_info_);
 
   // TODO(karimtera): limit number of nested includes, detect cycles? maybe.
```
hzeller, line 616: Below we should add a TODO that we we ideally get a MemBlock from the file opener, so that we don't have to create a second copy
(because the file opener now returns a string-view which then is copied again into another string in TextStructure).

TODO is ok for now, as I will have to refactor a little other stuff first before this is possible

verilog/tools/preprocessor/verilog_preprocessor.cc
karimtera: I thought the compiler would be able to deduce the type, but it couldn't, so I am leaving it as it is.

verilog/preprocessor/verilog_preprocess_test.cc
karimtera: Yes, It's a good idea, I changed it in this test, but in other tests I wanted to keep the project to test the include directories with relative paths.

hzeller (approved):
Nice!

PR #1521, opened 2022-10-20
Fix some clang-tidy complaints.

hzeller:


hzeller:
This were findings with a newer clang-tidy version we run in the CI.

Issue #1484, opened 2022-09-27
Support `line compiler directive in Verible


karimtera:
### Issue
Verible doesn't support `` `line`` directive.

#### Function
The `` `line`` compiler directive can be used to specify the original source code line number and file name.
This allows the location in the original file to be maintained if another process modifies the source.

#### Syntax
`` `line <number> "<file_name>" <level> ``

- The `<number>` parameter shall be a positive integer that specifies the new line number of the following text line.
- The `<file_name>` parameter shall be a string literal that is treated as the new name of the file, it can also be a full or relative path name. 
- The `<level>` parameter shall be 0, 1, or 2. The value 1 indicates that the following line is the first line after an include file has been entered. The value 2 indicates that the following line is the first line after an include file has been exited. The value 0 indicates any other line.

#### Notes

- Only white space may appear on the same line as the `` `line`` directive. 
- Comments are not allowed on the same line as a`line directive. 
- All parameters in the `` `line`` directive are required.

#### Example 
```systemverilog
`line 3 "orig.v" 2 
// This line is line 3 of orig.v after exiting include file
```

- The line number shall be incremented as each line is read, and the name shall be updated to the new current file being processed. 
- The line number shall be reset to 1 at the beginning of each file. 
- When beginning to read include files, the current line and file name shall be stored for restoration at the termination of the include file.

PR #1528, opened 2022-10-27
Support whitespaces in VerilogPreprocess

karimtera:
### Description
The goal is to support white-spaces in the prepreprocessor.
It's done by discarding some white-spaces (will not be able to regenerate them).
In `VerilogPreprocess` sometimes a token iterator is incremented to look for a macro name, or a include file name, in these situations we will not be able to retrieve the white-spaces.

### Standalone preprocessor
The support for white-spaces should be added to the  preprocessor standalone tool in all modes:
- [X] Multiple compilation units mode.
- [ ] Generate variants. (`flow_tree` need some work)

A discussion about this problem is at #1358

### DIFF
--- a/verilog/preprocessor/verilog_preprocess.cc
+++ b/verilog/preprocessor/verilog_preprocess.cc
@@ -58,10 +58,21 @@ VerilogPreprocess::VerilogPreprocess(const Config& config, FileOpener opener)
       BranchBlock(true, true, verible::TokenInfo::EOFToken()));
 }
 
+TokenStreamView::const_iterator VerilogPreprocess::GenerateBypassWhiteSpaces(
+    const StreamIteratorGenerator& generator) {
+  auto iterator =
+      generator();  // iterator should be pointing to a non-whitespace token;
+  while (verilog::VerilogLexer::KeepSyntaxTreeTokens(**iterator) == 0) {
+    iterator = generator();
+  }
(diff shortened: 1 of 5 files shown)


karimtera:
A current problem, consider this SV file:
```verilog
`define X value
`ifdef X
other_content
```
The output:
```verilog
`defineXvalue

other_content
```

The white-spaces in the macro definition are lost during `GenerateBypassWhiteSpaces()`.
I guess anyway the preprocessed content shouldn't contain preprocessor directives, right? but, maybe it's needed by other tools?
Also about the blank lines, I think maybe we can add a flag to either keep them or not, Verilator have a flag for this.

hzeller:
Another question: what happens if the input file has multiple newlines (or, a comment with multiple newlines).
Does the output preserve the additional newlines ?
Otherwise, we'll have a problem with line-number reporting further down (until we implement the `` `line`` feature (#1484)).

karimtera:
These white-spaces are the only ones that are lost:
- In macros definitions: 
```verilog
`define     X    1
// equivalent to
`defineX1
```
- In macro calls, between macro parameters. (won't matter in case `config.expand_macro = true`).
- In including files, between `` `include`` and the file path.

Other than those mentioned, everything should be preserved.

hzeller (approved):
LGTM

PR #1530, opened 2022-11-15
[LSP] Add support for providing definitions from the SymbolTable

glatosinski:
This PR is Work in Progress.

It adds support for creating a [SymbolTable](https://github.com/chipsalliance/verible/blob/master/verilog/analysis/symbol_table.cc) in the [Language Server](https://github.com/chipsalliance/verible/blob/master/verilog/tools/ls/verilog_ls.cc) and utilizing it for finding definitions.

The current status is:

- [X] Created an initial class that interfaces the symbol table with the LSP requests
- [X] Added parsing and reading the workspace directory from `textDocument/initialize` request
- [X] Added creating VerilogProject object used by symbol table
- [X] Added finding symbol based on `textDocument/definition` and current state of the buffers from LS client
- [X] Added filling the symbol table based on opened files
- [x] Added matching symbol from request to definition from SymbolTable
- [x] Added initial `textDocument/definition` support
- [X] Added providing project files via `verible.filelist` file, parsed by [verilog::FileList from verilog_filelist](https://github.com/chipsalliance/verible/blob/master/verilog/analysis/verilog_filelist.h)
- [X] Added integration for SymbolTable with LS API (to look for symbols in opened buffers)
- [X] Added providing additional include paths to the VerilogProject - it is done via `verible.filelist` file
- [X] Added creating complete SymbolTable for project and its includes - initial support is ready
- [x] Added taking symbol's scope into consideration when looking for definition
- [x] Implement tests for `SymbolTableHandler`
- [x] Implement tests for `VerilogLanguageServer` regarding `SymbolTableHandler` support

Currently, the workspace directory provided by the editor is used as VerilogProject root (however, emacs uses git repository root directory as the workspace directory, if available).

### DIFF
--- a/verilog/analysis/verilog_project.cc
+++ b/verilog/analysis/verilog_project.cc
@@ -197,7 +197,7 @@ std::string VerilogProject::GetRelativePathToSource(
   return relative.string();
 }
 
-absl::Status VerilogProject::UpdateFileContents(
+void VerilogProject::UpdateFileContents(
     absl::string_view path, const verible::TextStructureView* updatedtext) {
   std::string projectpath = GetRelativePathToSource(path);
   auto fileptr = files_.find(projectpath);
@@ -211,7 +211,6 @@ absl::Status VerilogProject::UpdateFileContents(
         projectpath, path, updatedtext,
         /*corpus=*/"");
   }
-  return absl::OkStatus();
 }
 
 VerilogSourceFile* VerilogProject::LookupRegisteredFileInternal(
--- a/verilog/analysis/verilog_project.cc
+++ b/verilog/analysis/verilog_project.cc
@@ -200,16 +200,17 @@ std::string VerilogProject::GetRelativePathToSource(
 void VerilogProject::UpdateFileContents(
     absl::string_view path, const verible::TextStructureView* updatedtext) {
   std::string projectpath = GetRelativePathToSource(path);
+  std::unique_ptr<VerilogSourceFile> contents = nullptr;
+  if (updatedtext)
+    contents = std::make_unique<ParsedVerilogSourceFile>(
+        projectpath, path, updatedtext, /*corpus=*/"");
+  else
+    contents = std::make_unique<VerilogSourceFile>(projectpath, path, "");
   auto fileptr = files_.find(projectpath);
   if (fileptr == files_.end()) {
-    files_.insert(
-        std::make_pair(projectpath, std::make_unique<ParsedVerilogSourceFile>(
-                                        projectpath, path, updatedtext,
-                                        /*corpus=*/"")));
+    files_.insert(std::make_pair(projectpath, std::move(contents)));
   } else {
-    fileptr->second = std::make_unique<ParsedVerilogSourceFile>(
-        projectpath, path, updatedtext,
-        /*corpus=*/"");
+    fileptr->second = std::move(contents);
   }
 }
 
--- a/verilog/analysis/verilog_project.cc
+++ b/verilog/analysis/verilog_project.cc
@@ -197,7 +197,7 @@ std::string VerilogProject::GetRelativePathToSource(
   return relative.string();
 }
 
-absl::Status VerilogProject::updateFileContents(
+absl::Status VerilogProject::UpdateFileContents(
     absl::string_view path, const verible::TextStructureView* updatedtext) {
   std::string projectpath = GetRelativePathToSource(path);
   auto fileptr = files_.find(projectpath);
--- a/verilog/analysis/verilog_project.cc
+++ b/verilog/analysis/verilog_project.cc
@@ -187,20 +187,30 @@ bool VerilogProject::RemoveRegisteredFile(
 }
 
 std::string VerilogProject::GetRelativePathToSource(
-    const std::string& absolute_filepath) {
+    const absl::string_view absolute_filepath) {
   // TODO add check if the absolute_filepath is out of the VerilogProject
-  std::filesystem::path absolutepath{absolute_filepath};
-  std::filesystem::path projectpath{std::string{TranslationUnitRoot()}};
+  std::filesystem::path absolutepath{absolute_filepath.begin(),
+                                     absolute_filepath.end()};
+  auto root = TranslationUnitRoot();
+  std::filesystem::path projectpath{root.begin(), root.end()};
   auto relative = std::filesystem::relative(absolutepath, projectpath);
   return relative.string();
 }
 
 absl::Status VerilogProject::updateFileContents(
     absl::string_view path, const verible::TextStructureView* updatedtext) {
-  auto projectpath = GetRelativePathToSource(static_cast<std::string>(path));
-  files_[projectpath] =
-      std::make_unique<ParsedVerilogSourceFile>(projectpath, updatedtext,
-                                                /*corpus=*/"");
+  std::string projectpath = GetRelativePathToSource(path);
+  auto fileptr = files_.find(projectpath);
+  if (fileptr == files_.end()) {
+    files_.insert(std::make_pair(
+        projectpath,
+        std::make_unique<ParsedVerilogSourceFile>(projectpath, updatedtext,
+                                                  /*corpus=*/"")));
+  } else {
+    fileptr->second =
+        std::make_unique<ParsedVerilogSourceFile>(projectpath, updatedtext,
+                                                  /*corpus=*/"");
+  }
   return absl::OkStatus();
 }
 
--- a/verilog/analysis/verilog_project.cc
+++ b/verilog/analysis/verilog_project.cc
@@ -14,6 +14,7 @@
 
 #include "verilog/analysis/verilog_project.h"
 
+#include <filesystem>
 #include <iostream>
 #include <memory>
 #include <string>
@@ -183,6 +184,15 @@ bool VerilogProject::RemoveRegisteredFile(
     }
   }
   return false;
+}
+
+std::string VerilogProject::GetRelativePathToSource(
+    const std::string& absolute_filepath) {
+  // TODO add check if the absolute_filepath is out of the VerilogProject
+  std::filesystem::path absolutepath{absolute_filepath};
+  std::filesystem::path projectpath{std::string{TranslationUnitRoot()}};
+  auto relative = std::filesystem::relative(absolutepath, projectpath);
+  return relative.string();
 }
 
 VerilogSourceFile* VerilogProject::LookupRegisteredFileInternal(
--- a/verilog/analysis/verilog_project.cc
+++ b/verilog/analysis/verilog_project.cc
@@ -202,14 +202,14 @@ absl::Status VerilogProject::UpdateFileContents(
   std::string projectpath = GetRelativePathToSource(path);
   auto fileptr = files_.find(projectpath);
   if (fileptr == files_.end()) {
-    files_.insert(std::make_pair(
-        projectpath,
-        std::make_unique<ParsedVerilogSourceFile>(projectpath, updatedtext,
-                                                  /*corpus=*/"")));
+    files_.insert(
+        std::make_pair(projectpath, std::make_unique<ParsedVerilogSourceFile>(
+                                        projectpath, path, updatedtext,
+                                        /*corpus=*/"")));
   } else {
-    fileptr->second =
-        std::make_unique<ParsedVerilogSourceFile>(projectpath, updatedtext,
-                                                  /*corpus=*/"");
+    fileptr->second = std::make_unique<ParsedVerilogSourceFile>(
+        projectpath, path, updatedtext,
+        /*corpus=*/"");
   }
   return absl::OkStatus();
 }
--- a/verilog/analysis/verilog_project.cc
+++ b/verilog/analysis/verilog_project.cc
@@ -195,6 +195,15 @@ std::string VerilogProject::GetRelativePathToSource(
   return relative.string();
 }
 
+absl::Status VerilogProject::updateFileContents(
+    absl::string_view path, const verible::TextStructureView* updatedtext) {
+  auto projectpath = GetRelativePathToSource(static_cast<std::string>(path));
+  files_[projectpath] =
+      std::make_unique<ParsedVerilogSourceFile>(projectpath, updatedtext,
+                                                /*corpus=*/"");
+  return absl::OkStatus();
+}
+
 VerilogSourceFile* VerilogProject::LookupRegisteredFileInternal(
     absl::string_view referenced_filename) const {
   const auto opened_file = FindOpenedFile(referenced_filename);
--- a/verilog/analysis/verilog_project.h
+++ b/verilog/analysis/verilog_project.h
@@ -297,7 +297,12 @@ class VerilogProject {
   const VerilogSourceFile* LookupFileOrigin(
       absl::string_view content_substring) const;
 
+  // Returns relative path to the VerilogProject
   std::string GetRelativePathToSource(const std::string& absolute_filepath);
+
+  // Updates file from external source, e.g. Language Server
+  absl::Status updateFileContents(
+      absl::string_view path, const verible::TextStructureView* updatedtext);
 
  private:
   absl::StatusOr<VerilogSourceFile*> OpenFile(
--- a/verilog/analysis/verilog_project.h
+++ b/verilog/analysis/verilog_project.h
@@ -306,7 +306,10 @@ class VerilogProject {
 
   // Adds include directory to the project
   void AddIncludePath(absl::string_view includepath) {
-    include_paths_.push_back({includepath.begin(), includepath.end()});
+    std::string path = {includepath.begin(), includepath.end()};
+    if (std::find(include_paths_.begin(), include_paths_.end(), path) ==
+        include_paths_.end())
+      include_paths_.push_back(path);
   }
 
  private:
--- a/verilog/analysis/verilog_project.h
+++ b/verilog/analysis/verilog_project.h
@@ -298,7 +298,7 @@ class VerilogProject {
       absl::string_view content_substring) const;
 
   // Returns relative path to the VerilogProject
-  std::string GetRelativePathToSource(const std::string& absolute_filepath);
+  std::string GetRelativePathToSource(absl::string_view absolute_filepath);
 
   // Updates file from external source, e.g. Language Server
   absl::Status updateFileContents(
--- a/verilog/analysis/verilog_project.h
+++ b/verilog/analysis/verilog_project.h
@@ -301,12 +301,12 @@ class VerilogProject {
   std::string GetRelativePathToSource(absl::string_view absolute_filepath);
 
   // Updates file from external source, e.g. Language Server
-  absl::Status updateFileContents(
+  absl::Status UpdateFileContents(
       absl::string_view path, const verible::TextStructureView* updatedtext);
 
   // Adds include directory to the project
-  void addIncludePath(const std::string& includepath) {
-    include_paths_.push_back(includepath);
+  void AddIncludePath(absl::string_view includepath) {
+    include_paths_.push_back({includepath.begin(), includepath.end()});
   }
 
  private:
--- a/verilog/analysis/verilog_project.h
+++ b/verilog/analysis/verilog_project.h
@@ -304,6 +304,11 @@ class VerilogProject {
   absl::Status updateFileContents(
       absl::string_view path, const verible::TextStructureView* updatedtext);
 
+  // Adds include directory to the project
+  void addIncludePath(const std::string& includepath) {
+    include_paths_.push_back(includepath);
+  }
+
  private:
   absl::StatusOr<VerilogSourceFile*> OpenFile(
       absl::string_view referenced_filename,
@@ -329,7 +334,7 @@ class VerilogProject {
 
   // The sequence of directories from which to search for `included files.
   // These can be absolute, or relative to the process's working directory.
-  const std::vector<std::string> include_paths_;
+  std::vector<std::string> include_paths_;
 
   // The corpus to which this project belongs (e.g.,
   // 'github.com/chipsalliance/verible').
--- a/verilog/analysis/verilog_project.h
+++ b/verilog/analysis/verilog_project.h
@@ -297,6 +297,8 @@ class VerilogProject {
   const VerilogSourceFile* LookupFileOrigin(
       absl::string_view content_substring) const;
 
+  std::string GetRelativePathToSource(const std::string& absolute_filepath);
+
  private:
   absl::StatusOr<VerilogSourceFile*> OpenFile(
       absl::string_view referenced_filename,
--- a/verilog/analysis/verilog_project.h
+++ b/verilog/analysis/verilog_project.h
@@ -310,8 +310,8 @@ class VerilogProject {
   std::string GetRelativePathToSource(absl::string_view absolute_filepath);
 
   // Updates file from external source, e.g. Language Server
-  absl::Status UpdateFileContents(
-      absl::string_view path, const verible::TextStructureView* updatedtext);
+  void UpdateFileContents(absl::string_view path,
+                          const verible::TextStructureView* updatedtext);
 
   // Adds include directory to the project
   void AddIncludePath(absl::string_view includepath) {
--- a/verilog/analysis/verilog_project.h
+++ b/verilog/analysis/verilog_project.h
@@ -187,6 +187,15 @@ class InMemoryVerilogSourceFile final : public VerilogSourceFile {
 // Doesn't require file-system access, nor create temporary files.
 class ParsedVerilogSourceFile final : public VerilogSourceFile {
  public:
+  // this constructor is used for updating file contents in the language server
+  // it sets referenced_path and resolved_path based on URI from language server
+  ParsedVerilogSourceFile(absl::string_view referenced_path,
+                          absl::string_view resolved_path,
+                          const verible::TextStructureView* text_structure,
+                          absl::string_view corpus = "")
+      : VerilogSourceFile(referenced_path, resolved_path, corpus),
+        text_structure_(text_structure) {}
+
   // filename can be fake, it is not used to open any file.
   // text_structure is a pointer to a TextStructureView object of
   //     already parsed file. Current implementation does _not_ make a
--- a/verilog/tools/ls/BUILD
+++ b/verilog/tools/ls/BUILD
@@ -80,6 +80,16 @@ cc_library(
 )
 
 cc_library(
+    name = "symbol-table-handler",
+    srcs = ["symbol-table-handler.cc"],
+    hdrs = ["symbol-table-handler.h"],
+    deps = [
+        "//verilog/analysis:symbol_table",
+        "//common/lsp:lsp-protocol",
+    ],
+)
+
+cc_library(
     name = "verilog-language-server",
     srcs = ["verilog-language-server.cc"],
     hdrs = ["verilog-language-server.h"],
--- a/verilog/tools/ls/BUILD
+++ b/verilog/tools/ls/BUILD
@@ -96,6 +96,7 @@ cc_library(
     visibility = ["//visibility:public"],
     deps = [
         ":lsp-parse-buffer",
+        ":symbol-table-handler",
         ":verible-lsp-adapter",
         "//common/lsp:json-rpc-dispatcher",
         "//common/lsp:lsp-protocol",
--- a/verilog/tools/ls/BUILD
+++ b/verilog/tools/ls/BUILD
@@ -86,7 +86,8 @@ cc_library(
     deps = [
         "//verilog/analysis:symbol_table",
         "//common/lsp:lsp-protocol",
-        "@com_google_absl//absl/container:flat_hash_set"
+        "@com_google_absl//absl/container:flat_hash_set",
+        ":lsp-parse-buffer"
     ],
 )
 
--- a/verilog/tools/ls/BUILD
+++ b/verilog/tools/ls/BUILD
@@ -84,10 +84,14 @@ cc_library(
     srcs = ["symbol-table-handler.cc"],
     hdrs = ["symbol-table-handler.h"],
     deps = [
-        "//verilog/analysis:symbol_table",
-        "//common/lsp:lsp-protocol",
         "@com_google_absl//absl/container:flat_hash_set",
         "@com_google_absl//absl/flags:flag",
+        "@com_google_absl//absl/strings",
+        "//common/lsp:lsp-protocol",
+        "//common/util:file_util",
+        "//verilog/analysis:symbol_table",
+        "//verilog/analysis:verilog_filelist",
+        "//verilog/analysis:verilog_project",
         ":lsp-parse-buffer"
     ],
 )
--- a/verilog/tools/ls/BUILD
+++ b/verilog/tools/ls/BUILD
@@ -88,6 +88,7 @@ cc_library(
         "@com_google_absl//absl/flags:flag",
         "@com_google_absl//absl/strings",
         "//common/lsp:lsp-protocol",
+        "//common/strings:line_column_map",
         "//common/util:file_util",
         "//verilog/analysis:symbol_table",
         "//verilog/analysis:verilog_filelist",
--- a/verilog/tools/ls/BUILD
+++ b/verilog/tools/ls/BUILD
@@ -97,6 +97,15 @@ cc_library(
     ],
 )
 
+cc_test(
+    name = "symbol-table-handler_test",
+    srcs = ["symbol-table-handler_test.cc"],
+    deps = [
+        ":symbol-table-handler",
+        "@com_google_googletest//:gtest_main",
+    ],
+)
+
 cc_library(
     name = "verilog-language-server",
     srcs = ["verilog-language-server.cc"],
--- a/verilog/tools/ls/BUILD
+++ b/verilog/tools/ls/BUILD
@@ -93,7 +93,7 @@ cc_library(
         "//verilog/analysis:symbol_table",
         "//verilog/analysis:verilog_filelist",
         "//verilog/analysis:verilog_project",
-        ":lsp-parse-buffer"
+        ":lsp-parse-buffer",
     ],
 )
 
@@ -115,6 +115,7 @@ cc_library(
         ":lsp-parse-buffer",
         ":symbol-table-handler",
         ":verible-lsp-adapter",
+        "//common/util:file_util",
         "//common/lsp:json-rpc-dispatcher",
         "//common/lsp:lsp-protocol",
         "//common/lsp:lsp-text-buffer",
--- a/verilog/tools/ls/BUILD
+++ b/verilog/tools/ls/BUILD
@@ -86,6 +86,7 @@ cc_library(
     deps = [
         "//verilog/analysis:symbol_table",
         "//common/lsp:lsp-protocol",
+        "@com_google_absl//absl/container:flat_hash_set"
     ],
 )
 
--- a/verilog/tools/ls/BUILD
+++ b/verilog/tools/ls/BUILD
@@ -87,6 +87,7 @@ cc_library(
         "//verilog/analysis:symbol_table",
         "//common/lsp:lsp-protocol",
         "@com_google_absl//absl/container:flat_hash_set",
+        "@com_google_absl//absl/flags:flag",
         ":lsp-parse-buffer"
     ],
 )
--- a/verilog/tools/ls/lsp-parse-buffer.cc
+++ b/verilog/tools/ls/lsp-parse-buffer.cc
@@ -68,9 +68,7 @@ BufferTrackerContainer::GetSubscriptionCallback() {
           const BufferTracker *tracker = Update(uri, *txt);
           // Now inform our listeners.
           for (const auto &change_listener : change_listeners_) {
-            if (change_listener) {
-              change_listener(uri, *tracker);
-            }
+            change_listener(uri, *tracker);
           }
         } else {
           Remove(uri);
--- a/verilog/tools/ls/lsp-parse-buffer.cc
+++ b/verilog/tools/ls/lsp-parse-buffer.cc
@@ -68,10 +68,13 @@ BufferTrackerContainer::GetSubscriptionCallback() {
           const BufferTracker *tracker = Update(uri, *txt);
           // Now inform our listeners.
           for (const auto &change_listener : change_listeners_) {
-            change_listener(uri, *tracker);
+            change_listener(uri, tracker);
           }
         } else {
           Remove(uri);
+          for (const auto &change_listener : change_listeners_) {
+            change_listener(uri, nullptr);
+          }
         }
       };
 }
--- a/verilog/tools/ls/lsp-parse-buffer.cc
+++ b/verilog/tools/ls/lsp-parse-buffer.cc
@@ -67,7 +67,11 @@ BufferTrackerContainer::GetSubscriptionCallback() {
         if (txt) {
           const BufferTracker *tracker = Update(uri, *txt);
           // Now inform our listeners.
-          if (change_listener_) change_listener_(uri, *tracker);
+          for (const auto &change_listener : change_listeners_) {
+            if (change_listener) {
+              change_listener(uri, *tracker);
+            }
+          }
         } else {
           Remove(uri);
         }
--- a/verilog/tools/ls/lsp-parse-buffer.h
+++ b/verilog/tools/ls/lsp-parse-buffer.h
@@ -87,11 +87,17 @@ class BufferTrackerContainer {
   verible::lsp::BufferCollection::UriBufferCallback GetSubscriptionCallback();
 
   // type for buffer change callback function
+  // The callback takes uri of the file, and the pointer to the BufferTracker
+  // The pointer can be nullptr, meaning that e.g. the file was closed.
+  // The nullptr case should be handled by callback.
   using ChangeCallback =
       std::function<void(const std::string &uri, const BufferTracker *tracker)>;
 
   // Add a change listener for clients of ours interested in updated fresly
   // parsed content.
+  // The callback takes uri of the file, and the pointer to the BufferTracker
+  // The pointer can be nullptr, meaning that e.g. the file was closed.
+  // The nullptr case should be handled by callback.
   void AddChangeListener(const ChangeCallback &cb) {
     change_listeners_.push_back(ABSL_DIE_IF_NULL(cb));
   }
--- a/verilog/tools/ls/lsp-parse-buffer.h
+++ b/verilog/tools/ls/lsp-parse-buffer.h
@@ -93,7 +93,7 @@ class BufferTrackerContainer {
   // Add a change listener for clients of ours interested in updated fresly
   // parsed content.
   void AddChangeListener(const ChangeCallback &cb) {
-    change_listeners_.push_back(cb);
+    change_listeners_.push_back(ABSL_DIE_IF_NULL(cb));
   }
 
   // Given the URI, find the associated parse buffer if it exists.
--- a/verilog/tools/ls/lsp-parse-buffer.h
+++ b/verilog/tools/ls/lsp-parse-buffer.h
@@ -86,11 +86,15 @@ class BufferTrackerContainer {
   // (internally, they exercise Update() and Remove())
   verible::lsp::BufferCollection::UriBufferCallback GetSubscriptionCallback();
 
-  // Add a change listener for clients of ours interested in updated fresly
-  // parsed content.
+  // type for buffer change callback function
   using ChangeCallback =
       std::function<void(const std::string &uri, const BufferTracker &tracker)>;
-  void SetChangeListener(const ChangeCallback &cb) { change_listener_ = cb; }
+
+  // Add a change listener for clients of ours interested in updated fresly
+  // parsed content.
+  void AddChangeListener(const ChangeCallback &cb) {
+    change_listeners_.push_back(cb);
+  }
 
   // Given the URI, find the associated parse buffer if it exists.
   const BufferTracker *FindBufferTrackerOrNull(const std::string &uri) const;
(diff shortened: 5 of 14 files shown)


verilog/tools/ls/verilog_ls.cc
```diff
@@ -146,7 +151,24 @@ int main(int argc, char *argv[]) {
   // -- Register JSON RPC callbacks
 
   // Exchange of capabilities.
-  dispatcher.AddRequestHandler("initialize", InitializeServer);
+  dispatcher.AddRequestHandler(
+      "initialize",
+      [&symboltablehandler](const verible::lsp::InitializeParams &p) {
+        // TODO add filling the include_paths and corpus
+        std::cerr << "Initialize workspace paths:  rootPath='" << p.rootPath
```
hzeller, line 158: Tip: you can use `LOG(INFO)` to output logging without necessarily spilling it into.
(you can influence what is logged by setting the `VERIBLE_LOGTHRESHOLD` environment variable)

verilog/tools/ls/verilog_ls.cc
```diff
@@ -129,6 +132,8 @@ int main(int argc, char *argv[]) {
   // It registers callbacks to receive the relevant events on the dispatcher.
   BufferCollection buffers(&dispatcher);
 
+  verilog::SymbolTableHandler symboltablehandler;
```
hzeller, line 135: the `symboltablehandler` gets called in the dispatcher. You want to avoid the dispatcher calling an out-of-scope object so its lifetime needs to be longer than that, i.e. you need to declare it before.

verilog/tools/ls/verilog_ls.cc
```diff
@@ -146,7 +151,24 @@ int main(int argc, char *argv[]) {
...
   // Exchange of capabilities.
-  dispatcher.AddRequestHandler("initialize", InitializeServer);
+  dispatcher.AddRequestHandler(
+      "initialize",
+      [&symboltablehandler](const verible::lsp::InitializeParams &p) {
+        // TODO add filling the include_paths and corpus
+        std::cerr << "Initialize workspace paths:  rootPath='" << p.rootPath
+                  << "' rootUri='" << p.rootUri << "'" << std::endl;
+        if (!p.rootUri.empty()) {
+          symboltablehandler.setProject(p.rootUri, {}, "");
```
hzeller, line 161: You probably also want it to be possible for the user to initialize a file-list beforehand via command-line flags; look for `AppendFileListFromCommandline()`.

Then, additional information can be added in this call.

verilog/tools/ls/symbol-table-handler.h
```diff
@@ -0,0 +1,57 @@
...
+  // message delivered i.e. in textDocument/definition message.
+  // Provides a list of locations with symbol's definitions.
+  std::vector<verible::lsp::Location> findDefinition(
+      const verible::lsp::DefinitionParams &params);
+
+ private:
+  std::shared_ptr<VerilogProject> currproject =
+      nullptr;  // current VerilogProject for which the symbol table is created
+  std::shared_ptr<SymbolTable> symboltable = nullptr;  // symbol table structure
+  std::unordered_set<std::string>
```
hzeller, line 50: prefer absl:: type containers as they are generally faster and less memory intensive.
`absl::flat_hash_set` would be a good choice here.

verilog/tools/ls/symbol-table-handler.h
```diff
@@ -0,0 +1,57 @@
...
+  void buildSymbolTableFor(VerilogSourceFile &file);
+
+  // Finds the definition for a symbol provided in the DefinitionParams
+  // message delivered i.e. in textDocument/definition message.
+  // Provides a list of locations with symbol's definitions.
+  std::vector<verible::lsp::Location> findDefinition(
+      const verible::lsp::DefinitionParams &params);
+
+ private:
+  std::shared_ptr<VerilogProject> currproject =
```
hzeller, line 47: Is there a particular reason you need `shared_ptr<>` ? Usually you only need these if you have unclear ownership handovers (e.g. temporarily multiple owners), but I don't see that here.
So unless really needed, use `std::unique_ptr<>`.

(also, no need to initialize them to `nullptr`, they do that already by themselves)

verilog/tools/ls/symbol-table-handler.cc
```diff
@@ -0,0 +1,41 @@
...
+void SymbolTableHandler::buildSymbolTableFor(VerilogSourceFile &file) {
+  auto result = BuildSymbolTable(file, symboltable.get(), currproject.get());
+}
+
+std::vector<verible::lsp::Location> SymbolTableHandler::findDefinition(
+    const verible::lsp::DefinitionParams &params) {
+  auto filepath = params.textDocument.uri;
+  if (checkedfiles.find(filepath) == checkedfiles.end()) {
+    // File hasn't been tracked yet in the symbol table, add it
+    auto openedfile = currproject->OpenTranslationUnit(filepath);
```
hzeller, line 26: Roughly here is where you want to open the file _not_ from the filesystem, but get it from the BufferTracker.
There you should alerady get some pre-parsed content.

For this use case it is probably sufficient to request the `BufferTracker::last_good()` buffer.

hzeller (commented):
Some first comments.

verilog/tools/ls/verilog_ls.cc
glatosinski: I replaced `std::cerr` in the current code with `LOG(...)` where possible, I'll keep that in mind in future commits

verilog/tools/ls/verilog_ls.cc
glatosinski: I fixed it now.

verilog/tools/ls/verilog_ls.cc
glatosinski: I will add reading the include paths from command-line flags

verilog/tools/ls/symbol-table-handler.h
glatosinski: OK, I'll check in future development for `absl` equivalents for containers

verilog/tools/ls/symbol-table-handler.h
glatosinski: I set it now to unique pointers

verilog/tools/ls/symbol-table-handler.cc
glatosinski: Yes, reading the files from VerilogProject was the initial step, since the base implementation of the SymbolTable relies on VerilogProject and its source containers.

In the next stage, the idea is to use existing parsed/raw buffers provided by the language server, and fallback to actual filesystem paths when necessary.

verilog/analysis/verilog_project.cc
```diff
@@ -185,6 +186,24 @@ bool VerilogProject::RemoveRegisteredFile(
   return false;
 }
 
+std::string VerilogProject::GetRelativePathToSource(
+    const std::string& absolute_filepath) {
```
hzeller, line 190: Does this need to be `const std::string&` ? Ideally we use absl::string_view wherever possible as parameter, so that we can avoid awkward (and expensive) conversions to strings such as in line 200

verilog/tools/ls/lsp-parse-buffer.h
```diff
@@ -90,7 +90,12 @@ class BufferTrackerContainer {
   // parsed content.
   using ChangeCallback =
       std::function<void(const std::string &uri, const BufferTracker &tracker)>;
-  void SetChangeListener(const ChangeCallback &cb) { change_listener_ = cb; }
+  void AddChangeListener(const std::string &name, const ChangeCallback &cb) {
+    change_listeners_[name] = cb;
+  }
+  void DeleteChangeListnener(const std::string &name) {
```
hzeller, line 96: We don't need `DeleteChangeListener()` currently, so maybe we should not have it until we actually need it (someone reading the code will wonder if/where it is needed if they see the funcion. So if we don't need it now we should also not necessarily have it).

verilog/tools/ls/lsp-parse-buffer.h
```diff
@@ -104,7 +109,7 @@ class BufferTrackerContainer {
   // Remove the buffer tracker for the given "uri".
   void Remove(const std::string &uri) { buffers_.erase(uri); }
 
-  ChangeCallback change_listener_ = nullptr;
+  std::unordered_map<std::string, ChangeCallback> change_listeners_;
```
hzeller, line 112: If we don't need the delete functionality of the change listener (see comment above) would a simple `std::vector<ChangeCallback>` be sufficient ?

verilog/tools/ls/symbol-table-handler.cc
```diff
@@ -0,0 +1,161 @@
+#include "verilog/tools/ls/symbol-table-handler.h"
```
hzeller, line 1: Don't forget to add the copyright header in the new files.

verilog/tools/ls/symbol-table-handler.cc
```diff
@@ -0,0 +1,161 @@
...
+
+#include <filesystem>
+
+#include "common/strings/line_column_map.h"
+
+namespace verilog {
+
+static const std::string fileschemeprefix = "file://";
+
+bool LSPUriToPath(absl::string_view uri, std::string &path) {
```
hzeller, line 11: For writable output parametes we prefer to use pointers instead of non-const references in Verible. So `std::string *path`.

verilog/tools/ls/symbol-table-handler.cc
```diff
@@ -0,0 +1,161 @@
...
+  auto found = uri.find(fileschemeprefix);
```
hzeller, line 12: This would probably be more readable and compact using `absl::StartsWith()`

verilog/tools/ls/symbol-table-handler.cc
```diff
@@ -0,0 +1,161 @@
...
+  }
+  if (found != 0) {
+    return false;
+  }
+  std::string res{uri.substr(found + fileschemeprefix.size())};
+  path = res;
+  return true;
+}
+
+bool PathToLSPUri(absl::string_view path, std::string &uri) {
```
hzeller, line 24: similar comment with writable output parameters to use a pointer (`std::string *uri`).

verilog/tools/ls/symbol-table-handler.cc
```diff
@@ -0,0 +1,161 @@
...
+}
+
+void SymbolTableHandler::buildSymbolTableFor(VerilogSourceFile &file) {
+  auto result = BuildSymbolTable(file, symboltable.get(), currproject.get());
+}
+
+void SymbolTableHandler::buildProjectSymbolTable() {
+  resetSymbolTable();
+  if (!currproject)
+  {
```
hzeller, line 49: this formatting looks funny. Did you use `.github/bin/run-clang-format.sh` ?

verilog/tools/ls/symbol-table-handler.cc
```diff
@@ -0,0 +1,177 @@
...
+
+#include "verilog/tools/ls/symbol-table-handler.h"
+
+#include <filesystem>
+
+#include "common/strings/line_column_map.h"
+
+namespace verilog {
+
+static const std::string fileschemeprefix = "file://";
```
hzeller, line 24: This can be a  `static constexpr absl::string_view`.

Ideally we want to avoid global objects such as `std::string`, because there is a non-trivial constructor that the program has to call at the end. A `string_view` just is a lightweight wrapper around a `const char*` that doesn't have to be deconstructed.

verilog/tools/ls/symbol-table-handler.cc
```diff
@@ -0,0 +1,177 @@
...
+  *path = res;
+  return true;
+}
+
+bool PathToLSPUri(absl::string_view path, std::string *uri) {
+  if (!uri) {
+    return false;
+  }
+  std::filesystem::path p = std::string(path);
+  *uri = fileschemeprefix + std::filesystem::absolute(p).string();
```
hzeller, line 44: For all string concatenation it is better to use `absl::StrCat()` (it works with string views and other random types, but also is usually faster)

verilog/tools/ls/symbol-table-handler.cc
```diff
@@ -0,0 +1,177 @@
...
+    auto parsestatus = (*openedfile)->Parse();
+    if (!openedfile.ok()) {
+      LOG(WARNING) << "Could not open [" << filepath << "] in project ["
+                   << currproject->TranslationUnitRoot() << "]";
+      return {};
+    }
+    auto buildstatus =
+        BuildSymbolTable(**openedfile, symboltable.get(), currproject.get());
+  }
+  auto parsedbuffer =
```
hzeller, line 129: If it is not too long, I'd suggest to use the full type
```
const BufferTracker *parsedbuffer = ...
```

While `auto` is quick to write and often shorter, it makes understanding the code often harder as the reader of the code does not see the type.

My personal way of dealing with it is to typically use the fully qualified type, unless it is somewhat bulky and somewhat clear from the context (e.g. say a for iterators).

verilog/analysis/verilog_project.cc
```diff
@@ -185,6 +186,25 @@ bool VerilogProject::RemoveRegisteredFile(
...
+      static_cast<std::string>(absolute_filepath)};
+  std::filesystem::path projectpath{std::string{TranslationUnitRoot()}};
+  auto relative = std::filesystem::relative(absolutepath, projectpath);
+  return relative.string();
+}
+
+absl::Status VerilogProject::updateFileContents(
+    absl::string_view path, const verible::TextStructureView* updatedtext) {
+  auto projectpath = GetRelativePathToSource(path);
+  files_[projectpath] =
```
hzeller, line 202: Instead of the `operator []` we usually prefer to explicitly call `insert()`, as it also allows from the return value to deduce if we had to replace a content (which might or might not be desired, but we have a chance to react to it).

verilog/analysis/verilog_project.h
```diff
@@ -297,6 +297,13 @@ class VerilogProject {
   const VerilogSourceFile* LookupFileOrigin(
       absl::string_view content_substring) const;
 
+  // Returns relative path to the VerilogProject
+  std::string GetRelativePathToSource(absl::string_view absolute_filepath);
+
+  // Updates file from external source, e.g. Language Server
```
hzeller, line 303: We pass the raw pointer to the file content. What are the lifetime and ownership expectations of this object ?
So this needs at least a documentation comment.

(In particular, the language server will replace the view relatively often, do we know that we never retain a stale object ?)

verilog/tools/ls/symbol-table-handler.cc
```diff
@@ -0,0 +1,177 @@
...
+
+#include <filesystem>
+
+#include "common/strings/line_column_map.h"
+
+namespace verilog {
+
+static const std::string fileschemeprefix = "file://";
+
+bool LSPUriToPath(absl::string_view uri, std::string *path) {
```
hzeller, line 26: Given that we just return a substr() of a string_view, we can also make the output parameter a `string_view`.

(std::string will have to allocate memory, but string_view does not)

hzeller (commented):
Some comments, mostly with focus on some style. It might look like a lot, but it is mostly minor.

  * Avoiding global complex destructing objects (e.g. use `absl::string_view` instead of `std::string` for global constant strings)
  * In general prefer `absl::string_view` to `std::string` if possible in particular for parameters but often in return values as well. This avoids unnecessary allocations and also provides a better way to trace where a particular string originates from.
  * Prefer `insert()` and `find()` to `operator[]` in maps. The latter does a lot more (e.g. creates a temporary object) that we usually not want (It is ok in containers such as vectors that only really return a reference to a particular pre-determined object; but map-like containers do more than we want).

An independent style-guide that we often use as reference is; it might not cover the above parts specifically, but in general it is worth-while reading
https://google.github.io/styleguide/cppguide.html

verilog/tools/ls/symbol-table-handler.cc
```diff
@@ -0,0 +1,170 @@
...
+
+#include "common/strings/line_column_map.h"
+
+namespace verilog {
+
+static constexpr absl::string_view fileschemeprefix = "file://";
+
+bool LSPUriToPath(absl::string_view uri, absl::string_view path) {
+  if (!absl::StartsWith(uri, fileschemeprefix)) return false;
+  path = uri.substr(fileschemeprefix.size());
```
hzeller, line 28: this will modify the local string_view, but will not return (string_view is passed by value).

But before considering a string_view pointer: I'd return it simply; The question is: do you need a boolean telling if the URI was modified ? If not, we could just return the string_view instead.

```
return absl::StartsWith(uri, fileschemeprefix) ? uri.substr(fileschemeprefix.size()) : uri;
```

verilog/analysis/verilog_project.cc
```diff
@@ -185,6 +186,25 @@ bool VerilogProject::RemoveRegisteredFile(
...
 }
 
+std::string VerilogProject::GetRelativePathToSource(
+    const absl::string_view absolute_filepath) {
+  // TODO add check if the absolute_filepath is out of the VerilogProject
+  std::filesystem::path absolutepath{absolute_filepath.begin(),
+                                     absolute_filepath.end()};
+  auto root = TranslationUnitRoot();
+  std::filesystem::path projectpath{root.begin(), root.end()};
+  auto relative = std::filesystem::relative(absolutepath, projectpath);
```
hzeller, line 196: Just a remark that we might consider revisiting the way we're resolving URIs at some point.

Right now, we essentially snip off the `file://` prefix, then implement the rest with standard with `std::filesystem` operations
(filesystem::relative, filesystem::absolute etc.): I suspect we might need to revisit that later as there are editors that might only have access files with some RPC mechanism or have some sort-of in-memory representation. So even a `file://` prefix might mean something else to them.

So at some point we might need some pluggable implementation that works on top of URIs (resolve relative, get content etc.). I think for now, the current implementation is ok, but it is worthwhile already thinking about some abstraction that goes beyond 'making filename from uri by removing prefix'.

hzeller:
What is the status here ? If you have finished a comment, is ok to `Resolve conversation` that particular comment, so that it is easier to see what is remaining.

You said that you'd like to put everything comprising the language server into an object. Are you working on that in a separate change ? It would be nice to get this in before this, as it then might simplify the integration here.

glatosinski:
I am currently working on utilizing the FileList object to obtain the list of include files and incdirs using the [`verilog::FileList`](https://github.com/chipsalliance/verible/blob/46b3e0b94ae6e7a97ddc0168f95e5f028761e1a2/verilog/analysis/verilog_filelist.h) struct.

I have added initial lookup for the file with file list (currently I called it `verible.filelist` but if you have a better name I can quickly change it) up in the directory hierarchy (starting from workspace directory up to root directory).

I am currently testing the solution and once I get everything to work I will push the update.

I will also work today on wrapping the language server logic into an object - I will create a separate PR for it.

verilog/tools/ls/symbol-table-handler.cc
glatosinski: The buffers from LS client are used by default (last_good for symbol table updating)

hzeller:
Ready for another review, or are you still working on the comments ?

hzeller:
Thanks, I'll have a look today.

Please also make sure to add unit test for every aspect. That would be a `symbol-table-handler_test.cc` to test the low-level aspects. And also an end-2-end test simulating an editor talking with the language server.

For end2end, there is currently the `verible-verilog-ls_test.sh`, but while it was good at the beginning, it gets harder and harder to modify and read.

Instead, it would be a good idea transfer the tests from that shell script into a `verilog-language-server_test.cc` unit test  that tests the `VerilogLanguageServer` class.
The nice thing with the abstraction of the `ReadFun` and `WriteFun` is, that you can instrument these callbacks directly in your unit-test (which is easier than the end-to-end testing with json through file descriptors as it is in the `verible-verilog-ls_test.sh`

Can you add such a test in a separate pull request (essentially just doing the same tests provided in the shell-script into the new unit test) ? That way, we can have a small review review just focused on that.

You can see if something was missed out by running coverage over that directory, generate HTML and look at the result:

```
MODE=coverage .github/bin/build-and-test.sh verilog/tools/ls/...
```

Since @kbieganski is working on another aspect of the language server and also needs testing capabilities, you two could probably coordinate on what a the test should look like.

hzeller (changes requested):
Looks already pretty good.

Still needs the unit tests, but here already a round of comments; mostly focusing on project preference style (e.g. use `k` prefix for global constants, or not have `Get` as prefix for getters. And a suggestion for a less leaky abstraction.

verilog/tools/ls/symbol-table-handler.cc
```diff
@@ -74,33 +77,27 @@ void SymbolTableHandler::LoadProjectFileList(absl::string_view current_dir) {
...
-    LOG(INFO) << "Checking existence of " << projectpath;
-    if (std::filesystem::exists(projectpath)) break;
-    // file not found, reached root directory
-    if (currentdirpath.parent_path() == currentdirpath) {
-      LOG(INFO) << filelistname << " not found";
-      return;
-    }
-    currentdirpath = currentdirpath.parent_path();
+  std::string projectpath;
+  if (verible::file::UpwardFileSearch(current_dir,
```
hzeller, line 82: Just a suggestion, feel free to ignore:

You can also use the `ok()` boolean return. So if-statements can read as `if (somethingsomething.ok())`

So here, this could be: 
```c++
if (!verible::file::UpdwardFileSearch(...).ok()) {
}
```

In general, it is probably useful to output the message from the status object in the logmessage. A typical code pattern would be

```c++
if (auto status = verible::file::UpwardFileSearch(...); !status.ok()) {
   LOG(WARNING) << "..." << status;
   return;
}
```

verilog/tools/ls/symbol-table-handler.cc
```diff
@@ -74,33 +77,27 @@ void SymbolTableHandler::LoadProjectFileList(absl::string_view current_dir) {
...
+                                      absl::GetFlag(FLAGS_file_list_path),
+                                      &projectpath) != absl::OkStatus()) {
+    LOG(WARNING) << "Could not find " << absl::GetFlag(FLAGS_file_list_path)
+                 << " file in the project";
+    return;
   }
   LOG(INFO) << "Found file list under " << projectpath;
   // fill the FileList object
-  absl::Status status = AppendFileListFromFile(projectpath.string(), &filelist);
+  absl::Status status = AppendFileListFromFile(projectpath, &filelist);
```
hzeller, line 91: Similar here, if we include the message returned in the status, the log-message might be more useful

```c++
if (auto status = AppendFileListFromFile(...); !status.ok()) {
   LOG(WARNING) << "..." << status;
   return;
}

File: verilog/tools/ls/symbol-table-handler.cc [hunk 25]
```diff
@@ -74,33 +77,27 @@ void SymbolTableHandler::LoadProjectFileList(absl::string_view current_dir) {
...
   // if failed to parse
   if (!status.ok()) {
-    LOG(WARNING) << "Failed to parse file list in " << projectpath.string();
+    LOG(WARNING) << "Failed to parse file list in " << projectpath;
     return;
   }
   // update include directories in project
   for (auto &incdir : filelist.preprocessing.include_dirs) {
```
hzeller, line 98: We're not modifying incdir, so maybe `const auto &incdir` ?

hzeller (commented):
Some suggestion how we can make use of the message that comes with the absl::Status, with a common pattern of

```c++
if (auto status = foo(); !status.ok()) {
   LOG(INFO) << status;
}
```

verilog/tools/ls/BUILD
```diff
@@ -62,13 +62,27 @@ cc_library(
     ],
 )
 
+cc_library(
+    name = "symbol-table-handler",
+    srcs = ["symbol-table-handler.cc"],
+    hdrs = ["symbol-table-handler.h"],
+    deps = [
+        "//verilog/analysis:symbol_table",
```
hzeller, line 70: Some dependencies are missing; all dependencies that we use in headers included in `symbol-table-handler.{h,cc}` should be mentioned here
```
"@com_google_absl//absl/container:flat_hash_set",
"@com_google_absl//absl/flags:flag",
"@com_google_absl//absl/strings",
"//common/lsp:lsp-protocol",
"//common/strings:line_column_map",
"//common/util:file_util",
"//verilog/analysis:symbol_table",
"//verilog/analysis:verilog_filelist",
"//verilog/analysis:verilog_project",
```
Also, `"@com_google_absl//absl/status:statusor",`, but I am not sure if we actually need that (it is included, but never used)

verilog/tools/ls/symbol-table-handler.h
```diff
@@ -0,0 +1,104 @@
...
+//
+
+#ifndef VERILOG_TOOLS_LS_SYMBOL_TABLE_HANDLER_H
+#define VERILOG_TOOLS_LS_SYMBOL_TABLE_HANDLER_H
+
+#include <memory>
+#include <vector>
+
+#include "absl/container/flat_hash_set.h"
+#include "absl/status/statusor.h"
```
hzeller, line 23: Looks like the `statusor.h` header is not needed anymore in this file.

verilog/tools/ls/BUILD
hzeller: Then you might also need to add `"//verilog/tools/ls:__pkg__",` to the visibility list in `//common/strings:line_column_map` .

verilog/tools/ls/symbol-table-handler.cc
```diff
@@ -0,0 +1,249 @@
...
+}
+
+void SymbolTableHandler::BuildSymbolTableFor(const VerilogSourceFile &file) {
+  auto result = BuildSymbolTable(file, symboltable.get(), currproject.get());
+}
+
+void SymbolTableHandler::BuildProjectSymbolTable() {
+  ResetSymbolTable();
+  if (!currproject) {
+    return;
```
hzeller, line 61: What if we do not _have_ a `currproject` set and it is `nullptr` ?
We should probably return a boolean here, so that we can use that to skip anything afterwards.

verilog/tools/ls/symbol-table-handler.cc
```diff
@@ -0,0 +1,249 @@
...
+  }
+  return nullptr;
+}
+
+std::vector<verible::lsp::Location> SymbolTableHandler::FindDefinition(
+    const verible::lsp::DefinitionParams &params,
+    const verilog::BufferTrackerContainer &parsed_buffers) {
+  const absl::Time finddefinition_start = absl::Now();
+  if (files_dirty_) {
+    BuildProjectSymbolTable();
```
hzeller, line 164: If we don't have a project, we should probably also finish here. So if `BuildProjectSymbolTable()` returns false (see comment above), then finish here with `return {};`

verilog/tools/ls/symbol-table-handler.cc
```diff
@@ -0,0 +1,249 @@
...
+void SymbolTableHandler::SetProject(
+    const std::shared_ptr<VerilogProject> &project) {
+  currproject = project;
+  ResetSymbolTable();
+  LoadProjectFileList(currproject->TranslationUnitRoot());
+}
+
+void SymbolTableHandler::ResetSymbolTable() {
+  checkedfiles.clear();
+  symboltable = std::make_unique<SymbolTable>(currproject.get());
```
hzeller, line 51: what if currproject is `nullptr` ?

verilog/tools/ls/symbol-table-handler.h
```diff
@@ -0,0 +1,104 @@
...
+      const verilog::BufferTrackerContainer &parsed_buffers);
+
+  absl::Status UpdateFileContent(absl::string_view path,
+                                 const verible::TextStructureView *content);
+
+ private:
+  // tells that symbol table should be rebuilt due to changes in files
+  bool files_dirty_ = true;
+  // current VerilogProject for which the symbol table is created
+  std::shared_ptr<VerilogProject> currproject;
```
hzeller, line 83: `currproject`, `symboltable` and `checkedfiles` miss an underscore at the end.

verilog/tools/ls/verilog-language-server.cc
```diff
@@ -88,66 +120,45 @@ void VerilogLanguageServer::SetRequestHandlers() {
...
+  if (!p.rootUri.empty()) {
+    absl::string_view path = verilog::LSPUriToPath(p.rootUri);
+    if (path.empty()) {
+      LOG(ERROR) << "Unsupported rootUri in initialize request:  " << p.rootUri
+                 << std::endl;
+      path = p.rootUri;
+    }
+    ConfigureProject(path);
+  } else if (!p.rootPath.empty()) {
+    ConfigureProject(p.rootPath);
```
hzeller, line 147: If neither `p.rootUri` is set or `p.rootPath`, we never call `ConfigureProject()` and end up with a `nullptr` project.

verilog/tools/ls/symbol-table-handler.cc
```diff
@@ -0,0 +1,249 @@
...
+  const absl::Time finddefinition_start = absl::Now();
+  if (files_dirty_) {
+    BuildProjectSymbolTable();
+  }
+  absl::string_view filepath = LSPUriToPath(params.textDocument.uri);
+  if (filepath.empty()) {
+    LOG(ERROR) << "Could not convert URI " << params.textDocument.uri
+               << " to filesystem path." << std::endl;
+    LOG(INFO) << "textDocument/definition processing time:  "
+              << absl::ToInt64Milliseconds(absl::Now() - finddefinition_start)
```
hzeller, line 171: The `absl::Duration` type can already print on a stream, so you don't have to wrap the whole thing in `absl::ToInt64Milliseconds()`.

```c++
LOG(...) << (absl::Now() - finddefinition_start);
```

would be sufficient. Advantage: it will automatically add the unit (e.g. `ms`), and if the time is short, say a few microseconds, it will actually print it as microseconds. So easier to read, and less work writing the log message.

verilog/tools/ls/symbol-table-handler.cc
```diff
@@ -0,0 +1,249 @@
...
+              << "ms";
+    return {};
+  }
+  std::string relativepath = currproject->GetRelativePathToSource(filepath);
+  const verilog::ParsedBuffer *parsedbuffer =
+      parsed_buffers.FindBufferTrackerOrNull(params.textDocument.uri)
```
hzeller, line 177: `FindBufferTrackerOrNull()` can return nullptr.
Here, we're relying on the editor sending us a valid `params.textDocument.uri` that we have store. But if not, we'd crash here. So I'd apply defensive programming here to test for that early.

(anything where we depend on external information that could make us crash ... doublechecking is good).

hzeller:
Above some robustness suggestions as we want to avoid crashing the language server due to any external data we get fed.

hzeller:
If you rebase, you get the change in #1615, that I had to do when playing around with my `vscode` installation.

glatosinski:
I rebased the code to the current state of `master`.

hzeller:
While working on the test, you can verify that you covered all  corner cases by running the test with coverage and generate the HTML output. 100% coverage is sometimes not possible, but aim for at least 90%.

```
MODE=coverage .github/bin/build-and-test.sh //verilog/tools/ls:symbol-table-handler_test
```

verilog/tools/ls/verilog-language-server.cc
```diff
@@ -33,14 +34,45 @@ VerilogLanguageServer::VerilogLanguageServer(const WriteFun &write_fun)
...
       });
   SetRequestHandlers();
 }
 
+verible::lsp::InitializeResult VerilogLanguageServer::GetCapabilities() {
+  // send response with information what we do.
+  verible::lsp::InitializeResult result;
+  result.serverInfo = {
+      .name = "Verible Verilog language server.",
+      .version = GetVersionNumber(),
```
hzeller, line 50: FYI, upstream in #1624  this one call changed from `GetVersionNumber()` changed to `verible::GetRepositoryVersion()`.

Because the block moved in the file, git can't resolve this and end up in merge conflict that looks more nasty than it actually is. So you'll have to manually change this function call.

verilog/tools/ls/symbol-table-handler.cc
```diff
@@ -0,0 +1,267 @@
...
+  // TODO (glatosinski): should we do this?
+  curr_project_->AddIncludePath(verible::file::Dirname(projectpath));
+  LOG(INFO) << "Adding \"" << projectpath << "\" to include directories";
+  // update include directories in project
+  for (const auto &incdir : filelist.preprocessing.include_dirs) {
+    LOG(INFO) << "Adding include path:  " << incdir;
+    curr_project_->AddIncludePath(incdir);
+  }
+  // add files from file list to the project
+  for (auto &incfile : filelist.file_paths) {
```
hzeller, line 112: these are not included files, just files that are part of the project. So `incfile` is not necessarily a good name. How about `file_in_project` ?

Also `OpenIncludedFile()` is of course not necessarily the right name for the function, but maybe add that as TODO to provide a better name. It also goes through the include paths and attempts to find them relative to that. which is not neccesarily what we need. We do want to find them relative to project root.
Maybe add a TODO what needs to be improved later.

verilog/tools/ls/verilog-language-server.cc
hzeller: Probably a good fallback would be to call ConfigureProject() with the directory where `verible.filelist` was found. And if _that_ doesn't exist, a fallback of `ConfigureProject(".")` probably.

But just the `LOG(WARNING)` below leaves the project unconfigured, so no symbol table functionality will be available.

hzeller:
While playing with the symbol table in vscode (I first had to add a `ConfigureProject(".")` fallback as vscode did not set `rootUri` or `rootPath`, see comment above), I found the following issue when using the file list.

Let's assume we are in a directory and have two files, foo.sv and bar.sv

```systemverilog
module foo();
  bar x;
endmodule;
```

```systemverilog
module bar();
   foo xy();
endmodule;
```

Now if we have a `verible.filelist` with the follwoing content in the same directory

```
bar.sv
foo.sv
```

... then opening `code foo.sv`, we can put the cursor on the `bar` instance, jump to definition and arrive at the `bar.sv` file on the symbol. Works perfectly.

Now, if the `verible.filelist` contains names with `./` in there

```
./bar.sv
./foo.sv
```

... then with the same operation the language server crashes:

```
F0123 22:02:32.715459 2585480 text_structure.cc:152] Check failed: from >= 0 (-1026659897 vs. 0) "foo"
    @     0x556aaf0f2dc9  absl::log_internal::LogMessage::SendToLog()
    @     0x556aaf0f3379  absl::log_internal::LogMessageFatal::~LogMessageFatal()
    @     0x556aaf08fb36  verible::TextStructureView::GetRangeForText()
    @     0x556aaef73c2e  verilog::SymbolTable::Builder::DiagnoseSymbolAlreadyExists()
    @     0x556aaef78a0d  verilog::SymbolTable::Builder::EmplaceElementInCurrentScope()
    @     0x556aaef78a46  verilog::SymbolTable::Builder::DeclareScopedElementAndDescend()
    @     0x556aaef7deef  verilog::SymbolTable::Builder::Visit()
    @     0x556aaf0d0ae1  verible::TreeContextVisitor::Visit()
    @     0x556aaef7dd6b  verilog::SymbolTable::Builder::Visit()
    @     0x556aaef77efb  verilog::BuildSymbolTable()
    @     0x556aaef78009  verilog::ParseFileAndBuildSymbolTable()
    @     0x556aaef7832b  verilog::SymbolTable::Build()
    @     0x556aaef3a513  verilog::SymbolTableHandler::BuildProjectSymbolTable()
    @     0x556aaef718f0  verilog::SymbolTableHandler::FindDefinition()
```

It does look like maybe the file that was opened vs. loaded via the language server protocol changed, but the symbol table still has the old version, pointing to a wrong version of the string view ? It sounds a bit like the symptoms you mentioned when we were chatting on the Christmas party.

verilog/tools/ls/symbol-table-handler_test.cc
```diff
@@ -0,0 +1,221 @@
...
+namespace verilog {
+namespace {
+
+constexpr absl::string_view  //
+    kSampleModuleA(
+        "module a;\n"
+        "  assign var1 = 1'b0;\n"
+        "  assign var2 = var1 | 1'b1;\n"
+        "endmodule\n"),
+    kSampleModuleB(
```
hzeller, line 33: Avoid using multiple variables in one declaration, split them up. In particular here, where things go over multiple lines and it is hard to even see at a glance that these are two.

```c++
// Instead of
absl::string_view a, b;

// ... make it
absl::string_view a;
absl::string_view b;
```

hzeller:
Did you find the issue with the crash when jumping to the definition ?

verilog/tools/ls/lsp-parse-buffer.cc
```diff
@@ -67,7 +67,9 @@ BufferTrackerContainer::GetSubscriptionCallback() {
         if (txt) {
           const BufferTracker *tracker = Update(uri, *txt);
           // Now inform our listeners.
-          if (change_listener_) change_listener_(uri, *tracker);
+          for (const auto &change_listener : change_listeners_) {
+            change_listener(uri, *tracker);
+          }
         } else {
           Remove(uri);
```
hzeller, line 74: On called back with this callback, you store references to the text structure stored in tracker data.

But if the file gets removed, there is no change listener that informs about a URI being removed (the `Remove(uri)` below does not call any listeners)...

Which means in case the URI is deleted (someone closes the buffer in the editor)  you'll store a dangeling reference to a removed text buffer in `VerilogLanguageServer::UpdateEditedFileInProject()` -> ` SymbolTableHandler::UpdateFileContent()`.

I suggest to have the change listener to take a pointer, which can inform about a `nullptr` buffer (i.e. inform about a removed URI), which you then can also use to properly clean up references.

Also, longer term it is probably better to have `BufferTracker`own the `ParsedVerilogSourceFile` which we then pass to `VerilogProject::UpdateFileContents()`. Right now, due to the chain of change listeners, the ownership is quite murky and will result in issues.

verilog/analysis/verilog_project.cc
```diff
@@ -185,6 +186,34 @@ bool VerilogProject::RemoveRegisteredFile(
...
+    files_.insert(
+        std::make_pair(projectpath, std::make_unique<ParsedVerilogSourceFile>(
+                                        projectpath, path, updatedtext,
+                                        /*corpus=*/"")));
+  } else {
+    fileptr->second = std::make_unique<ParsedVerilogSourceFile>(
+        projectpath, path, updatedtext,
+        /*corpus=*/"");
+  }
+  return absl::OkStatus();
```
hzeller, line 214: Currently, the only outcome of this function is `absl::OkStatus()`.
So until we actually have a status to return, let's just return `void` (simplifies all the places that call it).

hzeller:
The problem with the `./foo` in the [comment above](https://github.com/chipsalliance/verible/pull/1530#issuecomment-1401438974) can probably be fixed by canonicalizing the file coming from the filelist in `symbol-table-handler.cc`

```c++
 for (const auto &file_in_project : filelist.file_paths) {
    const std::string canoncialized = std::filesystem::path(file_in_project).lexically_normal().string();
   //... and use canonicalized instead of file_in_project below
```

hzeller:
For your testing: have a file where going to definition opens another file. Then close the other file (which internally should remove it, but if the remove listener from above is not there will point dangeling data).
Attempt to go to definition again - that will now go to the wrong place and possibly crash things.

hzeller:
The reacting to remove also would need to take into account that now we have to read the file from the filesystem again.

hzeller:
Robustness testing:  if there is one file in the filelist that has issues resolving symbols in the symbol table, _no_ symbol can be resolved, even though other symbols might still be there, in particular in the file.

But in that case, the symbol resolving bails out with `Could not find definition due to symbol table errors` and then does not symbols that might be there.

This is important, because of course if you have 5000 files in the file-list (typical scale project), there will be one or more that have some issues parsing or producing symbols. That should not prevent us from the rest of the 'go to definition' working.

hzeller:
With my VSCode that does not set the root path and URI the new logic that derives things from the file-list is kicking in (i really like that btw. I think that is a good fallback).

It does seem that it got confused in my testcase and attempts to resolve files relative to the filelist, not the filelist's directory - probably due to the dot in the filelist directory ? (/foo/bar/./verible.fileliist). Maybe some round of `lexically_normal()` needed somewhere (and some test that covers this use-case).

```
Verible Verilog Language Server built at v0.0-2644-g287db0e6
W0126 08:24:01.943099   18703 verilog-language-server.cc:178] Could not configure Verilog Project root
I0126 08:24:01.943285   18703 symbol-table-handler.cc:53] Found file list under /home/hzeller/tt/./verible.filelist
I0126 08:24:01.943310   18703 symbol-table-handler.cc:96] LoadProjectFileList
I0126 08:24:01.943339   18703 symbol-table-handler.cc:53] Found file list under /home/hzeller/tt/./verible.filelist
I0126 08:24:01.943343   18703 symbol-table-handler.cc:106] Found file list under /home/hzeller/tt/./verible.filelist
I0126 08:24:01.943350   18703 symbol-table-handler.cc:116] Updating the filelist
I0126 08:24:01.943464   18703 symbol-table-handler.cc:132] Adding "/home/hzeller/tt/." to include directories
I0126 08:24:01.943469   18703 symbol-table-handler.cc:135] Adding include path:  .
W0126 08:24:01.943493   18703 symbol-table-handler.cc:145] File included in /home/hzeller/tt/./verible.filelist not found:  bar.sv:  NOT_FOUND: /home/hzeller/tt/verible.filelist/bar.sv: Not a directory
W0126 08:24:01.943512   18703 symbol-table-handler.cc:145] File included in /home/hzeller/tt/./verible.filelist not found:  foo.sv:  NOT_FOUND: /home/hzeller/tt/verible.filelist/foo.sv: Not a directory
E0126 08:24:01.974014   18703 json-rpc-dispatcher.cc:58] Unhandled notification 'initialized'
```

verilog/tools/ls/verilog-language-server_test.cc
```diff
@@ -33,12 +36,27 @@ namespace {
 
 using namespace nlohmann;
```
hzeller, line 37: Don't use the generic namespace import, be specific in what you use (`using nlohmann::json`)

verilog/tools/ls/verilog-language-server_test.cc
```diff
@@ -33,12 +36,27 @@ namespace {
...
 
+// FIXME (glatosinski) use more plausible modules
+constexpr absl::string_view  //
```
hzeller, line 40: Similar to the comment in the symbol table handler test: make the variable separate. Also, use raw string literals.

verilog/tools/ls/symbol-table-handler_test.cc
```diff
@@ -0,0 +1,223 @@
...
+
+#include "absl/strings/string_view.h"
+#include "common/util/file_util.h"
+#include "gtest/gtest.h"
+#include "verilog/analysis/verilog_project.h"
+
+namespace verilog {
+namespace {
+
+static constexpr absl::string_view  //
```
hzeller, line 27: Use raw string literals for these multi-line things. Much more readable.

verilog/tools/ls/verilog-language-server.cc
```diff
@@ -124,35 +162,42 @@ void VerilogLanguageServer::PrintStatistics() const {
...
+    ConfigureProject();
+  }
+  return GetCapabilities();
+}
 
-  return result;
+void VerilogLanguageServer::ConfigureProject(absl::string_view project_root) {
+  std::string proj_root = {project_root.begin(), project_root.end()};
+  if (proj_root.empty()) {
+    proj_root = verible::file::Dirname(FindFileList(".")).data();
```
hzeller, line 187: data() just returns the pointer to the underlying data, so it is not null-terminated. So this will not return what you think it does.
Instead, if you need a string, convert it accordingly `std::string(verible::file::Dirname(FindFileList("."))`.

hzeller: This explains the issue from above, `/home/hzeller/tt/verible.filelist/bar.sv: Not a directory`

hzeller:
Maybe to reduce the impact of ['there is a single issue somewhere in building the symbol table and therefore we can't resolve any symbols'](https://github.com/chipsalliance/verible/pull/1530#issuecomment-1404485037)-issue there could be a quick fix for now: always build the symbol table for a single file, the current one, and try to resolve a symbol there (I suspect that is very common).
Only if the symbol is not defined there, start building a symbol table for the whole project.

(which also will help performance: in my case, building the symbol table for a single file takes a few 100 microseconds, while building for the whole project takes about 40 seconds).

Longer term we then can refine things of course: ignore issues in some files but at least build partial symbol table, incremental update etc.

verilog/analysis/verilog_project.cc
```diff
@@ -185,6 +186,34 @@ bool VerilogProject::RemoveRegisteredFile(
...
+  std::string projectpath = GetRelativePathToSource(path);
+  std::unique_ptr<VerilogSourceFile> contents = nullptr;
+  if (updatedtext)
+    contents = std::make_unique<ParsedVerilogSourceFile>(
+        projectpath, path, updatedtext, /*corpus=*/"");
+  else
+    contents = std::make_unique<VerilogSourceFile>(projectpath, path, "");
+  auto fileptr = files_.find(projectpath);
+  if (fileptr == files_.end()) {
+    files_.insert(std::make_pair(projectpath, std::move(contents)));
```
hzeller, line 211: For later: Probably `files_.emplace()` might be easier here.

verilog/tools/ls/symbol-table-handler.cc
```diff
@@ -0,0 +1,300 @@
...
+
+bool IsStringViewContained(absl::string_view origin, absl::string_view substr) {
+  const int from = std::distance(origin.begin(), substr.begin());
+  const int to = std::distance(origin.begin(), substr.end());
+  if (from < 0) return false;
+  if (to > static_cast<int>(origin.length())) return false;
+  return true;
+}
+
+const SymbolTableNode *ScanReferenceComponents(
```
hzeller, line 164: For future consideration: if scanning takes a while, maybe we need to build an index in the symbol table.

hzeller (approved):
LGTM

PR #1532, opened 2022-11-16
Update the preprocessor command to 'preprocess'.

hzeller:
Fix the unit test to actually fail on failures (there were a bunch of `exit 0` in there.)

Update documentation slightly and include in README.


hzeller:
@karimtera for you to have a look. Mostly renaming `multiple-compilation-unit` to `preprocess`. Also removing the filename at the top of the output and the additional trailing newline output.

But then I also noticed that the test was not failing when it should (because of `exit 0` instead of `exit 1`).

karimtera:
Nice, LGTM.

PR #1533, opened 2022-11-16
[WIP] adding exit code to the preprocessor

karimtera:
### Description:
Adding exit code to the preprocessor, to allow continuing the preprocessing even after some errors are met.
So, the preprocessed output might contain a macro call that couldn't be expanded, or an include directive that file can't be found, that way the output would still be usable.

### Exit code description:
It's a 4 bit code, with each bit shows if an error of a certain type is faced at least once or not.
First bit (rightmost significant bit): At least one System-Verilog file is not found.
2nd bit: At least one macro definition is not expanded.
3rd bit: At least one include file couldn't be found.
4-th bit is left for now.

### DIFF
--- a/verilog/preprocessor/verilog_preprocess.h
+++ b/verilog/preprocessor/verilog_preprocess.h
@@ -41,6 +41,7 @@
 #include <stack>
 #include <string>
 #include <vector>
+#include <bitset>
 
 #include "absl/status/status.h"
 #include "absl/status/statusor.h"
@@ -87,6 +88,10 @@ struct VerilogPreprocessData {
   // are two separate vectors.
   std::vector<VerilogPreprocessError> errors;
   std::vector<VerilogPreprocessError> warnings;
+
+  // An exit code of 4 bits is used to distinguish between errors met during 
+  // preprocessing.
+  std::bitset<4> exit_code;
 };
 
 // VerilogPreprocess transforms a TokenStreamView.


verilog/preprocessor/verilog_preprocess.h
```diff
@@ -87,6 +88,10 @@ struct VerilogPreprocessData {
   // are two separate vectors.
   std::vector<VerilogPreprocessError> errors;
   std::vector<VerilogPreprocessError> warnings;
+
+  // An exit code of 4 bits is used to distinguish between errors met during 
+  // preprocessing.
+  std::bitset<4> exit_code;
```
hzeller, line 94: You probably also want an enum that describes the bits:

```c++
enum ResultBitsBits {
     kFileErrorBit = 0;  // file not found or tokenization error
     kMacroUseNotExpandedBit = 1;  // At least one macro could not be expanded
     kIncludeFileNotFoundBit = 2;      // At least one include file could not be found
}
```

I'd probably not even use a `std::bitset`, just a `uint32_t` as this is the code we then exit `main()` with anyway.
Then your enum could be defined as:

```c++
enum ResultBitsBits {
     kFileErrorBit = (1 << 0);  // file not found or tokenization error
     kMacroUseNotExpandedBit = (1 << 1);  // At least one macro could not be expanded
     kIncludeFileNotFoundBit = (1 << 2);      // At least one include file could not be found
}
```
.. then `OR`-ing together the relevant values.
I think the `uint32_t` and `(1 << xxx)` thing would be easier to read to the average reader of the code.

hzeller:
I've created an issue for this #1535

verilog/preprocessor/verilog_preprocess.h
hzeller: I think for me it is easier to read as bitmap and we can use that number right in the return value.

Since it is very common to define these things as bit-patterns directly into `uint32_t` like that is is much quicker to read. Using a `std::bitset<>` would probably slow down readers of the code as they are wondering if there is a deeper meaning to do so.

PR #1541, opened 2022-11-28
Custom runners ci

AdamOlech:
This pull request modifies the CI pipeline definition and some related scripts so that the [custom runners](https://github.com/antmicro/runner) deployment for this repository is used.

What was changed:
* the `coverage` job will use the `@bazel_tools//tools/jdk:remote_jdk11` Java toolchain, as Java is not automatically installed in the container picked for the job.
* `sudo` is no longer necessary as the execution takes place in a container where privileges are already elevated.
* the `build` jobs install and start Docker explicitly. This way, the scripts that use Docker were left unchanged.
* all jobs run on `n2-highcpu-8` GCP machines, except for `ClangTidy` which runs on `n2-standard-4` (no need for a beefy machine as it doesn't do anything expensive).

The CI does not run on branches so during the development phase I used a "drop me" commit to test it: https://github.com/chipsalliance/verible/actions/runs/3549496319. I did drop the aforementioned commit before submitting this PR.

### DIFF
--- a/.github/bin/build-and-test.sh
+++ b/.github/bin/build-and-test.sh
@@ -94,6 +94,7 @@ case "$MODE" in
     bazel coverage \
           --combined_report=lcov \
           --coverage_report_generator=@bazel_tools//tools/test/CoverageOutputGenerator/java/com/google/devtools/coverageoutputgenerator:Main \
+          --java_runtime_version=remotejdk_11 \
           ${CHOSEN_TARGETS}
     # output will be in bazel-out/_coverage/_coverage_report.dat
     ;;
--- a/.github/settings.sh
+++ b/.github/settings.sh
@@ -32,7 +32,7 @@ export BAZEL_CXXOPTS="-std=c++17"
 export BAZEL_OPTS="-c opt --show_progress_rate_limit=10.0"
 
 # Used to fetch the BAZEL version where needed.
-export BAZEL_VERSION=4.0.0
+export BAZEL_VERSION=5.3.2
 
 # TODO(b/171989992): revert to using release version after upgrading system
 #   libraries/image (need: GLIBCXX_3.4.26, CXXABI_1.3.11, GLIBC_2.29).
--- a/.github/workflows/verible-ci.yml
+++ b/.github/workflows/verible-ci.yml
@@ -20,7 +20,8 @@ jobs:
 
 
   VerifyFormatting:
-    runs-on: ubuntu-20.04
+    container: ubuntu:20.04
+    runs-on: [self-hosted, Linux, X64]
 
     steps:
 
@@ -43,7 +44,8 @@ jobs:
       run: ./.github/bin/run-clang-format.sh
 
   ClangTidy:
-    runs-on: ubuntu-20.04
+    container: ubuntu:20.04
+    runs-on: [self-hosted, Linux, X64]
 
     steps:
 
@@ -82,7 +84,8 @@ jobs:
 
 
   Check:
-    runs-on: ubuntu-20.04
+    container: ubuntu:20.04
+    runs-on: [self-hosted, Linux, X64]
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/verible-ci.yml
+++ b/.github/workflows/verible-ci.yml
@@ -78,7 +78,7 @@ jobs:
         format: 'YYYY-MM-DD-HH-mm-ss'
 
     - name: Retrieve cached results
-      uses: actions/cache@v2
+      uses: actions/cache@v3
       with:
         path: |
           /tmp/clang-tidy-hashes.cache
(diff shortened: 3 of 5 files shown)


.github/settings.sh
```diff
@@ -32,7 +32,7 @@ export BAZEL_CXXOPTS="-std=c++17"
 export BAZEL_OPTS="-c opt --show_progress_rate_limit=10.0"
 
 # Used to fetch the BAZEL version where needed.
-export BAZEL_VERSION=4.0.0
+export BAZEL_VERSION=5.3.2
```
hzeller, line 35: Do we need a switch to latest bazel ?
I think we wanted to make sure to be able to run with slightly older versions, so keeping the lowest version that still works is probably good.

.github/workflows/verible-ci.yml
```diff
@@ -12,15 +12,17 @@ on:
...
   BOT_USER: "Deployment Bot"
   BOT_EMAIL: "verible-dev@googlegroups.com"
+  GHA_MACHINE_TYPE: "n2-highcpu-8"
 
 jobs:
 
 
   VerifyFormatting:
-    runs-on: ubuntu-20.04
+    #runs-on: ubuntu-20.04
```
hzeller, line 23: is this commented-out line a debugging leftover ?

hzeller (commented):
Nice

.github/settings.sh
AdamOlech: Just to reiterate what we've already discussed in person: because of the implicit existence of [bazelisk](https://github.com/bazelbuild/bazelisk) on the [GitHub Actions runner image](https://github.com/actions/runner-images/blob/9f92c57e695e8950765b3596e1add8b31d657df6/images/linux/scripts/installers/bazel.sh#L4), the CI in this project hasn't been using the version specified in the `BAZEL_VERSION` variable for quite some time now. What happened was that Bazelisk would update Bazel to the latest version. One particular run actually started failing and upon investigation the Bazel version was the only difference -- the `bazel version` command invoked in the "old" CI did in fact report `5.3.2`, hence I bumped it here and the build started to pass.

Considering that we no longer have Bazelisk (as we did not install it and the container that we use doesn't have it implicitly installed), bumping the version seems to be the good solution here. If need be, we can always install Bazelisk in the future.

.github/workflows/verible-ci.yml
AdamOlech: This is a leftover indeed. I will remove it shortly.

hzeller:
Even after restarting, the asan test failed. In the previous run, it was just timing out after hours, now some gcc just crashes.

AdamOlech:
I did not change anything apart from adding the performance graph upload stage but the pipeline passed so in order to verify if this is resource-dependent, we decided to force cache miss to see how much resources each job consumes when building afresh.

.github/workflows/verible-ci.yml
```diff
@@ -69,20 +85,28 @@ jobs:
...
+          /root/.cache/bazel
         key: clang-tidy-${{ steps.cache_timestamp.outputs.time }}
         restore-keys: clang-tidy-
 
     - name: Run clang tidy
       run: ./.github/bin/run-clang-tidy.sh
 
+    - name: 📤 Upload performance graphs
+      uses: actions/upload-artifact@v2
+      if: success() || failure()
```
hzeller, line 101: Couldn't we just use `always()`, or would that semantically be different ?

AdamOlech: It is different. As per last comment in the related [SO answer](https://stackoverflow.com/a/58859404), the `success() || failure() ` expression allows the user to cancel the build, while `always()` would insist on running the step upon cancellation.

hzeller: Ah, I see.

hzeller:
Did you also try if all the release builds will work ?

They are disabled by default in the pull request, but they are essential for the releases on push. In particular they are interesting as they are running various containers to build for ancient linux distros.

.github/bin/build-and-test.sh
```diff
@@ -94,6 +94,7 @@ case "$MODE" in
     bazel coverage \
           --combined_report=lcov \
           --coverage_report_generator=@bazel_tools//tools/test/CoverageOutputGenerator/java/com/google/devtools/coverageoutputgenerator:Main \
+          --java_runtime_version=remotejdk_11 \
```
hzeller, line 97: Could this be the issue we see with the remotejdk complaints ?

AdamOlech: Yes. This is the newer syntax for 5.x and it does not work on 4.x. I've just reworked the CI so that the `default-jdk` package is used (i.e. a system-wide Java installation), see https://github.com/chipsalliance/verible/actions/runs/3696121095/jobs/6259392848.

PR #1543, opened 2022-12-02
Wrap hanging_indentation calculation  to avoid negative indentation values

### LINKED ISSUE #1242: verible-verilog-format: crash on this fuzzer-generated input.
With the following fuzzer generated input...

```systemverilog
`J(D(`ifdef e))
```

The formatter crashes with
```
F0302 17:45:42.422962 1493663 layout_optimizer.cc:230] Check failed: indent >= 0 (-4 vs. 0) 
```

Though this is probably not something to encounter in real life, we must never crash.

jbylicki:
This PR wraps the calculation of hanging_indentation in
`TokenPartitionsLayoutOptimizer::CalculateOptimalLayout`'s `kWrap` case to never let the value be negative, as it caused a check failure down the line.

This PR fixes #1242

### DIFF
--- a/common/formatting/layout_optimizer.cc
+++ b/common/formatting/layout_optimizer.cc
@@ -570,10 +570,14 @@ LayoutFunction TokenPartitionsLayoutOptimizer::CalculateOptimalLayout(
           }
         }
       }
+      // std::max(hanging_indentation, 0) serves to resolve a case where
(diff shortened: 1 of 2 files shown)


hzeller:
Nice! Can you add a formatter unit test that fails (probably crashes) before this change and succeeds afterwards ?

PR #1545, opened 2022-12-06
Fix C++20-incompatible instances of aggregate initialization

hzeller:
In C++20, types that declare or delete any constructors are no longer aggregates, breaking compilation of many existing uses of aggregate initialization.

Fixes include:
  * removing defaultings or deletions of only the default constructor
  * removing defaultings of copy/move constructors
  * adding constructors that accept the arguments passed to existing aggregate initialization sites, leaving the type no longer an aggregate
  * removing a set of clearly-marked-as-boiler-plate copy/move/default constructor declarations

Also adds default member initializers for trivially default constructible members that can now be left uninitialized by construction of the form `Foo f;`


fangism:
Approved, thanks!

PR #1559, opened 2022-12-13
Propagate failures in releases setup + use Bazel 4.0.0

AdamOlech:


hzeller:
In #1561 , I am updating to the latest win flex bison, so that we have a clean baseline there.

hzeller:
alright, the windows thing is now merged. Let's see what happens if you rebase this change.

hzeller:
the `--java_runtime_version=remotejdk_11` in `.github/bin/build-and-test.sh` added in #1541 suspiciously looks like it could be the culprit in the coverage build.

AdamOlech:
@hzeller I flipped Bazel version back to `4.0.0` and I replaced the `--java_runtime_version` bit with the system-wide JDK installation (`apt install default-jdk`). This seems to have stabilized the pipeline -- I ran it a couple of times and I have yet to see it fail. I squashed some commits here and there and once it passes I'll remove the `[DROP ME]` commit for triggering the build here so that we can proceed with merging this one.

hzeller (approved):
Looks good. Ready to merge after run successful and `[drop me]` dropped.

AdamOlech:
The pipeline has successfully passed: https://github.com/chipsalliance/verible/actions/runs/3696838830. I removed the `[DROP ME]` commit so this is ready for merging.

PR #1563, opened 2022-12-14
ci: generate GH pages contents in a separate job

AdamOlech:


.github/bin/github-pages-setup.sh
```diff
@@ -27,7 +29,7 @@ echo "Git hash:    $GIT_HASH"
 echo "Git version: $GIT_VERSION"
 
 PAGES_DIR=/tmp/pages
-rm -rf $PAGES_DIR
+rm -rf $PAGES_DIR || true
```
hzeller, line 32: Do we need this  `|| true` ? The `-rf` option should make sure that we get an exit code that is 0 regardless if the directory already exists.

.github/bin/github-pages-setup.sh
```diff
@@ -17,6 +17,8 @@
 # ---------------
 # Generate the GitHub pages to deploy
 
+set -ex
```
hzeller, line 20: since we're also using a lot of variables and want to make sure there are no typos, maybe also add a `set -u`

.github/workflows/verible-ci.yml
```diff
@@ -176,14 +172,17 @@ jobs:
         verbose: true
       continue-on-error: true
 
-    - name: Deployment
-      if: github.ref == 'refs/heads/master' && github.event_name != 'pull_request' && matrix.mode == 'compile'
-      uses: JamesIves/github-pages-deploy-action@3.7.1
+    - name: Compress bazel-bin
```
hzeller, line 175: I'd probably call 'pack' instead of 'compress', as we only do that with the tar.

hzeller (commented):
Looks good. A few small comments.

.github/bin/github-pages-setup.sh
AdamOlech: Good catch. We don't need it. I added it "just in case" to be extra sure that my build doesn't fail here.

.github/bin/github-pages-setup.sh
AdamOlech: This is a very good idea, I'll add the `-u` flag.

.github/workflows/verible-ci.yml
```diff
@@ -424,7 +424,7 @@ jobs:
           name: bazel-bin
 
       - name: Extract bazel-bin
-        run: tar xvf bazel-bin.tar 
+        run: sudo tar xvf bazel-bin.tar -C /usr/bin
```
hzeller, line 427: I'd probably not unpack them somewhere in the public /usr/bin, but somewhere local as we don't want the the regular path get into the way (e.g. `/tmp/verible-bin`) Then use that path to invoke `github-pages-setup.sh /tmp/verible-bin`

Also, you don't need 'sudo', then :)

.github/bin/github-pages-setup.sh
```diff
@@ -79,7 +79,7 @@ wget https://raw.githubusercontent.com/IQAndreas/markdown-licenses/master/apache
 
 # Generate lint rules documentation
 # --------------------------------
-bazel-bin/verilog/tools/lint/verible-verilog-lint -generate_markdown > $PAGES_DIR/lint.md
+verible-verilog-lint -generate_markdown > $PAGES_DIR/lint.md
```
hzeller, line 82: We should not rely on the binaries being in the path, but rather we look for it in a directory we specify (makes everything easier to understand).

I suggest to have an optional parameter to `github-pages-setup.sh [<bindir>]` which then is used to assemble the binary path:

```sh
#!/usr/bin/env bash
BINDIR=${1:-/usr/bin}
# ...
${BINDIR}/verible-verilog-lint ...
```

.github/workflows/verible-ci.yml
AdamOlech: Done.

.github/bin/github-pages-setup.sh
AdamOlech: Oh, I like this one a lot. I was actually having a conundrum how to solve this in an elegant way and this suggestion seems to be it.

.github/workflows/verible-ci.yml
AdamOlech: Done.

.github/bin/github-pages-setup.sh
AdamOlech: Done.

.github/bin/github-pages-setup.sh
AdamOlech: I've removed this one.

AdamOlech:
@hzeller I addressed your remarks and I squashed the commits. The CI is green and seems to work as intended so I think we're good to merge.

.github/bin/github-pages-setup.sh
```diff
@@ -97,7 +100,7 @@ Verible tool suite.
 ## Command line arguments
 \`\`\`
 EOF
-bazel-bin/verilog/tools/syntax/verible-verilog-syntax -helpfull >> $SYNTAX_DOC
+${BINDIR}/verible-verilog-syntax -helpfull || true >> $SYNTAX_DOC
```
hzeller, line 103: Due to the `|| true`, if the binary does not exist we would actually not notice here and end up with an empty output here...

Maybe we should test the existence of every binary beforehand. Since we have `set -e`, something like the following should be sufficient to exit with an error early in that case:

```
test -x "${BINDIR}/verible-verilog-syntax"
```

AdamOlech: Done. I created a wrapper function for this.

hzeller:
... looks like the release scripts actually also use that github-pages-setup script:
https://github.com/chipsalliance/verible/actions/runs/3706325793/jobs/6281401447

PR #1565, opened 2022-12-16
[WIP] Fix build jobs

AdamOlech:


releasing/build.sh
```diff
@@ -18,5 +18,5 @@ set -e
 
 cd $(dirname "$0")/..
 
-./.github/bin/github-pages-setup.sh
-./.github/bin/github-releases-setup.sh /out
+source ./.github/bin/github-releases-setup.sh /out
```
hzeller, line 21: sourcing this to get access to the `PREFIX_BIN` variable looks very hacky to me.
Do you see another way ?

AdamOlech: This felt like a hack to me as well. I moved the invocation of `github-pages-setup` to `github-releases-setup` so that it is invoked after running the `install` target but before generating the documentation.

.github/bin/github-releases-setup.sh
```diff
@@ -19,6 +19,7 @@
 
 set -ex
 
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
```
hzeller, line 22: wouldn't just `SCRIPT_DIR=$(dirname $0)` work ?

releasing/build.sh
```diff
@@ -18,5 +18,4 @@ set -e
 
 cd $(dirname "$0")/..
 
-source ./.github/bin/github-releases-setup.sh /out
-./.github/bin/github-pages-setup.sh $PREFIX_BIN
+./.github/bin/github-releases-setup.sh /out
```
hzeller, line 21: Why do we need to create a `/out`, which we might not have permission to, vs. the default `-/tmp/releases` we get without the parameter ?

.github/bin/github-releases-setup.sh
AdamOlech: It wouldn't as `$0` would not contain the absolute path. Let's consider the following script:
```bash
#!/usr/bin/env bash                                                            
                                                                               
SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
echo "SCRIPT_DIR: $SCRIPT_DIR"                                                 
echo "Dollar zero: $0"                                                         
echo "Dirname dollar zero: $(dirname $0)"                                      
```
```shell
 ~/other/verible-bin .....................
> ./a/1.sh
SCRIPT_DIR: /home/adam/other/verible-bin/a
Dollar zero: ./a/1.sh
Dirname dollar zero: ./a
Hi there

```

hzeller: If you need the absolute path, you can still do `$(realpath $(dirname $0))`

But why do we need the absolute path ? You are not changing the directory.

releasing/build.sh
AdamOlech: This `/out` path is also referenced in https://github.com/chipsalliance/verible/blob/master/releasing/docker-run.sh#L136.

.github/bin/github-releases-setup.sh
AdamOlech: Now that I thought about it, we don't really need the absolute path as just like you said we're not changing the current directory anywhere. I have just replaced that bit with `SCRIPT_DIR=$(dirname $0)`.

AdamOlech:
Looks like the pipeline has passed successfully -- https://github.com/chipsalliance/verible/actions/runs/3713185953. I tidied up git history so this should be ready for merging.

hzeller (approved):

PR #1566, opened 2022-12-16
Make sure to include all the used depdnencies in the BUILD file.

hzeller:
The BUILD file was not containing all the dependencies yet that the header includes of the verilog-language-server needed.

While at it, move some headers that are only needed in the compilation unit there.

Encourage the formatter to format the closure wrapping the read() function to look more pleasing by adding a EOL comment at a strategic place.


glatosinski (approved):
LGTM

PR #1567, opened 2022-12-16
Fix line continuation wrapping

### LINKED ISSUE #1055: Fuzzing input crashes formatter
With a fuzzer-generated random code:

`cat /tmp/fuzz-crash.sv`
```SystemVerilog
bind bi i FP(\
`undef d\
`undef d,)
;
```
... we can crash the formatter

```
bazel-bin/verilog/tools/formatter/verible-verilog-format /tmp/fuzz-crash.sv
```
This check-fails with
```
[align.h:426] Check failed: leading_tokens.empty() || first_tree_token_it == ftokens.end() || first_tree_token_it->before.break_decision != SpacingOptions::MustWrap
```

We should deal with that more gracefully as user-input should never crash the formatter (use-case: if enabled in the Language Server, we should refuse to format, but not crash the server)

jbylicki:
This PR fixes #1055 

There was a problem where the `TK_LINE_CONT` tokens were not included while `LookAheadBeyondCurrentLeaf()` was running causing the `\` characters were not attached to their lines in the test case eg. `bind bi i FP(\` was trying to get converted to 
```
bind bi i FP(
    \
```
if the check that failed was to be disabled. This has made it attach itself to the line, no longer failing the condition `first_tree_token_it->before.break_decision != SpacingOptions::MustWrap`

There is also a test that had failed before the fix and now it doesn't.

### DIFF
--- a/verilog/formatting/tree_unwrapper.cc
+++ b/verilog/formatting/tree_unwrapper.cc
@@ -469,7 +469,8 @@ void TreeUnwrapper::LookAheadBeyondCurrentLeaf() {
   while (!NextUnfilteredToken()->isEOF()) {
     EatSpaces();
     VLOG(4) << "lookahead token: " << VerboseToken(*NextUnfilteredToken());
-    if (IsComment(verilog_tokentype(NextUnfilteredToken()->token_enum()))) {
+    if (auto next_tok = verilog_tokentype(NextUnfilteredToken()->token_enum());
+        IsComment(next_tok) || next_tok == verilog_tokentype::TK_LINE_CONT) {
       // TODO(fangism): or IsAttribute().  Basically, any token that is not
       // in the syntax tree and not a space.
       AdvanceLastVisitedLeaf();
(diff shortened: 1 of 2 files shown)


verilog/formatting/tree_unwrapper.cc
```diff
@@ -469,7 +469,9 @@ void TreeUnwrapper::LookAheadBeyondCurrentLeaf() {
   while (!NextUnfilteredToken()->isEOF()) {
     EatSpaces();
     VLOG(4) << "lookahead token: " << VerboseToken(*NextUnfilteredToken());
-    if (IsComment(verilog_tokentype(NextUnfilteredToken()->token_enum()))) {
+    if (IsComment(verilog_tokentype(NextUnfilteredToken()->token_enum())) ||
```
hzeller, line 472: To avoid having the complicated expression to get the enum evaluate twice (and for the reader: read twice), we could to it possibly this way:

```c++
 if (auto next_tok = verilog_tokentype(NextUnfilteredToken()->token_enum());
     IsComment(next_tok) || next_tok == verilog_tokentype::TK_LINE_CONT) {  ...
```

jbylicki: Thanks for the tip, it looks way better now

hzeller (approved):
Nice.
Once the CI is finished, ready to merge.

PR #1579, opened 2023-01-03
Emacs Verilog-Mode AUTOARG expansion in the language server

kbieganski:
Adds simple AUTO statement expansion support to the language server. This feature is modeled after [Emacs' Verilog-Mode](https://www.veripool.org/verilog-mode/) AUTO expansion. For now, it is limited to only the AUTOARG statement.

The AUTO expansion is implemented as LSP code actions. Two actions are currently available: a buffer-wide expansion, which works on the entire buffer, and a range-based one, which expands all AUTO statement within the range selected by the user.

Requires #1578.

### DIFF
--- a/verilog/tools/ls/autoexpand.cc
+++ b/verilog/tools/ls/autoexpand.cc
@@ -1,0 +1,289 @@
+// Copyright 2023 The Verible Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+#include "verilog/tools/ls/autoexpand.h"
+
+#include <algorithm>
+#include <regex>
+
+#include "common/text/text_structure.h"
+#include "verilog/CST/identifier.h"
+#include "verilog/CST/module.h"
+#include "verilog/CST/port.h"
+#include "verilog/formatting/format_style_init.h"
+#include "verilog/formatting/formatter.h"
+
+namespace verilog {
+using verible::LineColumn;
+using verible::StringSpanOfSymbol;
+using verible::Symbol;
+using verible::SymbolKind;
+using verible::SymbolPtr;
+using verible::SyntaxTreeLeaf;
+using verible::SyntaxTreeNode;
+using verible::TextStructureView;
+using verible::TokenInfo;
+using verible::lsp::CodeAction;
+using verible::lsp::CodeActionParams;
+using verible::lsp::TextEdit;
+using verilog::formatter::FormatStyle;
+using verilog::formatter::InitializeFromFlags;
+
+namespace {
+// Takes a TextStructureView and generates LSP TextEdits for AUTO expansion
+class AutoExpander {
+ public:
+  // Information about an AUTO comment
+  struct AutoComment {
+    LineColumn linecol;          // Line and column where the comment starts
+    int start_offset;            // Byte offset of the comment in the buffer
+    absl::string_view contents;  // The contents of the comment
+  };
+
+  AutoExpander(const TextStructureView &text_structure)
+      : text_structure_(text_structure), context_(text_structure.Contents()) {
+    // Get the indentation from the format style
+    FormatStyle format_style;
+    InitializeFromFlags(&format_style);
+    line_end_str_ = "\n" + std::string(format_style.indentation_spaces, ' ');
+  }
+
+  // Write given port names to the output stream, under the specified header
+  // comment
+  void EmitPortDeclarations(absl::string_view header,
+                            const std::vector<absl::string_view> &port_names,
+                            std::ostream &output);
+
+  // Find the location and span of an AUTO comment given
+  absl::optional<AutoComment> FindInSpan(const std::regex &re,
+                                         absl::string_view span);
+
+  // Retrieves port names from a module declared before the given offset
+  std::set<absl::string_view> GetPortsDeclaredBefore(const Symbol &module,
+                                                     int offset);
+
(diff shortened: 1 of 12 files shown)


hzeller:
Oh, I see, this contains part of #1578

hzeller:
#1578 is submitted, can you re-base ?

verilog/tools/ls/vscode/dist/extension.js
```diff
@@ -0,0 +1 @@
...=>{"use strict";e.exports=require("os")},1017:e=>{"use strict";e.exports=require("path")},3837:e=>{"use strict";e.exports=require("util")}},t={};function n(i){var r=t[i];if(void 0!==r)return r.exports;var o=t[i]={exports:{}};return e[i].call(o.exports,o,o.exports,n),o.exports}n.d=(e,t)=>{for(var i in t)n.o(t,i)&&!n.o(e,i)&&Object.defineProperty(e,i,{enumerable:!0,get:t[i]})},n.o=(e,t)=>Object.prototype.hasOwnProperty.call(e,t),n.r=e=>{"undefined"!=typeof Symbol&&Symbol.toStringTag&&Object.defineProperty(e,Symbol.toStringTag,{value:"Module"}),Object.defineProperty(e,"__esModule",{value:!0})};var i={};(()=>{"use strict";var e=i;Object.defineProperty(e,"__esModule",{value:!0}),e.deactivate=e.activate=void 0;const t=n(9496),r=n(2847);let o;function s(){const e={command:t.workspace.getConfiguration("verible").get("path")};o=new r.LanguageClient("verible","Verible Language Server",e,{documentSelector:[{scheme:"file",language:"system-verilog"}]}),o.start()}e.activate=function(e){return t.workspace.onDidChangeConfiguration((e=>{if(e.affectsConfiguration("verible"))return o?void o.stop().finally((()=>{s()})):s()})),s()},e.deactivate=function(){if(o)return o.stop()}})(),module.exports=i})();
```
hzeller, line 1: Are these changes in `vscode/dist/` intentional ? It looks like some accidental submit.
(Related: we should look out for what we can reduce more VSCode install boilerplate; #1139 )

verilog/tools/ls/autoexpand_test.cc
```diff
@@ -0,0 +1,258 @@
...
+  output logic o;
+endmodule
+module t2(/*AUTOARG*/);
+  input logic clk;
+  input rst;
+  output logic o;
+endmodule
+)");
+}
+
```
hzeller, line 256: Can you add another test that adds an `/*AUTOARG*/` in some situation that is not in the argument place ? So nothing should be expanded there and also not upset the expander.

hzeller:
If you rebase, this PR will be easier to review as it then incorporates the other function naming changes.

hzeller (commented):
LGTM, after comments addressed.

verilog/tools/ls/vscode/dist/extension.js
kbieganski: Sorry for this!

verilog/tools/ls/autoexpand.cc
```diff
@@ -0,0 +1,253 @@
...
+std::string AutoExpander::ExpandAutoarg(
+    const Symbol &module, std::set<absl::string_view> predeclared_ports) {
+  std::vector<absl::string_view> input_port_names;
+  std::vector<absl::string_view> inout_port_names;
+  std::vector<absl::string_view> output_port_names;
+  auto ports = FindAllModulePortDeclarations(module);
+  for (const auto &port : ports) {
+    auto *const dir_leaf = GetDirectionFromModulePortDeclaration(*port.match);
+    auto *const id_leaf = GetIdentifierFromModulePortDeclaration(*port.match);
+    auto dir = dir_leaf->get().text();
```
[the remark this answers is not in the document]
kbieganski: I think a port declaration needs a direction. Could be wrong. Maybe we should handle it gracefully so the server is more resilient?

kbieganski:
Rebased, and I think I addressed all your comments. Sorry about the VS Code extension mess.

verilog/tools/ls/autoexpand.cc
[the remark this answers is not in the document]
hzeller: What if the SystemVerilog code is faulty but is still passing the parsing ? Could we then run into such a situation ?
If we can't write a test that could break it, then we might be fine. If we suspect things could remotely be problematic, then handling it defensively is probably a good option: not getting a response from the language server is better than a crashing language server, which is a long-running process.

hzeller:
Thanks!

verilog/tools/ls/autoexpand.cc
```diff
@@ -0,0 +1,289 @@
...
+  if (!all_ports) return {};
+  for (const SymbolPtr &port : all_ports->children()) {
+    if (port->Kind() == SymbolKind::kLeaf) continue;
+    // Find the identifier of the port
+    const SyntaxTreeNode &port_node = SymbolCastToNode(*port);
+    const NodeEnum tag = NodeEnum(port_node.Tag().tag);
+    const SyntaxTreeLeaf *port_id_node = nullptr;
+    if (tag == NodeEnum::kPortDeclaration) {
+      port_id_node = GetIdentifierFromPortDeclaration(port_node);
+    } else if (tag == NodeEnum::kPort) {
```
hzeller, line 141: Mutant testing find:
Changing this 1 line to
    } else if (true) {
does not cause any test exercising them to fail.

verilog/tools/ls/autoexpand.cc
```diff
@@ -0,0 +1,289 @@
...
+  const auto all_ports = GetModulePortDeclarationList(module);
+  if (!all_ports) return {};
+  for (const SymbolPtr &port : all_ports->children()) {
+    if (port->Kind() == SymbolKind::kLeaf) continue;
+    // Find the identifier of the port
+    const SyntaxTreeNode &port_node = SymbolCastToNode(*port);
+    const NodeEnum tag = NodeEnum(port_node.Tag().tag);
+    const SyntaxTreeLeaf *port_id_node = nullptr;
+    if (tag == NodeEnum::kPortDeclaration) {
+      port_id_node = GetIdentifierFromPortDeclaration(port_node);
```
hzeller, line 140: This is the one line missing in the coverage test :)

Issue #1056, opened 2021-10-27
CI: run tests with -D_GLIBCXX_DEBUG


hzeller:
@mglb found that using `-D_GLIBCXX_DEBUG` could've detected issues from #1050 (fixed in #1051) and suggested to use this in general in our CI.

This would essentially mean to add `--copt '-D_GLIBCXX_DEBUG'` to the `bazel test` call in  [build-and-test.sh](https://github.com/chipsalliance/verible/blob/master/.github/bin/build-and-test.sh#L71) script.

But bfore that, there are a few test that currently are already reporting issues, that we need to address first. Often these are related to using iterator in an potentially undefined way.

 * [ ] //common/formatting:align_test
 * [ ] //common/text:text_structure_test
 * [ ] //common/text:tree_utils_test
 * [ ] //common/util:interval_map_test
 * [ ] //common/util:vector_tree_test
 * [ ] //verilog/formatting:formatter_test
 * [ ] //verilog/formatting:formatter_tuning_test
 * [ ] //verilog/formatting:tree_unwrapper_test
 * [ ] And then, finally, we can turn on the debug glibc in the unit test

PR #1588, opened 2023-01-09
Added empty check to the children node of argument_list_

vipsharmavip:


hzeller:
Can you add a unit test that fails before and passes after ?

hzeller:
Also, don't worry about the mac/windows failures in the CI. This has been fixed in #1590,

vipsharmavip:
> Can you add a unit test that fails before and passes after ?

Sure, will add them.

hzeller:
Any progress on adding the test ?

vipsharmavip:
> Any progress on adding the test ?

Working on it, trying to debug the code, I am aware of what type of test cases were failing before this change in the formatter_test file but not sure how that data is converted into tree format in the tree_unwrapper class.

hzeller:
The [syntax tool](https://github.com/chipsalliance/verible/tree/master/verilog/tools/syntax) has debug-output flags for printing the tokens and parse tree if that helps.

hzeller:
How is the test coming along ?

vipsharmavip:
> How is the test coming along ?

can you help me with the unit test, also I don't understand how to use this - https://github.com/vipsharmavip/verible/tree/master/verilog/tools/syntax

hzeller:
The syntax tool you can invoke with
```
verible-verilog-syntax --printtree testfile.sv
```
prints the syntax tree.

With `--printtokens` will print the first step, the lexing tokens.

So with that, you can play around with a file and see what arrives there.

vitalybuka:
I am not aware about how this code work, but making test which fails consistently seems hard or impossible.
&argument_list_->Children().back() on empty container will return pointer to memory and coder will work.
Unless you enable libc++ assertions with -D_LIBCPP_ENABLE_ASSERTIONS=1 (or other checks for your STL).

With assertions there is already formatter_test.cc  hits this coder and we see failures without the patch.

vitalybuka (commented):
Why code in MergeIdentifierAndLeftParen does not need this fix?

vitalybuka:
> I am not aware about how this code work, but making test which fails consistently seems hard or impossible. &argument_list_->Children().back() on empty container will return pointer to memory and coder will work. Unless you enable libc++ assertions with -D_LIBCPP_ENABLE_ASSERTIONS=1 (or other checks for your STL).
> 
> With assertions there is already formatter_test.cc hits this coder and we see failures without the patch.

Sorry, this could be very wrong, and it's possible observe by test behavior changed with this patch.

hzeller:
If it can not really be externally triggered, then I am also fine without test (we have #1056 open to prepare running a CI, as `-D_GLIBCXX_DEBUG` won't work right out of the box).
I have to explore `-D_LIBCPP_ENABLE_ASSERTIONS=1`.

Remains @vitalybuka 's other question about `MergeIdentifierAndLeftParen()`  - is there another situation where the empty() check needs to be added ?

vipsharmavip:
> If it can not really be externally triggered, then I am also fine without test (we have #1056 open to prepare running a CI, as `-D_GLIBCXX_DEBUG` won't work right out of the box). I have to explore `-D_LIBCPP_ENABLE_ASSERTIONS=1`.
> 
> Remains @vitalybuka 's other question about `MergeIdentifierAndLeftParen()` - is there another situation where the empty() check needs to be added ?

Based on debug, all of the failed test cases were triggering GroupIdentifierCommentsAndLeftParen(), so my suggestion is to do go with this change only, if we don't know what type of test case will fail in this method MergeIdentifierAndLeftParen() due this line

r_paren_ = r_paren_is_in_arg_list ? &argument_list_->Children().back()
                                        : NextSibling(*argument_list_);

hzeller (approved):
LGTM

PR #1574, opened 2022-12-29
Add more variants to ClassPropertyColumnSchemaScanner

msierszulski:
Fixes #1146

With this PR `ivtest/ivltests/packed_dims_invalid_class.v` no longer fails. Also, it is formatted correctly.
I also added some test cases for this case.

Issue #1146, opened 2022-01-15
ivtest/ivltests/packed_dims_invalid_class.v CHECK-fails in align.cc


hzeller:
A new version of ivtest causes crashes in the formatter:

```
bazel build -c dbg verilog/tools/formatter:verible-verilog-format
git clone https://github.com/steveicarus/ivtest /tmp/ivtest
(cd /tmp/ivtest ; git checkout -q 39ca0211074cff9aaf7500145c44410c17d40f29)

bazel-bin/verilog/tools/formatter/verible-verilog-format /tmp/ivtest/ivltests/packed_dims_invalid_class.v
F20220114 22:49:49.827275 136380 align.cc:438] Check failed: aggregate_column->Children().size() > index_entry->second.back() (0 vs. 1) 
*** Check failure stack trace: ***
    @     0x556eca1cb17e  google::LogMessage::SendToLog()
    @     0x556eca1c9416  google::LogMessage::Flush()
    @     0x556eca1cb669  google::LogMessageFatal::~LogMessageFatal()
    @     0x556eca1449ae  verible::ColumnSchemaAggregator::CollectColumnsTree()
    @     0x556eca14458c  verible::ColumnSchemaAggregator::CollectColumnsTree()
    @     0x556eca148988  verible::AlignablePartitionGroup::CalculateAlignmentSpacings()
    @     0x556eca14cd5e  verible::AlignablePartitionGroup::Align()
    @     0x556eca14d0ce  verible::TabularAlignTokens()
    @     0x556eca12268e  verilog::formatter::TabularAlignTokenPartitions()
    @     0x556eca10b3b2  verilog::formatter::Formatter::Format()
    @     0x556eca10d782  verilog::formatter::FormatVerilog()
    @     0x556eca10de5f  verilog::formatter::FormatVerilog()
    @     0x556eca0f6ce0  main
    @     0x7f07dc8c67ed  __libc_start_main
    @     0x556eca0fa61a  _start
    @              (nil)  (unknown)
Aborted
```

PR #1591, opened 2023-01-09
Document fixed bug in smoke test.

hzeller:
Bug #1146 was fixed in #1574. So we can remove the exemption in the smoke test.


hzeller:
cc: @msierszulski

PR #1595, opened 2023-01-13
Simplify syntax of attributes

### LINKED ISSUE #1333: formatter thinks always @( * ) gets lexically different output
verible_verilog_format thinks there are lexical differences between output and input when always @* statement has whitespace inside the parentheses (i.e. `always @( * )` vs `always @(*)`). Whitespace shouldn't matter here.

**Test case**

```systemverilog
module inv (
    input  logic a,
    output logic b
);

    always @( * ) begin
        b = ~a;
    end

endmodule

```

No options or configurations.
`verible_verilog_format test_case.sv`

**Actual output**

```shell
module inv (
    input  logic a,
    output logic b
);

    always @( * ) begin
        b = ~a;
    end

endmodule
test_case.sv: Formatted output is lexically different from the input.    Please file a bug.  Details:
Mismatch in token sequence lengths: 26 vs. 24
Mismatched token enums.  got: ('(') (#40: "(") vs. ('*') (#42: "(*)")
First mismatched token [14]: ('(') (#40: "(") vs. ('*') (#42: "(*)")
module inv (
    input  logic a,
    output logic b
);

  always @(*) begin
    b = ~a;
  end

endmodule
<<EOF>>
```

msierszulski:
The problem with #1333 was because syntax of 
```
always @(*) begin
```
were different from:
```
always @( * ) begin
```

CST from the first case:
```
Node @0 (tag: kEventControl) {
  Leaf @0 (#'@' @26-27: "@")
  Leaf @1 (#'*' @27-30: "(*)")
}
```
and from the second:
```
Node @0 (tag: kEventControl) {
  Leaf @0 (#'@' @26-27: "@")
  Node @1 (tag: kParenGroup) {
    Leaf @0 (#'(' @27-28: "(")
    Leaf @1 (#'*' @29-30: "*")
    Leaf @2 (#')' @31-32: ")")
  }
}
```

We want such cases to always split `()` into kParenGroup. So this PR removes the syntax case of `(*)` as `*` in `attributes`. This makes the syntax of such structures good and further formatting works well.
It also adds some tests that fail on the current master but work with this PR.

Fixes #1333.

### DIFF
--- a/verilog/CST/verilog_matchers.h
+++ b/verilog/CST/verilog_matchers.h
@@ -286,6 +286,10 @@ static const auto AlwaysStatementHasEventControlStar =
     verible::matcher::MakePathMatcher(
         {N(kProceduralTimingControlStatement), N(kEventControl), L('*')});
 
+static const auto AlwaysStatementHasParenthesis =
+    verible::matcher::MakePathMatcher({N(kProceduralTimingControlStatement),
+                                       N(kEventControl), N(kParenGroup)});
+
 // Matches occurrence of the 'always' keyword.
 // This is needed to distinguish between various kAlwaysStatement's.
 // This matches 'always', but not 'always_ff', nor 'always_comb'.
--- a/verilog/CST/verilog_matchers_test.cc
+++ b/verilog/CST/verilog_matchers_test.cc
@@ -511,13 +511,31 @@ TEST(VerilogMatchers, AlwaysStatementHasEventControlStarTests) {
       {AlwaysStatementHasEventControlStar(),
        EmbedInModule("always @* begin a = b; end"), 1},
       {AlwaysStatementHasEventControlStar(),
-       EmbedInModule("always @(*) begin a = b; end"), 1},
-      {AlwaysStatementHasEventControlStar(),
        EmbedInModule("always @(posedge foo) begin a <= b; end"), 0},
       {AlwaysStatementHasEventControlStar(),
        EmbedInModule("always_ff begin a <= b; end\n"
                      "always_comb begin a = b; end"),
        0},
+  };
+  for (const auto& test : tests)
+    verible::matcher::RunRawMatcherTestCase<VerilogAnalyzer>(test);
+}
+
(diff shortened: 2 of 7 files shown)


msierszulski:
CI fails, however, this is unrelated to the changes made in PR. It seems that http://archive.ubuntu.com/ubuntu/dists/focal/InRelease servers are not working.

msierszulski:
I am changing back to draft, as there are several tests that fail and I missed them.

hzeller:
Maybe @fangism can give a bit of context why this was there in the first place if he remembers: it looks deliberate.

msierszulski:
@hzeller This PR is now ready for review. I had to remove some tests as the syntax and the values expected by them changed. However, I also added new tests in the `verilog_matchers` and in the `formatter`.

I also had to remove some tests from `kAtrributeTests`, I am not sure if they were correct. Before the changes they were passing, with my changes they are not passing. I mean the tests `(* x)*)`, `(* ** *)`, `(** **)`. I did not find any such instructions in the manual from systemverilog. Those that I considered correct because they are in the manual I left and they pass the tests correctly.

fangism (commented):
I confess I I'm no authority on the correct way to treat this part of the language.  The initial implementation's roots can be traced back to iverilog's parser many years ago.  I don't remember making any conscious decision about it.  

The way I think about it is that `(*` _may be_ the start of an attribute, and looking ahead is the only way to resolve it.  The current parser doesn't even see attributes, they are filtered out like comments, because ignoring them entirely was the easiest thing to do at the time.  To avoid grammar conflicts, I think it is easiest to handle attributes separately in the lexical token stream, and attach them to the syntax tree in a separate pass.

hzeller:
Unfortunately, this creates a regression in at least one project I checked.

Attributes that contain parenthesis are affected, e.g.

```systemverilog
(* foo="bar()" *)
module top ();
endmodule
```

I'll revert this in #1623 and then we can go back and refine the solution

PR #1597, opened 2023-01-13
Added delay tracking to verilog_lexical_context

### LINKED ISSUE #1596: Right arrow token is misclassified when there is a delayed event trigger
**Describe the bug**
When using `verible-verilog-syntax`, `LexicalContext::_InterpretToken` misclassifies `_TK_RARROW` as a logical implication when having a delayed event trigger.

**To Reproduce**

```systemverilog
module foo;
initial
begin
  #5 -> event_bar;
end
endmodule
```

Include any options used.

**Actual behavior:**
```
test4.sv:4:6-7: syntax error at token "->"
```
I have found that it misinterprets that token as a logical implication

**Expected behavior**

In `A.6.5 Timing control statements`, a `procedural_timing_control_statement` is defined to have a `procedural_timing_control` (which `#5` fulfills later down the line) and a `statement_or_null` afterwards (which `event_trigger` staisfies). Cross-checking against `slang` shows that this is valid syntax as well.

jbylicki:
This PR implements tracking of the delays in the context of `initial/always/final` constructs, where when delaying a statement which was an event trigger, the `_TK_RARROW` would be misinterpreted as an implication. This adds another condition to the `ExpectingStatement()` method such that after a delay it returns `true`.

Fixes #1596

### DIFF
--- a/verilog/parser/verilog_lexical_context.cc
+++ b/verilog/parser/verilog_lexical_context.cc
@@ -567,6 +567,7 @@ void LexicalContext::_UpdateState(const TokenInfo& token) {
       if (in_initial_always_final_construct_) {
         // Exit construct for single-statement bodies.
         // e.g. initial $foo();
+        seen_delay_value_in_initial_always_final_construct_context = false;
         if (block_stack_.empty()) {
           in_initial_always_final_construct_ = false;
         }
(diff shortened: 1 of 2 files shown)


hzeller (approved):
LGTM

PR #1599, opened 2023-01-16
Test of Bazel installation on arm64 architecture

RRozak:
This PR is created only to test Bazel installation. Actual fix will be created later.


RRozak:
Bazel seems to be installed correctly, but all tests failed. I will continue the work in this PR: https://github.com/chipsalliance/verible/pull/1600/.

Issue #1549, opened 2022-12-06
A `//verilog_format:off` without newline will crash the formatter.


hzeller:
**Test case**

A `//verilog_format:off` without newline will crash the formatter.

See attached file 
[testcase-6208293258002432.gz](https://github.com/chipsalliance/verible/files/10168031/testcase-6208293258002432.gz)

```
bazel-bin/verilog/tools/formatter/verible-verilog-format <(zcat /tmp/testcase-6208293258002432.gz)
F1206 07:57:35.292728  124996 interval_set.h:235] Check failed: interval.valid() 
*** Check failure stack trace: ***
    @     0x55e9117cfb19  absl::log_internal::LogMessage::SendToLog()
    @     0x55e9117d00c9  absl::log_internal::LogMessageFatal::~LogMessageFatal()
    @     0x55e91170fcff  verible::IntervalSet<>::Add()
    @     0x55e911732da4  verilog::formatter::DisableFormattingRanges()
    @     0x55e91171c0af  verilog::formatter::Formatter::Format()
    @     0x55e91171dd17  verilog::formatter::FormatVerilog()
    @     0x55e91171e385  verilog::formatter::FormatVerilog()
    @     0x55e91170bbb0  main
    @     0x7f8c7682920a  (unknown)
Aborted
```

kbrunham-intel:
@hzeller I retested the reduced case from this issue with `verible-verilog-format` built from tag `v0.0-4294-gc1d8f5e8`. I could not reproduce the problem (formatter exits 0; output keeps the expected layout / converges). I would suggest closing this issue.

A file whose only contents are `//verilog_format:off` (no trailing newline) no longer CHECK-fails / aborts.

PR #1608, opened 2023-01-18
Fix adding disable range starting after EOF

j-kow:
Addresses the #1549 issue.
This PR fixes the case when `formatter::DisableFormattingRanges` started disabling range with the offset greater than text length, which happened when `//verilog_format:off` was the last line without newline character at the end.


tgorochowik (approved):

hzeller (approved):
Thanks

Issue #1598, opened 2023-01-15
`include compiler directive parse error


hankhsu1996:
**Describe the bug**

This is an example in [SystemVerilog-2017 Standard (LRM)](https://ieeexplore.ieee.org/document/8299595) page 675 (section 22.4 `include) but Verible is not able to parse it.

I'm using the latest version of Verible (v0.0-2577-g3daabcac)

**To Reproduce**

```systemverilog
`include <List.vh>
```

**Actual behavior:**
`-: <stdin>:13:10: syntax error at token "<"`

**Expected behavior**
Should be able to parse it.

PR #1612, opened 2023-01-18
Allow for angle brackets to be used around `include files.

hzeller:
Fixes: #1598


verilog/parser/verilog.lex
```diff
@@ -920,6 +925,21 @@ zi_zp { UpdateLocation(); return TK_zi_zp; }
   return TK_OTHER;
 }
 
+<PP_EXPECT_INCLUDE_FILE>{
+  {StringLiteral} {
```
fangism, line 929: Just as note, as an alternative, you could enter different string sub-states to accumulate string contents (`yymore`), as opposed to using a single expression.  The advantage of doing so is that it can give better diagnostics, such as "Unterminated string".  This could be follow-up work that includes lex-fail tests.

verilog/parser/verilog.y
```diff
@@ -230,6 +230,8 @@ is not locally defined, so the grammar here uses only generic identifiers.
 // gen_tokenizer start STRING_LITERAL
 %token TK_StringLiteral         // STRING
 %token TK_EvalStringLiteral     // STRING
+%token TK_AngleBracketInclude
```
fangism, line 233: add comment `// STRING`, just to make it more obvious this is intended as a string

fangism (approved):

verilog/parser/verilog.lex
hzeller: Right now, I am just re-using the already existing `StringLiteral`.
For now, I added a TODO where it is defined for future consideration.

verilog/parser/verilog.y
hzeller: Done.

PR #1615, opened 2023-01-20
Interpret 'null' json values as empty values in optional fields.

hzeller:
Otherwise, casting a 'null' value to the final object (e.g. a string) will result in an exception. This handles it as if it was a non-set value.


hzeller:
Reason: my  vscode was sending `null` for `rootPath` and `rootUri`, so the server was crashing while initializing.

TODO: this needs a test, but I wanted to submit it now so that it won't interrupt symbol table work.

hzeller:
@glatosinski : I am submitting now, so that you're not blocked when testing with editors that return null for the fields.

PR #1617, opened 2023-01-20
Emacs' Verilog-Mode's AUTOINST expansion in the language server

kbieganski:
Introduces Emacs' Verilog-Mode's `AUTOINST` expansion to Verible's language server. Builds on top of the AUTO expansion functionality introduced in #1579. Includes some improvements to and significant refactoring of the `AUTOARG` implementation.

Supports a basic version of `AUTO_TEMPLATE` (no regex matching and related features, no expression evaluation).

~~**Requires the symbol table handler from #1530; only the last two commits here are actually related to `AUTOINST` (will rebase once that is merged).**~~


hzeller:
Alright, #1530 is merged, this can be rebased.

kbieganski:
Rebased.

kbieganski:
Fixed formatting.

verilog/tools/ls/verible-lsp-adapter.h
```diff
@@ -37,7 +38,8 @@ std::vector<verible::lsp::CodeAction> GenerateLinterCodeActions(
 
 // Generate all available code actions.
 std::vector<verible::lsp::CodeAction> GenerateCodeActions(
-    const BufferTracker *tracker, const verible::lsp::CodeActionParams &p);
+    SymbolTableHandler &symbol_table_handler, const BufferTracker *tracker,
```
hzeller, line 41: In this project, the preference is to pass things that can be modified as pointer rather than a non-const reference. This improves readability as it is immediately clear that something can be modified.

(exception: `std::ostream` and related as this is somewhat established and expected in all kinds of operator<< situations).

verilog/tools/ls/autoexpand.cc
```diff
@@ -18,10 +18,14 @@
 #include <algorithm>
 #include <regex>
 
+#include "absl/container/flat_hash_map.h"
 #include "common/text/text_structure.h"
+#include "verilog/CST/declaration.h"
```
hzeller, line 23: Did you add all the necessary CST dependencies in the BUILD file ? It might be that they link by accident, but we should be explicit.

verilog/tools/ls/autoexpand.cc
```diff
@@ -89,33 +154,156 @@ class AutoExpander {
   // Token context created from the buffer contents
   const TokenInfo::Context context_;
 
+  // Symbol table wrapper for the language server
+  SymbolTableHandler &symbol_table_handler;
```
hzeller, line 158: Since this is something we potentially modify, make it a pointer.  In Verible we like to emphasize if some field points to something that can be modified (pointer) or something only read (const reference).

verilog/tools/ls/autoexpand.cc
```diff
@@ -89,33 +154,156 @@ class AutoExpander {
...
+//              opening paren
+// \s*\(\s*   – optional whitespace, opening paren, optional whitespace again
+// ([^\s(]+?) – second group, same as the first one
+// \s*\)*     – optional whitespace, closing paren
+const std::regex AutoExpander::autotemplate_conn_re_{
+    R"(\.\s*([^\s(]+?)\s*\(\s*([^\s(]+?)\s*\))"};
+
+const AutoExpander::Template *AutoExpander::Template::from(
+    const Map &templates, const absl::string_view type_id,
+    const long int instance_offset) {
```
hzeller, line 206: (same comment as above: instead of using long int, use `int64_t` from `<cstdint>` to emphasize that you want to potentially use large values (`long int` is defined differently on different platforms).

verilog/tools/ls/autoexpand.cc
```diff
@@ -89,33 +154,156 @@ class AutoExpander {
...
+  const auto templates_it = templates.find(type_id);
+  const Template *matching_tmpl = nullptr;
+  if (templates_it != templates.end()) {
+    // Linear search for the matching template (there should be very few
```
hzeller, line 210: Thanks for adding annotations like that, it directly addresses the concerns someone reading the code has, improving reading speed.

verilog/tools/ls/autoexpand.cc
```diff
@@ -89,33 +154,156 @@ class AutoExpander {
...
+void AutoExpander::Module::EmitPortDeclarations(
+    std::ostream &output, const absl::string_view indent,
+    const absl::string_view header,
+    const std::function<bool(const Port &)> &pred) const {
+  bool first = true;
+  for (const Port &port : ports_) {
+    if (!pred(port)) continue;
+    if (first) {
+      if (output.tellp() != 0) output << ',';
+      output << '\n' << indent << "// " << header << '\n' << indent;
```
hzeller, line 229: Looks like you try to properly match the formatting style.

It might be worthwhile (maybe later) to add a unit test that pulls in the formatter from other parts of Verible and runs the resulting template code through it to verify that input and output is the same. Low priority at this point I think, so in this PR adding a TODO might be sufficient.

verilog/tools/ls/autoexpand.cc
```diff
@@ -89,33 +154,156 @@ class AutoExpander {
...
+           << indent << '.' << port.name << '(' << connected_name << ")";
   }
 }
 
-absl::optional<AutoExpander::AutoComment> AutoExpander::FindInSpan(
-    const std::regex &re, const absl::string_view span) {
-  std::match_results<const char *> match;
+void AutoExpander::Module::RetrieveModuleHeaderPorts(const Symbol &module) {
+  const auto module_ports = GetModulePortDeclarationList(module);
+  if (module_ports) {
```
hzeller, line 264: Since this is embraching the whole rest of the function, reducing the cyclomatic complexity with an early out here will improve readability.

```c++
if (!module_ports) return;
```

verilog/tools/ls/autoexpand.cc
```diff
@@ -89,33 +154,156 @@ class AutoExpander {
...
+    for (const SymbolPtr &port : module_ports->children()) {
+      if (port->Kind() == SymbolKind::kLeaf) continue;
+      const SyntaxTreeNode &port_node = SymbolCastToNode(*port);
```
hzeller, line 267: Thanks for using `const` where possible! That helps the reader to not worry if values will be re-assigned later.

hzeller (commented):
I had a first pass. Looks very good so far!

verilog/tools/ls/autoexpand.cc
kbieganski: I get the indentation from the formatting style. It would actually be better to get it from the file itself, but I'm not sure it's worth it.

verilog/tools/ls/autoexpand.cc
kbieganski: Missed it, I added them now

verilog/tools/ls/autoexpand_test.cc
```diff
@@ -20,50 +20,88 @@
...
   BufferTracker tracker;
-  tracker.Update("<<test>>", buffer);
-  std::vector<TextEdit> edits = editFun(&tracker);
+  tracker.Update(TESTED_FILENAME, buffer);
+  // Create a Verilog project with the given project file contents
+  const std::shared_ptr<VerilogProject> proj =
+      std::make_shared<VerilogProject>(".", std::vector<std::string>());
+  size_t i = 0;
+  for (const absl::string_view file_contents : project_file_contents) {
+    auto filename = absl::StrCat("<<project-file-", std::to_string(i), ">>");
```
hzeller, line 48: You don't need `std::to_string()` when using StrCat() - it already does all necessary conversion.

hzeller (commented):
LGTM, ready to submit after one slight comment change and removing `std::to_string()`.

Separate: we should probably find a place to describe these AUTO* features.

verilog/tools/ls/autoexpand_test.cc
```diff
@@ -163,118 +198,715 @@ module t(/*AUTOARG*/
...
+  input logic i2;
   output logic o1;
   output logic o2;
 endmodule)");
 }
 
+TEST(Autoexpand, AUTOINST_ExpandEmpty) {
+  TestTextEdits(GenerateAutoExpandTextEdits,
+                R"(
+module bar(input i1, output o1);
```
hzeller, line 234: What happens if we have two conflicting versions of 'bar' , say one with two and one with three parameters. What happens ? (a) see the conflict and refuse to AUTOINST, (b) take the first or last known one (c) crash ?
Might be worthwhile a test.
I think (a) would be most desirable, though (b) is probably an initially acceptable solution as well.

verilog/tools/ls/autoexpand.cc
```diff
@@ -42,45 +47,108 @@ using verilog::formatter::FormatStyle;
...
+    static const Template *from(const Map &templates, absl::string_view type_id,
+                                int64_t instance_offset);
   };
 
-  AutoExpander(const TextStructureView &text_structure)
-      : text_structure_(text_structure), context_(text_structure.Contents()) {
+  // Module information relevant to AUTO expansion
+  class Module {
+   public:
+    // Writes all port names to the output stream, under the specified heading
```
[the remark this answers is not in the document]
kbieganski: Sorry, missed it

verilog/tools/ls/autoexpand_test.cc
kbieganski: Added a test. (b) is what happens, due to `SymbolTableHandler` only being able to find one definition for now.

I think we can revisit this once we're able to return multiple locations for `textDocument/definition`. We could use LSP's `window/showMessage` to display an error for the user.

verilog/tools/ls/autoexpand.cc
```diff
@@ -89,33 +157,155 @@ class AutoExpander {
...
+// \s*\)*     – optional whitespace, closing paren
+const std::regex AutoExpander::autotemplate_conn_re_{
+    R"(\.\s*([^\s(]+?)\s*\(\s*([^\s(]+?)\s*\))"};
+
+const AutoExpander::Template *AutoExpander::Template::from(
+    const Map &templates, const absl::string_view type_id,
+    const int64_t instance_offset) {
+  const auto templates_it = templates.find(type_id);
+  const Template *matching_tmpl = nullptr;
+  if (templates_it != templates.end()) {
```
hzeller, line 212: Here is another example of potentially early-out, reducing the indent level from below.
```
  if (templates_it == templates.end()) return nullptr;
```
(good for now, just to keep in mind for future passes over this code)

hzeller (approved):

PR #1619, opened 2023-01-20
Simplify language server shellscript test.

hzeller:
Since now the functional tests moved to verilog-language-server_test, the shellscript test can be reduced to a simple round-trip test. This removes a lot of hard-to-understand spaghetti-json and simplifies it a lot.

### DIFF
--- a/verilog/tools/ls/verible-verilog-ls_test.sh
+++ b/verilog/tools/ls/verible-verilog-ls_test.sh
@@ -13,6 +13,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# This test provides an end-to-end test for the server.
+# Simple round-trip
+#  - 1: initialize
+#  - event: open file
+#  - 2: request document symbol on that file
+#  - event: close file
+#  - 3: attempt to request documenet symbol on that closed file.
+#  - 100: shutdown.
+#
+#
+# Functional tests should go to verilog-language-server_test.cc
+
 set -u
 
 [[ "$#" == 2 ]] || {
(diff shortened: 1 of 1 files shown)


verilog/tools/ls/verible-verilog-ls_test.sh
```diff
@@ -43,38 +55,17 @@ MSG_OUT=${TEST_TMPDIR:-/tmp/}/test-lsp-out-msg.txt
...
-{"jsonrpc":"2.0","method":"textDocument/didOpen","params":{"textDocument":{"uri":"file://mini.sv","text":"module mini();\nendmodule"}}}
-
-# Requesting a code-action exactly at the position the EOF message is reported.
-# This is an interesting special case, as the missing EOF-newline is an empty
-# range, yet it should be detected as overlapping with that diagnostic message.
-{"jsonrpc":"2.0", "id":10, "method":"textDocument/codeAction","params":{"textDocument":{"uri":"file://mini.sv"},"range":{"start":{"line":1,"character":9},"end":{"line":1,"character":9}}}}
-{"jsonrpc":"2.0","method":"textDocument/didChange","params":{"textDocument":{"uri":"file://mini.sv"},"contentChanges":[{"range":{"start":{"character":9,"line":1},"end":{"character":9,"line":1}},"text":"\n"}]}}
-{"jsonrpc":"2.0", "id":11, "method":"textDocument/documentSymbol","params":{"textDocument":{"uri":"file://mini.sv"}}}
+# Close the file
 {"jsonrpc":"2.0","method":"textDocument/didClose","params":{"textDocument":{"uri":"file://mini.sv"}}}
```
glatosinski, line 65: According to the description above we should do `didOpen`, `documentSymbol`, `didClose` and `documentSymbol` on the same file. Currently, we open and request symbol for `simple.sv` file, while closing and re-requesting symbol for `mini.sv`. I guess we should use the same file name for all requests.

glatosinski (changes requested):
In my opinion we should change the names of files in last requests. Except for this, the rest looks OK

verilog/tools/ls/verible-verilog-ls_test.sh
hzeller: Oh, indeed. Good find, thanks for the thorough review.
Fixed.

PR #1620, opened 2023-01-21
Update DocumentSymbolRequest to cover all document filler features.

hzeller:
The coverage of the document-symbol-filler had been poor due to it being hard-to-test with the shell-script test. Now this can be tested much nicer programmatically.

While at it, replace a few ASSERT with EXPECT


hzeller:
Submitting now, so it will not block you to rebase on Monday.

PR #1625, opened 2023-01-24
Refine test and documentation of JoinPath().

hzeller:
Also fix a case in which an existing absolute path would become relative if the base directory to prepend was an empty string.


hzeller:
I was playing with the symbol table and noticed a crash depending on if the files in the file list were written as `foo.sv` or `./foo.sv`. I suspected an issue with JoinPath(), but turns out that is not the problem.

Anyway, while at it, I've improved the documentation and testing of that function. Also found and fixed a corner-case that was handled poorly.

fangism (approved):

PR #1626, opened 2023-01-24
Utilize absl::Duration formatting to print times in human readable form.

hzeller:
Code that runs faster than a millisecond was otherwise just printed as 0ms.


hzeller:
Since we came across this logfile readability improvement recently, here are some other places I found.

glatosinski (approved):
LGTM

PR #1627, opened 2023-01-24
Don't use Kate symbol mapping by default. Choose a mapping for VSCode.

hzeller:
VSCode is a larger user-base, so by default don't enable symbol mapping from the document outline to match Kate.

However, to look good in vscode, another hack was needed: the icon for 'module' looks like a generic namespace icon, which is already pretty common with labelled blocks. So instead, emit a symbol whose representation looks 'nicer' in vscode.

TODO: at some point, add some editor-specific mappings that then attempt to map correctly from the initailzation string sent to the language server.


glatosinski (approved):
Seems legit

PR #1623, opened 2023-01-24
Revert #1595 as it creates regressions.

hzeller:
Attributes that contain parenthesis are affected. Example:

```systemverilog
(* foo="bar()" *)
module top ();
endmodule
```

Signed-off-by: Henner Zeller <hzeller@google.com>

PR #1628, opened 2023-01-24
Simplify syntax of attributes

msierszulski:
Follow-up with #1595 and #1623
I changed the regex and added new tests.
cc @hzeller @fangism


verilog/parser/verilog_lexer_unittest.cc
```diff
@@ -79,13 +79,16 @@ static std::initializer_list<LexerTestData> kCommentTests = {
 // treating attributes lists as C-style comments,
 // except they are not returned as comment blocks.
 static std::initializer_list<SimpleTestData> kAttributeTests = {
-    {"(**)"},        {"(*     *)"},
```
hzeller, line 82: which of the original tests don't work anymore and why ?

We shoud be careful with removing any tests mentioned here as they probably come from a real-world example. Err on the side of adding tests but never remove them unless the original test was clearly faulty.

msierszulski: `(* x)`, `(* **  *)`, `(***)`, `(** **)` - these tests have been removed. I am not sure about their correctness because I have not found such cases anywhere. The systemverilog manual also does not mention such cases.

hzeller: I found some (small) regression in one of our internal tests. Will construct a test-case out of it tomorrow.

hzeller: Alright, in #1695 I added an example.

msierszulski: Thanks. Will look into that.

msierszulski:
@hzeller I changed the regular expression. Currently all tests are passing, I also restored some previously removed ones - `(* ** *)`, `(***)`, `(** **)`. Only the `(* x) *)` test is removed, it does not pass the tests with the current regex and I am not sure of its correctness. The unclosed parenthesis looks at least weird.

verilog/CST/verilog_matchers_test.cc
```diff
@@ -510,8 +510,6 @@ TEST(VerilogMatchers, AlwaysStatementHasEventControlStarTests) {
        EmbedInModule("always_comb begin a = b; end"), 0},
       {AlwaysStatementHasEventControlStar(),
        EmbedInModule("always @* begin a = b; end"), 1},
-      {AlwaysStatementHasEventControlStar(),
```
hzeller, line 513: Isn't `AlwaysStatementHasEventControlStar` technical still true ?

hzeller (approved):
Nice!
One question, but other than that, looks like ready to submit.

verilog/CST/verilog_matchers_test.cc
msierszulski: `AlwaysStatementHasEventControlStar` is not true because it only searches for `*` without parentheses. Previously, it looked like `(*)` was simplified in the tree to just `*`. So this test is no longer true (I can restore it and set the expected value to `false`), instead I added a new test `AlwaysStatementHasParenthesis` that checks if `(*)` and its various variations parse into parentheses and `*` separately.

hzeller: Yes, with 'technical true' I essentially meant that the meaning of event-star is still the same even though it is a slightly different syntax.

Is this something where we need to adapt the [always_comb.cc](https://github.com/chipsalliance/verible/blob/master/verilog/analysis/checkers/always_comb_rule.cc) lint rule ?

hzeller: Ready to submit, after always_comb rule question is sorted.

msierszulski:
I think I corrected what you meant. However, I was not able to do it with one matcher and so I had to add another one that checks the same thing but with parentheses. In always_comb_rule.cc I added double checking instead, (with brackets as well as without) and restored the one test case removed by this.

hzeller:
Does the always_comb_rule test needs to be updated to check for all the cases ?

msierszulski:
Sure, better to have more tests. Will add

msierszulski:
I added some tests to always_comb_rule. @hzeller can you please take a look?

hzeller (approved):

hzeller:
Thanks!

PR #1636, opened 2023-01-26
Don't use std::move through using alias, be explicit.

hzeller:
It looks like newer compiler start warning about it:

https://github.com/chipsalliance/verible/actions/runs/4012371146/jobs/6890758390#step:7:196


hzeller:
No more complaints in the corresponding build:
https://github.com/chipsalliance/verible/actions/runs/4017969412/jobs/6903003079#step:7:41

fangism (approved):
Very moving.

PR #1644, opened 2023-01-27
Split the Kythe CI action in two. Enable the Kythe verification tests.

ivan444:
New actions:
- verification tests (enabled)
- producing C++ xrefs and uploading them to the action artifacts (disabled -- still broken)


hzeller (approved):
LGTM

PR #1645, opened 2023-01-27
Fix a syntax tree problem when processing a single `verilog_syntax` directive

### LINKED ISSUE #1552: CHECK failure text_structure_view_.SyntaxTree()' Must be non-null in ``//verilog_syntax: ` line
**Test case**

Adding a single line (no newline) of `//verilog_syntax: parse-as-class-body` results in a CHECK failure.

Given this input file (fuzzer finding):
[testcase-5183768403771392.gz](https://github.com/chipsalliance/verible/files/10168286/testcase-5183768403771392.gz)

```
bazel-bin/verilog/tools/formatter/verible-verilog-format <(zcat /tmp/testcase-5183768403771392.gz)
F1206 08:22:36.305406  129456 tree_unwrapper.cc:78] Check failed: 'text_structure_view_.SyntaxTree()' Must be non-null
    @     0x55a40ec6bb19  absl::log_internal::LogMessage::SendToLog()
    @     0x55a40ec6c0c9  absl::log_internal::LogMessageFatal::~LogMessageFatal()
    @     0x55a40eba5502  absl::log_internal::DieBecauseNull()
    @     0x55a40ebf5f10  verible::TreeUnwrapper::Unwrap()
    @     0x55a40ebb815e  verilog::formatter::Formatter::Format()
    @     0x55a40ebb9d17  verilog::formatter::FormatVerilog()
    @     0x55a40ebba385  verilog::formatter::FormatVerilog()
    @     0x55a40eba7bb0  main
    @     0x7f76e822920a  (unknown)
Aborted
```

msierszulski:
Fixes a syntax tree problem when processing a single `verilog_syntax` directive returning `syntax_tree_` as `nullptr`.

Fixes #1552

### DIFF
--- a/common/text/text_structure.cc
+++ b/common/text/text_structure.cc
@@ -194,6 +194,13 @@ void TextStructureView::FocusOnSubtreeSpanningSubstring(int left_offset,
   VLOG(2) << __FUNCTION__ << " at " << left_offset << " +" << length;
   const int right_offset = left_offset + length;
   TrimSyntaxTree(left_offset, right_offset);
+
+  // Don't let the syntax tree be empty.
+  // Always return a tree with one node.
+  // This can happen when TrimSyntaxTree() yields a nullptr syntax tree
+  if (syntax_tree_ == nullptr) {
+    syntax_tree_ = MakeNode();
+  }
   TrimTokensToSubstring(left_offset, right_offset);
   TrimContents(left_offset, length);
   lazy_lines_info_.valid = false;
--- a/verilog/analysis/verilog_analyzer_test.cc
+++ b/verilog/analysis/verilog_analyzer_test.cc
@@ -353,6 +353,70 @@ TEST(AnalyzeVerilogAutomaticMode, NormalModeModule) {
   EXPECT_OK(ABSL_DIE_IF_NULL(analyzer_ptr)->ParseStatus());
 }
 
+TEST(AnalyzeVerilogAutomaticMode, StatementsModeSingleDirective) {
+  std::unique_ptr<VerilogAnalyzer> analyzer_ptr =
+      VerilogAnalyzer::AnalyzeAutomaticMode(
+          "// verilog_syntax: parse-as-statements", "<file>",
+          kDefaultPreprocess);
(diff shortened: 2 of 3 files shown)


common/text/text_structure.cc
```diff
@@ -194,6 +194,11 @@ void TextStructureView::FocusOnSubtreeSpanningSubstring(int left_offset,
   VLOG(2) << __FUNCTION__ << " at " << left_offset << " +" << length;
   const int right_offset = left_offset + length;
   TrimSyntaxTree(left_offset, right_offset);
+  // Don't let the syntax tree be empty.
+  // Always return a tree with one node.
+  if (syntax_tree_ == nullptr) {
+    syntax_tree_ = MakeNode();
```
hzeller, line 200: This _looks_ like it it is a quick fix that works around a problem that probably  could've been caught earlier if some test for `syntax_tree_ == nullptr` had happend more upstream.
Did you explore this possibility ? If it was not feasible, why ?

This does look like a very specific situation that is fixed by this, but looking more upstream could probably eliminate a whole class of potential issues.

msierszulski: Yes, you're right.
In this case (that is, the single `verilog_syntax` directive), the problem starts with the `TextStructureView::TrimSyntaxTree` function, which removes the prolog and epilog from the `syntax_tree`. In that case, there is no node or leaf associated with the original code in the tree, so the `TextStructureView::TrimSyntaxTree` function that uses `verible::TrimSyntaxTree` returns `nullptr` (because it removes the entire epilog and prolog). I can change the `verible::TrimSyntaxTree` function so that it is the one that never returns the tree as nullptr, only if `ZoomSyntaxTree` returns nullptr then this one will return a tree with one node (kUntagged).

What do you think about it @hzeller ?

hzeller: mmh, changing `TrimSyntaxTree()`, maybe even returning a bool would probably not improve the situation much.

So maybe for now, your solution is ok. Maybe add a comment pointing out that this can happen when TrimSyntaxTree() yields a nullptr syntax tree.

msierszulski: Added comment

hzeller (approved):
LGTM

PR #1656, opened 2023-01-30
Smallish cleanup: use const where possible in verilog project classes.

hzeller:


glatosinski:
Understood, the PR looks legit to me.

glatosinski (approved):
Understood, the PR looks legit to me.

PR #1657, opened 2023-01-31
Provide a simple platform-independent thread_pool.

hzeller:
This is in preparation to using it for processing files in parallel, initially for large file-lists to symbolize in the language server.

### DIFF
--- a/common/util/thread_pool.h
+++ b/common/util/thread_pool.h
@@ -1,0 +1,75 @@
+// Copyright 2017-2023 The Verible Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef VERIBLE_COMMON_UTIL_THREAD_POOL_H
+#define VERIBLE_COMMON_UTIL_THREAD_POOL_H
+
+#include <deque>
+#include <functional>
+#include <future>
+#include <thread>
+#include <vector>
+
+namespace verible {
+// Simple thread-pool.
+// Passing in functions, returning futures.
+//
+// Why not use std::async() ? That standard is so generic and vaguely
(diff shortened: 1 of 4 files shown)


common/util/thread_pool.h
```diff
@@ -0,0 +1,103 @@
...
+  ThreadPool(int thread_count) {
+    for (int i = 0; i < thread_count; ++i) {
+      threads_.push_back(new std::thread(&ThreadPool::Runner, this));
+    }
+  }
+
+  // Exit and drop all work currently being worked on.
+  ~ThreadPool() {
+    CancelAllWork();
+    for (auto t : threads_) {
```
fangism, line 45: `auto*` so it is immediately clear that it is a pointer

common/util/thread_pool.h
```diff
@@ -0,0 +1,103 @@
...
+    std::promise<T> *p = new std::promise<T>();
+    std::future<T> future_result = p->get_future();
+    auto promise_fulfiller = [p, f]() {
+      p->set_value(f());
+      delete p;
+    };
+    if (threads_.empty()) {
+      promise_fulfiller();
+    } else {
+      lock_.lock();
```
fangism, line 67: should lock/unlock here go inside a try-catch? or use RAII? (same question elsewhere)

common/util/thread_pool.h
```diff
@@ -0,0 +1,103 @@
...
+//
+// Why not use std::async() ? That standard is so generic and vaguely
+// specified that in practice there is no implementation of a policy that
+// provides a thread-pool behavior with a guaranteed upper bound of cores used
+// on all platforms.
+class ThreadPool {
+ public:
+  // Create thread pool with "thread_count" threads.
+  // If that count is zero, functions will be executed synchronously.
+  ThreadPool(int thread_count) {
```
fangism, line 36: since this is not a template class, maybe implement the ctor and dtor in the .cc file?  (These are not likely to be inline-critical)

fangism (approved):

common/util/thread_pool.h
hzeller: Done.

common/util/thread_pool.h
hzeller: Using RAII now (not super-important, as if things throw (std::verctor::push_back), we're anyway in good trouble :))

Originally done with explicit lock/unlock as we need the separate `notify_one()` outside that block, so this increases the nestedness.

hzeller: w.r.t. exceptions, I've now extended the promise fulfiller to also catch exceptions and properly pass them on to the future. We typically avoid exceptions, but there are unfortunately not all things under our control.
Also added unit test to confirm that.

PR #1662, opened 2023-01-31
Readability improvement: Put methods first; separate fields.

hzeller:
No functional change.

Having a newline between members improves readability.


glatosinski (approved):
OK, noted. Changes look legit

Issue #1669, opened 2023-02-01
Build error on Windows


LoveMyPillow:
The issue is an extension from the existed issue #958 . However, it seems that the build guidance on windows is still not released, so I open a new issue here.

I try to compile verible on windows10, developer mode as #796 mentioned, and use wsl2 to run bash on windows. When I run the command:
`bazel build -c opt --enable_runfiles //...`
or
`bazel test --test_output=errors //...`
 
the console print as below:
//================================================================
ERROR: D:/s-verible/verible/common/util/BUILD:86:8: Executing genrule //common/util:version_header [for tool] failed: (Exit 127): bash.exe failed: error executing command (from target //common/util:version_header) C:\Windows\system32\bash.exe -c ... (remaining 1 argument skipped)
/usr/bin/env: ‘bash\r’: No such file or directory
//================================================================
It seems like something about CRLF and CR(I guess, as mention on #796 ), and I don't think I am on the right way to compile this project on windows...
Thanks!

hzeller:
I've simplified a few things, can you try the latest version in git and see if the compile proceeds further ?

LoveMyPillow:
> I've simplified a few things, can you try the latest version in git and see if the compile proceeds further ?

yes, the compile proceeds further, but it meets new errors. And as to the bash shell and LLVM toolchain, I have installed them before this issue opened.

The error log is attached.

//=============================================
D:\S-Verible\verible>bazel build -c opt --enable_runfiles //...
INFO: Analyzed 1894 targets (126 packages loaded, 6330 targets configured).
INFO: Found 1894 targets...
INFO: From Executing genrule //verilog/parser:verilog_lex:
verilog/parser/verilog.lex:305: warning, rule cannot be matched
verilog/parser/verilog.lex:971: warning, rule cannot be matched
ERROR: D:/s-verible/verible/verilog/parser/BUILD:110:24: Executing genrule //verilog/parser:verilog_y_moved failed: (Exit 1): bash.exe failed: error executing command (from target //verilog/parser:verilog_y_moved) C:\Windows\system32\bash.exe -c ... (remaining 1 argument skipped)
write failed 32: Broken pipe
write failed 32: Broken pipe
write failed 32: Broken pipe
write failed 32: Broken pipe
write failed 32: Broken pipe
write failed 32: Broken pipe
write failed 32: Broken pipe
write failed 32: Broken pipe
write failed 32: Broken pipe

LoveMyPillow:
[...]

PR #1670, opened 2023-02-01
Reformulate version_header to use sed instead.

hzeller:
This way, we're less dependent on system paths of the shell.

Issue #1669


fangism (approved):

common/util/BUILD
```diff
@@ -86,9 +86,8 @@ cc_library(
 genrule(
     name = "version_header",
     outs = ["generated_verible_build_version.h"],
-    cmd = "./$(location create_version_header.sh) > $@",
+    cmd = "sed 's/^/#define VERIBLE_/' bazel-out/volatile-status.txt > $@",
```
fangism, line 89: Recommend specifying -e: `sed -e '...'`

fangism (approved):

common/util/BUILD
hzeller: Done.

PR #1676, opened 2023-02-02
Add severity to language server's diagnostics

kbieganski:
Adds support for diagnostic severity in the language server.

Before:
before

After:
after

Signed-off-by: Krzysztof Bieganski <kbieganski@antmicro.com>


hzeller (approved):
Nice!

PR #1677, opened 2023-02-02
Changed VSCode plugin language capabilities

### LINKED ISSUE #1666: Missing color scheme for verilog and fails while working with Verilog-HDL plugin
A couple of questions first

 - Your IDE/editor (e.g. vscode, emacs,...) you use with verible LSP ?
    vscode 
 - IDE version: 1-65-1
 - What other SystemVerilog plugins are active alongside ?
 : Verilog-HDL/SystemVerilog/Bluespec SystemVerilog
v1.7.0
-----

**What activity failed**
_(what functionality did not work as expected (e.g. Linting, Formatting, AUTO-expansion, go-to-definition, outline, hover,...)_
1. Without the verilog-HDL plugin the code coloring is missing.
2. While enabling the  verilog-HDL plugging the highlighting of the verilog style stops working.
**Expectation**
To support theme plugins or to provide color verilog theme.

**What actually happened**

**Test case**

```
module buff #(WIDTH=1) ();
endmodule
```

If needed, include SystemVerilog sample file(s) that show the code you want to do an operation on
```systemverilog
// Sample SystemVerilog file in case
```

**Logfiles**

If you invoke your editor with the environment variable `VERIBLE_LOGTHRESHOLD` set to 0, then useful log information will
be generated that you can include here

```bash
export VERIBLE_LOGTHRESHOLD=0
code foo.sv
```

***How to see the log output***
 - _vscode_: go to `View -> Output`, then choose `Verible Language Server` in the dropdown
 - _emacs_: there is a buffer `*verible-ls::stderr*` that contains the log output
 - _kate_: output is printed on the terminal

```txt
(paste log output here)
```

jbylicki:
This PR changes how SystemVerilog capability is being communicated to VSCode, from `system-verilog` to just `systemverilog` - it is the default way, otherwise it produces conflicts with other packages and does not load the plugin alongside them.

Fixes #1666

### DIFF
--- a/verilog/tools/ls/vscode/package.json
+++ b/verilog/tools/ls/vscode/package.json
@@ -15,14 +15,14 @@
   "contributes": {
     "languages": [
       {
-        "id": "system-verilog",
+        "id": "systemverilog",
         "extensions": [
           ".sv",
           ".v"
         ],
         "aliases": [
           "System Verilog",
-          "systemverilog",
+          "system-verilog",
           "verilog"
         ]
       }
@@ -41,7 +41,7 @@
     }
   },
   "activationEvents": [
-    "onLanguage:system-verilog",
+    "onLanguage:systemverilog",
     "onLanguage:verilog"
   ],
   "categories": [
--- a/verilog/tools/ls/vscode/src/extension.ts
+++ b/verilog/tools/ls/vscode/src/extension.ts
@@ -17,7 +17,7 @@ function initLanguageClient() {
     // Options to control the language client
     const clientOptions: vscodelc.LanguageClientOptions = {
         // Register the server for (System)Verilog documents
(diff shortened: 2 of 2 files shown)


verilog/tools/ls/vscode/package.json
```diff
@@ -15,14 +15,14 @@
   "contributes": {
```
hzeller, line 15: Can you also change the version above from 0.0.2700 to 0.0.2800 ? That way it is easy to determine what version is active when looking at the plugin settings.

hzeller (changes requested):
Wonderful!
Just one comment to make sure it is easy to identify what runs on a particular VSCode innstallation.

hzeller:
In #1680 I added automatic patching of that version number, so we can do this later.
(idea is to roughly follow the version numbering of git in case someone builds it manually.).

Issue #1682, opened 2023-02-03
SymbolTable: update incrementally ?


hzeller:
Given a large file list with only a few files changed and the need to update the SymbolTable: can we do it without having to re-index everything, but just update symbols incrementally ?

This could help for larger projects, in particular when we have to wait in the language server when calling 'go to definition' / 'find references'.

There are multiple parts to this question

  * How big of a problem is this ? Given a huge project, what percentage of time is spent on parsing vs. SymbolTable update.
  * Given a large project, what is the absolute time in each of these spent ? (Ideally, we would like to minimize 99% time for user queries to < 500ms).
  * How much complexity would this add - is it worthwhile vs. maintenance overhead ?

hzeller:
CC @fangism in case he has some ideas w.r.t. complexity as he implemented the symbol table.

hzeller:
Asked David, and his first assessment is that it might be a larger task, probably requires some careful measurement first how much this is a problem.

fangism:
One difficulty with this is that the symbol table is a monolithic accumulated structure, whose result is the accumulation of changes from processing multiple files.  There could be interactions between files as a result of one file being updated, one moved definition could alter the implicit dependency graph (of files).  For this idea to work, you would need some well-defined state of the symbol-table between processing each translation unit, and either transitions between states have to be reversible, or you'd keep around copies of old state for reversibility.  This problem maps naturally to incremental builds systems, whose primary purpose is to know precisely what subgraph needs to be redone for any update.

PR #1687, opened 2023-02-06
Reduce logging noise: Only aggregate to LOG(INFO), file-level VLOG(1).

hzeller:
For projects with many files, this filled the logs with all kinds of LOG(WARNING) that are not really relevant unless debugging. So moved all these to VLOG(1). Keep aggregate logs ("3000 files took 20 seconds to process") as LOG(INFO).

While at it:
 * Don't call BuildSymbolTable() while in LoadProjectFileList().
   It is confusing as it doesn't do anything, because it requires
   the file to be analyzed first. So just do what is needed to
   open the files into memory.
 * Remove methods only needed internally to the private section
   oof SymbolTableHandler header. Also remove unused method.
  * Separate out parsing from symbol table building to log timing independently.
  
These changes are improving timing information needed to make relevant implementation decisions in #1682.

### DIFF
--- a/verilog/analysis/verilog_project.cc
+++ b/verilog/analysis/verilog_project.cc
@@ -77,7 +77,12 @@ absl::Status VerilogSourceFile::Parse() {
 
   const absl::Time start = absl::Now();
   status_ = analyzed_structure_->Analyze();
-  VLOG(1) << "Analyzed " << ResolvedPath() << " in " << (absl::Now() - start);
+  const absl::Duration analyze_time = absl::Now() - start;
+  if (analyze_time > absl::Milliseconds(500)) {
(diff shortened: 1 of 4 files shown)


hzeller:
With these logging messages, we now have separate timings for Parse(), Build() and Resolve():

```
symbol-table-handler.cc:96] VerilogSourceFile::Parse() for 2714 files: 22.058365507s
symbol-table-handler.cc:111] SymbolTable::Build() took 5.200470899s
symbol-table-handler.cc:117] SymbolTable::Resolve() took 1.415504288s
symbol-table-handler.cc:272] Looking for symbol:  foobar
symbol-table-handler.cc:313] textDocument/definition processing time:  28.822581166s
```

glatosinski (approved):
Looks legit to me

PR #1691, opened 2023-02-07
Support AUTO expansion of declarations

kbieganski:
Adds the following Emacs Verilog-Mode features to the language server:
* `AUTOINPUT`
* `AUTOOUTPUT`
* `AUTOINOUT`
* `AUTOWIRE`
* `AUTOREG`

Also refactors existing code in `autoexpand.cc`:
* enums are now in proper case,
* got rid of `AutoExpander::Span` and `AutoExpander::FindInSpan` to simplify code,
* moved `AUTO_TEMPLATE` handling into `AutoExpander::Module`,
* in general module symbols are now usually wrapped into `AutoExpander::Module`s instead of being passed to functions directly.


verilog/tools/ls/autoexpand.cc
```diff
@@ -204,32 +312,38 @@ const std::regex AutoExpander::autotemplate_type_re_{
...
+// (?:\s*//.*\n)?                           – optional starting comment
+//                                            ("Beginning of automatic...")
+// (?: ... )?                               – an optional non-capturing group:
+//   ([\s\S]*?)?                            – any text (usually port
+//                                            declarations)
+//   (?:[^\S\r\n]*// End of automatics.*\n) – ended by an "End of automatics"
+//                                            comment
+constexpr const char *auto_declaration_re_fmt =
+    R"((/\*\s*%s\s*\*/\s*?\n)(?:\s*//.*\n)?(?:([\s\S]*?)?(?:[^\S\r\n]*// End of automatics.*\n))?)";
+const std::regex AutoExpander::autoinput_re_{
```
hzeller, line 326: mmh, global constants that require complicated set-up calls are often murky in the way things are sequenced and available (here, we assemble a global value ideally before main() with StrFormat(). And also create a complicated regex in addition to that). (the compiler/linker might not always be able to properly find a good sequence for these "Static Initialization Order Fiasco").

To solve I suggest to use the pattern where we have a function to call that then does the initialization (static variables in functions are only initialized on the first call):

```c++
const std::regex& autoinput_re() {
   static const std::regex re{absl::StrFormat(auto_declaration_re_fmt, "AUTOINOUT")};
  return re;
}
```

(also the other places where regex are statically initalized)

verilog/tools/ls/autoexpand.cc
```diff
@@ -204,32 +312,38 @@ const std::regex AutoExpander::autotemplate_type_re_{
...
+// AUTOINPUT/OUTPUT/INOUT/WIRE/REG regex breakdown:
+// (/\*\s* ... \s*\*/\s*?\n)                – starting comment
+// (?:\s*//.*\n)?                           – optional starting comment
+//                                            ("Beginning of automatic...")
+// (?: ... )?                               – an optional non-capturing group:
+//   ([\s\S]*?)?                            – any text (usually port
+//                                            declarations)
+//   (?:[^\S\r\n]*// End of automatics.*\n) – ended by an "End of automatics"
+//                                            comment
+constexpr const char *auto_declaration_re_fmt =
```
hzeller, line 324: that can be a `static constexpr absl::string_view` - that way, the length is already compile-time initialized and doesn't have to be determined with a `strlen()` at runtime.

hzeller (commented):
Nice.
Still have to go through all the test cases for a full review. Here a first comment.

verilog/tools/ls/autoexpand.cc
kbieganski: I think this is problematic only for dependencies between translation units? Also, the `StrFormat` is part of the same expression as the regex construction. It depends on a string constant which needs no initialization.

(compile-time regex compilation would be nice)

kbieganski:
Heads up that I'll have to push a fix to this code (or open a separate pull) as I missed the case of AUTOINST signals being renamed. Currently this results in new declarations being appended instead of replacing the old ones.

kbieganski:
Fixed the re-declaration issue. Also simplified the code a bit.

verilog/tools/ls/autoexpand.cc
hzeller: Unless `StrFormat()` is constexpr, I'd be wary to call it before everything is initialized.

It is a function provided by a different library that is not under our control (`absl::`), so we can't assume that it does not depend on some other global state that might not be ready yet.

You could make the string concatenation at safe without calling `StrFormat()`, something along the lines of:

```c++
#define MAKE_AUTO_REGEX(whatmatch) R"((/\*\s*)" whatmatch R"(\s*\*/\s*?\n)(?:\s*//.*\n)?(?:[\s\S]*?[^\S\r\n]*// End of automatics.*\n)?)"

const std::regex AutoExpander::autoinput_re_{MAKE_AUTO_REGEX("AUTOINPUT")};
//... more inits
#undef MAKE_AUTO_REGEX
```
One of the few cases where macros are actually kindof warranted...

Using global static std::regex is technically also not ok, because it is not `is_trivially_destructible`, which is outside of our [style guide suggestions](https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables). Though I'd say it is benign and probably something we can ignore in this case for now.

So if you can avoid the StrFormat() call, I think it's ok.

(And yes, it would be good if `std::regex` was constexpr, but alas, it throws. So unless that is fixed in the standard, we won't see that... (though I have seen people made impressive `constexpr` regular expression matchers).)

kbieganski: Seems like local statics are also disallowed by the guidelines. I think the only way to conform to them is to make `AutoExpander` a persistent object that is constructed by the language server entry point, and the regexes its fields.

I think I'll use the macro as it's more concise and consistent with the other regexes.

verilog/tools/ls/autoexpand.cc
```diff
@@ -20,18 +20,21 @@
 #include <regex>
 
 #include "absl/container/flat_hash_map.h"
+#include "absl/container/node_hash_map.h"
```
hzeller, line 23: also, flat_hash_set ?

verilog/tools/ls/autoexpand.cc
```diff
@@ -243,27 +362,265 @@ void AutoExpander::Module::EmitPortConnections(
...
+
+void AutoExpander::Module::EmitUndeclaredOutputWireDeclarations(
+    std::ostream &output, const absl::string_view indent,
+    const absl::string_view auto_span) const {
+  absl::flat_hash_set<absl::string_view> declared_wires;
+  for (const auto &reg : FindAllNetVariables(symbol)) {
+    const SyntaxTreeLeaf *const net_name_leaf =
+        GetNameLeafOfNetVariable(*reg.match);
+    const absl::string_view net_name = net_name_leaf->get().text();
+    if (net_name.end() <= auto_span.begin() &&
```
hzeller, line 412: Maybe a little helper function `static bool OutsideSpan(absl::string_view piece, absl::string_view span);` for this test might be useful. Improves readability, and you need it around 448 again.

verilog/tools/ls/autoexpand.cc
```diff
@@ -243,27 +362,265 @@ void AutoExpander::Module::EmitPortConnections(
...
+    std::ostream &output, const absl::string_view indent,
+    const absl::string_view auto_span) const {
+  absl::flat_hash_set<absl::string_view> declared_regs;
+  for (const auto &reg : FindAllRegisterVariables(symbol)) {
+    const SyntaxTreeLeaf *const reg_name_leaf =
+        GetNameLeafOfRegisterVariable(*reg.match);
+    const absl::string_view reg_name = reg_name_leaf->get().text();
+    if (reg_name.end() <= auto_span.begin() &&
+        reg_name.begin() >= auto_span.end()) {
+      declared_regs.insert(reg_name);
```
hzeller, line 450: Does any unit test run into this condition ?

verilog/tools/ls/autoexpand.cc
```diff
@@ -243,27 +362,265 @@ void AutoExpander::Module::EmitPortConnections(
...
     }
-    output << '\n'
-           << indent << '.' << port.name << '(' << connected_name << ")";
+    fun(port, connected_name);
+  }
+}
+
+void AutoExpander::Module::AddGeneratedConnection(
+    const absl::string_view port_name, const Port::Direction direction,
+    const Connection connection) {
```
hzeller, line 481: Connection is a fairly small struct, but I'd probably still pass it by `const &` not by value (mostly to not slow down readability flow - if it is passed by value the reader might stop a second and wonder if there is a particular reason).

verilog/tools/ls/autoexpand_test.cc
```diff
@@ -832,14 +832,926 @@ endmodule
...
+    .i2(i2),
+    // Inouts
+    .io(io),
+    // Outputs
+    .o1(o1),
+    .o2(o2));
+endmodule
+)");
+}
+
```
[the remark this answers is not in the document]
hzeller: Ah, I see, these are mostly the `*_ SkipPreConnected` tests.
But what if someone has written the fields after the `/*AUTOINST*/` ?

hzeller (approved):
Very nice, this is some high-quality code!
Mostly just some smallish style comments (like `k` prefix in enum constants).

The nice and exhaustive tests cover well what should work.
Have you also considered all kinds of cases where things could break, or wrong input ?For instance a user having some fields filled out already. or there are typos in fields ..
So knowing the implementation, finding all the things that might possibly hit corner cases, but still 'behave'.

So testing if broken inputs things work, or at least refuse to modify the input (and don't crash), but also demonstrate that the output is behaving in a documented way - because for future changes of that very behavior there is then already a test that we can modify.

But in general, I think this is already in a good state for submit. We always can think of more tests later.

Ah, and don't worry about `static` fields for now, I think they are ok. I was most worried in calling an external function (the strfmt) in the middle of initialization. With the macro we're good for now.

verilog/tools/ls/autoexpand_test.cc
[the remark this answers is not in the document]
kbieganski: As in Emacs, everything up to the `);` gets replaced.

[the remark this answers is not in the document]
hzeller: Perfect. Is there a test that documents that ?

verilog/tools/ls/autoexpand.cc
kbieganski: I thought I had tested this, but it seems not. And the condition is wrong, of course.

kbieganski:
Thanks.

> Have you also considered all kinds of cases where things could break, or wrong input ?

I'm sure there is some input I haven't considered that will cause trouble. The cases that should work fine are:
* Incorrect syntax, which AFAIK results in no syntax tree being generated. In that case we don't even try doing any AUTO expansion.
* putting `/*AUTOARGS*/` outside a module header, and `/*AUTOINST*/` outside a module instantiation. These have no effect outside of these contexts.
* Anything the user does between `/*AUTO<whatever>*/` and `// End of automatics`. All that is simply discarded and should not affect the current expansion. AUTO expansion is done based on only the current buffer state, so even if the user writes the whole thing by hand, we treat it the same.
* Any time the user changes something that we use for AUTO expansion (e.g. a signal that is auto-connected and then a matching signal is generated in the parent module), we regenerate anything dependent on it.

Probably not all of these are tested thoroughly, so it would still be good to expand the test suite.

One thing comes to mind that is somewhat dangerous. Let's say we have:
```
/*AUTOINPUT*/
// some auto-generated inputs
// End of automatics

... // some stuff written by the user

/*AUTOOUTPUT*/
// some auto-generated outputs
// End of automatics
```
If the user deletes the first `End of automatics`, or even part of it, anything up to the second `End of automatics` would be considered part of `/*AUTOINPUT*/`, so it would get replaced on AUTO expansion. We should probably protect against it somehow.

hzeller (approved):
Nice. More test cases of things we know should work (e.g. no changes on syntax error), and how corner cases are handled can be added as we think about them and as they come up in testing.

PR #1692, opened 2023-02-07
Fix some trailing whitespace.

hzeller:
While at it, add needed, remove unneeded dependency in  BUILD file.


hzeller:
(My emacs is configured to follow along with `${VERIBLE_ROOT}/.editorconfig` so opening and saving generated this diff.)

kbieganski (approved):

PR #1693, opened 2023-02-07
Test also with c++23

hzeller:
While we only require c++17, make sure that we don't accidentally use future-deprecated features.


hzeller:
FYI @AdamOlech

PR #1702, opened 2023-02-08
Add sv-tests to smoke tests.

hzeller:
The very small syntax tests in sv-tests make it
particularly suited to find specific issues.


hzeller:
(submitting now already to skip timezone shift; might already be useful now)

Issue #1690, opened 2023-02-07
LSP fails go-to-definition in Windows environment


viktortomov:
A couple of questions first

 - Your IDE/editor (e.g. vscode, emacs,...) you use with verible LSP ? 
 - IDE version: neovim v0.8.3
 - What other SystemVerilog plugins are active alongside ?
 - mason with verible lsp installed
-----

**What activity failed**
_(what functionality did not work as expected (e.g. Linting, Formatting, AUTO-expansion, go-to-definition, outline, hover,...)_
go-to-definition

**Expectation**
go to the specific line of the definition

**What actually happened**
Seems that the nvim lsp is giving an error due to incorrect url from the verible language server. See the last debug message in the log file. The url starts with ```file://C:\\C:\\``` where it should be  ```file:///C:/workspace/.....```

**Test case**
The cursor is under the cntr assignment and the go to function is executed.

If needed, include SystemVerilog sample file(s) that show the code you want to do an operation on
```systemverilog
// Sample SystemVerilog file in case
// Simple counter for verification purposes.

module smpl_cntr
(
    input logic clk,
    output logic cntr
);

always_ff @(posedge clk) begin
    cntr <= cntr + 1'b1;
end

endmodule
```

**Logfiles**

```txt
[START][2023-02-07 18:26:30] LSP logging initiated
[ERROR][2023-02-07 18:26:30] .../vim/lsp/rpc.lua:733	"rpc"	"C:\\Users\\victor.tomov\\AppData\\Local\\nvim-data\\mason\\bin\\verible-verilog-ls.CMD"	"stderr"	"Verible Verilog Language Server built at v0.0-2832-gffdd6671\r\n"
[ERROR][2023-02-07 18:26:30] .../vim/lsp/rpc.lua:733	"rpc"	"C:\\Users\\victor.tomov\\AppData\\Local\\nvim-data\\mason\\bin\\verible-verilog-ls.CMD"	"stderr"	"E0207 16:26:30.297135   23252 json-rpc-dispatcher.cc:58] Unhandled notification 'initialized'\r\n"
[DEBUG][2023-02-07 18:26:40] .../lua/vim/lsp.lua:1381	"LSP[verible]"	"client.request"	1	"textDocument/definition"	{  position = {    character = 6,    line = 9  },  textDocument = {    uri = "file:///C:/workspace/git/tool_validation/rtl/smpl_cntr/smpl_cntr.sv"  }}	<function 1>	1
[...]

PR #1705, opened 2023-02-09
Use proper log severity for ignored notifications.

hzeller:


hzeller:
Issues #1690 - Don't log benign situations with LOG(ERROR) as it looks like it detracts from actual potential issues.

PR #1707, opened 2023-02-09
Work around an overload that becomes ambiguous with C++20

hzeller:
Starting with c++20, there is a new constructor #5 for string_view in town:
https://en.cppreference.com/w/cpp/string/basic_string_view/basic_string_view

Together with concepts the initailization with {} can be ambiguous for either a std::vector<std::string_view> or a std::string_view{} initialization with two parameters that conceptually allow iterating.

Both functions would match, in the latter case the last string would be interpreted as the optional boolean ...

Unfortunately, I could not reproduce this in the CI (#1706 was for that), I suspect this only shows up with libstdc++ (which we're not using in the CI).

Long story short: dismbiguiated it by renaming the functions. An alternative would be to just remove one of the functions.

### DIFF
--- a/verilog/tools/ls/autoexpand_test.cc
+++ b/verilog/tools/ls/autoexpand_test.cc
@@ -29,12 +29,12 @@ using verible::lsp::TextEdit;
 
 // Generate text edits using the given function and test if they had the desired
 // effect
-void TestTextEdits(const std::function<std::vector<TextEdit>(
-                       SymbolTableHandler*, BufferTracker*)>& edit_fun,
-                   const std::vector<absl::string_view>& project_file_contents,
-                   const absl::string_view text_before,
-                   const absl::string_view text_golden,
-                   const bool repeat = true) {
+void TestTextEditsWithProject(
+    const std::function<std::vector<TextEdit>(SymbolTableHandler*,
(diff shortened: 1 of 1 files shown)


kbieganski (approved):
Rename is a good call, and I get the ambiguity arising from that constructor. But can you elaborate on the optional boolean part? Isn't it just the case of two string literals being interpreted as iterators? Which is silly BTW but that's C++.

hzeller:
It took me a while as well to see:

We have two overloaded functions, simplified:
```c++
foo(vector<string_view>, string_view, string_view, optional_bool);  // variant 1
foo(                     string_view, string_view, optional_bool);  // variant 2
```

The ambiguity is that
```
foo({ "string", "string"}, "string", "string");
```

Can be interpreted as 
```c++
foo(vector<string_view>, "string", "string");  // calling variant 1, using the default value for bool
```

or 
```c++
 // calling variant2 with superflous const char* converted to the optional boolean at the end...
foo(stringivew_constructed_from_curly_initializer, "string", implicit_cast_to<bool>("string")); 
```

The human of course would not even think of variant2 ... but the compiler is easily confused.

kbieganski:
Oh, I see. Well, like I said, renaming the function is fine. Nice catch.

PR #1715, opened 2023-02-16
Fix broken clang-tidy config; fix new warnings showing up.

hzeller:
* TIL that a #-comment in the .clang-tidy configuration silently messes up the config, resulting in ignored rules.
  * Remove the comment in our clang tidy config. It has been there since 065904fa4ac24858a96fbf055d595ee5ac34c879
  * Harden the run-clang-tidy.sh script to make sure such comments are not added.
  * Fix the fall-out of not having complete clang-tidy rules active for a few months.

### DIFF
--- a/.github/bin/run-clang-tidy.sh
+++ b/.github/bin/run-clang-tidy.sh
@@ -14,16 +14,21 @@
 # limitations under the License.
 
 set -u
+
+readonly CLANG_CONFIG_FILE=.clang-tidy
+
 TMPDIR="${TMPDIR:-/tmp}"
+readonly NAME_PREFIX=verible-clang-tidy  # Make files easy to tab-complete-find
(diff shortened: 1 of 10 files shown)


hzeller:
FIY @kbieganski as I had to modify some non-const references to be pointers (we prefer to pass mutable objects by pointer, so that it is more visible to the reader that this is something mutable; parameters that are not modified are passed as `const &`).

.github/bin/run-clang-tidy.sh
```diff
@@ -40,6 +45,32 @@ for version in 11 12 13 14 ; do
 done
 hash ${CLANG_TIDY} || exit 2  # make sure it is installed.
 
+# Explicitly test the configuration, otherwise clang-tidy will silently
+# fall back to some minimal default config.
+
+# A particular annoying thing: if there is a comment in the configuration,
+# things break. Silently.
+awk '
```
fangism, line 53: inline awk.  love it!

fangism (approved):

PR #1717, opened 2023-02-16
Some mostly mechanical cleanups.

hzeller:
Following reasonable modernize-* clang tidy suggestions. Enabling all of them, then disabling the ones that produce overzealous or possibly less readable results.

Also, fix some of google-readability-braces-around-statements suggestions, but there are more.

A lot based on clang-tidy suggestions.


hzeller:
Some `modernize-return-braced-init-list` already as well, but not entirely yet.

fangism (approved):

PR #1723, opened 2023-02-17
Use absl::string_views for lint messages.

hzeller:
LintViolation() recently had its constructor changed to accept an absl::string_view instead of a std::string; so let's use these natively when definiting the messages.


hzeller:
`s|static const char \(.*\)\[\]|static constexpr absl::string_view \1|g`, essentially.

fangism (approved):

PR #1724, opened 2023-02-17
Create path matchers as constexpr.

hzeller:
First step: make PathMatchBuilder a constexpr to build a consexpr matcher.


fangism (approved):
awesome!

fangism (approved):

PR #1727, opened 2023-02-17
Fix transitive dependencies in AUTO expansions

kbieganski:
Module sorting in AUTO expansion does not handle transitive dependencies between modules. This fix makes `DependsOn` recursively traverse all dependencies.

### DIFF
--- a/verilog/tools/ls/autoexpand.cc
+++ b/verilog/tools/ls/autoexpand.cc
@@ -187,15 +187,17 @@ class AutoExpander {
     void RetrieveAutoTemplates();
 
     // Gets all dependencies of the module (modules instantiated within it)
-    void RetrieveDependencies();
+    void RetrieveDependencies(
+        const absl::node_hash_map<absl::string_view, Module> &modules);
 
     // Retrieves the matching template from a typename -> template map
     const Template *GetAutoTemplate(absl::string_view type_id,
                                     int64_t instance_offset) const;
 
     // Returns true if the module depends on (uses) a given module
-    bool DependsOn(const absl::string_view type_name) const {
-      return dependencies_.contains(type_name);
+    bool DependsOn(const Module *module) const {
+      absl::flat_hash_set<const Module *> visited;
+      return DependsOn(module, &visited);
     }
 
     // Returns the Symbol representing this module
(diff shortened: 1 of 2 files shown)


kbieganski:
CI run failed due to GitHub issues :(

verilog/tools/ls/autoexpand.cc
```diff
@@ -220,6 +222,16 @@ class AutoExpander {
...
 
+    // Recurses into dependencies to check if we depend on a given module.
+    // Stores visited modules in a set to avoid infinite loops. For big
+    // dependency graphs one should build a proper graph and do a
+    // topological sort. However, these dependencies (and the dependent) are all
+    // from a single file, and usually there is only one module per file. This
+    // should be fast enough for unusual cases where there are multiple modules
+    // in a single file.
+    bool DependsOn(const Module *module,
+                   absl::flat_hash_set<const Module *> &visited) const;
```
hzeller, line 233: Mutable objects like the `visited` shall be passed by pointer not mutable reference.

kbieganski: Right, sorry.

verilog/tools/ls/autoexpand_test.cc
```diff
@@ -27,14 +27,22 @@ using verible::lsp::Range;
 using verible::lsp::TextDocumentContentChangeEvent;
 using verible::lsp::TextEdit;
 
+// Determines how TextTextEdits* should test a function
+enum class TestStrategy {
```
hzeller, line 31: Nice making this an enum!

hzeller (approved):

verilog/tools/ls/autoexpand.cc
```diff
@@ -775,6 +790,18 @@ void AutoExpander::Module::PutDeclaredPort(const SyntaxTreeNode &port_node) {
                     std::move(unpacked_dimensions)});
 }
 
+bool AutoExpander::Module::DependsOn(
```
hzeller, line 793: This does not seem to always adhere to strict weak ordering.
In the [`Autoexpand.AUTO_ExpandPortsMultipleConnectionsWithAUTO_TEMPLATE` ](https://github.com/chipsalliance/verible/blob/398a8505af01f61e930bd95fc2a4f1131dc0d467/verilog/tools/ls/autoexpand_test.cc#L1850)
test, a [strict weak ordering check](https://discourse.llvm.org/t/rfc-strict-weak-ordering-checks-in-the-debug-libc/70217) shows that DependsOn() does not always result in a strict weak ordering in the sort using it:

https://github.com/chipsalliance/verible/blob/398a8505af01f61e930bd95fc2a4f1131dc0d467/verilog/tools/ls/autoexpand.cc#L1448-L1451

PR #1728, opened 2023-02-17
Fix AUTO port expansion in module header

kbieganski:
Before this patch, AUTO port expansion in module headers generates incorrect syntax, as port declarations are being separated by `;`, which is only correct in module body. This change fixes that by choosing the correct delimiter based on the context.


hzeller:
If we do the `repeat` in the `TestTextEditsWithProject()` - shoudn't this catch invalid syntax as it builds a VerilogProject, which internally does a parse of the file ?

kbieganski:
It does, but AUTO ports in module headers weren't tested.

kbieganski:
I pushed an assert to check if the file was actually parsed.

hzeller (approved):
Maybe having a testing fixture manually parse all inputs and generated outputs with `VerilogAnalyzer::AnalyzeAutomaticMode()` might be a good secondary test in future refinements of the test.

hzeller:
Ha, LOL, so you already added some extra verification while I was writing my last message :)

PR #1729, opened 2023-02-17
Don't use reserved identifiers with prefix underscore

hzeller:


hzeller:
`internal::Affairs`, even.

PR #1730, opened 2023-02-17
Add loading linter rules from .rules.verible_lint

glatosinski:
This PR:

* Moves conversions of LSP URI to paths and vice versa to a separate library
* Enables checking `.rules.verible_lint` by default
* Adds mechanism in parsed buffers to check for `.rules.verible_lint` up in directory hierarchy, starting from linted file, to load linter rules
* Adds tests for verifying reactions of `VerilogLanguageServer` for `.rules.verible_lint` file

@hzeller - I am not sure about changing the default value of the `--rules-config-search` to `true`. This flag is available from the level of `verible-verilog-ls` - not sure whether I should change the flag value to `true` via `SetFlag` in a function delivered with `VerilogLanguageServer`, or change this default value directly in `verilog_linter`.

### DIFF
--- a/verilog/tools/ls/verilog-language-server.cc
+++ b/verilog/tools/ls/verilog-language-server.cc
@@ -21,6 +21,7 @@
 #include "common/lsp/lsp-protocol.h"
 #include "common/util/file_util.h"
 #include "common/util/init_command_line.h"
+#include "verilog/tools/ls/lsp-file-utils.h"
 #include "verilog/tools/ls/verible-lsp-adapter.h"
 
 namespace verilog {
--- a/verilog/analysis/verilog_linter.h
+++ b/verilog/analysis/verilog_linter.h
@@ -19,6 +19,7 @@
 #include <string>
 #include <vector>
 
+#include "absl/flags/declare.h"
 #include "absl/status/status.h"
 #include "absl/status/statusor.h"
 #include "absl/strings/string_view.h"
@@ -33,6 +34,10 @@
 #include "common/text/text_structure.h"
 #include "verilog/analysis/lint_rule_registry.h"
 #include "verilog/analysis/verilog_linter_configuration.h"
+
+// Flag is declared for testing purposes (used e.g. in
+// verilog/tools/ls/verilog-language-server_test.cc)
+ABSL_DECLARE_FLAG(bool, rules_config_search);
 
 namespace verilog {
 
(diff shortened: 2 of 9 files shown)


verilog/analysis/verilog_linter.cc
```diff
@@ -62,7 +62,7 @@ ABSL_FLAG(verilog::RuleBundle, rules, {},
 ABSL_FLAG(std::string, rules_config, "",
           "Path to lint rules configuration file. "
           "Disables --rule_config_search if set.");
-ABSL_FLAG(bool, rules_config_search, false,
+ABSL_FLAG(bool, rules_config_search, true,
```
hzeller, line 65: We can't switch this on by default, as this also affects the stand-alone tool where the default expectation currently is that this is not searching for such rules.
(we can probably change that expectation long-term, but we should not introduce a behavioral change in passing).

If you want to force this flag on in the language server (which is more reasonable), there is `absl::SetFlag()`...

glatosinski: ok, will fix it.

glatosinski: I declared the flag for `--rules-config-search` and set it to true for `VerilogLanguageServer`. We can also have it disabled by default and tell users to set this flag in their configuration of language server

verilog/tools/ls/verilog-language-server.cc
```diff
@@ -17,14 +17,20 @@
...
 #include "common/lsp/lsp-protocol.h"
 #include "common/util/file_util.h"
 #include "common/util/init_command_line.h"
+#include "verilog/tools/ls/lsp-file-utils.h"
 #include "verilog/tools/ls/verible-lsp-adapter.h"
 
 namespace verilog {
 
+void VerilogLanguageServer::UpdateFlags() {
+  absl::SetFlag(&FLAGS_rules_config_search, true);
```
hzeller, line 31: I am not convinced actually that we should do the upward search. It feels 'too magic'. And it also can fail in surprising ways in various occasions (e.g. if people have constructed their projects with symbolic links).

So for the language server I would suggest to not use the automatic searching, but look at the exact same directory  where the project root is. Either the project root configured coming from the editor or the project root established by finding `verible.filelist`.

glatosinski: I disabled automatic searching (I left a flag declaration, because I use it in LS tests).

In my opinion, having two mechanisms searching for lint configuration (enabling it with `--rules-config-search` and additional check in project root) may be also prone to errors, maybe we should highlight in README that it should be enabled in the configuration for launching this language server?

hzeller: Yes, having  multiple mechanisms will be problematic.

Anyway, I think it is good for now to merge, we can refine the behavior as we go.

PR #1733, opened 2023-02-20
Add directories to the kzip archive. Without it, some tools fail to decompress it

ivan444:


verilog/tools/kythe/kzip_creator.cc
```diff
@@ -50,7 +50,12 @@ KzipCreator::KzipCreator(absl::string_view output_path)
     : zip_file_(fopen(std::string(output_path).c_str(), "wb"), &fclose),
       archive_(kKZipCompressionLevel, [this](absl::string_view s) {
         return fwrite(s.data(), 1, s.size(), zip_file_.get()) == s.size();
-      }) {}
+      }) {
+  // Create the directory structure.
+  archive_.AddFile("root/", verible::zip::MemoryByteSource(""));
```
hzeller, line 55: Cool, this works ? I would've expected that we need to create a separate file type of sorts in our zip implementation.

hzeller (approved):

verilog/tools/kythe/kzip_creator.cc
ivan444: It works (confirmed by `unzip`, `kzip info` and `zipinfo -vh`). It's important that the name ends with "/". Othewise it's not recognized as a directory.

I also thought we need to set "this is a dir" attribute. The Zip file format has something called "external attributes". But the docs are unclear (mostly focus on MS DOS). I tried to use the values from libzip. That didn't help much (`zipinfo` didn't show correct attributes). Most likely that's needed for MS DOS, and maybe some windows zip implementations (would be nice to support Windows, but it's not crucial -- if needed, I'd be more in favor of rewriting this tool in a language that has zip library as part of its standard library).

PR #1734, opened 2023-02-21
Fix the formatting of AUTO expansions

kbieganski:
AUTO expansions currently do not follow formatting settings. This results in the formatter and AUTO expansion modifying each other's outputs. Not a great experience for the user, with text constantly being moved around.

This patch mostly fixes the formatting of AUTO expansion output. It makes AUTO expansion follow the Verible formatter rules. It's not perfect, as the user might not be using the formatter, but it's better than not being consistent with it. Also, not everything in the format style is supported yet: it doesn't follow alignments of port connections and module nets.

This PR also adds formatting and syntax checking to the AUTO expansion test. There's also a Python script that automatically formats all Verilog sources in the test.


verilog/tools/ls/autoexpand.cc
```diff
@@ -305,8 +308,10 @@ class AutoExpander {
   // Symbol table wrapper for the language server
   SymbolTableHandler *symbol_table_handler;
 
-  // String to add at the end of each generated line
+  // Indentation in various contexts
```
hzeller, line 311: Instead of the strings containing indentation spaces, you could consider using `verible::Spacer()`, (common/util/spacer.h). Up to you.

verilog/tools/ls/autoexpand_test_format
```diff
@@ -0,0 +1,47 @@
+#!/bin/env python3
+# Formats all raw strings in autoexpand_test.cc using Verible's formatter.
```
hzeller, line 2: This looks like a script that was helpful in the first transition, but now using the formatter directly from within the test should be sufficient. So I'd not include this script here.

hzeller (commented):
Looks great. One suggestion for the indentation an possibly not including the transition script.

verilog/tools/ls/autoexpand_test_format
kbieganski: I think it could be useful if a developer adds/edits a test, or the default formatting rules change. The formatter check gives you a diff, but you still have to format the code manually.

kbieganski: That makes sense. I'll get rid of the script.

kbieganski:
Pushed fixes for your comments. I'm working on actually using the formatter on the expanded code, but this can be merged for now if you're OK with this changeset.

hzeller (approved):
LGTM. Ok to add the formatter-call in a separate PR.

PR #1740, opened 2023-02-23
Invoke grail compilation db script directly.

hzeller:
Creating the compilation DB from within bazel has
its issues, e.g. we can't include the test targets.

In the project for creating a compilation database for baze l(https://github.com/grailbio/bazel-compilation-database), there is also an alternative way: a script that can be used to create a compilation DB for all targets by invoking bazel
with some aspects enabled.

Let's use that instead: with that, we get all targets, including test targets into the compilation database.

To avoid having to download the script manually, hash-sum check  and set up some temporary place on the users' workstation to host the script and the files it needs, we let bazel download it in the WORKSPACE file and then call into it.

While this assumes a certain layout of the bazel directory structure, this does seem to work fine with bazel 4, 5, and 6.

### DIFF
--- a/.github/bin/make-compilation-db.sh
+++ b/.github/bin/make-compilation-db.sh
@@ -15,16 +15,11 @@
 
 set -u
 set -e
-set -o pipefail
 
-readonly OUTPUT_BASE=$(bazel info output_base)
+bazel fetch ...  # no double-slash: seems to problematic on Windows
 
-# First, build the compilation database baseline with placeholders for exec-root
-bazel build :compdb > /dev/null 2>&1
+readonly OUTPUT_BASE="$(bazel info output_base)"
+python3 "${OUTPUT_BASE}/external/com_grail_bazel_compdb/generate.py"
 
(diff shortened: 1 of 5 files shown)


.github/bin/make-compilation-db.sh
```diff
@@ -15,16 +15,11 @@
...
 set -u
 set -e
-set -o pipefail
 
-readonly OUTPUT_BASE=$(bazel info output_base)
+bazel fetch ...  # no double-slash: seems to problematic on Windows
 
-# First, build the compilation database baseline with placeholders for exec-root
-bazel build :compdb > /dev/null 2>&1
+readonly OUTPUT_BASE=$(bazel info output_base)
```
fangism, line 21: nit: recommend double-quoting `"$(bazel info output_base)"`

fangism (approved):

.github/bin/make-compilation-db.sh
hzeller: Done.

PR #1747, opened 2023-02-24
Remember all files in the clang-tidy cache that were clean.

hzeller:
Previously, we discarded the whole list of files when there was a single clang-tidy complaint in any of them.
Now, only exclude the files from the cache that had issues, but keep all the rest. That way it is possible to incrementally fix all the clang-tidy complatins without having to reprocess all the files that didn't have any issues.

clang-tidy is very expensive: running it over all verible files takes about 3-4 core hours. This will help reduce duplicate work.

### DIFF
--- a/.github/bin/run-clang-tidy.sh
+++ b/.github/bin/run-clang-tidy.sh
@@ -102,7 +102,7 @@ for f in $(find . -name "*.cc" -and -not -name "*test*.cc" \
            )
 do
   (${CLANG_TIDY} --version; cat ${CLANG_CONFIG_FILE} WORKSPACE $f) \
-    | md5sum | sed "s|-|$f|g"
+    | md5sum | sed -e "s|-|$f|g"
 done | sort > ${CLANG_TIDY_SEEN_CACHE}.new
 
 # Only the files with different hashes are the ones we want to process.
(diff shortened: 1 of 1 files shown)


.github/bin/run-clang-tidy.sh
```diff
@@ -128,17 +129,35 @@ if [ -s ${FILES_TO_PROCESS} ]; then
...
+        /^[a-zA-Z_./]*:/ { seen[$1]=1; }
+        END {
+          is_first=1;
+          for (k in seen) {
+            printf("%s%s", is_first ? "/" : "\\|", gensub(/\//, "\\\\/", "g", k));
+            is_first=0;
+          }
+          printf("/d");
+        }' < ${TIDY_OUT})
+  sed "${EXCLUDE_EXPRESSION}" < ${CLANG_TIDY_SEEN_CACHE}.new > ${CLANG_TIDY_SEEN_CACHE}
```
fangism, line 156: `sed -e` (to be explicit)

Alternatively, you could send the output of `awk` to a file and use `sed -f` on that.  Might run less risk of hitting string length limits in the shell.

.github/bin/run-clang-tidy.sh
```diff
@@ -128,17 +129,35 @@ if [ -s ${FILES_TO_PROCESS} ]; then
...
+  # in the clang-tidy output.
+  # Use that to filter the file list cache to leave out files with failures,
+  # but keep the ones that had no issues.
+  # That way, only files that had issues will have to be processed in
+  # subsequent clang-tidy runs.
+  EXCLUDE_EXPRESSION=$(awk -F: '
+        /^[a-zA-Z_./]*:/ { seen[$1]=1; }
+        END {
+          is_first=1;
+          for (k in seen) {
```
fangism, line 150: If there are concerns about ordering/stability, due to looping over a hashmap, consider sorting the output.

fangism (approved):

.github/bin/run-clang-tidy.sh
```diff
@@ -128,17 +129,35 @@ if [ -s ${FILES_TO_PROCESS} ]; then
...
 else
   echo "Skipping clang-tidy run: nothing to do"
 fi
 
-# No complaints. We can cache this list now as baseline for next time.
-cp ${CLANG_TIDY_SEEN_CACHE}.new ${CLANG_TIDY_SEEN_CACHE}
 
-echo "No clang-tidy complaints.😎"
-exit 0
+if [ -s ${TIDY_OUT} ]; then
```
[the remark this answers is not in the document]
hzeller: Done here and partially elsewhere. Will do that separately.

.github/bin/run-clang-tidy.sh
hzeller: Good idea. Done.

.github/bin/run-clang-tidy.sh
hzeller: No concerns.

PR #1750, opened 2023-02-24
compile-db generation: only fetch workspace if needed.

hzeller:
There is only a need to call bazel fetch on a fresh workspace, so only do it if the compilation database script is not there yet.

### DIFF
--- a/.github/bin/make-compilation-db.sh
+++ b/.github/bin/make-compilation-db.sh
@@ -16,10 +16,12 @@
 set -u
 set -e
 
-bazel fetch ...  # no double-slash: seems to problematic on Windows
-
 readonly OUTPUT_BASE="$(bazel info output_base)"
-python3 "${OUTPUT_BASE}/external/com_grail_bazel_compdb/generate.py"
+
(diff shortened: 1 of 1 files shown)


.github/bin/make-compilation-db.sh
```diff
@@ -16,10 +16,14 @@
 set -u
 set -e
 
-bazel fetch ...  # no double-slash: seems to problematic on Windows
-
 readonly OUTPUT_BASE="$(bazel info output_base)"
-python3 "${OUTPUT_BASE}/external/com_grail_bazel_compdb/generate.py"
+
+readonly COMPDB_SCRIPT="${OUTPUT_BASE}/external/com_grail_bazel_compdb/generate.py"
+if [ ! -r "${COMPDB_SCRIPT}" ]; then
```
fangism, line 22: if you like a bit more compact form this could also be written:

```sh
[ -r "${COMPDB_SCRIPT}" ] || bazel fetch ...
```

It's fine as-is though.

fangism (approved):

.github/bin/make-compilation-db.sh
hzeller: Nice, I like that idea. Done.

PR #1752, opened 2023-02-24
Better document the path canonicalization expression.

hzeller:
With it, also fix the missing slash at the end of the `$(pwd)`, which
we also want to remove to have pure relative paths.

### DIFF
--- a/.github/bin/run-clang-tidy.sh
+++ b/.github/bin/run-clang-tidy.sh
@@ -15,6 +15,11 @@
 
 set -u
 
+if [ ! -f WORKSPACE ]; then
+  echo "This script needs to be invoked in the root of the bazel project"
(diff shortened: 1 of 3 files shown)


.github/bin/run-clang-tidy.sh
```diff
@@ -15,6 +15,11 @@
 
 set -u
 
+if [ ! -e WORKSPACE ]; then
```
fangism, line 18: or `[ ! -f WORKSPACE ]` (test if exists and is file)

fangism (approved):

.github/bin/run-clang-tidy.sh
hzeller: Thanks. Done.

PR #1759, opened 2023-02-25
clang-tidy: content-addressable results in config-dependent dir

hzeller:
Name the directory based on the active configuration, so that
it is possible to play with multiple configurations without
loosing cached information. In particular this means that files with
issues don't have to be reprocessed unless changed; the corresponding
cached clang-tidy output is just re-used.
    
For each cc-file, store the output of clang-tidy content-addressed,
so that a minimum of reprocessing is needed: if a file with a
particular content has been processed with a particular clang-tidy,
that content is cached and retrieved quickly.

Make sure that cc files are reprocessed if any of their headers
change.

Not only does this improve performance for unchanged files, the
implementation is also much simpler as we now use named files
instead of working with lines in a cache file.

Provide a symbolic link that points to the last called
clang-tidy output file; thus this can be used as a stable location.
    
If `CLANG_TIDY` environment variable is set, use its content
as the clang-tidy binary to call, otherwise fall back to search
in system path.

### DIFF
--- a/.github/bin/run-clang-tidy.sh
+++ b/.github/bin/run-clang-tidy.sh
@@ -19,7 +19,10 @@
 
 set -u
 
-readonly PROJECT_NAME=verible           # Prefix for directory in ${TMPDIR}
+# If enviornment variable CACHE_DIR is set, uses that as base to assemble
+# a path for the cache. Otherwise ~/.cache, ${TMPDIR}, or /tmp whatever first.
+
+readonly PROJECT_NAME=verible           # To assemble name inside ${CACHE_DIR}
 readonly EXIT_CODE_ON_WARNINGS=1        # Should fail on warnings found ?
 readonly CLANG_CONFIG_FILE=.clang-tidy  # config file used
 
@@ -51,15 +54,26 @@ fi
 
 hash "${CLANG_TIDY}" || exit 2  # make sure it is installed.
 
-TMPDIR="${TMPDIR:-/tmp}"
 EXIT_CODE=0
+
+# Find best place to put the cache.
+TMPDIR="${TMPDIR:-/tmp}"
+CACHE_DIR="${CACHE_DIR:-empty_env}"
+if [ "${CACHE_DIR}" == "empty_env" ]; then
+  if [ -d "${HOME}/.cache" ]; then
+    CACHE_DIR="${HOME}/.cache/clang-tidy"
+    mkdir -p "${CACHE_DIR}"
+  else
+    CACHE_DIR="${TMPDIR}"
+  fi
+fi
 
 # Assemble unique name depending on the configuration. This allows to play with
 # various clang-tidy versions and have outputs cached in separate directories.
 readonly CLANG_SHORT_VERSION="$(${CLANG_TIDY} --version | sed 's/.*version \([0-9]\+\).*/\1/p;d')"
 readonly CONFIG_NAME="v${CLANG_SHORT_VERSION}_$((${CLANG_TIDY} --version; cat ${CLANG_CONFIG_FILE} WORKSPACE; echo ${ADDITIONAL_TIDY_OPTIONS}) | md5sum | cut -b1-8)"
 
-readonly BASE_DIR="${TMPDIR}/${PROJECT_NAME}-clang-tidy-${CONFIG_NAME}"
+readonly BASE_DIR="${CACHE_DIR}/${PROJECT_NAME}-clang-tidy-${CONFIG_NAME}"
 readonly BASE_DIR_TMP="${BASE_DIR}/tmp"
 readonly CONTENT_DIR="${BASE_DIR}/contents"
 mkdir -p "${BASE_DIR}" "${BASE_DIR_TMP}" "${CONTENT_DIR}"
@@ -73,7 +87,8 @@ readonly FILE_HASHES_TO_PROCESS="${BASE_DIR_TMP}/files-hash.process"
 trap 'rm -rf -- "${BASE_DIR_TMP}"' EXIT
 
 readonly TIDY_OUT="${BASE_DIR}/tidy.out"
-readonly CURRENT_TIDY_OUT="${TMPDIR}/${PROJECT_NAME}-clang-tidy-current.out"
+readonly SUMMARY_OUT="${BASE_DIR}/tidy.out.summary"
+readonly CURRENT_TIDY_OUT="${PROJECT_NAME}-clang-tidy-current.out"
 
 # Explicitly test the configuration, otherwise clang-tidy will silently
 # fall back to some minimal default config.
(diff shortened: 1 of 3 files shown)


hzeller:
This PR will require #1760, fixing clang-tidy 15 issues, to get this CI to pass.

.github/bin/run-clang-tidy.sh
```diff
@@ -13,60 +13,79 @@
...
 fi
 
-# The caller can supply additional clang-tidy flags that are passed
-# to it as-is.
+# The caller can supply additional clang-tidy flags that are passed as-is
 # Example:
-# .github/bin/run-clang-tidy.sh --checks="-*,modernize-use-equals-default" --fix
-readonly ADDITIONAL_TIDY_OPTIONS=("$@")
+#   run-clang-tidy.sh --checks="-*,modernize-use-equals-default" --fix
+readonly ADDITIONAL_TIDY_OPTIONS="$@"
```
hzeller, line 34: FYI, I switched that back as regular scalar, not array. I couldn't get the quoting to work around line 170.

.github/bin/run-clang-tidy.sh
```diff
@@ -76,96 +110,109 @@ fi
...
+  # as parameter. Split the argument into ${args[0]} (hash value)
+  # and ${args[1]} (cc-file to process) and call clang-tidy on it. So each
+  # clang-tidy result is stored to its content-addressed file.
+  # (shell-quoting nightmare ahead:)
+  echo "Processing in ${PARALLEL_COUNT} parallel invocations of ${CLANG_TIDY}"
+  cat "${FILE_HASHES_TO_PROCESS}" | \
+    xargs -P"${PARALLEL_COUNT}" -d'\n' -n1 -- \
+          bash -c "read -a args <<< \$1 ;
+              ${CLANG_TIDY} --config=\"$(cat ${CLANG_CONFIG_FILE})\" \
+              --quiet ${ADDITIONAL_TIDY_OPTIONS} \${args[1]} 2>/dev/null \
```
fangism, line 186: For arrays, normally this needs to be expanded as `"${ADDITIONAL_TIDY_OPTIONS[@]}"`.

Since you are inside a here-doc, you want just `${ADDITIONAL_TIDY_OPTIONS[@]}` or with additional quote protection, `${ADDITIONAL_TIDY_OPTIONS[@]@Q}`.  (I just learned this latter variant recently.)

hzeller: Done. The quoting-hell prevented that to work inside the shell script, so generated a helper shell-script in a here document that then is executed.

.github/bin/run-clang-tidy.sh
```diff
@@ -76,96 +111,110 @@ fi
...
+# we simply replace references to headers with hashes of their content.
+# If content of these files change, the hashes of the includers will change.
+#   #include "foo/bar/baz.h" -> #include "a46b89dd7a7f8f108b836b50762fb6e1"
+# Ideally loop until no changes in hashes, but one level good enough.
+
+# Create a sed expression that replaces header filenames with content hash.
+for f in $(cat ${FILES_OF_INTEREST}); do
+  if [ "${f##*.}" == "h" ]; then     # only headers expected to be relevant
+    c_hash="$(cat $f | md5sum | awk '{print $1}')"
+    echo "s|$f|$c_hash|g" | sed 's|\.|\\.|g' >> "${INCLUDE_REPLACE}"
```
fangism, line 141: Instead of the append redirection inside the loop, you could just print to stdout and then do a single redirection outside of the loop:
```
done >> "${INCLUDE_REPLACE}"
```

.github/bin/run-clang-tidy.sh
```diff
@@ -76,96 +111,110 @@ fi
...
   echo "::endgroup::"
-else
-  echo "Skipping clang-tidy run: nothing to do"
+
+  # The files might be reported with various absolute prefixes that
+  # we remove to canonicalize and report as relative paths from project
+  # root.
+  readonly EXEC_ROOT="$(bazel info execution_root)"
+  readonly CURRENT_DIR="$(pwd)"
+  readonly CANONICALIZE_SOURCE_PATH_REGEX="^\(${EXEC_ROOT}\|${CURRENT_DIR}\|\.\)/"
```
fangism, line 175: Alternative: `realpath --relative-to=...`

.github/bin/run-clang-tidy.sh
```diff
@@ -76,96 +111,110 @@ fi
...
+  # Run clang-tidy on <filename> and store result in ${CONTENT_DIR}/<hash>
+  cat > "${CLANG_TIDY_HELPER}" <<EOF
+read OUTPUT_FILE INPUT_FILE <<< \$1
+${CLANG_TIDY} --quiet --config="$(cat ${CLANG_CONFIG_FILE})" \
+  ${ADDITIONAL_TIDY_OPTIONS[@]@Q} "\${INPUT_FILE}" 2>/dev/null \
+  | sed -e "s@${CANONICALIZE_SOURCE_PATH_REGEX}@@g" \
+  > "${CONTENT_DIR}/\${OUTPUT_FILE}.tmp"
+mv "${CONTENT_DIR}/\${OUTPUT_FILE}".tmp "${CONTENT_DIR}/\${OUTPUT_FILE}"
+EOF
+
```
fangism, line 187: `chmod +x "${CLANG_TIDY_HELPER}"`
and call it directly?

.github/bin/run-clang-tidy.sh
```diff
@@ -76,96 +111,110 @@ fi
...
+  # we remove to canonicalize and report as relative paths from project
+  # root.
+  readonly EXEC_ROOT="$(bazel info execution_root)"
+  readonly CURRENT_DIR="$(pwd)"
+  readonly CANONICALIZE_SOURCE_PATH_REGEX="^\(${EXEC_ROOT}\|${CURRENT_DIR}\|\.\)/"
+
+  # Helper script gets "<hash-value> <filename>" as single argument.
+  # Run clang-tidy on <filename> and store result in ${CONTENT_DIR}/<hash>
+  cat > "${CLANG_TIDY_HELPER}" <<EOF
+read OUTPUT_FILE INPUT_FILE <<< \$1
```
fangism, line 180: need a she-bang?  `!#/bin/sh`?

.github/bin/run-clang-tidy.sh
```diff
@@ -76,96 +111,110 @@ fi
...
-  sed -f "${CLANG_TIDY_SEEN_CACHE}".exclude_expression \
-    < "${CLANG_TIDY_SEEN_CACHE}".new > "${CLANG_TIDY_SEEN_CACHE}"
+  EXIT_CODE="${EXIT_CODE_ON_WARNINGS}"
+  echo "::warning::There were clang-tidy warnings. Please fix."
+  echo "::group::clang-tidy detailed output"
+  cat "${TIDY_OUT}"
+  echo "Find this output in ${CURRENT_TIDY_OUT}"
+  echo "::endgroup::"
+
+  echo "Summary (in ${SUMMARY_OUT})"
```
fangism, line 210: You can also use `| tee filename` to save stdout to a file.

fangism (approved):

.github/bin/run-clang-tidy.sh
hzeller: I want to live with as little as possible external different binaries.

.github/bin/run-clang-tidy.sh
hzeller: I prefer not having things executable if possible, and it would also mean that I have to add a `#!`.

.github/bin/run-clang-tidy.sh
hzeller: Since I call it directly with bash, not needed (see above).

.github/bin/run-clang-tidy.sh
hzeller: then I rely on the existence of `tee`. I always have macs or windows in mind, which have an abysmal set of working standard tools available.

.github/bin/run-clang-tidy.sh
hzeller: Good point. Done.
This was a leftover from an earlier version of the loop that wrote into two different files.

PR #1761, opened 2023-02-27
Show AUTO expansion code actions only for relevant lines

### LINKED ISSUE #1741: auto-expand code-action shown for every line
A couple of questions first

 - Your IDE/editor (e.g. vscode, emacs,...) you use with verible LSP ? vscode
-----

With the auto-expand option enabled, the code action offering (the little ':bulb:' in vscode) shows for every line.
This is somewhat distracting to the user (and also it apparently requires the auto-expand to re-evaluate everything on every cursor move.)

**Expectation**

The code action should only show up in a line in which an actual `AUTO*` comment is residing.

Ideally, it should offer two code-actions to choose from: 'only expand this instance', 'expand all auto comments in the same file' (maybe also 'expand all fromt he same kind' ? e.g.  if I am in a line with AUTOINST, only expand other AUTOINST in the rest of the file).

Advantage: less distraction for the user, and potentially cheaper to calculate for the language server if it knows where the cursor is.

kbieganski:
AUTO expansion is now only available if the user invokes the code action request on a line or range of lines that contains any AUTOs.

Adds a new code action that expands AUTOs only of the same kind as the ones pointed at/in the selected range.

Adds a separate function that detects if any AUTOs exist in a given range (`FindAutoKinds()`). This will let us split AUTO detection and expansion into separate code action and command calls in the future.

Refactors a lot of the code. Now `AutoExpander` only deals with string views instead of a mix of string views, line ranges, and byte offsets. In general now the code should be simpler and easier to follow.

Also modifies the test process so that the golden file is formatted before comparing it to the expanded code.

Fixes #1741.


hzeller (approved):
Nice, the expanded use of the string view to represent ranges improves things a lot as the reader needs fewer concepts keeping in mind.

I'll test the latest version tomorrow locally, but I think we're ready to merge now.

Thanks!

PR #1767, opened 2023-03-02
Rewrite run-clang-tidy script in C++ for readability.

hzeller:
The run-clang-tidy script provides efficient caching
of checked files and their dependencies (as a full run
takes about 400 core minutes: caching is vital).
As such, it is bit more compliated than simply invoking
clang-tidy.
    
The shell script to deal with that was growing beyond
healthy and was assuming certain features in the
various tools involved (sed, awk, xargs, sort...) that
might not be available everywhere. This, and
complicated quoting rules and esoteric scripting
features also made it fragile, hard to read, and maintain.
    
This is a rewrite in portable c++17. As such, it is
also much easier to maintain by the C++ authors of
Verible. Tested on Linux and MacOS.
    
File is designed as C++ script, so can be invoked
directly.


.github/bin/run-clang-tidy.cc
```diff
@@ -0,0 +1,226 @@
+#if 0  // Invoke with /bin/sh or simply add executable bit on this file on Unix.
+BIN=${0%%.cc}; [ $BIN -nt $0 ] || c++ -std=c++17 $0 -o $BIN && exec -a$0 $BIN $@
```
fangism, line 2: cute.

.github/bin/run-clang-tidy.cc
```diff
@@ -0,0 +1,226 @@
...
+  std::vector<std::thread> workers;
+  for (auto i = 0; i < kJobs; ++i) workers.emplace_back(clang_tidy_runner);
+  for (auto &t : workers) t.join();
+  fprintf(stderr, "     \n");  // Clean out progress counter.
+}
+
+int main(int argc, char *argv[]) {
+  std::error_code ignored_error;  // filesystem errors we don't care about.
+  if (!fs::exists("compile_commands.json")) {
+    std::cerr << "No compilation db found. First, run make-compilation-db.sh\n";
```
fangism, line 127: `std::endl` is more canonical.

fangism (approved):

.github/bin/run-clang-tidy.cc
hzeller: I try to avoid as much as possible of the stream operator stuff as it mostly only adds clutter.
And in this case, it would break the line over 80 characters, so would occupy another line...

PR #1769, opened 2023-03-03
Format AUTO expansions

kbieganski:
Use Verible's formatter to output nicely formatted AUTO expansions. This means AUTO expand code doesn't need to try and imitate the formatter anymore, and some parts of it are simplified (no more passing of `indent` everywhere).


verilog/tools/ls/autoexpand.cc
```diff
@@ -1402,25 +1368,77 @@ std::optional<absl::string_view> AutoExpander::FindSpanToReplace(
...
+  if (format_status.ok()) {
+    new_text = formatted;
+  } else {
+    LOG(ERROR) << "Failed to format AUTO expanded code";
+  }
+
+  return {{.range =
+               {
+                   .start = {.line = 0, .character = 0},
+                   .end = {.line = (int)text_structure.Lines().size() - 1,
```
hzeller, line 1437: To avoid fiddling with the lines and columns manually too much, you can also use the [GetRangeForText()](https://github.com/chipsalliance/verible/blob/56ec2f9938b39491ef0bee01d9718a19e88abe81/common/text/text_structure.h#L118-L120)  function to get the information, just passing it the full content.

(this is what happens in the formatter LSP call:)
https://github.com/chipsalliance/verible/blob/56ec2f9938b39491ef0bee01d9718a19e88abe81/verilog/tools/ls/verible-lsp-adapter.cc#L303-L310

verilog/tools/ls/autoexpand.cc
```diff
@@ -1402,25 +1368,77 @@ std::optional<absl::string_view> AutoExpander::FindSpanToReplace(
...
+    std::vector<AutoExpander::Expansion> expansions) {
+  std::sort(expansions.begin(), expansions.end(),
+            [](const AutoExpander::Expansion &first,
+               const AutoExpander::Expansion &second) {
+              return first.replaced_span.begin() < second.replaced_span.begin();
+            });
+
+  // To format expansions, we need to apply them to the source. Then we only
+  // call the formatter once, and produce a single TextEdit that replaces the
+  // entire file. This is orders of magnitude faster than formatting individual
```
hzeller, line 1384: Maybe clarify in the comment, that you still limit the formatter ranges to the affected line ranges generated by the auto expander. Otherwise this comment sounds like the user ends up with the full file reformatted, but all we want is the snippets that have been auto-expand modified.

verilog/tools/ls/autoexpand_test.cc
```diff
@@ -2377,8 +2379,8 @@ module foo (  /*AUTOARG*/
       .clk(clk),
       .rst(rst),
       // Outputs
-      .o1(out_a[63:0]),
-      .o2(out_b  /*.[16]*/)
+      .o1 (out_a[63:0]),
+      .o2 (out_b  /*.[16]*/)
   );
 endmodule
```
hzeller, line 2386: If not already, it might be worthwhile testing a file with multiple autoexpand sections and some, 'freestyle' formatting by the user in-between these (that does not conform to the formatter), then apply auto expand.

What we then expect to see is

   * sections between the auto-expanded parts are still featuring the original formatting
   * the formatting does the right thing of not messing up regions - since the expanded areas are typically longer than before we still want to see that the resulting document looks nice.

verilog/tools/ls/autoexpand_test.cc
```diff
@@ -142,12 +139,12 @@ void TestTextEditsWithProject(
   // way we can apply them one by one and have the following ones still be
   // valid.
   // Note: according to the spec, TextEdits should never overlap.
-  std::sort(edits.begin(), edits.end(),
+  std::sort(edits.rbegin(), edits.rend(),
```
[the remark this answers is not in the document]
kbieganski: I think I changed the first `>` (which is slightly incorrect) into an `>=`, and then I thought the `rbegin..rend` approach would be more readable (the comparator is slightly easier to parse). Now I realized that I should also swap `start` and `end` in the comparator. :(

verilog/tools/ls/autoexpand.cc
kbieganski: Great idea, thanks

verilog/tools/ls/autoexpand_test.cc
kbieganski: I get the first dot, but what do you mean by *not messing up regions*?

hzeller: _messing up regions_: since we're replacing sections of the input with new output that might be more lines, the whole lower part of the file changes down. Previously, you made sure to apply edits from the bottom to the top, but since now the formatter is involved which also looks at sub-ranges and emits other subranges, there is a chance that for whatever reason we create overlapping regions that 'eat' up output. At least this is something we should confirm with a test that it does not happen.

kbieganski: Added a test for it. Seems like the formatter preserves some weird formatting in the ranges that AUTO expansion puts in declarations and connections (they are copied verbatim from the user's code).

verilog/tools/ls/autoexpand_test.cc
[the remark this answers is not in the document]
kbieganski: Fixed

hzeller (approved):
Nice!

PR #1771, opened 2023-03-06
Canonicalize paths emitted from clang-tidy.

hzeller:
The paths of source files reported are not always relative to the project root, but some are prefixed with the bazel root or the current directory.

Canonicalize them so that files are always reported the same way.

Also: better quote variables in the 'shellscript' compiler invocation line; update and refine some comments.

### DIFF
--- a/.github/bin/run-clang-tidy-cached.cc
+++ b/.github/bin/run-clang-tidy-cached.cc
@@ -1,5 +1,5 @@
 #if 0  // Invoke with /bin/sh or simply add executable bit on this file on Unix.
-BIN=${0%%.cc}; [ $BIN -nt $0 ] || c++ -std=c++17 $0 -o $BIN && exec -a$0 $BIN $@
+B=${0%%.cc}; [ "$B" -nt "$0" ] || c++ -std=c++17 -o"$B" "$0" && exec "$B" "$@";
 #endif
 // Copyright 2023 The Verible Authors.
 //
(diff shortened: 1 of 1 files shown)


.github/bin/run-clang-tidy-cached.cc
```diff
@@ -83,6 +83,24 @@ fs::path GetCacheDir() {
...
+    if (!root.empty()) {
+      root.pop_back();  // remove newline.
+      canonicalize_expr += root + "/|";
+    }
+    canonicalize_expr += fs::current_path().string() + "/";  // $(pwd)/
+    canonicalize_expr += ")?(\\./)?";  // Some start with, or have a trailing ./
+    return std::regex{canonicalize_expr};
+  }();
+  const auto in_content = GetContent(infile);
+  std::fstream out_stream(outfile, std::ios::out);
```
fangism, line 100: or just `std::ofstream`?

.github/bin/run-clang-tidy-cached.cc
```diff
@@ -83,6 +83,24 @@ fs::path GetCacheDir() {
   return fs::path(getenv("TMPDIR") ?: "/tmp");
 }
 
+// Fix filename paths that are not emitted relative to project root.
```
fangism, line 86: Perhaps include a few example transformation in comments?

fangism (approved):
Can you also include in a comment near the top where this script is expected to run?  e.g. linux/posix environment only.

.github/bin/run-clang-tidy-cached.cc
hzeller: Done.

.github/bin/run-clang-tidy-cached.cc
hzeller: Good idea. Added a comment at the top.

PR #1776, opened 2023-03-07
Re-enable stack trace printing.

hzeller:
When switching to absl logging, that was skipped.


hzeller:
While debugging issues, a stacktrace can be quite useful :) (also `--config=asan` will typically do the trick while developing, but 'in the field' that would not be there)

kbieganski (approved):

PR #1784, opened 2023-03-09
json-rpc-dispatcher: Make calling with no parameters more robust.

hzeller:
I noticed that vscode calls the shutdown method without actually sending a "params" object - which is fine, as there are none. Make sure that we handle the absense of a "params" object gracefully.


glatosinski (approved):
Looks LGTM

PR #1785, opened 2023-03-09
Prepare code to be ready for asynchronous buffer updates.

hzeller:
Handing out the shared_ptr<> of the current()/last_good() buffer will make sure that ongoing operations will have a valid copy even if it is replaced asynchronously.

Just API update, no actual asynchronous operations happening yet.

While at it: add documentation to the function calls as well as thread guarantees.

### DIFF
--- a/verilog/tools/ls/symbol-table-handler.cc
+++ b/verilog/tools/ls/symbol-table-handler.cc
@@ -234,7 +234,7 @@ std::vector<verible::lsp::Location> SymbolTableHandler::FindDefinitionLocation(
     LOG(ERROR) << "Could not find buffer with URI " << params.textDocument.uri;
     return {};
   }
-  const verilog::ParsedBuffer *parsedbuffer = tracker->current();
+  const auto parsedbuffer = tracker->current();
   if (!parsedbuffer) {
     LOG(ERROR) << "Buffer not found among opened buffers:  "
                << params.textDocument.uri;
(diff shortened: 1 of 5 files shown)


hzeller:
Requesting review from everyone working with the language server. Right now, there is no asynchronous operation going on yet, and I need to add more infrastructure for that to happen.

But this prepares to set the expectation for callers of the `current()` and `last_good()` functions. Mostly: as long as the shared pointer returned from these functions is still in scope, everything should be fine as-is and no worries about thread-safety are needed.

verilog/tools/ls/symbol-table-handler.cc
```diff
@@ -234,7 +234,7 @@ std::vector<verible::lsp::Location> SymbolTableHandler::FindDefinitionLocation(
     LOG(ERROR) << "Could not find buffer with URI " << params.textDocument.uri;
     return {};
   }
-  const verilog::ParsedBuffer *parsedbuffer = tracker->current();
+  const auto parsedbuffer = tracker->current();
```
hzeller, line 237: Given that  finding any symbol is better than finding none at all, using `last_good()` might probably be a better choice. Leaving it to @glatosinski to consider.

glatosinski: `parsedbuffer` in general should always exist when a file was properly opened in the LS context.
We use it to obtain the current state of the buffer,  its `TextStructureView` to be more specific.

The reason for using `current()` instead for `last_good()` is that the user can make some extensive changes in the file which are not necessarily good from the parser perspective, but can significantly change the text structure.

Due to the fact that `textDocument/definition` request sends only URI to file and symbol coordinates, the usage of `last_good` could result in extracting the wrong symbol (because the contents changed).

glatosinski (approved):
In general switching to `shared_ptr`s seems legit.

fangism (approved):

PR #1788, opened 2023-03-09
Handle incorrect instance types in `AUTOINST`

### LINKED ISSUE #1775: Autoexpand crash when module name not found.
Playing around in some files (with possibly incomplete symbol table as the file list was not complete) I noticed crashes of the language server when it came around a situation in which `GetModuleName()` returned a nullptr. When given such symbol to a `Module()` object, it crashes as it dereferences that around line 145 in the `Module` constructor.

Testing for that situation beforehand helped that case (quick hack patch below), but it then crashes later (still investigating).
So it would be good to have a test that creates a situation like this and test that things are robust afterwards.

```patch
--- a/verilog/tools/ls/autoexpand.cc
+++ b/verilog/tools/ls/autoexpand.cc
@@ -981,6 +981,10 @@ std::optional<AutoExpander::Expansion> AutoExpander::ExpandAutoinst(
     LOG(ERROR) << "AUTOINST: No definition found for module type: " << type_id;
     return {};
   }
+  if (!GetModuleName(*type_def)) {
+    LOG(ERROR) << "AUTOINST: No module name " << type_id;
+    return {};
+  }
   if (!modules_.contains(type_id)) {
     modules_.insert(std::make_pair(type_id, Module(*type_def)));
   }
```

kbieganski:
The `AUTOINST` expansion function assumes the symbol found by searching for the instance type name is a module declaration. However, if it's not a module declaration, Verible crashes due to theie assumption. This patch fixes the issue by checking the type of the node found by the symbol table.

Fixes #1775.


verilog/tools/ls/autoexpand.cc
```diff
@@ -981,6 +981,12 @@ std::optional<AutoExpander::Expansion> AutoExpander::ExpandAutoinst(
     LOG(ERROR) << "AUTOINST: No definition found for module type: " << type_id;
     return {};
   }
+  if (NodeEnum(type_def->Tag().tag) != NodeEnum::kModuleDeclaration) {
+    LOG(ERROR) << "AUTOINST: Instance type " << type_id
+               << " is not a module, but a '" << NodeEnum(type_def->Tag().tag)
```
hzeller, line 986: So ideally, we should work towards that we can ask the symbol table 'do you have something with this name that is also module?'. Not sure if the symbol table well can keep track of these if they have multiple names, but that would be ideal.
@glatosinski do you know if that can be implemented in a simple way ?

PR #1789, opened 2023-03-09
Fix overlapping AUTO expansions

kbieganski:
LSP `TextEdit`s have to be disjoint, but that is not necessarily the case with the AUTO expander (see added test case). This patch fixes the issue by simply ignoring overlapping expansions. In the future, we should notify the user about such problems, preferably via diagnostics.


verilog/tools/ls/autoexpand.cc
```diff
@@ -1387,10 +1387,24 @@ std::vector<TextEdit> ConvertAutoExpansionsToFormattedTextEdits(
   // TextEdits.
   std::string text;
   auto begin = text_structure.Contents().begin();
-  for (const AutoExpander::Expansion &expansion : expansions) {
-    text.append(begin, expansion.replaced_span.begin());
-    text.append(expansion.new_text);
-    begin = expansion.replaced_span.end();
+  for (auto it = expansions.begin(), next = it; it != expansions.end();
```
hzeller, line 1390: this new snippet looks like it wants to be a separate function with a name that explains what it does.

verilog/tools/ls/autoexpand.cc
```diff
@@ -1369,6 +1369,43 @@ std::optional<absl::string_view> AutoExpander::FindSpanToReplace(
   return replaced_span;
 }
 
+// Returns an iterator pointing at the next expansion not overlapping with *it
+auto GetNextNonOverlappingExpansion(
+    std::vector<AutoExpander::Expansion>::const_iterator it, decltype(it) end) {
```
hzeller, line 1374: I'd probably rename `it`  to `start`, that might make the code below read more naturally.

Can we reformulate this whole function as a template, so that we don't have to use the awkward decltype() stuff ?

kbieganski: Is a template less awkward? I think I'd rather just use the explicit type then.

hzeller: Yeah,but now we describe the same type in three different ways. `auto` for the return type, `std::vector<AutoExpander::Expansion>::const_iterator` and `decltype(it)`.

I think there is a value in readability if the same things read the same.

kbieganski: Good point, changed it to templates, added a comment to make it better once we switch to C++20

hzeller:
Thanks!

PR #1790, opened 2023-03-09
Prevent expansion of misplaced AUTOs

kbieganski:
This patch addresses situations where AUTOs are expanded in incorrect contexts, such as expanding `AUTOINST` under a port actual, instead of directly in module instantiation parens.

The syntax tree is traversed until the deepest node that encompasses a given AUTO expansion is found. If the node kind is expected for the expansion, it can be applied. Otherwise, the expansion is discarded. For example, `AUTOINST` expansion can only be done directly under module instantiation, but not any deeper in the tree, e.g. in a port actual.

### DIFF
--- a/common/text/tree_utils.cc
+++ b/common/text/tree_utils.cc
@@ -220,6 +220,31 @@ class FirstSubtreeFinder : public SymbolVisitor {
   // Contains first matching result found or nullptr if no match is found.
   const Symbol* result_ = nullptr;
 };
+
+// A visitor that finds the last matching node. Inherits from
+// TreeVisitorRecursive, as it needs to visit all nodes in the tree.
+class LastSubtreeFinder : public TreeVisitorRecursive {
+ public:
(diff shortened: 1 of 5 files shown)


hzeller:
Can you also add a corresponding test in tree_utils_test.cc ? That way, we test the functionality as close as possible to the source (which makes it also easier to expand, debug, find more specific corner cases etc.)

You can observe just the outcome of a single test by passing the blaze target to the build-and-test.sh script:

```
MODE=coverage .github/bin/build-and-test.sh  common/text:tree_utils_test
```

then opening `coverage-html/text/tree_utils.cc.gcov.html` in the browser.

kbieganski:
I'll push tests after https://github.com/chipsalliance/verible/pull/1795

kbieganski:
Pushed tests, split the PR into two commits.

hzeller (approved):
Nice.

PR #1793, opened 2023-03-10
Start using RE2 instad of std:;regex

hzeller:
std::regex is not optimized for performance and can be quite slow on different platforms.
While it does not matter in this particular tool, this is to test it in a small context so that we then can apply it in other places where performance matters.

### DIFF
--- a/common/tools/BUILD
+++ b/common/tools/BUILD
@@ -66,6 +66,7 @@ cc_binary(
         "@com_google_absl//absl/container:flat_hash_map",
         "@com_google_absl//absl/flags:flag",
         "@com_google_absl//absl/strings",
+        "@com_googlesource_code_re2//:re2",
     ],
 )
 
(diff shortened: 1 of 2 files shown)


hzeller:
I've compared `std::regex` with [RE2](https://github.com/google/re2) and std::regex is shockingly slow (10x and more). We already have pulled RE2 into this project in the WORKSPACE file (mostly because gtest uses it), but we can also utilize it in our code and avoid the slow and exception throwing `std::regex`.

In this particular tool, it does not matter at all, but it does in the autoexpand uses; this is why I added you as reviewer @kbieganski so that you can see that the changes are probably pretty mechanical and simple.

They also have the so-called [LazyRE](https://github.com/google/re2/blob/main/re2/re2.h#L943) which is a little helper that makes static initialization of regular expressions safe.

hzeller:
Also created #1794

kbieganski (approved):
Looks good! Though I believe AUTO expand execution time is currently dominated by the formatter :smile:

PR #1799, opened 2023-03-13
File opening status messages: always include filename.

hzeller:
The status messages due to opening files might be created deeply in the code, so it is good if issues with files percolate up with the filename included.

### DIFF
--- a/common/util/file_util.cc
+++ b/common/util/file_util.cc
@@ -67,36 +67,44 @@ absl::string_view Stem(absl::string_view filename) {
              : filename.substr(0, last_dot_pos);
 }
 
-// This will always return an error, even if we can't determine anything
-// from errno. Returns "fallback_msg" in that case.
-static absl::Status CreateErrorStatusFromSysError(const char *fallback_msg,
-                                                  int sys_error) {
-  using absl::StatusCode;
+// Create an error message derived from "sys_error" (which contains an errno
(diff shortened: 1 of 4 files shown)


common/util/file_util.cc
```diff
@@ -69,34 +69,39 @@ absl::string_view Stem(absl::string_view filename) {
 
 // This will always return an error, even if we can't determine anything
 // from errno. Returns "fallback_msg" in that case.
-static absl::Status CreateErrorStatusFromSysError(const char *fallback_msg,
-                                                  int sys_error) {
-  using absl::StatusCode;
+static absl::Status CreateErrorStatusFromSysError(absl::string_view filename,
+                                                  int sys_error,
+                                                  const char *fallback_msg) {
```
fangism, line 74: Perhaps add a comment about `fallback_msg` not needing to be stable because it will be copied?

fangism (approved):

common/util/file_util.cc
hzeller: Done, refined comment
```c++
// Create an error message derived from "sys_error" (which contains an errno
// number). The message includes the filename as prefix.
// If "sys_error" can not be resolved, creates an UNKNOWN message.
// The "filename" and "fallback_msg" will be copied, no need for them to
// stay alive after the call.
```

PR #1800, opened 2023-03-14
Revise an old filesystem function to return StatusOr<>

hzeller:
When we started, we did not have StatusOr<> available, but it makes a lot of sense for file::GetContents() similar to the newer GetContentAsMemBlock().

Provide a new function, similar to it sibling
GetContentAsMemBlock()
  absl::StatusOr<std::string> GetContentAsString();

Provide an adapter function that behaves just as the old file::GetContents(). To be replaced in subsequent change(s) until finally removed.


hzeller:
Also refined the tests a bit, let's see if the directory tests work well on Windows.

fangism (approved):

PR #1803, opened 2023-03-14
Use `RE2` instead of `std::regex` in AUTO expansion

### LINKED ISSUE #1794: Use RE2 instead of std::regex in autoexpand
std::regex is much slower than RE2. In the autoexpand feature, we are called very regularly on potentially very large files, so it makes sense to make sure we're not unnecessarily slow.

kbieganski:
* Replaces all usage of `std::regex` with `RE2` in AUTO expansion code,
* Simplifies some of the code that used to pass `std::cmatch`es around.

Closes #1794.

### DIFF
--- a/verilog/tools/ls/BUILD
+++ b/verilog/tools/ls/BUILD
@@ -36,6 +36,7 @@ cc_library(
         "@com_google_absl//absl/container:flat_hash_map",
         "@com_google_absl//absl/container:flat_hash_set",
         "@com_google_absl//absl/container:node_hash_map",
+        "@com_googlesource_code_re2//:re2",
     ],
 )
 
--- a/verilog/tools/ls/autoexpand.cc
+++ b/verilog/tools/ls/autoexpand.cc
@@ -17,12 +17,12 @@
 
 #include <algorithm>
 #include <cstdint>
-#include <regex>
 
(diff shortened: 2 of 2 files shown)


kbieganski:
I think the `clang-tidy` failure is from `absl::call_once` in `LazyRE2`. Not sure how to disable this lint locally (`//NOLINTBEGIN` and `//NOLINTEND` don't seem to work).

hzeller:
Usually often `//NOLINT` helps in the same line, but maybe it is a bit more involved. I'll download your patch and see what I can do locally.

If you have a machine with a bunch of cores, you can run clang-tidy on the whole project with
 
```
.github/bin/make-compilation-db.sh && .github/bin/run-clang-tidy-cached.cc
```

hzeller:
You might need to rebase so that we get the updated CI output

hzeller (approved):
Nice

hzeller:
Update: [Fixed the absl warning upstream](https://github.com/abseil/abseil-cpp/commit/e5b65f172a2fbf488dc1927d86c8f3671d6b032f). So next time we update absl, we get that included, until then we rely on our [local patch](https://github.com/chipsalliance/verible/blob/master/bazel/fix-uninitialized-var.patch).

PR #1807, opened 2023-03-14
clang-tidy: reprocesses troublesome files if older than build env.

hzeller:
If the compilation DB was re-created, it might contain changes that improve the outcome, e.g. an include directory reference that didn't work before and made clang-tidy fail on a particular *.cc file (typically due to BUILD) changes.

Similarly if there are changes in the WORKSPACE file since last time a file was looked at.

Check timestamps of compile_commands.json and WORKSPACE to decide if a file that otherwise is unchanged but generated messges last time (i.e. file size != 0) shall be reprocessed.


hzeller:
This will make sure that once #3476 is in, the files where an issue has been reported the last time will be properly reprocessed and not the old cached value be displayed.

PR #1809, opened 2023-03-14
Make more use of status macro.

hzeller:
There are some patterns that are eaiser done using RETURN_IF_ERROR(), improving readability.


hzeller:
For discussion. Still to be decided if we want to use the macro or mabye switch to not using it at all.

Pros:
  * More compact code
  * Readable once developer is used to it

Cons:
  * It is a macro
  * Readable _only_ once developer is used to it.

fangism (approved):

PR #1816, opened 2023-03-16
AUTO expansion – connections to/between multiple instances

kbieganski:
1.  Support multiple connections to an AUTO-generated port
    
    This patch makes it possible to connect a single module port to multiple
    instances during AUTO expansion. For generated ports, this is not a full
    solution, as the port's dimensions are taken from the first connected instance.
    Instead, it should be the maximum of dimensions of all ports connected to it.
    
2.  Handle `AUTOWIRE` between instances
    
    During AUTO expansion, whenever:
    
    - an instance has a port of a certain direction,
    - another instance has a port of a different direction, but the same name,
    - and there is an `AUTOWIRE` in the module,
    
    a wire is generated that connects these instances (instead of a port).


hzeller (approved):
Nice!

(With all these features, maybe we should make this also a stand-alone utility within the Verible suite at some point)

kbieganski:
Good idea, especially since running the original Emacs AUTO stuff in batch mode is a real use case, I believe.

PR #1817, opened 2023-03-17
Comment updates; remove unused function.

hzeller:


common/lsp/dummy-ls.cc
```diff
@@ -61,23 +61,25 @@ InitializeResult InitializeServer(const nlohmann::json &params) {
...
+  // Windows messes with newlines by default. Fix this here.
   _setmode(_fileno(stdin), _O_BINARY);
+  _setmode(_fileno(stdout), _O_BINARY);
 #endif
 
   std::cerr << "Note: this dummy-ls is for testing." << std::endl;
 
   // Input and output is stdin and stdout
-  static constexpr int in_fd = 0;  // STDIN_FILENO
+  constexpr int kInputFD = 0;  // STDIN_FILENO, but Win does not have that macro
```
fangism, line 72: would an `#ifndef STDIN_FILENO ...` help here?

verilog/tools/ls/verilog_ls.cc
```diff
@@ -48,11 +41,18 @@ int main(int argc, char *argv[]) {
...
+  // Output: provided write-function is called with entire response messages.
+  verilog::VerilogLanguageServer server([](absl::string_view reply) {
+    // Output formatting as header/body chunk as required by LSP spec to stdout.
+    std::cout << "Content-Length: " << reply.size() << "\r\n\r\n";
+    std::cout << reply << std::flush;
+  });
 
+  // Input: Messages received from the read function are dispatched and
+  // processed until shutdown message received.
+  constexpr int kInputFD = 0;  // STDIN_FILENO, but Win does not have that macro
```
fangism, line 55: same question: `#ifndef STDIN_FILENO ...`?

fangism (approved):

common/lsp/dummy-ls.cc
hzeller: but then we have 3 lines of code that clutter up things.

PR #1818, opened 2023-03-17
Match `AUTO_TEMPLATE` by instance name regex

kbieganski:
1.  Match `AUTO_TEMPLATE` by instance name regex

    The instance name regex in `AUTO_TEMPLATE` has been ignored up until this patch.
    Now it is actually being used and considered for matching the correct template
    to a given instance. The way it works is straightforward: if a template matches
    an instance by type, the instance's name is checked against the template's
    regex. If it matches, the template is selected. Otherwise, the next template
    that matches by type is considered.

2.  Replace `@` in `AUTO_TEMPLATE` with module instance name

    This change makes it so that the symbol `@` in signal names in `AUTO_TEMPLATE`
    is replaced with the name of the instance matched to a given template.

### DIFF
--- a/verilog/tools/ls/autoexpand.cc
+++ b/verilog/tools/ls/autoexpand.cc
@@ -144,8 +144,10 @@ class AutoExpander {
   struct Template {
     using Map = absl::flat_hash_map<absl::string_view, std::vector<Template>>;
 
-    absl::string_view::const_iterator it;  // Location of the template in the
-                                           // source file
+    absl::string_view::const_iterator it;   // Location of the template in the
+                                            // source file
+    std::shared_ptr<RE2> instance_name_re;  // Regex for matching the instance
+                                            // name
     absl::flat_hash_map<absl::string_view, Connection>
         connections;  // Map of instance ports ports to connected module ports
   };
@@ -221,6 +223,7 @@ class AutoExpander {
     // Retrieves the matching template from a typename -> template map
     const Template *GetAutoTemplate(
         absl::string_view type_id,
+        std::optional<absl::string_view> instance_name,
         absl::string_view::const_iterator instance_it) const;
 
     // Returns true if the module depends on (uses) a given module
@@ -335,8 +338,7 @@ class AutoExpander {
   // Expands AUTOINST for the given module instance
   std::optional<Expansion> ExpandAutoinst(Module *module,
                                           const Symbol &instance,
-                                          absl::string_view type_id,
-                                          const Template *tmpl);
+                                          absl::string_view type_id);
 
   // Expands AUTO<port-direction/data-type> for the given module
   // Limitation: this only detects ports from AUTOINST. This limitation is also
@@ -430,11 +432,11 @@ const LazyRE2 AutoExpander::autoinst_re_{R"((/\*\s*AUTOINST\s*\*/))"};
 // (?:\s*\S+\s+AUTO_TEMPLATE\s*\n)*  – optional other AUTO_TEMPLATE types, end
 //                                     with newline
 // \s*\S+\s+AUTO_TEMPLATE            – at least one AUTO_TEMPLATE is required
-// \s*(".*")?                        – optional instance name regex
-// \s*\([\s\S]*?\);                  – parens with port connections
+// \s*(?:"([^"])*")?                 – optional instance name regex
+// \s*\(?:[\s\S]*?\);                – parens with port connections
 // \s*\*/                            – end of comment
 const LazyRE2 AutoExpander::autotemplate_re_{
-    R"((/\*(?:\s*\S+\s+AUTO_TEMPLATE\s*\n)*\s*\S+\s+AUTO_TEMPLATE\s*(".*")?\s*\([\s\S]*?\);\s*\*/))"};
+    R"((/\*(?:\s*\S+\s+AUTO_TEMPLATE\s*\n)*\s*\S+\s+AUTO_TEMPLATE\s*(?:"([^"]*)\")?\s*\([\s\S]*?\);\s*\*/))"};
 
 // AUTO_TEMPLATE type regex: the first capturing group is the instance type
 const LazyRE2 AutoExpander::autotemplate_type_re_{R"((\S+)\s+AUTO_TEMPLATE)"};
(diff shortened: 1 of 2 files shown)


verilog/tools/ls/autoexpand.cc
```diff
@@ -87,15 +87,16 @@ class AutoExpander {
 
   // Represents a port connection
   struct Connection {
-    absl::string_view port_name;  // The name of the port in the module instance
-    bool emit_dimensions;  // If true, when emitted, the connection should be
-                           // annotated with the signal dimensions
+    std::string port_name;  // The name of the port in the module instance
```
hzeller, line 90: Did you run into issues with the lifetime of strings, so you had to switch from string_view to std::string ?

verilog/tools/ls/autoexpand.cc
```diff
@@ -144,8 +145,10 @@ class AutoExpander {
   struct Template {
     using Map = absl::flat_hash_map<absl::string_view, std::vector<Template>>;
 
-    absl::string_view::const_iterator it;  // Location of the template in the
-                                           // source file
+    absl::string_view::const_iterator it;   // Location of the template in the
+                                            // source file
+    std::shared_ptr<RE2> instance_name_re;  // Regex for matching the instance
```
hzeller, line 150: Woulnd't unique_ptr<> be sufficient ? If not, it is probably a good idea to add a comment explaining why.

hzeller (commented):
Not fully through yet, but here first comments.

verilog/tools/ls/autoexpand.cc
```diff
@@ -996,29 +1022,38 @@ std::optional<AutoExpander::Expansion> AutoExpander::ExpandAutoinst(
   }
   const Module &inst_module = modules_.at(type_id);
 
+  // Find an AUTO_TEMPLATE that matches this instance
+  std::optional<absl::string_view> instance_name;
+  if (const auto instance_name_token =
+          GetModuleInstanceNameTokenInfoFromGateInstance(instance)) {
```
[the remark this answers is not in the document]
kbieganski: At this point we're sure this is a `GateInstance`, so it should have an identifier. Wouldn't get parsed otherwise. But after #1788 I don't want to say anything for certain, I guess.

verilog/tools/ls/autoexpand.cc
kbieganski: Added a comment

verilog/tools/ls/autoexpand.cc
```diff
@@ -416,11 +421,11 @@ const LazyRE2 AutoExpander::autoinst_re_{R"((/\*\s*AUTOINST\s*\*/))"};
 // (?:\s*\S+\s+AUTO_TEMPLATE\s*\n)*  – optional other AUTO_TEMPLATE types, end
 //                                     with newline
 // \s*\S+\s+AUTO_TEMPLATE            – at least one AUTO_TEMPLATE is required
-// \s*(".*")?                        – optional instance name regex
-// \s*\([\s\S]*?\);                  – parens with port connections
+// \s*(?:"(.*)")?                      – optional instance name regex
```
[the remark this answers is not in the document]
kbieganski: Replaced the dot in the regex with `[^"]`

verilog/tools/ls/autoexpand.cc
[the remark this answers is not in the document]
kbieganski: Made it so it emits an error if there's no instance name and bails. I don't think it's possible for the name not being there, but at least it won't crash if it happens.

verilog/tools/ls/autoexpand.cc
```diff
@@ -1,4 +1,4 @@
-// Copyright 2023 The Verible Authors.
+// Copyright 2023 The Verible Authors
```
hzeller, line 1: Lost a dot :)

verilog/tools/ls/autoexpand.cc
```diff
@@ -1052,29 +1077,40 @@ std::optional<AutoExpander::Expansion> AutoExpander::ExpandAutoinst(
   }
   const Module &inst_module = modules_.at(type_id);
 
+  // Find an AUTO_TEMPLATE that matches this instance
+  const verible::TokenInfo *instance_name_token =
+      GetModuleInstanceNameTokenInfoFromGateInstance(instance);
+  if (!instance_name_token) {
+    LOG(ERROR) << "AUTOINST: Instance with no name, aborting";
+    return {};
```
hzeller, line 1085: In a separate PR, but I think we should change all the `std::nullopt` returns from `{}` to actual `std::nullopt` .
Reading the code it can be confusing as it could be mistaken for an empty struct initialization of the std::optional value. Being explicit is nicer there.

verilog/tools/ls/autoexpand.cc
```diff
@@ -709,9 +724,17 @@ void AutoExpander::Module::SortPortsByLocation() {
...
   absl::string_view autotmpl_span;
+  absl::string_view autotmpl_inst_name;
   while (RE2::FindAndConsume(&autotmpl_search_span, *autotemplate_re_,
-                             &autotmpl_span)) {
+                             &autotmpl_span, &autotmpl_inst_name)) {
     Template tmpl{.it = autotmpl_span.begin()};
+    if (!autotmpl_inst_name.empty()) {
+      tmpl.instance_name_re = std::make_shared<RE2>(autotmpl_inst_name);
+      if (!tmpl.instance_name_re->ok()) {
+        LOG(ERROR) << "Invalid regex in AUTO template: " << autotmpl_inst_name;
```
hzeller, line 734: Given that this is user-supplied data: Do you have a test that explicitly passes a broken regular expression to see if  everything else around handles this situation gracefully ?

hzeller (commented):
LGTM, but I'd add a test with a broken regex

verilog/tools/ls/autoexpand.cc
kbieganski: Good point, I'll add a test

kbieganski: Done

verilog/tools/ls/autoexpand.cc
kbieganski: https://github.com/chipsalliance/verible/pull/1841

verilog/tools/ls/autoexpand.cc
```diff
@@ -648,20 +656,27 @@ void AutoExpander::Module::EmitUnconnectedOutputRegDeclarations(
...
       const auto it = tmpl->connections.find(port.name);
       if (it != tmpl->connections.end()) connected = it->second;
     }
+    size_t pos = connected.port_name.find('@');
+    while (pos != std::string::npos) {
+      connected.port_name.replace(pos, 1, instance_name.begin(),
+                                  instance_name.length());
+      pos = connected.port_name.find('@', pos);
+    }
+    connected.port_name = connected.port_name;
```
hzeller, line 673: this looks ... unnecessary

kbieganski: It does :rofl: not sure what happened there.

Issue #382, opened 2020-08-06
forbid negative constant integer literals in declared array dimensions


fangism:
e.g. `logic [-1:0] foo;`

Suggestion:

* Create test cases with arrays in various places: module instances, ports, class members, local variables...
* Examine where they appear in `verible-verilog-syntax --printtree`
* Implement a `SyntaxTreeLintRule` to find offending locations.
* As needed use/write CST helper functions to extract substructure.
* Only focus on expressions that are constant literals.

fangism:
b/160002985

IEncinas10:
Hello, I would like to try solving this one.

I have made a bit of progress (I'm able to build, made the .h, .cc, test, I'm able to run them) but it looks like negative numbers are not detected by  https://github.com/chipsalliance/verible/blob/a1534abbd83b9ca5b84bc577e303724a9ebb5994/verilog/CST/expression.cc#L65

I guess the problem is already stated in the function declaration.

```c++
// Unwrap layers of only-child nodes until reaching a leaf or a node with
// multiple children.
const Symbol* DescendThroughSingletons(const Symbol& symbol);

```

GetDimensionRange[Left/Right]Bound gives us the node of each bound, but if we add a negative sign the node has a *kUnaryPrefixExpression*, that makes *DescendThroughSingletons* give up and return false (it detects that it has two childs).

```
Node @0 (tag: kUnaryPrefixExpression) {
  Leaf @0 (#'-' @37-38: "-")
  Node @1 (tag: kNumber) {
    Leaf @0 (#TK_DecNumber @38-39: "1")
  }
}
```

I guess the fix would involve being able to evaluate *kUnaryPrefixExpressions*. Should I try to add that functionality? Any pointers would be greatly appreciated!



On a side note, just to check whether I'm on the right path: I have basically inspired the new rule on the [unpacked dimensions rule](https://github.com/chipsalliance/verible/blob/master/verilog/analysis/checkers/unpacked_dimensions_rule.cc) (and packed one) so that we check *kDimensionRange*S inside both Packed and UnpackedDimensions.

```c++
static const Matcher& DimensionValueMatcher() {
  static const Matcher matcher(NodekDimensionRange());
  return matcher;
}

void ForbidNegativeArrayDim::HandleSymbol(
    const verible::Symbol& symbol, const verible::SyntaxTreeContext& context) {
  if (!ContextIsInsidePackedDimensions(context) && !ContextIsInsideUnpackedDimensions(context)) return;

[...]

Issue #159, opened 2020-01-29
hide direct indexing into subtree structures behind functions in CST library


fangism:
In numerous places in the code-base, including linter analyzers and the formatter, we are directly `[indexing]` into syntax tree nodes' children (sometimes referencing `front()`/`back()`).  This is fragile, and prone to breakage through syntax tree restructuring.  Instead we should be defining/testing small helper functions to access subtrees (as if we had proper AST classes and accessors) in the CST library.

https://github.com/google/verible/tree/master/verilog/CST

Imagine if you will limiting use of the `GetSubtree*` functions in
https://github.com/google/verible/blob/master/common/text/tree_utils.h
and `operator[]` in https://github.com/google/verible/blob/master/common/text/concrete_syntax_tree.h

to only the CST library.

PR #1835, opened 2023-03-26
Add new lint rule for negative array dimensions

IEncinas10:
Attempt to fix https://github.com/chipsalliance/verible/issues/382. I'm sure the code could use some improvements, so happy to accomodate any feedback!

Edit:

Sorry for the several updates, after making the pull request I realized that there are some corner cases (parenthesis, ...) so I tried to fix some of those. It still won't pick up things like +(-5) but I can't see an easy way to fix that.

### DIFF
--- a/verilog/CST/expression.cc
+++ b/verilog/CST/expression.cc
@@ -100,7 +100,9 @@ const verible::Symbol* GetConditionExpressionFalseCase(
 
 const verible::TokenInfo* GetUnaryPrefixOperator(
     const verible::Symbol& symbol) {
-  const SyntaxTreeNode* node = &verible::SymbolCastToNode(symbol);
+  const SyntaxTreeNode* node = symbol.Kind() == SymbolKind::kNode
+                                   ? &verible::SymbolCastToNode(symbol)
+                                   : nullptr;
   if (!node || !MatchNodeEnumOrNull(*node, NodeEnum::kUnaryPrefixExpression)) {
     return nullptr;
   }
@@ -110,7 +112,9 @@ const verible::TokenInfo* GetUnaryPrefixOperator(
 }
 
 const verible::Symbol* GetUnaryPrefixOperand(const verible::Symbol& symbol) {
-  const SyntaxTreeNode* node = &verible::SymbolCastToNode(symbol);
+  const SyntaxTreeNode* node = symbol.Kind() == SymbolKind::kNode
+                                   ? &verible::SymbolCastToNode(symbol)
+                                   : nullptr;
   if (!node || !MatchNodeEnumOrNull(*node, NodeEnum::kUnaryPrefixExpression)) {
     return nullptr;
   }
--- a/verilog/CST/expression.cc
+++ b/verilog/CST/expression.cc
@@ -1,4 +1,4 @@
-// Copyright 2017-2020 The Verible Authors.
+// Copyright 2017-2023 The Verible Authors.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/verilog/CST/expression.cc
+++ b/verilog/CST/expression.cc
@@ -100,15 +100,21 @@ const verible::Symbol* GetConditionExpressionFalseCase(
 
 const verible::TokenInfo* GetUnaryPrefixOperator(
     const verible::Symbol& symbol) {
-  const auto& children = verible::SymbolCastToNode(symbol).children();
-  const auto leaf_symbol = children.front().get();
+  const SyntaxTreeNode* node = &verible::SymbolCastToNode(symbol);
+  if (!node || !MatchNodeEnumOrNull(*node, NodeEnum::kUnaryPrefixExpression)) {
+    return nullptr;
+  }
+  const verible::Symbol* leaf_symbol = node->children().front().get();
   return &verible::down_cast<const verible::SyntaxTreeLeaf*>(leaf_symbol)
               ->get();
 }
 
 const verible::Symbol* GetUnaryPrefixOperand(const verible::Symbol& symbol) {
-  const auto& children = verible::SymbolCastToNode(symbol).children();
-  return children.back().get();
+  const SyntaxTreeNode* node = &verible::SymbolCastToNode(symbol);
+  if (!node || !MatchNodeEnumOrNull(*node, NodeEnum::kUnaryPrefixExpression)) {
+    return nullptr;
+  }
+  return node->children().back().get();
 }
 
 std::vector<verible::TreeSearchMatch> FindAllBinaryOperations(
--- a/verilog/CST/expression.cc
+++ b/verilog/CST/expression.cc
@@ -98,6 +98,19 @@ const verible::Symbol* GetConditionExpressionFalseCase(
   return GetSubtreeAsSymbol(condition_expr, NodeEnum::kConditionExpression, 4);
 }
 
+const verible::TokenInfo* GetUnaryPrefixOperator(
+    const verible::Symbol& symbol) {
+  const auto& children = verible::SymbolCastToNode(symbol).children();
+  const auto leaf_symbol = children.front().get();
+  return &verible::down_cast<const verible::SyntaxTreeLeaf*>(leaf_symbol)
+              ->get();
+}
+
+const verible::Symbol* GetUnaryPrefixOperand(const verible::Symbol& symbol) {
+  const auto& children = verible::SymbolCastToNode(symbol).children();
+  return children.back().get();
+}
+
 std::vector<verible::TreeSearchMatch> FindAllBinaryOperations(
     const verible::Symbol& root) {
   return verible::SearchSyntaxTree(root, NodekBinaryExpression());
--- a/verilog/CST/expression.h
+++ b/verilog/CST/expression.h
@@ -81,6 +81,12 @@ const verible::Symbol* GetConditionExpressionTrueCase(const verible::Symbol&);
 // Returns the false-case expression of a kConditionExpression.
 const verible::Symbol* GetConditionExpressionFalseCase(const verible::Symbol&);
 
+// Returns the operator of a kUnaryPrefixExpression
+const verible::TokenInfo* GetUnaryPrefixOperator(const verible::Symbol&);
+
+// Returns the operand of a kUnaryPrefixExpression
+const verible::Symbol* GetUnaryPrefixOperand(const verible::Symbol&);
+
 // From binary expression operations, e.g. "a + b".
 // Associative binary operators may span more than two operands, e.g. "a+b+c".
 std::vector<verible::TreeSearchMatch> FindAllBinaryOperations(
--- a/verilog/CST/expression.h
+++ b/verilog/CST/expression.h
@@ -1,4 +1,4 @@
-// Copyright 2017-2020 The Verible Authors.
+// Copyright 2017-2023 The Verible Authors.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/verilog/CST/expression_test.cc
+++ b/verilog/CST/expression_test.cc
@@ -1,4 +1,4 @@
-// Copyright 2017-2020 The Verible Authors.
+// Copyright 2017-2023 The Verible Authors.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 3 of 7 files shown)


verilog/analysis/checkers/forbid_negative_array_dim.cc
```diff
@@ -0,0 +1,97 @@
...
+  if (!context.IsInsideFirst(
+          {NodeEnum::kPackedDimensions, NodeEnum::kUnpackedDimensions},
+          {NodeEnum::kBinaryExpression, NodeEnum::kUnaryPrefixExpression}))
+    return;
+
+  verible::matcher::BoundSymbolManager manager;
+  if (UnaryPrefixExprMatcher().Matches(symbol, &manager)) {
+    const auto& node = SymbolCastToNode(symbol);
+    const auto& children = node.children();
+    const auto& leaf_symbol = *children[0].get();
```
fangism, line 73: I recommend defining a named function for anything that indexes into the list of childre.
e.g. add a `GetUnaryPrefixExprOperator` to https://cs.opensource.google/verible/verible/+/master:verilog/CST/expression.h

verilog/analysis/checkers/forbid_negative_array_dim.cc
```diff
@@ -0,0 +1,97 @@
...
+    const auto& term =
+        verible::down_cast<const verible::SyntaxTreeLeaf&>(leaf_symbol);
+
+    // mandatory? {leaf, node}
+    if (children.size() <= 1) return;
+
+    int value = -1;
+    const auto& child_node = children[1];
```
fangism, line 81: Similar suggestion, maybe call this `GetUnaryPrefixExprOperand` in https://cs.opensource.google/verible/verible/+/master:verilog/CST/expression.h

verilog/analysis/checkers/forbid_negative_array_dim.cc
```diff
@@ -0,0 +1,97 @@
...
+    const bool is_constant = ConstantIntegerValue(*child_node.get(), &value);
+
+    const verible::TokenInfo token(TK_OTHER,
```
fangism, line 84: This can be declared inside the following `if` body, because it is only used there.

verilog/analysis/checkers/forbid_negative_array_dim.h
```diff
@@ -0,0 +1,48 @@
...
+#include "common/analysis/lint_rule_status.h"
+#include "common/analysis/syntax_tree_lint_rule.h"
+#include "common/text/symbol.h"
+#include "common/text/syntax_tree_context.h"
+#include "verilog/analysis/descriptions.h"
+
+namespace verilog {
+namespace analysis {
+
+class ForbidNegativeArrayDim : public verible::SyntaxTreeLintRule {
```
fangism, line 30: Recommend documenting what this class does here, with some simple examples.

verilog/analysis/checkers/forbid_negative_array_dim.h
```diff
@@ -0,0 +1,48 @@
+// Copyright 2017-2020 The Verible Authors.
```
fangism, line 1: might as well make this 2023 (other files too)

verilog/analysis/checkers/forbid_negative_array_dim_test.cc
```diff
@@ -0,0 +1,52 @@
...
+      {""},
+      {"logic l [", {kScalar, "-10"}, ":0];"},
+      {"logic [", {kScalar, "-10"}, ":-0] l;"},
+      {"logic l [0:", {kScalar, "-10"}, "];"},
+      {"logic l [+1:0];"},
+      {"logic l [-p:0];"},
+      {"logic l [10+(-5):0];"},
+      {"logic l [(", {kScalar, "-5"}, "):0];"},
+      {"logic l [-(-5):0];"},
+  };
```
fangism, line 46: Suggested cases:
* `foo_type [a:b] name [c:d]`
* declaration inside module ports
* declaration inside module body
* declaration inside class
* local declaration inside function

fangism (changes requested):
Thank you for contributing this!

verilog/analysis/checkers/forbid_negative_array_dim.cc
fangism: Incidentally, what I'm requesting here is aligned with #159

verilog/analysis/checkers/forbid_negative_array_dim.h
IEncinas10: Done! Updated also expression.cc and expression.h files

verilog/analysis/checkers/forbid_negative_array_dim.cc
IEncinas10: Went ahead and added the suggested functions. Let me know if they're ok or if you think they could use some more work.

verilog/CST/expression.cc
```diff
@@ -98,6 +98,19 @@ const verible::Symbol* GetConditionExpressionFalseCase(
   return GetSubtreeAsSymbol(condition_expr, NodeEnum::kConditionExpression, 4);
 }
 
+const verible::TokenInfo* GetUnaryPrefixOperator(
+    const verible::Symbol& symbol) {
+  const auto& children = verible::SymbolCastToNode(symbol).children();
```
hzeller, line 103: if symbol does not happen to be a node, then this will crash. You probably want to return `nullptr` in case `symbol.Kind()` is not a node.
Or you could test that it is a `kUnaryPrefixExpression`, which is a better, more specific, test.

(Also add a test for it to see that other nodes than kUnaryPrefixExpressions always return nullptr).

verilog/CST/expression.cc
```diff
@@ -98,6 +98,19 @@ const verible::Symbol* GetConditionExpressionFalseCase(
...
+  const auto leaf_symbol = children.front().get();
```
hzeller, line 104: In general, we try to avoid using `auto` unless whatever comes back is a huge templated type which does not add any benefit for the reader. So use the type that is returned.

verilog/CST/expression.cc
```diff
@@ -98,6 +98,19 @@ const verible::Symbol* GetConditionExpressionFalseCase(
...
+  return &verible::down_cast<const verible::SyntaxTreeLeaf*>(leaf_symbol)
```
hzeller, line 105: Do we know that `leaf_symbol` we get is a `SyntaxTreeLeaf` so that we can safely cast ?

verilog/CST/expression.h
```diff
@@ -81,6 +81,12 @@ const verible::Symbol* GetConditionExpressionTrueCase(const verible::Symbol&);
 // Returns the false-case expression of a kConditionExpression.
 const verible::Symbol* GetConditionExpressionFalseCase(const verible::Symbol&);
 
+// Returns the operator of a kUnaryPrefixExpression
```
hzeller, line 84: Please add tests for the new functions in expression_test.cc.

verilog/analysis/checkers/forbid_negative_array_dim.cc
```diff
@@ -0,0 +1,92 @@
...
+  verible::matcher::BoundSymbolManager manager;
+  if (UnaryPrefixExprMatcher().Matches(symbol, &manager)) {
+    int value = 0;
+
+    // Extract operand and operator from kUnaryPrefixExpression
+    const auto u_operator = verilog::GetUnaryPrefixOperator(symbol);
+    const auto operand = verilog::GetUnaryPrefixOperand(symbol);
+
+    const bool is_constant = verilog::ConstantIntegerValue(*operand, &value);
+    if (is_constant > 0 && value > 0 && u_operator->text() == "-") {
```
hzeller, line 79: is_constant is a bool, so using `>` on it is confusing.

verilog/analysis/checkers/forbid_negative_array_dim.cc
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  verible::matcher::BoundSymbolManager manager;
+  if (UnaryPrefixExprMatcher().Matches(symbol, &manager)) {
+    int value = 0;
+
+    // Extract operand and operator from kUnaryPrefixExpression
+    const auto u_operator = verilog::GetUnaryPrefixOperator(symbol);
+    const auto operand = verilog::GetUnaryPrefixOperand(symbol);
+
```
hzeller, line 77: If we're at this point, did the matcher make sure that u_operator and operand are never nullptr ?

verilog/CST/expression.cc
IEncinas10: I assumed every kUnaryPrefixExpression has the following form

```
Node @0 (tag: kUnaryPrefixExpression) {
                        Leaf @0 /* ... */
                        Node @1 /* ... /*
```

So I guess that if we're sure the current node is a kUnaryPrefixExpression we can assume the cast will work always? Let me know if I'm wrong.

verilog/CST/expression.h
IEncinas10: Done. Let me know if they're thorough enough.

verilog/CST/expression.cc
IEncinas10: Got it, thanks! Used auto because I saw it in use and thought that was the general style, looks much better now.

verilog/analysis/checkers/forbid_negative_array_dim.cc
IEncinas10: Yes, sorry about that. Removed "value > 0" check, noticed it was needed and when I added it back I somehow  made a mess.

verilog/analysis/checkers/forbid_negative_array_dim.cc
IEncinas10: As the matcher ensures that *symbol* is a kUnaryPrefixExpression, I would expect that they would never be nullptr (if my asumption about the general format for every kUnaryPrefixExpression is correct). If I'm wrong, I would just add an early return if either *operand* or *u_operator* are nullptr.

verilog/CST/expression.cc
IEncinas10: Added the checks in both functions, thanks pointing it out.

verilog/CST/expression.cc
hzeller: At the time of the writing of this comment, we only got `verible::Symbol`, so didn't know that yet.

verilog/CST/expression.cc
```diff
@@ -98,6 +98,25 @@ const verible::Symbol* GetConditionExpressionFalseCase(
   return GetSubtreeAsSymbol(condition_expr, NodeEnum::kConditionExpression, 4);
 }
 
+const verible::TokenInfo* GetUnaryPrefixOperator(
+    const verible::Symbol& symbol) {
+  const SyntaxTreeNode* node = &verible::SymbolCastToNode(symbol);
```
hzeller, line 103: this will crash if symbol is not a node.

verilog/CST/expression_test.cc
```diff
@@ -423,6 +423,59 @@ TEST(GetConditionExpressionFalseCaseTest, Various) {
...
+      EXPECT_EQ(NodeEnum(unary_expr.Tag().tag),
+                NodeEnum::kUnaryPrefixExpression);
+      EXPECT_TRUE(GetUnaryPrefixOperand(unary_expr));
+    } else {
+      EXPECT_NE(NodeEnum(last_node->Tag().tag),
+                NodeEnum::kUnaryPrefixExpression);
+    }
+  }
+}
+
```
hzeller, line 478: Also add tests where Symbol is _not_ a unary prefix expression. What if symbol is a leaf ?
Having test cases with broken input helps to make sure we handle that gracefully.

verilog/analysis/checkers/forbid_negative_array_dim.cc
```diff
@@ -0,0 +1,93 @@
...
+void ForbidNegativeArrayDim::HandleSymbol(
+    const verible::Symbol& symbol, const verible::SyntaxTreeContext& context) {
+  if (!context.IsInsideFirst(
+          {NodeEnum::kPackedDimensions, NodeEnum::kUnpackedDimensions},
+          {NodeEnum::kBinaryExpression, NodeEnum::kUnaryPrefixExpression})) {
+    return;
+  }
+
+  verible::matcher::BoundSymbolManager manager;
+  if (UnaryPrefixExprMatcher().Matches(symbol, &manager)) {
```
hzeller, line 71: If this is not matching, we can early out here. Saves one level of indentation and reduces the stack-size the reader of this code needs :)

```
if (!UnaryPrefixExprMatcher().Matches(symbol, &manager)) return;
```
... everything that follows implies that we have a unary prefix expression.

hzeller (commented):
Nice. A few more comments.

verilog/CST/expression_test.cc
IEncinas10: I added calls to the GetUnaryPrefix... functions to check they return a null pointer

verilog/CST/expression.cc
IEncinas10: Check added.

verilog/analysis/checkers/forbid_negative_array_dim_test.cc
```diff
@@ -0,0 +1,71 @@
...
+      {"logic l [", {kScalar, "-10"}, ":0];"},
+      {"logic l [+1:0];"},
+      {"logic l [-p:0];"},
+      {"logic l [10+(-5):0];"},
+      {"logic l [(", {kScalar, "-5"}, "):0];"},
+      {"logic l [-(-5):0];"},
+      /* Can't detect this at the moment */
+      {"logic l [+(-5):0];"},
+      /* Packed */
+      {"logic [", {kScalar, "-10"}, ":-0] l;"},
```
hzeller, line 48: I am wondering, should there be a configuration for the packed and unpacked ranges ?
Maybe a particular style allows to set a bitwidth with negative values but not array sizes.

Something to discuss in #382, then a potential implementation that distinguishes the ranges in a follow-up PR though.

hzeller (commented):
LGTM, @fangism any additional comments ?

fangism (approved):
This looks great, thank you for your hard work!

IEncinas10:
Thank you both! My pleasure, I'll search for other issues feasible for my with my little knowledge of the codebase. Feel free to point me in any direction that might be of use to you!

Thanks again, enjoy the weekend and thanks for the great project you manage!

PR #1836, opened 2023-03-27
[LSP] textDocument/references: Initial implementation

glatosinski:
This PR adds the implementation of the `textDocument/references` request handling, returning list of references for a given symbol

Tasks:

- [ ] Implemented finding references based on a SymbolTable
- [ ] Added support for `textDocument/references` request in VerilogLanguageServer
- [ ] Added tests for `textDocument/references` request


verilog/tools/ls/symbol-table-handler.cc
```diff
@@ -248,7 +251,43 @@ std::vector<verible::lsp::Location> SymbolTableHandler::FindDefinitionLocation(
   const verible::TextStructureView &text = parsedbuffer->parser().Data();
 
   const verible::TokenInfo cursor_token = text.FindTokenAt(cursor);
-  auto symbol = cursor_token.text();
+  return cursor_token.text();
+}
+
+std::optional<verible::lsp::Location>
+SymbolTableHandler::GetLocationFromSymbolTableNode(
+    absl::string_view symbol, const VerilogSourceFile *file_origin) {
```
hzeller, line 259: the word `symbol` is confusing here as it is often usef for a `verible::Symbol*`. Maybe `symbol_name` ?

verilog/tools/ls/symbol-table-handler.cc
```diff
@@ -248,7 +251,43 @@ std::vector<verible::lsp::Location> SymbolTableHandler::FindDefinitionLocation(
...
+  // TODO (glatosinski) add iterating over multiple definitions
+  verible::lsp::Location location;
```
hzeller, line 261: You need `location` much later. I suggest declaring it when you first need it.

verilog/tools/ls/symbol-table-handler.cc
```diff
@@ -248,7 +251,43 @@ std::vector<verible::lsp::Location> SymbolTableHandler::FindDefinitionLocation(
...
+    return {};
+  }
+  location.uri = PathToLSPUri(file_origin->ResolvedPath());
+  auto *textstructure = file_origin->GetTextStructure();
+  if (!textstructure) {
+    LOG(ERROR) << "Origin file's text structure is not parsed";
+    return {};
+  }
+  verible::LineColumnRange symbollocation =
+      textstructure->GetRangeForText(symbol);
```
hzeller, line 276: That symbol_name getting from the text structure must be coming from that text structure before (because it must be a substring), otherwise you can not look it up.

Do we know that is the case ?
And if so: why do we have to check that GetTextStructure() returned an object as we must've gotten the symbol from there.

verilog/tools/ls/symbol-table-handler.cc
```diff
@@ -248,7 +251,43 @@ std::vector<verible::lsp::Location> SymbolTableHandler::FindDefinitionLocation(
...
+  location.range.start = {.line = symbollocation.start.line,
+                          .character = symbollocation.start.column};
+  location.range.end = {.line = symbollocation.end.line,
+                        .character = symbollocation.end.column};
+  return location;
+}
+
+std::vector<verible::lsp::Location> SymbolTableHandler::FindDefinitionLocation(
```
hzeller, line 284: why does this function returns an array of locations when all it ever does is return a single location ?
I suspect it should in the future return possibly multiple locations, but the corresponding comment got lost.
 `// TODO add iterating over multiple definitions?`

verilog/tools/ls/symbol-table-handler.cc
```diff
@@ -248,7 +251,43 @@ std::vector<verible::lsp::Location> SymbolTableHandler::FindDefinitionLocation(
...
+    const verible::lsp::DefinitionParams &params,
+    const verilog::BufferTrackerContainer &parsed_buffers) {
+  Prepare();
+  const absl::string_view filepath = LSPUriToPath(params.textDocument.uri);
+  std::string relativepath = curr_project_->GetRelativePathToSource(filepath);
+  auto symbol = GetTokenAtTextDocumentPosition(params, parsed_buffers);
```
hzeller, line 290: Try to reduce the use of `auto`, it makes it really hard after a while to read the code as the human is not as good as the computer to see what `auto` is.

verilog/tools/ls/symbol-table-handler.h
```diff
@@ -48,6 +48,20 @@ class SymbolTableHandler {
   // Once the project's root is set, a new SymbolTable is created.
   void SetProject(const std::shared_ptr<VerilogProject> &project);
 
+  // prepares structures for symbol-based requests
+  void Prepare();
+
+  // Returns the token (symbol) of the file at a given position, specified
+  // in the TextDocumentPositionParams
+  absl::string_view GetTokenAtTextDocumentPosition(
```
hzeller, line 56: There is no parameter called 'token' or 'symbol' the comment refers to.
What happens if it is not found ?

verilog/tools/ls/symbol-table-handler.h
```diff
@@ -48,6 +48,20 @@ class SymbolTableHandler {
...
+      const verible::lsp::TextDocumentPositionParams &params,
+      const verilog::BufferTrackerContainer &parsed_buffers);
+
+  // Returns the location of the symbol from SymbolInfo in a
+  // Location JSON format
```
hzeller, line 61: Location is not yet json format, it is a `verible::lsp::Location` struct. So probably a good description could be  `... as a lsp::Location`

`symbol` is a bit of an overloaded term (often used as verible::Symbol*), so maybe `symbol_name` ?

verilog/tools/ls/symbol-table-handler.h
```diff
@@ -48,6 +48,20 @@ class SymbolTableHandler {
   // Once the project's root is set, a new SymbolTable is created.
   void SetProject(const std::shared_ptr<VerilogProject> &project);
 
+  // prepares structures for symbol-based requests
+  void Prepare();
```
hzeller, line 52: Does this need to be public ? We should be very specific in what should be visible in the API.
If there are functions needed in tests, it is ok to make public, but add a comment.

verilog/tools/ls/verilog-language-server_test.cc
```diff
@@ -580,6 +580,19 @@ std::string DefinitionRequest(absl::string_view file, int id, int line,
   return formattingrequest.dump();
 }
 
+// Creates a textDocument/definition request
+std::string ReferencesRequest(absl::string_view file, int id, int line,
```
hzeller, line 584: Can you assemble these from the verible::lsp:: structs ? The json is quickly getting pretty hard to read.
There is verible::lsp::Reference which we can use.

verilog/tools/ls/verilog-language-server_test.cc
```diff
@@ -580,6 +580,19 @@ std::string DefinitionRequest(absl::string_view file, int id, int line,
   return formattingrequest.dump();
 }
 
+// Creates a textDocument/definition request
```
hzeller, line 583: This comment looks like it has been copy/pasted from somewhere :)
It also indicates that we might have an opportunity to combine these.

hzeller (commented):
Nice!
A few first remarks.

In the Test, I suggest to avoid json as much as possible but use the verible::lsp:: structs as much as possible. They allow a type-safe way to set values in C++, and they seamlessly convert to and from json.

verilog/tools/ls/symbol-table-handler.cc
```diff
@@ -248,7 +251,43 @@ std::vector<verible::lsp::Location> SymbolTableHandler::FindDefinitionLocation(
...
+  }
+  location.uri = PathToLSPUri(file_origin->ResolvedPath());
+  auto *textstructure = file_origin->GetTextStructure();
+  if (!textstructure) {
+    LOG(ERROR) << "Origin file's text structure is not parsed";
+    return {};
+  }
+  verible::LineColumnRange symbollocation =
+      textstructure->GetRangeForText(symbol);
+  location.range.start = {.line = symbollocation.start.line,
```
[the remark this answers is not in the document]
glatosinski: Added TODO for that

verilog/tools/ls/symbol-table-handler.cc
glatosinski: I brought back the TODO comment. Originally the `std::vector<verible::lsp::Location>` was supposed to reflect how the `textDocument/definition` output looks like (a list of found definitions). In the original implementation it returns at most one definition. In the https://github.com/chipsalliance/verible/pull/1678, this method will return all places contributing to symbol definition (for multiline definitions).

glatosinski: I'd leave it as a vector for now, since it should be updated soon.

verilog/tools/ls/symbol-table-handler.cc
```diff
@@ -217,20 +218,22 @@ const SymbolTableNode *SymbolTableHandler::ScanSymbolTreeForDefinition(
...
   LoadProjectFileList(curr_project_->TranslationUnitRoot());
   if (files_dirty_) {
     BuildProjectSymbolTable();
   }
+}
+
+absl::string_view SymbolTableHandler::GetTokenAtTextDocumentPosition(
+    const verible::lsp::TextDocumentPositionParams &params,
+    const verilog::BufferTrackerContainer &parsed_buffers) {
   const absl::string_view filepath = LSPUriToPath(params.textDocument.uri);
```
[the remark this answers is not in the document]
glatosinski: Right, I removed the unnecessary filepath

verilog/tools/ls/symbol-table-handler.cc
glatosinski: Moved location declaration

verilog/tools/ls/symbol-table-handler.cc
glatosinski: Fixed

verilog/tools/ls/symbol-table-handler.cc
glatosinski: I fixed this in the `symbol-table-handler.cc` where possible (except for for-each loops)

verilog/tools/ls/symbol-table-handler.h
glatosinski: I updated the documentation, I think it should be better now.

verilog/tools/ls/symbol-table-handler.h
glatosinski: I made all internally used functions `private` for now (except for methods used in AUTO expansions).

verilog/tools/ls/verilog-language-server_test.cc
glatosinski: Yes, I updated the comment, used `verible::lsp`-based structure and moved most of the implementation to a common function used by `textDocument/references` and `textDocument/definition`.

hzeller (approved):
Nice! Just one suggestion.
(also, makes probably sense to update verilog/tools/ls/README.md and verilog/tools/ls/vscode/README.md with the added feature.

common/text/text_structure.cc
```diff
@@ -154,6 +154,14 @@ LineColumnRange TextStructureView::GetRangeForText(
   return {GetLineColAtOffset(from), GetLineColAtOffset(to)};
 }
 
+bool TextStructureView::ContainsText(absl::string_view text) const {
+  const int from = std::distance(Contents().begin(), text.begin());
```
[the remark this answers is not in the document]
glatosinski: Fixed, also added updates in READMEs.

common/text/text_structure.cc
```diff
@@ -154,6 +154,14 @@ LineColumnRange TextStructureView::GetRangeForText(
   return {GetLineColAtOffset(from), GetLineColAtOffset(to)};
 }
 
+bool TextStructureView::ContainsText(absl::string_view text) const {
+  const auto from = std::distance(Contents().begin(), text.begin());
+  const auto to = std::distance(Contents().begin(), text.end());
+  if (from < 0) return false;
+  if (to > static_cast<int>(Contents().length())) return false;
```
hzeller, line 161: If you still need a static cast here (I suspect because of signed, unsigned), better used `int64_t`

glatosinski: Fixed

Issue #1839, opened 2023-03-28
[Discussion] Smoke Test improvement roadmap


jbylicki:
# Smoke test improvement roadmap

## Introduction
The smoke-tests have around 30000 lines of error output, mostly regarding syntax errors, macro problems and header inclusion issues. Around 1400 of those lines are not yet categorized. As an example, the `opentitan` test analysis report shows:

|Name|Count|
| :--- | :--- |
|unresolved-macro|21|
|slang-verified-error|16|
|related-to-slang-validated-error|1056|
|standalone-header|571|
|misc-preprocessor|3|
|likely-unhandled-macro-call|7|
|related-to-likely-unhandled-macro-call|272|
|misc-preprocessor-related|19|
|undefined|13|
|All|1978|

To explain briefly the more miscellaneous categories, `slang-verified-error` means that the issue detected in smoke-tests had also been found with `slang`. `related-to-*` means that error `*` had subsequent issues that were related to the existence of a syntax error previously. This behavior was tested manually and I have found that the subsequent issues disappear when the root cause was dealt with. 

## Problems
As it can be clearly seen, the main issues are related to the preporcessor, mostly:

1. Mishandling macro calls, which in some parts of the syntax cause the issues to break the understanding of the rest of the file, eg. in `ibex` there is an enum with a macro that produces syntax errors at the macro itself:
```systemverilog
  typedef enum bit [5:0] {
    LOAD = 0,
    STORE,
    SHIFT,
    ARITHMETIC,
    LOGICAL,
    COMPARE,
    BRANCH,
    JUMP,
    SYNCH,
    SYSTEM,
    COUNTER,
    CSR,
    CHANGELEVEL,
    TRAP,
    INTERRUPT,
    `VECTOR_INCLUDE("riscv_instr_pkg_inc_riscv_instr_category_t.sv")
    AMO // (last one)
  } riscv_instr_category_t;

``` 

 Another use case where macro calls produce lots of syntax errors is visible in `black-parrot`, where parameters are declared using macros. There, the syntax tool throws 95 errors in a file starting with this declaration (`bp_be/src/v/bp_be_calculator/bp_be_calculator_top.sv`):

```systemverilog
module bp_be_calculator_top
 import bp_common_pkg::*;
 import bp_be_pkg::*;
 #(parameter bp_params_e bp_params_p = e_bp_default_cfg
    `declare_bp_proc_params(bp_params_p)
[...]

PR #1840, opened 2023-03-28
Ripgrep based output analysis in smoke-test-analyzer

jbylicki:
Added a last-resort method to detect include statements containing the filename of the tested file, such that if it was included, the syntax errors would be attributed to the file being taken out of context. This change was made during the preparation for #1839.


hzeller:
Looks like we also need to install ripgrep in the CI

jbylicki:
Yes, I have missed that at first glance, should be all green now.

hzeller:
Thanks!

PR #1842, opened 2023-03-28
Adding flag for portable build of Verible binaries

eszpotanski:
This PR adds `--//bazel:create_portable_executables` flag, which allows to generate portable binaries. It can be useful for e.g. creating VS Code extension where only generated binaries will be needed -- without additional shared libraries.

This flag can be used with `build` command, e.g.:
```
bazel build -c opt --//bazel:create_portable_executables //...
```

### DIFF
--- a/releasing/docker-run.sh
+++ b/releasing/docker-run.sh
@@ -28,6 +28,7 @@ export TARGET=`echo "$1" | sed 's#:#-#g'`
 
 TARGET_VERSION=`echo $TARGET | cut -d- -f2`
 TARGET_OS=`echo $TARGET | cut -d- -f1`
+export TARGET_LINK=${LINK_TYPE:-"dynamic"}
 
 export TAG=${TAG:-$(git describe --match=v*)}
 
@@ -72,6 +73,10 @@ case "$TARGET_OS" in
     cat ${TARGET_OS}/common/compiler.dockerstage >> ${OUT_DIR}/Dockerfile
   ;;
 esac
+
+if [ "${TARGET_LINK}" = static ]; then
+  BAZEL_OPTS="${BAZEL_OPTS} --//bazel:create_static_linked_executables"
+fi
 
 # Bazel
 cat bazel.dockerstage >> ${OUT_DIR}/Dockerfile
@@ -135,6 +140,7 @@ docker run --rm \
   -e GIT_VERSION="${GIT_VERSION:-$(git describe --match=v*)}" \
   -e GIT_DATE="${GIT_DATE:-$(git show -s --format=%ci)}" \
   -e GIT_HASH="${GIT_HASH:-$(git rev-parse HEAD)}" \
+  -e TARGET_LINK="${TARGET_LINK:-""}" \
   -v $(pwd)/out:/out \
   $IMAGE \
   ./releasing/build.sh
--- a/releasing/docker-run.sh
+++ b/releasing/docker-run.sh
@@ -74,8 +74,8 @@ case "$TARGET_OS" in
   ;;
 esac
 
-if [ "${TARGET_LINK}" = static ]; then
-  BAZEL_OPTS="${BAZEL_OPTS} --//bazel:create_static_linked_executables"
+if [ "${TARGET_LINK}" = "static" ]; then
+  BAZEL_OPTS="${BAZEL_OPTS} --config=create_static_linked_executables"
 fi
 
 # Bazel
--- a/releasing/docker-run.sh
+++ b/releasing/docker-run.sh
@@ -76,6 +76,9 @@ esac
 
 if [ "${TARGET_LINK}" = "static" ]; then
   BAZEL_OPTS="${BAZEL_OPTS} --config=create_static_linked_executables"
+  # Bazel link options with static libs cause error with
+  # --config=create_static_linked_executables and they are redundant
+  sed -i '/ENV BAZEL_LINK/d' ${OUT_DIR}/Dockerfile
 fi
 
 # Bazel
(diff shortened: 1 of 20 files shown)


hzeller:
Nice. I'd call the flag something that explains that this is static linking, e.g. `create_static_linked_executables`.

The word 'portable' has a different meaning which might indicate that the binary works on Linux, Windows and Mac and on any processor architecture.

hzeller:
Just tried, and I did have success compiling statically on a debian with bazel 6.1.1. I remember I had issues when I tried back in the day with older versions of bazel.

Can you test how far back this feature would work ? Currently, we specify as minimum bazel version 4.0.0 to compile Verible; if there is a cut-off point before it does not work, we can probably use that as recommendation as minimum, or we add that in when someone adds the `--//bazel:*` flag that if true it bails out if the version number does not match.

To test that the compile works, it needs to be added to the CI. Also, this then can be used to provide static binaries in the release (and we will also see if it works with ARM). 

So, next steps

  * [ ] Rename flag to better reflect the 'static building' aspect instead of 'portable'
  * [ ] Verify `bazel` version range the static linking works 
  * [ ] Document the flag in the `README`  (somewhere in the [build section]( https://github.com/chipsalliance/verible#build))
  * [ ] Add a static build for X86 and Arm to the CI
  * [ ] _(could be a follow-up PR):_ release the static built artifact on the [release page](https://github.com/chipsalliance/verible/releases)

releasing/docker-run.sh
```diff
@@ -73,6 +74,10 @@ case "$TARGET_OS" in
   ;;
 esac
 
+if [ "${TARGET_LINK}" = static ]; then
```
hzeller, line 77: Maybe quote the `"static"` string ?

hzeller (commented):
Nice! Just one comment.

I'd also like to have @umarcor have a look as he has experience with packaging.

When it turns out that the static binaries are not too big in the end, and we can show that the binaries work on all target distros, we could in a later pull request just consolidate and only provide two releases for Linux: one static linux x86_64 and one static linux Arm64.

hzeller:
Also adding Unai as FYI.
@eszpotanski Did you make sure that the static compilation works for all the systems ? By default, the artifacts are not built, but if you temporarily disable the condition on the `Build:` job (around line 320ish), you can make it build inside the CI:

```
if: ${{github.event_name == 'push'}}
```

hzeller:
Are you looking into working around the build-failures in the CI ?

eszpotanski:
It seems that Ubuntu introduced some changes to libc, which don't work well with `ld.gold` (on Debian it works just fine). I created config `create_static_linked_executables` with `ld.bfd` linker and disabled [supports_start_end_lib](https://bazel.build/docs/cc-toolchain-config-reference#wellknown-features) feature. I tested it on the `ubuntu:jammy` image (the same as in CI) and it worked, so now `compile-static` and `compile-static-clang` should pass.

I didn't enabled `Build` job, as there is still some errors. For now, I know that `BAZEL_LINKOPTS` and `BAZEL_LINKLIBS` cannot be set during static build, because it causes errors like undefined reference to functions from `cmath`, but unsetting these variables haven't resolved all issues.

releasing/docker-run.sh
```diff
@@ -135,6 +140,7 @@ docker run --rm \
   -e GIT_VERSION="${GIT_VERSION:-$(git describe --match=v*)}" \
   -e GIT_DATE="${GIT_DATE:-$(git show -s --format=%ci)}" \
   -e GIT_HASH="${GIT_HASH:-$(git rev-parse HEAD)}" \
+  -e TARGET_LINK="${TARGET_LINK:-""}" \
```
umarcor, line 143: I think that `-e _TARGET_LINK` would also work. You don't need to specify the value if you want it to be inherited from the host.

hzeller (approved):
LGTM. Can you also try @umarcor 's suggestion before we merge ?

hzeller:
Ah, if I read your comment correctly @eszpotanski it wouldn't work right now as it misses thigns such as cmath ? Maybe we need to explicitly add `--linkopt=-lm --host_linkopt=-lm` ? Or we also need to add  `apt-get install glibc-static` to the dependencies ?

eszpotanski:
This error was caused by `BAZEL_LINKOPTS` and `BAZEL_LINKLIBS`, as flags from these variables are placed almost at the beginning of the flags used during linking, so linker ignored them, because there weren't reference to this libraries yet. Moreover, these variables contains options connected with static libraries -- they are redundant during building statically linked executables, so I removing them if `"${TARGET_LINK}" = "static"`.

On the other hand, Centos couldn't find some libraries (e.g. for `-lm`) and I had to add `yum install -y glibc-static` to `centos/common/compiler.dockerstage`.

Moreover, on Ubuntu, `//verilogs/tools/lint:verible-verilog-lint` target ignores features and `linkopt` from `.bazelrc` or build's flags and it leads to the same errors like when `ld.gold` was used. I added required options in `cc_binary` of this target and it worked. Unfortunately, I haven't found better solution.

Right now, building statically linked executables works fine on `centos:7` and `ubuntu:jammy`. But on other supported Ubuntu versions generated `verible-verilog-lint` causes segmentation fault and I still have to check it.

Size of locally generated TAR file with statically linked executables is approximately 16MB, so static version is around 1.6 times bigger than dynamically linked one.

hzeller:
I think as long as we get one of the systems compile a static binary we should be good. Say we compile the static binaries always on `ubuntu:jammy`, then we know it will run on any of the ubuntu and centos systems as it is fully static.

So if you single out just one of the systems to build the static binaries, we're good.

releasing/docker-run.sh
eszpotanski: changed to `"static"`

releasing/docker-run.sh
eszpotanski: changed to `-e TARGET_LINK`

eszpotanski:
I changed Build job, so statically linked libraries are built only on `ubuntu:jammy`. I also changed generated TAR file name to `verible-VERSION-linux-static-ARCH.tar.gz`.

I pushed additional commit with commented `if: ${{github.event_name == 'push'}}` in Build job, so introduced changes can be tested on real CI.

hzeller:
This looks great!
I am going to merge this now so that we see how things end up in the release
(and in a separate PR, the comment on `if {github.event_name == 'push'}` can be removed again)

PR #1848, opened 2023-04-03
Avoid performing filename lint rules for stdin

IEncinas10:
Fixes https://github.com/chipsalliance/verible/issues/1359. Went ahead and fixed the package_filename_rule too. 

As I'm not familiar with the tool, I don't know if every time we consume stdin input the filename is always "-", or if we should have an array of possible stdin filenames to match against. For example: 
```c++
static const char *STDIN_FILENAMES[NUM_STDIM_NAMES] = {
  "-",
  "stdin",
  "something something"
};
```

I'm not sure if adding tests for this would be an easy task, but I suspect it isn't.

### DIFF
--- a/common/util/file_util.cc
+++ b/common/util/file_util.cc
@@ -152,7 +152,7 @@ absl::StatusOr<std::string> GetContentAsString(absl::string_view filename) {
   std::string content;
   std::ifstream fs;
   std::istream *stream = nullptr;
-  const bool use_stdin = filename == "-";  // convention: honor "-" as stdin
+  const bool use_stdin = IsStdin(filename);
   if (use_stdin) {
     stream = &std::cin;
   } else {
@@ -288,6 +288,11 @@ absl::StatusOr<Directory> ListDir(absl::string_view dir) {
   std::sort(d.files.begin(), d.files.end());
   std::sort(d.directories.begin(), d.directories.end());
   return d;
+}
+
+bool IsStdin(absl::string_view filename) {
+  static constexpr absl::string_view kStdinFilename = "-";
+  return filename == kStdinFilename;
 }
 
 namespace testing {
--- a/common/util/file_util.h
+++ b/common/util/file_util.h
@@ -33,6 +33,12 @@
 namespace verible {
 namespace file {
 
+// Check if the input filename corresponds to stdin.
+// Convention: honor "-" as stdin
+// Added to avoid performing lint filename rules such as
+// {package,module}_filename_rule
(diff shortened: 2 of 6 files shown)


common/util/file_util.cc
```diff
@@ -290,6 +290,9 @@ absl::StatusOr<Directory> ListDir(absl::string_view dir) {
   return d;
 }
 
+static const char *STDIN_FILENAME = "-";
+bool IsStdin(absl::string_view filename) { return filename == STDIN_FILENAME; }
```
hzeller, line 294: Can you also use this new function in the `GetContentAsString()` function, around line 155 ?

common/util/file_util.cc
```diff
@@ -290,6 +290,9 @@ absl::StatusOr<Directory> ListDir(absl::string_view dir) {
   return d;
 }
 
+static const char *STDIN_FILENAME = "-";
```
hzeller, line 293: To limit the scope, I suggest to put this constant in the `IsStdin()` function.

Constants should have a `k` prefix and otherwise be camel-case according [to the styleguide we use](https://google.github.io/styleguide/cppguide.html#Constant_Names).

Also, if you make it an `absl::string_view` we save the strlen() call at runtime.

```
bool IsStdin(absl::string_view filename) {
  static constexpr absl::string_view kStdinFilename = "-";
  return filename == kStdinFilename;
}
```

hzeller (commented):
Nice! We should also use that function in the one place where we decide to read from stdin in the same file_uttil.cc, see comments below.

We have end-to=end tests for the linter in
verilog/tools/lint/lint_tool_test.sh
... maybe adding a test there would be relatively easy.

common/util/file_util.cc
IEncinas10: Thanks for the feedback! Sorry about the naming convention

hzeller: I've fixed the smoke-test expecations in #1849 so if you do a rebase, that should work then.

common/util/file_util.cc
IEncinas10: Sure! Didn't notice that part of the code. This answers my doubt about stdin being uniquely identified by "-". Applied suggested changes too.

common/util/file_util.cc
IEncinas10: Tried to avoid having a merge commit with my own branch and ended up making a mess. I'll try going back and just adding the merge commit.

hzeller (approved):
Thanks!

PR #1858, opened 2023-04-06
Make generated signal width the max of connected signal widths

kbieganski:
If a generated wire or port is connected to multiple instance ports, its width
is taken as the maximum of widths of all connected ports. This only works if the
dimensions are given as numbers. Otherwise, the fallback is to just use the
first relevant dimensions.


hzeller (approved):
Nice

PR #1863, opened 2023-04-13
line-length-rule: Be more robust in walking token iterators.

hzeller:
Making sure to not wander outside of range. Discovered by fuzzing.

b/269962677


verilog/analysis/checkers/line_length_rule.cc
```diff
@@ -84,12 +84,19 @@ static bool AllowLongLineException(TokenSequence::const_iterator token_begin,
   // TODO(b/134180314): Preserve all text in lexer.
   if (token_begin == token_end) return true;  // Conservatively ignore.
   auto last_token = token_end - 1;            // Point to last token.
-  if (last_token->token_enum() == verible::TK_EOF) --last_token;
+  if (last_token > token_begin && last_token->token_enum() == verible::TK_EOF) {
```
fangism, line 87: (No action needed). Just an observation: the use of greater/less than comparisons implies that the iterator is `RandomAccess`ible, but I think that was already the case with `token_end - 1`.

fangism (approved):

Issue #526, opened 2020-10-07
[verible-patch-tool] color the diffs/hunks presented to the user


fangism:
For the underlying patch-tool described here:
https://github.com/google/verible#interactive-formatting

It would be nice if the diff could be colorized for readability.  Currently it is monochrome.

Example mock-up using (github markdown's default coloring):

```diff
$ verible-verilog-format-changed-lines-interactive.sh 
--- /home/fangism/local/src/lowRISC/opentitan/hw/ip/aes/rtl/aes_core.sv
+++ NEW/home/fangism/local/src/lowRISC/opentitan/hw/ip/aes/rtl/aes_core.sv
@@ -32,3 +32,3 @@
   // Signals
-  logic                 foo_bar;
+  logic foo_bar;
   logic                 ctrl_qe;
Apply this hunk? [y,n,a,d,s,q,?] n
@@ -107,3 +107,3 @@
 
-      logic   illogical;
+  logic illogical;
 
Apply this hunk? [y,n,a,d,s,q,?] n
```

Would also color the prompt `Apply this hunk?` a little differently like this (dark-themed):

![color-patch-tool](https://user-images.githubusercontent.com/607886/95371982-ca299380-088f-11eb-82ef-3a9814c71c5a.png)

Task: 

* Detect when running in interactive mode, e.g. using `isatty()`, and only apply colors in this case.
  * Possibly detect terminal type as well. 
* Make sure existing tests still work (ok that they all run non-interactively)

fangism:
b/161930379

DhairyaBahl:
@fangism I would like to work on this issue. 

Thanks :)

fangism:
Thank you.

fangism:
@DhairyaBahl had any time to look into this?

DhairyaBahl:
@fangism  I am extremely sorry for delay towards the issue. I will try my best to push a PR in a day or so. 

Thank you :) sorry again :)

DhairyaBahl:
@fangism Work Update: I am trying to understand how verible works, about its various features to resolve this issue and to first understand the issue. I get that i have to change the color of the text or make it look more interactive but i am confused that how to change the colordiff if its installed by default in the system by the apt-get install commands. Kindly guide me a bit to how to approach this problem. 

Thanks a lot for having patience towards me :)

fangism:
The patch-tool is a small piece of the Verible project that is generic and only operates on patch files.

[...]

PR #1866, opened 2023-04-15
Color the diffs/hunks presented to the user #526

IEncinas10:
Hello, as the issue seemed stale I took the liberty of making a pull request, hope that's ok.

This would be my preferred approach to add coloring to the terminal output, but let me know if you think other solution is better.

Once we agree on how to add the colors, I'd be happy to make it configurable (ON/OFF, Different color palletes) if you want.

Current output:
image
* the typo in '---' NEW instead of '+++' has already been corrected

The issue says "Possibly detect terminal type as well". Could you ellaborate on what that means?

TODO:

- [x] Add tests to coloring functions once we have a stable api
- [ ] Detect terminal type (?)
- [ ] Make it configurable
- [x] Document

### DIFF
--- a/common/strings/patch.cc
+++ b/common/strings/patch.cc
@@ -82,7 +82,11 @@ std::ostream& operator<<(std::ostream& stream, const MarkedLine& line) {
   const term::Color c = line.IsDeleted() ? term::Color::kRed
                         : line.IsAdded() ? term::Color::kCyan
                                          : term::Color::kNone;
-  return stream << StartColor(stream, c) << line.line << EndColor(stream, c);
+  return term::Colored(stream, line.line, c);
+}
+
+std::string HunkIndices::ToString() const {
+  return absl::StrCat(start, ",", count);
 }
 
 absl::Status HunkIndices::Parse(absl::string_view text) {
@@ -100,7 +104,7 @@ absl::Status HunkIndices::Parse(absl::string_view text) {
 }
 
 std::ostream& operator<<(std::ostream& stream, const HunkIndices& indices) {
-  return stream << indices.start << ',' << indices.count;
+  return stream << indices.ToString();
 }
 
 absl::Status HunkHeader::Parse(absl::string_view text) {
@@ -149,10 +153,11 @@ absl::Status HunkHeader::Parse(absl::string_view text) {
 }
 
 std::ostream& operator<<(std::ostream& stream, const HunkHeader& header) {
-  return stream << term::StartColor(stream, term::Color::kGreen) << "@@ -"
-                << header.old_range << " +" << header.new_range << " @@"
-                << header.context
-                << term::EndColor(stream, term::Color::kGreen);
+  return term::Colored(
+      stream,
+      absl::StrCat("@@ -", header.old_range.ToString(), " +",
+                   header.new_range.ToString(), " @@", header.context),
+      term::Color::kGreen);
 }
 
 // Type M could be any container or range of MarkedLines.
@@ -408,10 +413,10 @@ absl::Status FilePatch::PickApply(std::istream& ins, std::ostream& outs,
 
   if (!hunks_.empty()) {
     // Display the file being processed, if there are any hunks.
-    outs << term::StartColor(outs, term::Color::kRed) << "--- "
-         << old_file_.path << term::EndColor(outs) << '\n';
-    outs << term::StartColor(outs, term::Color::kCyan) << "+++ "
-         << new_file_.path << term::EndColor(outs) << '\n';
+    term::Colored(outs, absl::StrCat("--- ", old_file_.path, "\n"),
+                  term::Color::kRed);
+    term::Colored(outs, absl::StrCat("+++ ", new_file_.path, "\n"),
+                  term::Color::kCyan);
   }
 
   const std::vector<absl::string_view> orig_lines(SplitLines(*orig_file_or));
--- a/common/strings/patch.cc
+++ b/common/strings/patch.cc
@@ -79,10 +79,10 @@ absl::Status MarkedLine::Parse(absl::string_view text) {
 }
 
 std::ostream& operator<<(std::ostream& stream, const MarkedLine& line) {
-  term::Color c = line.IsDeleted() ? term::Color::kRed
-                  : line.IsAdded() ? term::Color::kCyan
-                                   : term::Color::kNone;
-  return stream << StartColor(c) << line.line << EndColor(c);
+  const term::Color c = line.IsDeleted() ? term::Color::kRed
+                        : line.IsAdded() ? term::Color::kCyan
+                                         : term::Color::kNone;
+  return stream << StartColor(stream, c) << line.line << EndColor(stream, c);
 }
 
 absl::Status HunkIndices::Parse(absl::string_view text) {
@@ -149,9 +149,10 @@ absl::Status HunkHeader::Parse(absl::string_view text) {
 }
 
 std::ostream& operator<<(std::ostream& stream, const HunkHeader& header) {
-  return stream << term::StartColor(term::Color::kGreen) << "@@ -"
+  return stream << term::StartColor(stream, term::Color::kGreen) << "@@ -"
                 << header.old_range << " +" << header.new_range << " @@"
-                << header.context << term::EndColor(term::Color::kGreen);
+                << header.context
+                << term::EndColor(stream, term::Color::kGreen);
 }
 
 // Type M could be any container or range of MarkedLines.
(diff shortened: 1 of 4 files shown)


common/strings/patch.cc
```diff
@@ -363,7 +367,8 @@ LineNumberSet FilePatch::AddedLines() const {
 static char PromptHunkAction(std::istream& ins, std::ostream& outs) {
   // Suppress prompt in noninteractive mode.
   if (IsInteractiveTerminalSession()) {
-    outs << "Apply this hunk? [y,n,a,d,s,q,?] ";
+    outs << term::StartColor(term::Color::kYellow)
```
hzeller, line 370: I think I'd avoid yellow, as it is really hard to see on terminals with white background. How about we use `term::bold()` here ?

common/strings/patch.cc
```diff
@@ -79,7 +79,10 @@ absl::Status MarkedLine::Parse(absl::string_view text) {
 }
 
 std::ostream& operator<<(std::ostream& stream, const MarkedLine& line) {
-  return stream << line.line;
+  term::Color c = line.IsDeleted() ? term::Color::kRed
```
hzeller, line 82: I'd make this `const`.

hzeller (commented):
Nice!
Did you check that we only use the patch output in the context of cout/cerr ? Because only then the auto-detection of the interactive session (the result of `IsInteractiveTerminalSession()`) make sense. If there is a situation in which the patch logic is used to write to an std::ostream that is a fstream-opened file, then we will end up writing colored output to a file just because we happened to be in an interactive session.

common/strings/patch.cc
```diff
@@ -79,7 +79,10 @@ absl::Status MarkedLine::Parse(absl::string_view text) {
...
+                  : line.IsAdded() ? term::Color::kCyan
```
hzeller, line 83: (The common color would be green here, but I like cyan as it is more color-blind friendly.)

IEncinas10:
> Nice! Did you check that we only use the patch output in the context of cout/cerr ? Because only then the auto-detection of the interactive session (the result of `IsInteractiveTerminalSession()`) make sense. If there is a situation in which the patch logic is used to write to an std::ostream that is a fstream-opened file, then we will end up writing colored output to a file just because we happened to be in an interactive session.

Code says "**outs** is the stream that displays text and prompts to the user".

https://github.com/chipsalliance/verible/blob/525ffaf7cdc273e41a9b5c04b199bc0a1275e731/common/strings/patch.h#L74

```c++
  // Interactively prompt user to select hunks to apply in-place.
  // 'ins' is the stream from which user-input is read,
  // and 'outs' is the stream that displays text and prompts to the user.
  absl::Status PickApplyInPlace(std::istream& ins, std::ostream& outs) const;
```

 Of course we can trick the tool by doing

```
verible-verilog-format-changed-lines-interactive.sh > file_output
```

and that would write "colored" strings into a file, so we would see the escape characters. I guess this is ok.

common/strings/patch.cc
IEncinas10: You're right, missed this. Thanks!

common/strings/patch.cc
IEncinas10: I agree, didn't think about this. Bold seems right for every type of background.

common/util/user_interaction.cc
```diff
@@ -72,5 +82,21 @@ std::string inverse(absl::string_view s) {
...
+absl::string_view StartColor(const std::ostream& stream, Color c) {
+  if (!IsInteractiveTerminalSession() || stream.rdbuf() != std::cout.rdbuf()) {
+    return "";
+  }
+  return kColorsStart[c];
+}
+
+absl::string_view EndColor(const std::ostream& stream, Color c) {
+  if (!IsInteractiveTerminalSession() || c == Color::kNone ||
+      stream.rdbuf() != std::cout.rdbuf()) {
```
hzeller, line 95: I like this idea of testing if the stream we're talking about is also indeed the stdout connection. I think this should be part of `IsInteractiveTerminalSession()`.

I've opened
https://github.com/chipsalliance/verible/pull/1870
to prepare that. After that is submitted, you'd need to rebase your change to that.

common/strings/patch.cc
```diff
@@ -146,8 +149,10 @@ absl::Status HunkHeader::Parse(absl::string_view text) {
 }
 
 std::ostream& operator<<(std::ostream& stream, const HunkHeader& header) {
-  return stream << "@@ -" << header.old_range << " +" << header.new_range
-                << " @@" << header.context;
+  return stream << term::StartColor(stream, term::Color::kGreen) << "@@ -"
+                << header.old_range << " +" << header.new_range << " @@"
+                << header.context
+                << term::EndColor(stream, term::Color::kGreen);
```
fangism, line 155: Can we provide an interface that uses RAII to automatically change back to the previous color setting, so no one has to ever write `EndColor`?

common/util/user_interaction.cc
IEncinas10: Nice catch! :+1:

IEncinas10:
Sorry about the mess, tried to fix the changes very quickly. I'll pick it up later in the afternoon

common/strings/patch.cc
IEncinas10: What you suggest was my initial approach, but this snippet of code made me change my mind as I didn't want to change many things.

The other alternative I see is doing something like this:

```c++

return stream << term::Color(stream, kGreen, StrCat("@@ -", header.old_range, ...));
```

I don't have a clear preference, so happy to adapt it either way you find cleaner

IEncinas10: Sorry, I missremembered this earlier. The reason I went with Start/EndColor is because in this function we rely on the overloaded operator<< for HunkIndices, so I didn't want to mess with that. 

The easiest fix I see is removing operator<< for HunkIndices and substituting it for a to_string() method so we could do the following

```c++
std::ostream& operator<<(std::ostream& stream, const HunkHeader& header) {
  return term::Colored(
      stream,
      absl::StrCat("@@ -", header.old_range.to_string(), " +",
                   header.new_range.to_string(), " @@", header.context),
      term::Color::kGreen);
}
```

common/strings/patch.h
```diff
@@ -147,6 +147,8 @@ struct HunkIndices {
   }
   bool operator!=(const HunkIndices& other) const { return !(*this == other); }
 
+  std::string ToString() const;
```
hzeller, line 150: something named `ToString()` sounds a bit like a DebugString, but here it is actually formatted as expected in a patch, so I'd try to see if there can be a better name for this.
(Or even: remove it entirely and keep it with the original `operator<<` implementation).

Anyway, not a biggy, but if you have a good idea for different naming, this can also be done in a later PR.

hzeller (commented):
Nice, I think we're essentially ready to go.
Just two remarks: enum -> enum class we should do now; the naming issue could be deferred to a later PR if needed.

common/strings/patch.h
IEncinas10: I agree naming is not great. The need for the separate function comes from 
```c++
std::ostream& operator<<(std::ostream& stream, const HunkHeader& header) {
  return term::Colored(
      stream,
      absl::StrCat("@@ -", header.old_range.ToString(), " +",
                   header.new_range.ToString(), " @@", header.context),
      term::Color::kGreen);
}
```
I don't see a way around having that explicit "ToString" function because we require term::Colored to receive strings. Perhaps something like: Format, FormatToString is clearer?

common/util/user_interaction.h
```diff
@@ -25,6 +25,19 @@ namespace term {
 // is an interactive session connected to a terminal. Otherwise just plain.
 std::ostream& bold(std::ostream& out, absl::string_view s);
 std::ostream& inverse(std::ostream& out, absl::string_view s);
+
+enum Color {
```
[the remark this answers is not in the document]
IEncinas10: Just out of curiosity, what would be most appropiate here?

```suggestion
enum class Color : uint8_t {
```

or

```suggestion
enum class Color {
```

Defaulting to 'int' underlying type to use the value as an index into an array seems weird to me, although I know the difference is insignificant in this case.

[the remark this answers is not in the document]
hzeller: I think the 'class' part is more important (type-safety), the fact that it internally is an int I would only worry when wrinting things for a microcontroller :)

common/strings/patch.h
hzeller: Yes, maybe `FormatToString()` would be nicer (it has less of the 'this is a debug-output' vibe).

common/util/user_interaction.h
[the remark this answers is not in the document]
IEncinas10: Right, makes sense. Sorry about the 'Suggested change' stuff, I thought it was just a visual thing, didn't realize it is mean to do a quick fix. The static cast was missing form the suggestion :facepalm: I'll fix it now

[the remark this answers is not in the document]
hzeller: Hah, and I thought it was the 'choose your own poison' button :)

I kindof like the suggestion change thing, how did you manage to include it ? Just adding a diff ?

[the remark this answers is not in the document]
IEncinas10: It shows in my web UI (probably experimental) but it should work for anyone by annotating a ``` block with sugestion (instead of c++ or whatever to get highlighting)

PR #1870, opened 2023-04-17
Refine interactive terminal session detecation.

hzeller:
Not only stdin is queried, but also if the output stream we attempt to write to is also a terminal (as this is used to determine if we use terminal code to output bold or inverse text).


hzeller:
CC: @IEncinas10 
I've refined  the interactive session testing as well as bold and inverse to always test if the stream we're about to write to is a terminal.

fangism (approved):

PR #1875, opened 2023-04-20
Unify names for bazel targets.

hzeller:
This was a bit mixed. Now standardized to
  * short names as-is, longer names with `lib-with-dash-separator`
  * tests and unittest end with _test or _unittest. So `foo-bar_test`

For later: rename `*_unittest` to `*_test`, but deferred until there is no PR open anymore that modifies such a file.


fangism (approved):
ok, i suppose, though test target names read a little weird to me, like `format-file-lex-error_test`.  `-`feels like a lower precedence operator to me than `_` because`-` is less universally accepted in identifiers. I couldn't find definitive style guidance in my one minute of searching around.

PR #1878, opened 2023-04-20
Move function to convert verible source range to lsp::Range in header.

hzeller:
We use the same conversion in multiple places, so let's give it a name and put in a header.


glatosinski (approved):
LGTM

PR #1880, opened 2023-04-21
Added support for macro calls inside struct bodies

### LINKED ISSUE #1873: Using a call-style preprocessor macro in a struct trips the formatter
**Test case**

```systemverilog
module foo
#(
  parameter type baz_t = struct packed { `BAZ(); }
)
();
endmodule
```

This results in an error:
```
/tmp/a.sv: /tmp/a.sv:3:42-45: syntax error at token "`BAZ"
/tmp/a.sv:3:50: syntax error at token "}"
/tmp/a.sv:6:1-9: syntax error at token "endmodule"
```

The main problem here is, that verible has limited preprocessing features.
But having said that, this should not trip the formatter that in this case and ideally just regards this as some opaque token.

In fact, the following (without parenthesis at the `BAZ` invocation) works:

```systemverilog
module foo #(
    parameter type baz_t = struct packed { `BAZ; }
) ();
endmodule
```

b/275731119

jbylicki:
This PR implements support for macro calls in the struct body. The syntax allows the flexibility to use or skip the semicolon.

Fixes #1873

### DIFF
--- a/verilog/formatting/formatter_test.cc
+++ b/verilog/formatting/formatter_test.cc
@@ -8258,6 +8258,19 @@ static constexpr FormatterTestCase kFormatterTestCases[] = {
      "endmodule : foo  // C\n"
      "                 // C.1\n"
      "                 // C.2\n"},
+    {"module foo\n"
+     "#(\n"
+     "  parameter type baz_t = struct packed { `BAZ(); }\n"
+     ")\n"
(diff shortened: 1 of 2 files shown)


hzeller:
I am going to merge this now to play around with it, but please add the non-semicolon test in a separate PR.

PR #1881, opened 2023-04-21
Added option to break apart end and else tokens

jbylicki:
This PR implements a new flag to the formatter: `wrap_end_else_clauses`. It allows the control over if end-else pairs should be formatted as `[...] end else if...` or
```
end
else if (...)
```

### DIFF
--- a/verilog/formatting/token_annotator.cc
+++ b/verilog/formatting/token_annotator.cc
@@ -849,9 +849,12 @@ static WithReason<SpacingOptions> BreakDecisionBetween(
   if (right.TokenEnum() == TK_else) {
     // TODO(fangism): feels like this should be the responsibility of
     // tree_unwrapper, handled by kElseClause, kGenerateElseClause, etc.
-    if (left.TokenEnum() == TK_end) {
+    if (left.TokenEnum() == TK_end && !style.wrap_end_else_clauses) {
       return {SpacingOptions::kMustAppend,
               "'end'-'else' and should be together on one line."};
+    }
+    if (left.TokenEnum() == TK_end && style.wrap_end_else_clauses) {
+      return {SpacingOptions::kMustWrap, "'end'-'else' Should be split."};
     }
     if (left.TokenEnum() == '}') {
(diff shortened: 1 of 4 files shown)


hzeller:
Nice!
Can you also update [verilog/tools/formatter/README.md](https://github.com/chipsalliance/verible/blob/master/verilog/tools/formatter/README.md) with the updated flags ?

jbylicki:
I've added the flag description to the usage snippet

verilog/tools/formatter/README.md
```diff
@@ -74,6 +74,8 @@ To pipe from stdin, use '-' as <file>.
     --try_wrap_long_lines (If true, let the formatter attempt to optimize line
       wrapping decisions where wrapping is needed, else leave them unformatted.
       This is a short-term measure to reduce risk-of-harm.); default: false;
+    --wrap_end_else_clauses (Split end and else keywords into separate lines);
+      default: false; currently: true;
```
hzeller, line 78: Where is the `curerently: true;` text annotation coming from ? Were you passing the flag at the time you created the --helpfull output ?

verilog/formatting/token_annotator.cc
```diff
@@ -849,10 +849,13 @@ static WithReason<SpacingOptions> BreakDecisionBetween(
   if (right.TokenEnum() == TK_else) {
     // TODO(fangism): feels like this should be the responsibility of
     // tree_unwrapper, handled by kElseClause, kGenerateElseClause, etc.
-    if (left.TokenEnum() == TK_end) {
+    if (left.TokenEnum() == TK_end && !style.wrap_end_else_clauses) {
       return {SpacingOptions::kMustAppend,
               "'end'-'else' and should be together on one line."};
     }
+    if (left.TokenEnum() == TK_end && style.wrap_end_else_clauses) {
```
hzeller, line 856: Test opportunity: if you change this line to
```
if (left.TokenEnum() == TK_end || style.wrap_end_else_clauses) 
```

... then no test actually notices that and triggers...

Issue #1524, opened 2022-10-20
Mismatched token enum in formatter output: ')' vs. MacroCallCloseToEndLine


anjanaskumar27:
**Test case**

```systemverilog
module a (
  input wire [`clog2(2):0] b
);
endmodule
```

**Actual output**

```systemverilog
// This doesn't look right.
module a (
    input wire [
    `clog2(2)
    :0] b
);
endmodule
```

Include any possible diagnostic messages from the formatter.

**Expected or suggested output**

```systemverilog
Formatted output is lexically different from the input.    Please file a bug.  Details:
Mismatched token enums.  got: (')') (#41: ")") vs. (MacroCallCloseToEndLine) (#711: ")")
First mismatched token [1974]: (')') (#41: ")") vs. (MacroCallCloseToEndLine) (#711: ")")
```

Citations to published style guides would help.

hzeller:
Looks like a closing parenthesis of a macro call at the end of a line gets its own token `MacroCallCloseToEndLine`, possibly to be a useful distinguishing factor in the [uvm semicolon rule](https://github.com/chipsalliance/verible/blob/39ed2e3f737b8f8bc61dc5fbe99be09091f63e40/verilog/analysis/checkers/uvm_macro_semicolon_rule.cc#L89).

Two things come to mind here
  * Maybe for comparison of token enums we should (probably only a few ?) equivalence classes. That way the triggering [enum comparison](https://github.com/chipsalliance/verible/blob/39ed2e3f737b8f8bc61dc5fbe99be09091f63e40/verilog/analysis/verilog_equivalence.cc#L170) could see this as equivalent.
  * Independent of that: Fix the formatter to actually not add a newline after a macro in these circumstances ``wire [`foo():0]``

hzeller:
b/230396830

hzeller:
Since you're already working towards a solution, assigning to you @jbylicki

hzeller:
With the recent changes, it at least does not complain anymore, but the formatting result is ... not super pretty, indicating that some of these tokens are not well classified in the formatter

```systemverilog
module a (
    input wire [
    `clog2(2)
    :0] b
);
endmodule
```

kbrunham-intel:
@hzeller I retested the reduced case from this issue with `verible-verilog-format` built from tag `v0.0-4294-gc1d8f5e8`. I could not reproduce the problem (formatter exits 0; output keeps the expected layout / converges). I would suggest closing this issue.

[...]

Issue #1883, opened 2023-04-22
verible-verilog-format fails verification with macro


nathanbeckmann:
**Test case**

```
module test

    (
     input logic x,
     output logic y
     );

    always_comb begin
        if (`IS_ZERO(x)) begin
            y <= x;
        end
    end

endmodule
```

Include any options or configuration used.

**Actual output**

```
✗ verible-verilog-format macro/mem/behavioral/test.sv 
module test

    (
     input logic x,
     output logic y
     );

    always_comb begin
        if (`IS_ZERO(x)) begin
            y <= x;
        end
    end

endmodule
macro/mem/behavioral/test.sv: Formatted output is lexically different from the input.    Please file a bug.  Details:
Mismatched token enums.  got: (')') (#41: ")") vs. (MacroCallCloseToEndLine) (#712: ")")
First mismatched token [19]: (')') (#41: ")") vs. (MacroCallCloseToEndLine) (#712: ")")
; problematic formatter output is
module test (
    input  logic x,
    output logic y
);

  always_comb begin
    if (
        `IS_ZERO(x)
        ) begin
      y <= x;
    end
  end

endmodule
<<EOF>>
```

**Expected or suggested output**

The macro should probably not have newlines surrounding it, and verification should succeed because (as far as I can tell) the output is equivalent to the input.

nathanbeckmann:
Observations
 * Removing the backtick from the input fixes the problem.
 * Moving the macro out of the condition and into the assignment fixes the problem. That is, ```  always_comb begin
    if (x) begin
      y <= `IS_ZERO(x);
    end
  end```

nathanbeckmann:
I am seeing a possibly related failure on macros in declarations:

```
✗ verible-verilog-format macro/mem/behavioral/test.sv
module test

    (
     input logic x,
     output logic y
     );

    logic [`WIDTH(x):0] z;

    always_comb begin
        if (x) begin
            y <= x;
        end
    end

endmodule
macro/mem/behavioral/test.sv: Formatted output is lexically different from the input.    Please file a bug.  Details:
Mismatched token enums.  got: (')') (#41: ")") vs. (MacroCallCloseToEndLine) (#712: ")")
First mismatched token [17]: (')') (#41: ")") vs. (MacroCallCloseToEndLine) (#712: ")")
; problematic formatter output is
module test (
    input  logic x,
    output logic y
);

  logic [
  `WIDTH(x)
  :0] z;

  always_comb begin
    if (x) begin
[...]

PR #1885, opened 2023-04-24
Added macro call in struct test without a semicolon.

jbylicki:
As I was asked in https://github.com/chipsalliance/verible/pull/1880, I have added the test without a semicolon.

However, this test required a change to how formatted code is being lexically compared to the original: Macro Calls are whitespace sensitive when there are no semicolons. The lexer might find a generic macro call with a "normal" closing bracket, or `MacroCallCloseToEndLine` when it is on a newline. As they are syntactically equivalent, I have added support for it in Yacc and changed the comparator to look for that similarity in this particular use case.

### DIFF
--- a/verilog/analysis/verilog_equivalence.cc
+++ b/verilog/analysis/verilog_equivalence.cc
@@ -167,7 +167,11 @@ DiffStatus LexicallyEquivalent(
   DiffStatus diff_status = DiffStatus::kEquivalent;
   auto recursive_comparator = [&](const TokenSequence::const_iterator l,
                                   const TokenSequence::const_iterator r) {
-    if (l->token_enum() != r->token_enum()) {
+    if (l->token_enum() != r->token_enum() &&
+        !((l->token_enum() == verilog_tokentype::MacroCallCloseToEndLine &&
+           r->text() == ")") ||
+          (r->token_enum() == verilog_tokentype::MacroCallCloseToEndLine &&
+           l->text() == ")"))) {
       if (errstream != nullptr) {
         *errstream << "Mismatched token enums.  got: ";
         token_printer(*l, *errstream);
@@ -246,6 +250,14 @@ DiffStatus FormatEquivalent(absl::string_view left, absl::string_view right,
         return IsWhitespace(verilog_tokentype(t.token_enum()));
       },
       [=](const TokenInfo& l, const TokenInfo& r) {
+        // MacroCallCloseToEndLine should be considered equivalent to ')', as
+        // they are whitespace dependant
+        if (((r.token_enum() == verilog_tokentype::MacroCallCloseToEndLine) &&
+             (l.text() == ")")) ||
+            ((l.token_enum() == verilog_tokentype::MacroCallCloseToEndLine) &&
+             (r.text() == ")"))) {
+          return true;
+        }
         return l.EquivalentWithoutLocation(r);
       },
       errstream);
--- a/verilog/parser/verilog.y
+++ b/verilog/parser/verilog.y
@@ -3685,11 +3685,13 @@ struct_union_member
     { $$ = MakeTaggedNode(N::kStructUnionMember, $1, $2, $3, $4, $5, $6); }
   | preprocessor_directive
     { $$ = std::move($1); }
-  | MacroCall
-    { $$ = std::move($1);}
   // since the semicolon is optional, it is better to have both cases covered
   | MacroCall ';'
     { $$ = ExtendNode($1, $2);}
+  | MacroCall
+    { $$ = std::move($1);}
+  | MacroCallId '(' macro_args_opt MacroCallCloseToEndLine
+    { $$ = MakeTaggedNode(N::kMacroCall, $1, MakeParenGroup($2, $3, $4)); }
   ;
 
 case_item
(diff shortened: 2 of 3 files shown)


hzeller:
Almost - testing  #1524  with this patch now results in a non-converging situation.  
```
module a (
  input wire [`clog2(2):0] b
);
endmodule
/tmp/a.sv: Re-formatted text does not match formatted text; formatting failed to converge!  Please file a bug.
module a (
  input wire [`clog2(2):0] b
);
endmodule
============== Formatted: ==============
module a (
    input wire [
    `clog2(2)
    :0] b
);
endmodule
============= Re-formatted: ============
module a (
    input wire [
    `clog2(2)
    : 0] b
);
endmodule
 module a (
     input wire [
     `clog2(2)
-    :0] b
+    : 0] b
 );
 endmodule
module a (
    input wire [
    `clog2(2)
    :0] b
);
endmodule
```

jbylicki:
There was a problem where after a `MacroCallCloseToEndLine` there was a line wrap, next the `:` was checked for spacing and then all was applied later. On the second pass the code noticed spaces used for indentation, and adjusting afterwards to 1 space as well (as many spaces are condensed to be one if many are present). It is now fixed and #1524 does not seem to produce the problem. 

Still, the decision of `)` vs `MacroCallCloseToEndLine` is whitespace dependent, so I am not sure whether to close #1524. I have made those tokens equivalent in the comparator but still either one may appear in places. I think this is not an issue, but it may produce edge cases in the future. What do you think @hzeller?

verilog/formatting/token_annotator.cc
```diff
@@ -436,7 +437,13 @@ static WithReason<int> SpacesRequiredBetween(
     if (InRangeLikeContext(right_context)) {
       int spaces = right.OriginalLeadingSpaces().length();
       if (spaces > 1) {
-        spaces = 1;
+        // If ExcessSpaces returns 0 if there was a newline - prevents
+        // counting indentation as spaces
+        if (!right.ExcessSpaces()) {
```
hzeller, line 442: I'd write this as ternary operator for improved readability:
```c++
  spaces = right.ExcessSpaces() ? 1 : 0;
```

verilog/parser/verilog.y
```diff
@@ -3685,11 +3685,13 @@ struct_union_member
     { $$ = MakeTaggedNode(N::kStructUnionMember, $1, $2, $3, $4, $5, $6); }
   | preprocessor_directive
     { $$ = std::move($1); }
-  | MacroCall
-    { $$ = std::move($1);}
   // since the semicolon is optional, it is better to have both cases covered
   | MacroCall ';'
     { $$ = ExtendNode($1, $2);}
+  | MacroCall
```
hzeller, line 3691: can these recent changes here be covered by tests in `verilog_parser_unittest.cc` ?

verilog/formatting/token_annotator.cc
```diff
@@ -436,7 +437,13 @@ static WithReason<int> SpacesRequiredBetween(
     if (InRangeLikeContext(right_context)) {
       int spaces = right.OriginalLeadingSpaces().length();
       if (spaces > 1) {
-        spaces = 1;
+        // If ExcessSpaces returns 0 if there was a newline - prevents
```
hzeller, line 440: If we have changes here, we should also have a test in `token_annotator_test.cc` that covers both situations.

verilog/analysis/verilog_equivalence.cc
```diff
@@ -167,7 +167,11 @@ DiffStatus LexicallyEquivalent(
   DiffStatus diff_status = DiffStatus::kEquivalent;
   auto recursive_comparator = [&](const TokenSequence::const_iterator l,
                                   const TokenSequence::const_iterator r) {
-    if (l->token_enum() != r->token_enum()) {
+    if (l->token_enum() != r->token_enum() &&
```
hzeller, line 170: There should be a test covering the code changes. In particular formulate the tests in a way that if the condition is not given here, it would fail, so that we truly make sure that it has the expected effect.

hzeller (commented):
Looks pretty good! Some tests are missing, so we need them before merge.
And one suggestion for improved readability (use ternary operator instead of bulky if/else).

W.r.t. the `MacroCallCloseToEndLine` I think this is good for now, but maybe we should open an issue for further consideration of the equivalence; e.g. is there a good reason why these are two different tokens anyway ?

In general, this addresses now indeed #1524 and also #1883. In both cases, the resulting formatting is not very pretty with a newline before and after the macro. But that should be addressed in a separate pull request that focuses on that before we can close these issues.

#1524 now formats as:
```systemverilog
module a (
    input wire [
    `clog2(2)
    :0] b
);
endmodule

```

#1883 now formats as (except):
```systemverilog
    if (
        `IS_ZERO(x)
        ) begin
      y <= x;
    end
```

verilog/formatting/token_annotator.cc
jbylicki: There are already tests that look at the case of having no newlines. I have added a test for newline-containing spacing and it indeed passes, but only on the introduced changes, before it would fail

hzeller (approved):
Thanks!

PR #1886, opened 2023-04-24
Language Server: Added update for README.md

glatosinski:
This PR:

* Adds information about configuring `verible-verilog-ls` in Kakoune editor
* Adds information about configuring linting in Language Server
* Adds information about `verible.filelist`


verilog/tools/ls/README.md
```diff
@@ -236,3 +271,55 @@ code --install-extension verible.vsix
 ```
 
 [release]: https://github.com/chipsalliance/verible/releases
+
+## Configuring the Language Server for a project
```
hzeller, line 275: I think it would be good to have these above the editor configuration, as it is probably the information that people most come back to: they only set up their editor once, but they might need to re-read what the name of the file list is etc.

verilog/tools/ls/README.md
```diff
@@ -236,3 +271,55 @@ code --install-extension verible.vsix
...
+
+### Adding `verible.filelist` for project-wide symbol discovery
+
+`verible-verilog-ls` by default loads and analyses only currently edited files in the editor.
+To be able to utilize such features as going to definition or references project-wide, add a `verible.filelist` file to the project.
+`verible.filelist` is a file containing a list of design files to collect data necessary to build a symbol table and other helper structures to allow Language Server to provide definitions, references, information on hover and more.
+It is a simple file containing one file per line - it can be easily created using such tools as `find`, e.g.:
```
hzeller, line 282: Most people will already have a file list in some shape as this is what their other tooling is using.

So here I'd first describe that this is just a standard file list and if the tooling already has a file list, simply adding a symbolic link with the name `verible.filelist` pointing to the original file list should be sufficient.

Then, if that is not available, having the `find` method is a good fallback way of creating a file list.

verilog/tools/ls/README.md
```diff
@@ -236,3 +271,55 @@ code --install-extension verible.vsix
...
+
+```bash
+find . -name "*.sv" -o -name "*.svh" -o -name "*.v" | sort > verible.filelist
+```
+
+With it, the Language Server creates a symbol table for the entire project, allowing to find definitions and references for a symbol accross multiple files.
```
hzeller, line 288: allowing to find definitions is now mentioned the third time in this paragraph, so this feels redundant.

verilog/tools/ls/README.md
```diff
@@ -236,3 +271,55 @@ code --install-extension verible.vsix
...
+
```
hzeller, line 289: Important: where is the file located ? We need to describe what we look for and what that means. Toplevel can come from the editor, if it provides a root.uri (and so here you need to find what this is commonly called in these editors). Or we fall back to the current working directory the editor has been opened in (but: what does that mean for Windows, where people just click on an icon somewhere and the root directory is not so clearly defined ?).

So here, we need to explain where it is and potentially figuring out with various edtiors what that means in their context.

Maybe we need a separate paragraph about the root dir situation, as then we can refer to it in the context of the file list and the linter configuration.

verilog/tools/ls/README.md
```diff
@@ -236,3 +271,55 @@ code --install-extension verible.vsix
...
+### Adding project-specific linter configuration
+
+By default, Language Server publishes all possible linting issues for a given document.
+It is possible to disable certain linter warnings for a given project using the `.rules.verible_lint` file.
+It is a simple file that consists of comma-separated or newline-separated settings for rules, e.g.:
```
hzeller, line 294: Add a link to the corresponding README file in the linter.

hzeller (commented):
Nice!
I think we also need some context on where we find the file list and linter config file.

verilog/tools/ls/README.md
glatosinski: Ok, I suggested linking , and also provided a `--file_list_path` flag as a possibility to change the default name of the list file

verilog/tools/ls/README.md
glatosinski: I added some description regarding the workspace directory and where it can be located

verilog/tools/ls/README.md
glatosinski: I initially provided it in the end of the section, but now it is moved up where the example is provided

hzeller (approved):
Nice!

PR #1888, opened 2023-04-24
Add test for PR 1884

IEncinas10:
PR 1884 solved a crash when the user asked the VerilogLanguageServer to format a buffer with syntax errors. This test checks we don't go back on this incorrect behaviour

The test doesn't assume anything for the response, it just checks the Language Server doesn't crash


hzeller:
thanks!

PR #1895, opened 2023-04-26
VSCode extension: Download necessary binary

eszpotanski:
This PR adds downloading `verible-verilog-ls` (from GitHub release with tag matching plugin's version) to VSCode extension. 

Binary is downloaded only when language server's executable (from configs `verible.path`) is not available.

### DIFF
--- a/verilog/tools/ls/vscode/src/download-ls.ts
+++ b/verilog/tools/ls/vscode/src/download-ls.ts
@@ -1,0 +1,108 @@
+import { platform, arch } from "os";
+import * as fs from "fs";
+import * as path from "path";
+import { IncomingMessage } from "http";
+import { https as httpsFr } from "follow-redirects";
+import { execSync } from "child_process";
+import * as decompress from "decompress";
+const decompressTargz = require("decompress-targz");
+const TAG = require("../package.json").repository.tag;
+
+
+function checkIfBinaryExists(binaryPath: string) {
+  let whichCommand: string, binaryExists: boolean;
+  if (platform() == "win32") whichCommand = "where";
+  else whichCommand = "which";
+
+  binaryExists = true;
+  try {
+    execSync(`${whichCommand} ${binaryPath}`, { windowsHide: true });
+  } catch {
+    binaryExists = false;
+  }
+
+  return binaryExists;
+}
+
+export async function checkAndDownloadBinaries(
+  binaryPath: string
+): Promise<string> {
+  if (platform() === "darwin") {
+    // There is no static binaries for MacOS -- aborting
+    return binaryPath;
+  }
+  if (checkIfBinaryExists(binaryPath)) {
+    // Language server binary exists -- nothing to do
+    return binaryPath;
+  }
+
+  const pluginDir = path.join(__dirname, "..");
+  const binDir = path.join(pluginDir, "bin");
+  const pluginBinaryPath = path.join(
+    binDir,
+    "verible-verilog-ls" + (platform() === "win32" ? ".exe" : "")
+  );
+  if (checkIfBinaryExists(pluginBinaryPath)) {
+    // Language server binary already downloaded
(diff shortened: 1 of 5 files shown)


hzeller:
Nice!
Since I don't know how any of the json interacts with vs-code, I am inclined to just merge and and see, then iterate if needed.

I think important would be for the user to see if they are using the binary downloaded by the plug-in or if they use the system-installed one. I suspect it is visible in the extension settings looking at the verible path ?

verilog/tools/ls/vscode/src/download-ls.ts
```diff
@@ -0,0 +1,108 @@
...
+import { execSync } from "child_process";
+import * as decompress from "decompress";
+const decompressTargz = require("decompress-targz");
+const TAG = require("../package.json").repository.tag;
+
+
+function checkIfBinaryExists(binaryPath: string) {
+  let whichCommand: string, binaryExists: boolean;
+  if (platform() == "win32") whichCommand = "where";
+  else whichCommand = "which";
```
hzeller, line 15: I think these days it is recommended to use `command -v`
e.g. debian warns about `which` being deprecated, on other systems you have to explicitly install `which`; while `command -v` should work with all typical bourne-shells.

verilog/tools/ls/vscode/src/download-ls.ts
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  return binaryExists;
+}
+
+export async function checkAndDownloadBinaries(
+  binaryPath: string
+): Promise<string> {
+  if (platform() === "darwin") {
+    // There is no static binaries for MacOS -- aborting
```
hzeller, line 31: Maybe file an issue that we don't forget to supply one if possible.
There is also https://github.com/chipsalliance/homebrew-verible maybe we can make that work with regular updates.

verilog/tools/ls/vscode/src/download-ls.ts
```diff
@@ -0,0 +1,108 @@
...
+
+  const pluginDir = path.join(__dirname, "..");
+  const binDir = path.join(pluginDir, "bin");
+  const pluginBinaryPath = path.join(
+    binDir,
+    "verible-verilog-ls" + (platform() === "win32" ? ".exe" : "")
+  );
+  if (checkIfBinaryExists(pluginBinaryPath)) {
+    // Language server binary already downloaded
+    return pluginBinaryPath;
```
hzeller, line 47: Is the path dependend on the version of the plug-in ? Otherwise, if we have downloaded it a while ago, it will be stale.
If not, maybe also checking if the binary is older than, say, a month and if so regard it too old.

hzeller (commented):
Just some smallish comments, which we can look at asynchronously.

verilog/tools/ls/vscode/src/download-ls.ts
eszpotanski: Yes, this path is depended on the version, because downloaded executable (with tag matching the plug-in's version) is saved in the plug-in's directory and VSCode creates separate folder for each versions.

Moreover, updating extension will also update executable -- assuming `verible.path` from settings doesn't point to installed `verible-verilog-ls`

verilog/tools/ls/vscode/src/download-ls.ts
eszpotanski: I added information that Verible can be installed with `brew`.

I also created issue #1897

verilog/tools/ls/vscode/src/download-ls.ts
eszpotanski: VSCode claims to remove old or uninstalled extensions right after restart and when I tested it, old folders was removed a few minutes after restarting app. Thus, I didn't implement any cleanup.

hzeller (approved):
Thanks!

PR #1903, opened 2023-04-29
Add autofix for always_comb rule

IEncinas10:
Note: clang-format introduced many changes in the files. 

First attempt on adding an AutoFix for the always_comb rule. It looks somewhat decent but there are some things I'm not completely satisfied with.

After inspecting the syntax tree with the cli tool
```
Node @0 (tag: kAlwaysStatement) {
        Leaf @0 (#"always" @266-272: "always")
        Node @1 (tag: kProceduralTimingControlStatement) {
          Node @0 (tag: kEventControl) {
             "..."
          }
```

My approach has been to replace the *StringSpanOfSymbol* for kEvenControl for the empty string, as the syntax tree for an 'always_comb' doesnt present this element. With this, I also noticed that when performing this transformation we're changing the syntax tree, so I don't know whether this is ok or could cause problems elsewhere.

Other issues:

- The end result contains an 'extra' space. I don't know if it is possible to remove that extra space.
```diff
-always @(*) begin
+always_comb  begin
```
- Auxilary functions: For clarity, I added two functions to get to the kProceduralTimingControlStatement and the kEventControl nodes inside an AlwaysStatement. I'm not sure if the functions could be generalized to extract them in every possible appearence or if we keep them AlwaysStatement-specific and rename them (although GetProceduralTimingControlStatementInAlwaysStatement seems a bit long :laughing:  )

Edit: kAlwaysStatement nodes might not even present a kProceduralTimingStatement/kEventControl, so I don't know what to do with this auxiliary functions.

- TODO: 
  - [ ] Add tests for the auxiliary functions after agreeing on how they should look like

### DIFF
--- a/verilog/CST/statement.cc
+++ b/verilog/CST/statement.cc
@@ -1,4 +1,4 @@
-// Copyright 2017-2020 The Verible Authors.
+// Copyright 2017-2023 The Verible Authors.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


hzeller:
Nice!
Don't worry about the clang-format changes, they are intended from a recent change in the config file.

Using the StringSpanOfSymbol is good, and in this case you probably want to extend that:
The superfluous space, you can probably erase by taking the whole string-view range between begin of `always` and end of `@*` instead of replacing always and the event separately. Since everything is in a contiguous string-view, you can replace from the `begin()` of the always to the `end()` of the event control.

IEncinas10:
Right! Missed the very nice
```c++
absl::string_view StringSpanOfSymbol(const Symbol& lsym, const Symbol& rsym);
```

Looks cleaner now. One edit, no extra spaces as before. I fixed the tests and removed extra comments. Let me know if the naming of the new helper functions in statement.cc is acceptable. The only thing missing is adding tests for those functions.

hzeller (approved):
Thanks!

PR #1906, opened 2023-05-03
Unify naming of files: adapt to the scheme of the bazel rule.

hzeller:
The bazel rules are generally with dashes, and so are many files. Adapt the files that are still using underscore to match their corresponding bazel rules.

This is _not_ changing files that are currently being modified in active pull requests. Remaining files will be addressed later.


hzeller:
(works, but will do at a later time)

PR #1908, opened 2023-05-04
Add autofix for always_comb_blocking rule

IEncinas10:
Not a great deal as the manual fix involves removing one character, but might make the user experience better.


hzeller (approved):
Even if it is only one character - if we can provide some automatic  edit help for the user, it is still valuable.

PR #1912, opened 2023-05-08
Provide autofixes for always_ff_non_blocking_rule

IEncinas10:
This PR adds autofixes for every case currently flagged in always_ff_non_blocking_rule. I added a separate formatting commit just in case it made review easier.

Quick summary:

- Added getters for kAssignModifyStatements
- Added getters for kIncrementDecrementExpressions
- Added a getter for kNetVariableAssignments
  - [ ] I'm not sure if this should be placed in statement{.h,.cc,_test.cc}

As an example, applying autofixes would transform 
```verilog
  c = b - a;
  c++;
  ++c;
  c &= (k + 1);
end
```

into this

```verilog
  c <= b - a;
  c <= c + 1;
  c <= c + 1;
  c <= c & (k + 1);
end
```


dbesse:
Does it add parentheses if needed for priority of operation?
```systemverilog
c *= k + 1;
```
into
```systemverilog
c <= c * (k + 1);
```
whereas without it would mean something logically different
```systemverilog
c <= c * k + 1; // (c*k)+1
```

IEncinas10:
Missed that, thanks for the catch! I guess wrapping the whole right hand side expression into a parenthesis would be always correct, right? 

The fix is very simple. I'll add a check and avoid adding two parenthesis levels if the rhs is already between parenthesis.

To clarify:

```verilog
a *= b + 1;
a *= (b + 1);
```

would both become

```verilog
a <= a * (b + 1);
a <= a * (b + 1);
```

dbesse:
Actually, even just changing the blocking to non-blocking (or reversely) may change the logic. I fully understand that it would report as a lint message, but this
```systemverilog
    x++;
    y <= x;
  end
```
would do something different than the fixed version
```systemverilog
    x <= x + 1;
    y <= x;
  end
```
Depending on the policy for autofix, we might actually need the user to provide the correct fix

```systemverilog
    x <= x + 1;
    y <= x + 1;
  end
```

IEncinas10:
Yeah, I see could be a problem. My initial thought was to just provide the quick fix and rely on the user to fix the logic changes.

Otherwise, if we modify an blocking assignment ( x=..., ++a, a++, x &=...) we would need to search the whole always_ff block and substitute every reference to x for the appropiate update, depending on if it is before the blocking assignment or after. I don't know how hard that could become.

Are blocking assignments even well defined inside always_ff blocks? If they're not, I guess we could provide the naive autofix and suggest the user to check the logic?

verilog/analysis/checkers/always_ff_non_blocking_rule_test.cc
```diff
@@ -120,13 +121,63 @@ TEST(AlwaysFFNonBlockingRule, LocalWaiving) {
...
+      // Check that we're correctly waiving the 'p = 0' as it is a local
+      // variable
+      {"module m;\nalways_ff begin reg p; p = 0; k = 1; end\nendmodule",
+       "module m;\nalways_ff begin reg p; p = 0; k <= 1; end\nendmodule"},
+  };
+
+  RunApplyFixCases<VerilogAnalyzer, AlwaysFFNonBlockingRule>(
+      kTestCases, "waive_for_locals:on");
+}
+
```
hzeller, line 180: Plus tests for `a++` becoming `a <= a + 1` ?

verilog/analysis/checkers/always_ff_non_blocking_rule.cc
```diff
@@ -97,31 +100,71 @@ void AlwaysFFNonBlockingRule::HandleSymbol(const verible::Symbol& symbol,
...
+
+      autofixes.emplace_back(
+          AutoFix("Substitute assignment operator for equivalent "
+                  "nonblocking assignment",
+                  {original, fix}));
     } else if (asgn_incdec_matcher.Matches(symbol, &symbol_man)) {
-      check_root = &symbol;
+      const verible::SyntaxTreeNode *operand =
+          GetIncrementDecrementOperand(symbol);
+      const verible::SyntaxTreeLeaf *operator_ =
```
hzeller, line 148: why underscore in the `operator_` name ?

verilog/analysis/checkers/always_ff_non_blocking_rule.cc
```diff
@@ -97,31 +100,71 @@ void AlwaysFFNonBlockingRule::HandleSymbol(const verible::Symbol& symbol,
...
+      const absl::string_view identifier =
+          verible::StringSpanOfSymbol(*operand);
+      // Extract just the operation. Just '+' from '++'
+      const absl::string_view op =
+          verible::StringSpanOfSymbol(*operator_).substr(0, 1);
+
+      // Equivalent nonblocking assignment
+      // {'x++', '++x'} become 'x <= x + 1'
+      // {'x--', '--x'} become 'x <= x - 1'
+      std::string fix =
```
hzeller, line 161: Can be const: `const std::string fix`

hzeller:
I think we should leave it up to the user to recognize that they have some more work to do when they apply a fix mentioned in https://github.com/chipsalliance/verible/pull/1912#issuecomment-1540321474 - so I think it is fine if we don't try to over-analyze the whole situation and fix everything.

The only thing we could do maybe later is to recognize that `x` (in this case) is used later in the same block as RHS and thus we might not offer the auto-fix at all as it would be wrong, just reporting the lint error.

hzeller:
Adding @snsokolov for verilog expertise: do you think the above auto-fix suggestions would be reasonable and correct ?

snsokolov:
> Adding @snsokolov for verilog expertise: do you think the above auto-fix suggestions would be reasonable and correct ?

As it was noted above, changing blocking to non-blocking could be harmless, but could also change the user's logic in many cases. I don't have a good intuition about what percentage of lint violations the proposed autofix would break the code. I feel that it could be relatively high. When there is an unary operator or same variable is assigned and used - the chances of breaking the code will be very high.

So answering your question: The suggestion to auto-fix these violations is reasonable in most simple cases, though in some percentage of cases the auto-fixed code will stop behaving the same way.

IEncinas10:
Thanks everyone for the reviews!

I'll work on the parenthesis thing, looked at it and shouldn't be too difficult.

Regarding the autofix problems, I agree that the safest way to proceed is to detect if the change (blocking -> nonblocking) would change code behaviour by checking if the affected variable is read somewhere else inside the always block, and in that case don't suggest the autofix. Not 100% sure how easy this will be, but I guess it won't be too much work.

```systemverilog
  x++;
  y <= x;
end
```

I think doing the correct autofix in every cornercase could be worth it although I will probably try it in a separate PR.

verilog/analysis/checkers/always_ff_non_blocking_rule.cc
IEncinas10: ```operator``` is a reserved keyword, so I couldn't use it.

verilog/analysis/checkers/always_ff_non_blocking_rule.cc
IEncinas10: Right, thanks!

IEncinas10:
Had time to clean things up, but I encountered a problem with AutoFix tests. Right now, it doesn't allow for test cases
with several linting violations. But in order to test the 'safety' of autofixes, I need it.

The easiest workaround I see is to default to the first violation with autofixes, let me know what you think. If you think it is acceptable, let me know whether to submit a separate PR or include it in this one. @hzeller 

Basically, the 'fix' would be:

```diff
-  CHECK_EQ(violations.size(), 1) << "TODO: apply multi-violation fixes";
-  CHECK_GT(violations.begin()->autofixes.size(), test.fix_alternative);
-  const verible::AutoFix& fix =
-      rule_status.violations.begin()->autofixes[test.fix_alternative];
+  // Default to first violation with available autofix
+  const LintViolation &violation =
+      *std::find_if(begin(violations), end(violations),
+                    [](const LintViolation &v) { return v.autofixes.size(); });
+
+  CHECK_GT(violation.autofixes.size(), test.fix_alternative);
+  const verible::AutoFix &fix = violation.autofixes[test.fix_alternative];
```

The need comes from this test:

```c++
  const std::initializer_list<verible::AutoFixInOut> kTestCases = {
      // We can't fix 'k &= 1', because it would affect
      // 'p <= k'
      {"module m;\nalways_ff begin\n"
       "k &= 1;\n"
       "p <= k;\n"
       "a++;"
       "end\nendmodule",
       "module m;\nalways_ff begin\n"
       "k &= 1;\n"
       "p <= k;\n"
       "a <= a + 1;"
       "end\nendmodule"},
};
```

Where I basically check that the first violation doesn't have an autofix available. It could be done changing the test but it is easier this way.

verilog/analysis/checkers/always_ff_non_blocking_rule_test.cc
IEncinas10: Added, thanks!

IEncinas10:
Sorry for the mess, let me know if I should open another PR to ease review or not mess with the commits

hzeller:
Let me run clang-format separately, then if you re-base the changes should not look so big.
Does that sound good ?

IEncinas10:
Would it also work if I just force push after cleaning up the mess locally? I rebased needlessly and somehow my commits appear replicated.

Thanks!

hzeller:
to rebase, you'd first

```
git fetch upstream
git checkout master
git merge upstream/master
git push
```

... so that your fork is up-to-date with the master.

Then, on your command line you can go into your branch
```
git checkout autofix_always_ff_non_blocking
git rebase master
```

Then, if there are any conflicts, resolve them locally. The git CLI tool guides you what to do.

Then, you can push your branch and then only your changes should show up.
```
git push
```

Don't worry about the changes introduced by the formatter, we can ignore them by adding `?w=1` to the review URL.
https://github.com/chipsalliance/verible/pull/1912/files?w=1

Maybe, the rebase attempt breaks even more though. In that case it is probably easier to create a copy of the files that you did change and create an entirely new pull request with that.

hzeller:
Just looking through old PRs, I guess this also fell along the wayside.
It looked ready to merge after some rebase situation, but there was this one open question on a test. Had that been resolved ?

IEncinas10:
Good old times where I didn't know my git 😄 

There were several problems:

1. There were changes required for both this PR and the `dff-name-style` one (helpers for IncrementDecrementExpression: `var++`, `++var`)
2. Some issues with test helpers as you point out which I hacked away but wasn't sure if it was a good change
3. Potentially offering this autofix when it would change code behaviour

(1) will hopefully soon go away. (2) Isn't that bad but perhaps there is a better solution (I'll think about it). (3) is probably not perfect but I don't think it can be, I don't know verilog that well but probably non-automatic functions calls can modify local variables inside `always_ff` blocks which we can't possibly catch to avoid offering the autofix (or now that I think about it perhaps we can just get really scared and avoid offering the autofix if we see macros/function calls)

IEncinas10:
I took a new (hopefully better) approach to handle the issues this PR with current autofix test utils. 

This one was a bit tricky, so there might be things to improve. Naming could also improve, as there are a couple of places where `references_` and `references` exist. I'll take care of this along with the potential requests

PR #1917, opened 2023-05-12
Fix macro formatting divergence and splitting issues

### LINKED ISSUE #1894: Mismatched token enums (macro with argument)
**Test case**

```systemverilog
module m1;
    task automatic t1 ();
        t2(`R1(1) + 8);
    endtask : t1
endmodule
```
Options:
```
verible-verilog-format --inplace=true --compact_indexing_and_selections=false
```

**Actual output**

```systemverilog
./example.sv: Formatted output is lexically different from the input.    Please file a bug.  Details:
Mismatched token enums.  got: (')') (#41: ")") vs. (MacroCallCloseToEndLine) (#712: ")")
First mismatched token [14]: (')') (#41: ")") vs. (MacroCallCloseToEndLine) (#712: ")")

module m1;
  task automatic t1();
    t2(
    `R1(1)
    + 8);
  endtask : t1
endmodule
<<EOF>>
```

```
verible-verilog-format --version
v0.0-3222-gb19cdf44
Commit  2023-04-25
Built   2023-04-26T09:02:38Z
```

### LINKED ISSUE #1924: Ugly output for type declarations with macro + argument
**Test case**

A type declaration like the following:

```
    logic [WIDTH-1:0] data[
    `NONNEGATIVE(COUNT-1)
    :0];
```

gets formatted with newlines. (Until recently, this crashed -- thanks for fixing that!) I would prefer:

```
    logic [WIDTH-1:0] data[`NONNEGATIVE(COUNT-1):0];
```

Is there an option to control this that I have missed?

jbylicki:
This PR removes a unnecessary check when dealing with kReferenceCallBase where the r_paren token should be added regardless of partition policy (as the style dictates that the bracket is always connected to the last token)

Fixes #1894 
Fixes #1924

### DIFF
--- a/verilog/formatting/tree_unwrapper.cc
+++ b/verilog/formatting/tree_unwrapper.cc
@@ -849,6 +849,7 @@ void TreeUnwrapper::SetIndentationsAndCreatePartitions(
     case NodeEnum::kCasePatternItem:
     case NodeEnum::kGenerateCaseItem:
     case NodeEnum::kGateInstance:
+    case NodeEnum::kRegisterVariable:
     case NodeEnum::kGenerateIfClause:
     case NodeEnum::kGenerateElseClause:
     case NodeEnum::kGenerateIfHeader:
@@ -863,17 +864,6 @@ void TreeUnwrapper::SetIndentationsAndCreatePartitions(
       VisitIndentedSection(node, 0, PartitionPolicyEnum::kFitOnLineElseExpand);
       break;
     }
-    case NodeEnum::kRegisterVariable: {
-      if (node.children().size() > 1) {
-        VisitIndentedSection(node, 0,
-                             PartitionPolicyEnum::kAppendFittingSubPartitions);
-      } else {
-        VisitIndentedSection(node, 0,
-                             PartitionPolicyEnum::kFitOnLineElseExpand);
-      }
-      break;
-    }
-
       // The following cases will always expand into their constituent
       // partitions:
     case NodeEnum::kModuleDeclaration:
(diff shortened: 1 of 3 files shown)


hzeller:
Nice! Can you add a unit-test that fails before and passes after ?

jbylicki:
Renamed the PR since the added case for dimensions having formatting issues with macros fitted nicely into the same problem as the original cause in this PR. That being said, I have converted it to draft since during the addition of the unit test there still was a convergence problem. Dimensions are formatted correctly after adding one case to its `is_standalone` function

jbylicki:
Fixes #1924

jbylicki:
Original issues have been fixed, however long ranges still produce weird outputs. I have found one divergence in case:
```
module foo();
function bar();
  logic [12:34] data [`AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA(1) : `BBBBBBBBBBBBBBBBBBBBBBBBBBBBBB(2) + 3];
endfunction
endmodule

```

The output:
```
module foo();
function bar();
  logic [12:34] data [`AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA(1) : `BBBBBBBBBBBBBBBBBBBBBBBBBBBBBB(2) + 3];
endfunction
endmodule
============== Formatted: ==============
module foo ();
  function bar();
    logic [12:34] data[
    `AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA(1)
    :
    `BBBBBBBBBBBBBBBBBBBBBBBBBBBBBB(2)
+ 3];
  endfunction
endmodule
============= Re-formatted: ============
module foo ();
  function bar();
    logic [12:34] data[`AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA(1):`BBBBBBBBBBBBBBBBBBBBBBBBBBBBBB(2)
+ 3];
  endfunction
endmodule
 module foo ();
   function bar();
-    logic [12:34] data[
-    `AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA(1)
-    :
-    `BBBBBBBBBBBBBBBBBBBBBBBBBBBBBB(2)
+    logic [12:34] data[`AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA(1):`BBBBBBBBBBBBBBBBBBBBBBBBBBBBBB(2)
 + 3];
   endfunction
 endmodule
module foo ();
  function bar();
    logic [12:34] data[
    `AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA(1)
    :
    `BBBBBBBBBBBBBBBBBBBBBBBBBBBBBB(2)
+ 3];
  endfunction
endmodule
```

verilog/formatting/tree_unwrapper.cc
```diff
@@ -864,6 +863,16 @@ void TreeUnwrapper::SetIndentationsAndCreatePartitions(
...
       break;
     }
+    case NodeEnum::kRegisterVariable: {
+      if (node.children().size() > 1) {
+        VisitIndentedSection(node, 0,
+                             PartitionPolicyEnum::kAppendFittingSubPartitions);
+      } else {
+        VisitIndentedSection(node, 0,
+                             PartitionPolicyEnum::kFitOnLineElseExpand);
+      }
```
kbieganski, line 873:
```suggestion
      VisitIndentedSection(node, 0, node.children().size() > 1 ?
                                    PartitionPolicyEnum::kAppendFittingSubPartitions : 
                                    PartitionPolicyEnum::kFitOnLineElseExpand);
```
I think this could use a comment why this decision is being made.

verilog/formatting/tree_unwrapper.cc
```diff
@@ -3123,6 +3131,36 @@ void TreeUnwrapper::ReshapeTokenPartitions(
       break;
     }
 
+    case NodeEnum::kRegisterVariable:
+    case NodeEnum::kInstantiationType: {
+      size_t partition_size = partition.Children().size();
+      if (partition_size >= 4) {
```
kbieganski, line 3137:
```suggestion
      if (partition_size < 4) break;
```
Early returns (or breaks) help the reader focus on the important part of the code. They also help avoid rightward drift.
More importantly, why is this `4` here? Could use an explanation. I know that you later access children `0..3`, but what does this mean in relation to actual Verilog code?

verilog/formatting/tree_unwrapper.cc
```diff
@@ -3123,6 +3131,36 @@ void TreeUnwrapper::ReshapeTokenPartitions(
...
+                              static_cast<int>(NodeEnum::kUnpackedDimensions)) {
+            break;
+          }
+        } else {
+          subnode = verible::GetSubtreeAsNode(*subnode, subnode->Tag().tag, 3);
+          if (!subnode || subnode->Tag().tag !=
+                              static_cast<int>(NodeEnum::kPackedDimensions)) {
+            break;
+          }
+        }
```
kbieganski, line 3151: I think this could be a bit more readable if we extracted either of the dimension nodes first, and only after that checked if `subnode` is `kUnpackedDimensions` or `kPackedDimensions`. That gives one fewer `if` and less nesting.

Also I think the Verible idiom is to do `NodeEnum(subnode->Tag().tag) != NodeEnum::kUnpackedDimensions` instead of the `static_cast`. I know this is not always the case, but personally I find that easier to parse.

verilog/formatting/tree_unwrapper.cc
```diff
@@ -3123,6 +3131,36 @@ void TreeUnwrapper::ReshapeTokenPartitions(
...
+        auto& colon = partition.Children()[partition_size - 3];
+        if (colon.Value().TokensRange().begin()->Text() != ":") break;
+        AttachSeparatorToPreviousOrNextPartition(&colon);
+        verible::MergeLeafIntoPreviousLeaf(
+            &partition.Children()[partition_size - 2]);
+        verible::FlattenOneChild(partition, partition_size - 3);
+        verible::MergeLeafIntoPreviousLeaf(
+            &partition.Children()[partition_size - 2]);
```
kbieganski, line 3159: I trust that this does the job but it's not obvious to me why we're doing this. I mean I kind of get it from the PR description but personally I prefer most of the necessary info directly in the code.

kbieganski (changes requested):

verilog/formatting/tree_unwrapper.cc
jbylicki: It actually got left over from a previous attempt at a solution for the inconsistent line splitting, I have restored it to the previous state.

kbieganski (approved):

PR #1922, opened 2023-05-16
[LSP] Initial support for textDocument/hover request

### LINKED ISSUE #1187: Language Server: implement textDocument/hover
The Hover request should return some information that can be useful in the context

Ideas:

 * Variables: type if available (maybe requires symbol table, but could probably be more simply derived; register/wire/parameter/localparam; if possible to determine without going too much in elaboration-weeds: width. Input, output ?)
 * Blocks: when hovering over the `end` block - if this is a labelled block, name of the label
 * generalized to `endmodule`, `endfunction`... name of the corresponding thing.

Some simple things can probably be relatively easily be derived, others might need symbol table.

glatosinski:
Resolves #1187

This PR adds initial support for `textDocument/hover` request.

It adds a HoverBuilder that constructs hover messages for the current position.

Tasks:

- [X] Adds hover feature to LS capabilities
- [X] Prints metatype for symbol
- [X] Prints type for symbol
- [ ] Prints documentation for symbol
- [X] Provides label for `end`
- [ ] Provides label for `endmodule`, `endfunction` of block (currently WIP)
- [ ] Tests for the `textDocument/hover` method (currently WIP)

### DIFF
--- a/verilog/tools/ls/verilog-language-server_test.cc
+++ b/verilog/tools/ls/verilog-language-server_test.cc
@@ -1490,6 +1490,48 @@ endmodule
       absl::StrContains(response["result"]["contents"]["value"], "reg [31:0]"));
 }
 
+// Checks if the hover appears on "end" token when block name is available
+TEST_F(VerilogLanguageServerSymbolTableTest, HoverOverEnd) {
+  absl::string_view filelist_content = "mod.v\n";
+  static constexpr absl::string_view  //
+      module_content(
+          R"(module mod(
+    input clk,
+    input reg [31:0] a,
+    input reg [31:0] b,
+    output reg [31:0] sum);
+  always @(posedge clk) begin : addition
+    assign sum = a + b;
+  end
+endmodule
+)");
+
+  const verible::file::testing::ScopedTestFile filelist(
+      root_dir, filelist_content, "verible.filelist");
+  const verible::file::testing::ScopedTestFile module(root_dir, module_content,
+                                                      "mod.v");
+
+  const std::string module_open_request =
+      DidOpenRequest("file://" + module.filename(), module_content);
+  ASSERT_OK(SendRequest(module_open_request));
+
+  GetResponse();
+
+  std::string hover_request =
+      HoverRequest("file://" + module.filename(), 2, 7, 3);
+
+  ASSERT_OK(SendRequest(hover_request));
+  json response = json::parse(GetResponse());
+
+  ASSERT_EQ(response["id"], 2);
+  ASSERT_EQ(response["result"].size(), 1);
+  ASSERT_EQ(response["result"]["contents"]["kind"], "markdown");
+  ASSERT_TRUE(absl::StrContains(response["result"]["contents"]["value"],
+                                "End of block"));
+  ASSERT_TRUE(absl::StrContains(response["result"]["contents"]["value"],
+                                "Name: addition"));
+}
+
 // Tests correctness of Language Server shutdown request
 TEST_F(VerilogLanguageServerTest, ShutdownTest) {
   const absl::string_view shutdown_request =
--- a/verilog/tools/ls/verilog-language-server_test.cc
+++ b/verilog/tools/ls/verilog-language-server_test.cc
@@ -612,6 +612,13 @@ std::string ReferencesRequest(absl::string_view file, int id, int line,
                                           line, character);
 }
 
+// Creates a textDocument/hover request
+std::string HoverRequest(absl::string_view file, int id, int line,
+                         int character) {
+  return TextDocumentPositionBasedRequest("textDocument/hover", file, id, line,
+                                          character);
+}
+
 // Performs simple textDocument/definition request with no VerilogProject set
 TEST_F(VerilogLanguageServerSymbolTableTest, DefinitionRequestNoProjectTest) {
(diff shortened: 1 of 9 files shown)


verilog/tools/ls/verilog-language-server_test.cc
```diff
@@ -593,6 +593,13 @@ std::string ReferencesRequest(absl::string_view file, int id, int line,
                                           line, character);
 }
 
+// Creates a textDocument/hover request
+std::string HoverRequest(absl::string_view file, int id, int line,
```
hzeller, line 597: is the `id` actually a relevant value we need to pass for the tests ? If not maybe we just replace it with some `static constexpr int kDummyId = 1;` to not distract from the other values.

verilog/tools/ls/verilog-language-server_test.cc
```diff
@@ -1423,6 +1430,51 @@ TEST_F(VerilogLanguageServerSymbolTableTest, CheckReferenceUnknownSymbol) {
...
+                                                      "mod.v");
+
+  const std::string module_open_request =
+      DidOpenRequest("file://" + module.filename(), module_content);
+  ASSERT_OK(SendRequest(module_open_request));
+
+  // obtain diagnostics for both files
+  GetResponse();
+
+  // find definition for "var1" variable of a module in b.sv file
```
hzeller, line 1460: the _find definition..._ comment sounds like a copy from some other location :)

verilog/tools/ls/verilog-language-server_test.cc
```diff
@@ -1423,6 +1430,51 @@ TEST_F(VerilogLanguageServerSymbolTableTest, CheckReferenceUnknownSymbol) {
...
+      HoverRequest("file://" + module.filename(), 2, 6, 12);
+
+  ASSERT_OK(SendRequest(hover_request));
+  json response = json::parse(GetResponse());
+
+  // For now when the file is invalid we will not be able to obtain symbols
+  // from it if it was incorrect from the start
+  ASSERT_EQ(response["id"], 2);
+  ASSERT_EQ(response["result"].size(), 1);
+  ASSERT_EQ(response["result"]["contents"]["kind"], "markdown");
```
hzeller, line 1471: could you cast  `response["result"]` to `verible::lsp::Hover`, so that we have a nice struct to inspect ? The sequence of nested `["..."]` accesses makes it harder to read.

verilog/tools/ls/verilog-language-server_test.cc
```diff
@@ -1423,6 +1430,51 @@ TEST_F(VerilogLanguageServerSymbolTableTest, CheckReferenceUnknownSymbol) {
...
+  const std::string module_open_request =
+      DidOpenRequest("file://" + module.filename(), module_content);
+  ASSERT_OK(SendRequest(module_open_request));
+
+  // obtain diagnostics for both files
+  GetResponse();
+
+  // find definition for "var1" variable of a module in b.sv file
+  std::string hover_request =
+      HoverRequest("file://" + module.filename(), 2, 6, 12);
```
hzeller, line 1462: Maybe adding some comments that indicate what we're looking at: `..., /* line */ 6, /* col */ 12`

verilog/tools/ls/verilog-language-server_test.cc
```diff
@@ -1423,6 +1430,51 @@ TEST_F(VerilogLanguageServerSymbolTableTest, CheckReferenceUnknownSymbol) {
...
+  absl::string_view filelist_content = "mod.v\n";
+  static constexpr absl::string_view  //
+      module_content(
+          R"(module mod(
+    input clk,
+    input reg [31:0] a,
+    input reg [31:0] b,
+    output reg [31:0] sum);
+  always @(posedge clk) begin : addition
+    assign sum = a + b;
```
hzeller, line 1443: Maybe add a comment here that this is roughly here where we're hovering for the reader of the code matching the line with the content (as it is zero-indexed, it might be surprising at first)

```systemverilog
...
   assign sum = a + b;  // hovering over this sum
...
```

verilog/tools/ls/verilog-language-server_test.cc
```diff
@@ -1423,6 +1430,51 @@ TEST_F(VerilogLanguageServerSymbolTableTest, CheckReferenceUnknownSymbol) {
...
+  GetResponse();
+
+  // find definition for "var1" variable of a module in b.sv file
+  std::string hover_request =
+      HoverRequest("file://" + module.filename(), 2, 6, 12);
+
+  ASSERT_OK(SendRequest(hover_request));
+  json response = json::parse(GetResponse());
+
+  // For now when the file is invalid we will not be able to obtain symbols
```
hzeller, line 1467: Is this comment relevant here ?

hzeller (commented):
Nice! A few comments so far.

verilog/tools/ls/verilog-language-server_test.cc
glatosinski: It should be unique per request - we can go with the static value that is incremented per request

hzeller: Yep, but is any place even checking if it is unique per request ?

hzeller:
Closing this one, as it was rebased and merged in #2087

PR #1927, opened 2023-05-19
Changed how system calls are handled when in conditional begin-end blocks

### LINKED ISSUE #1914: crashes on $finish and $fatal
if(!failed) begin
      $fwrite(fd, "Test passed!\n");
      $finish;
    end else begin
      $fatal(1, "Test failed!");
    end
  end

jbylicki:
This PR adds a case for system calls in conditional begin-end blocks, as the current options provide a non-conditional stack layout and don't split the calls (they exist within the same UWLine) when formatting them.

Fixes #1914


hzeller (approved):
LGTM!

PR #1930, opened 2023-05-22
ci: fix gh-pages deployment condition

umarcor:
The documentation branch (`gh-pages`) was last updated in December 2022: https://github.com/chipsalliance/verible/tree/gh-pages. That's probably because the deployment step in the CI job does depend on `matrix.mode == 'compile'` but `matrix` is not defined in the job (https://github.com/chipsalliance/verible/blob/master/.github/workflows/verible-ci.yml#L575-L599), so the condition can never be met.

This PR removes the `matrix.mode` condition.

### DIFF
--- a/.github/workflows/verible-ci.yml
+++ b/.github/workflows/verible-ci.yml
@@ -590,7 +590,7 @@ jobs:
         run: ./.github/bin/github-pages-setup.sh $VERIBLE_BINDIR
 
       - name: Deployment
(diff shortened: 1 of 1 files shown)


umarcor:
@AdamOlech, I think this was an oversight in https://github.com/chipsalliance/verible/commit/cb7ac7dca176e727337f912257362d0c7798c91c.

PR #1936, opened 2023-05-30
Remove unnecessary CHECK when adding duplicate virtual files

ivan444:


verilog/analysis/verilog_project.cc
```diff
@@ -338,7 +338,10 @@ void VerilogProject::AddVirtualFile(absl::string_view resolved_filename,
       std::make_unique<InMemoryVerilogSourceFile>(
           resolved_filename, std::make_shared<verible::StringMemBlock>(content),
           /*corpus=*/""));
-  CHECK(inserted.second);
+  if (!inserted.second) {
+    LOG(WARNING) << "The virtual file " << resolved_filename
+                 << " is already registered with the project!";
```
hzeller, line 343: Would probably be interesting to know if the content is also the same
```
const bool same_content = inserted.first->second->GetContent() == content;
LOG ..."already registered with the project (with " << (same_content ? "same" : "different") << " content)";
```

ivan444: Done

hzeller (approved):

PR #1937, opened 2023-05-31
Update RE2.

hzeller:
This now uses absl directly in the main branch, no need to reference a particular branch.


hzeller:
CC @junyer

junyer:
Thanks!

PR #1941, opened 2023-06-06
Add proper handling of escape characters in LSP URIs

glatosinski:
This PR adds methods for encoding and decoding URIs and filesystem paths in the LSP (so that URIs follow RFC 3986 URI specification).


common/lsp/lsp-file-utils_test.cc
```diff
@@ -0,0 +1,83 @@
...
+
+#include "gtest/gtest.h"
+
+TEST(URIDecodingTest, SimplePathDecoding) {
+  std::string exemplar = "/home/user/test-project/add.sv";
+  std::string result =
+      verible::lsp::LSPUriToPath("file:///home/user/test-project/add.sv");
+
+  ASSERT_EQ(exemplar, result);
+}
```
hzeller, line 26: These tests would probably more compact and faster to read if you do the test in-line: 
```
EXPECT_EQ(LSPUriToPath("file:///home/user/test-project/add.sv"),
          "/home/user/test-project/add.sv");
```

common/lsp/lsp-file-utils_test.cc
```diff
@@ -0,0 +1,83 @@
...
+
+TEST(URIEncodingTest, PathWithConsecutiveEscapes) {
+  std::string exemplar =
+      "file:///home/user/test-project/my%20%20%20%20module.sv";
+  std::string result =
+      verible::lsp::PathToLSPUri("/home/user/test-project/my    module.sv");
+
+  ASSERT_EQ(exemplar, result);
+}
```
hzeller, line 83: Also add tests with broken escape sequences `/foo/bar/%xyz` ... they should come out as-is.

common/lsp/lsp-file-utils_test.cc
```diff
@@ -0,0 +1,83 @@
...
+
+TEST(URIDecodingTest, SimplePathDecoding) {
+  std::string exemplar = "/home/user/test-project/add.sv";
+  std::string result =
+      verible::lsp::LSPUriToPath("file:///home/user/test-project/add.sv");
+
+  ASSERT_EQ(exemplar, result);
+}
+
+TEST(URIDecodingTest, WindowsPathDecoding) {
```
hzeller, line 28: Instead of first having all the decoding tests, then the encoding tests, I'd probably group the tests so that encoding and decoding are close together; that makes it easier to look at them in one go.

WindowsPathDecoding
WindowsPathEncoding

PathWithSpaceDecoding
PathWithSpacesEncoding
...

hzeller (commented):
Looks good in general; a few comments.

hzeller:
It does look like the Windows installation or at least the interaction with bazel is hosed on the CI, that probably has to be fixed first.

hzeller:
#1945 is merged, so after a rebase, this might compile.

common/lsp/lsp-file-utils_test.cc
```diff
@@ -0,0 +1,86 @@
...
+
+#include "absl/strings/str_cat.h"
+#include "gtest/gtest.h"
+
+namespace verible {
+namespace lsp {
+
+std::string PathPrefix(const std::string &path) {
+#ifdef _WIN32
+  return absl::StrCat("c:/", path);
```
hzeller, line 26: Is this something that we can only test on Windows, because the `std::filesystem::path` behaves differently there ?

I suggest to use a different drive letter so that we don't accidentally rely on existence of the drive letter. So instead of `c:/` maybe `y:/`.

hzeller (commented):
Looks like we now also need to modify the tests in language server test that use the `ScopedTestFile`.

Previously they were just assuming that the output paths are `"file://" + module_b.filename()`, but now we will have to look if they are `PathToLSPUri(module_b.filename())`.

common/lsp/lsp-file-utils_test.cc
glatosinski: Fixed it

glatosinski:
The tests are updated now

hzeller (approved):
LGTM

PR #1948, opened 2023-06-15
[LSP] Added Symbol Renaming to the language server

jbylicki:
This PR implements functionality for `textDocument/prepareRename` and `textDocument/rename` capabilities.

### DIFF
--- a/common/lsp/lsp-protocol.yaml
+++ b/common/lsp/lsp-protocol.yaml
@@ -203,13 +203,13 @@ DocumentLink:
   range: Range
   target?: string  # DocumentUri
 
-# -- textDocument/prepareRename   (TODO(jbylicki): check that requires project + active symbol table #1189)
+# -- textDocument/prepareRename
 PrepareRenameParams:
   <: TextDocumentPositionParams
 
-# Response: Range[]
+# Response: Range
 
-# -- textDocument/rename   (TODO(jbylicki): check that requires project + active symbol table #1189)
+# -- textDocument/rename
 RenameParams:
   <: TextDocumentPositionParams
   newName: string
--- a/common/lsp/lsp-protocol.yaml
+++ b/common/lsp/lsp-protocol.yaml
@@ -209,3 +209,10 @@ PrepareRenameParams:
 
 # Response: Range[]
 
+# -- textDocument/rename   (TODO(jbylicki): check that requires project + active symbol table #1189)
+RenameParams:
+  <: TextDocumentPositionParams
+  newName: string
+
+# Response: Range[]
+
--- a/common/lsp/lsp-protocol.yaml
+++ b/common/lsp/lsp-protocol.yaml
@@ -202,3 +202,10 @@ DocumentLinkParams:
 DocumentLink:
   range: Range
   target?: string  # DocumentUri
+
+# -- textDocument/prepareRename   (TODO(jbylicki): check that requires project + active symbol table #1189)
+PrepareRenameParams:
+  <: TextDocumentPositionParams
+
+# Response: Range[]
+
--- a/verilog/tools/ls/symbol-table-handler.cc
+++ b/verilog/tools/ls/symbol-table-handler.cc
@@ -376,7 +376,7 @@ SymbolTableHandler::FindRenameableRangeAtCursor(
   }
   std::optional<verible::TokenInfo> symbol =
       GetTokenInfoAtTextDocumentPosition(params, parsed_buffers);
-  if (symbol.has_value()) {
+  if (symbol) {
     verible::TokenInfo token = symbol.value();
     const SymbolTableNode &root = symbol_table_->Root();
     const SymbolTableNode *node =
@@ -411,10 +411,10 @@ SymbolTableHandler::FindRenameLocationsAndCreateEdits(
   if (locations.empty()) return {};
   std::map<absl::string_view, std::vector<verible::lsp::TextEdit>>
       file_edit_pairs;
-  for (auto &loc : locations) {
+  for (const auto &loc : locations) {
     file_edit_pairs[loc.uri].reserve(locations.size());
   }
-  for (auto &loc : locations) {
+  for (const auto &loc : locations) {
     // TODO(jbylicki): Remove this band-aid fix once #1678 is merged - it should
     // fix
     //  duplicate definition/references appending in modules and remove the need
--- a/verilog/tools/ls/symbol-table-handler.cc
+++ b/verilog/tools/ls/symbol-table-handler.cc
@@ -248,22 +248,23 @@ absl::string_view SymbolTableHandler::GetTokenAtTextDocumentPosition(
 
 verible::LineColumnRange
 SymbolTableHandler::GetTokenRangeAtTextDocumentPosition(
-    const verible::lsp::TextDocumentPositionParams &params,
+    const verible::lsp::TextDocumentPositionParams &document_cursor,
     const verilog::BufferTrackerContainer &parsed_buffers) {
   const verilog::BufferTracker *tracker =
-      parsed_buffers.FindBufferTrackerOrNull(params.textDocument.uri);
+      parsed_buffers.FindBufferTrackerOrNull(document_cursor.textDocument.uri);
   if (!tracker) {
-    VLOG(1) << "Could not find buffer with URI " << params.textDocument.uri;
+    VLOG(1) << "Could not find buffer with URI "
+            << document_cursor.textDocument.uri;
     return {};
   }
   std::shared_ptr<const ParsedBuffer> parsedbuffer = tracker->current();
   if (!parsedbuffer) {
     VLOG(1) << "Buffer not found among opened buffers:  "
-            << params.textDocument.uri;
+            << document_cursor.textDocument.uri;
     return {};
   }
-  const verible::LineColumn cursor{params.position.line,
-                                   params.position.character};
+  const verible::LineColumn cursor{document_cursor.position.line,
+                                   document_cursor.position.character};
   const verible::TextStructureView &text = parsedbuffer->parser().Data();
 
   const verible::TokenInfo cursor_token = text.FindTokenAt(cursor);
@@ -342,18 +343,16 @@ std::vector<verible::lsp::Location> SymbolTableHandler::FindReferencesLocations(
   return locations;
 }
 
-verible::lsp::Range SymbolTableHandler::FindRenameLocations(
+verible::lsp::Range SymbolTableHandler::FindRenameableRangeAtCursor(
     const verible::lsp::PrepareRenameParams &params,
     const verilog::BufferTrackerContainer &parsed_buffers) {
   Prepare();
+  if (files_dirty_) {
+    BuildProjectSymbolTable();
+  }
   verible::LineColumnRange symbol =
       GetTokenRangeAtTextDocumentPosition(params, parsed_buffers);
-  verible::lsp::Range range;
-  range.start.line = symbol.start.line;
-  range.start.character = symbol.start.column;
-  range.end.line = symbol.end.line;
-  range.end.character = symbol.end.column;
-  return range;
+  return RangeFromLineColumn(symbol);
 }
 
 verible::lsp::WorkspaceEdit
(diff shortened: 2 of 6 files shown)


verilog/tools/ls/symbol-table-handler.cc
```diff
@@ -317,6 +342,59 @@ std::vector<verible::lsp::Location> SymbolTableHandler::FindReferencesLocations(
   return locations;
 }
 
+verible::lsp::Range SymbolTableHandler::FindRenameLocations(
```
hzeller, line 345: What comes back is not plural, but returns a single range. And it is optional.
Maybe `RenameableRangeAtCursor()` or something like that ?

common/lsp/lsp-protocol.yaml
```diff
@@ -202,3 +202,17 @@ DocumentLinkParams:
 DocumentLink:
   range: Range
   target?: string  # DocumentUri
+
+# -- textDocument/prepareRename   (TODO(jbylicki): check that requires project + active symbol table #1189)
+PrepareRenameParams:
+  <: TextDocumentPositionParams
+
+# Response: Range[]
```
hzeller, line 210: Isn't the response just a `Range`, not `Range[]`  ( https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_prepareRename ) ?

verilog/tools/ls/symbol-table-handler.cc
```diff
@@ -317,6 +342,59 @@ std::vector<verible::lsp::Location> SymbolTableHandler::FindReferencesLocations(
...
+    const verilog::BufferTrackerContainer &parsed_buffers) {
+  Prepare();
+  verible::LineColumnRange symbol =
+      GetTokenRangeAtTextDocumentPosition(params, parsed_buffers);
+  verible::lsp::Range range;
+  range.start.line = symbol.start.line;
+  range.start.character = symbol.start.column;
+  range.end.line = symbol.end.line;
+  range.end.character = symbol.end.column;
+  return range;
```
hzeller, line 356: What if we are over a token where a rename would not work ? The protocol definition says that in that case `null` is to be returned.
I guess that means we should return here a `std::optional<verible::lsp::Range>`  - essentially figuring out if the cursor is even hovering over something that can be rennamed (in the first simple implementation we could just look that we're not over a whitespace or keyword token, later we can look if we are actually over a symbol known to the symbol table).

And then in the `textDocument/prepareRename` request handler something that checks if the return `has_value()` and if not a nullptr.

verilog/tools/ls/verilog-language-server.cc
```diff
@@ -138,6 +139,16 @@ void VerilogLanguageServer::SetRequestHandlers() {
         return symbol_table_handler_.FindReferencesLocations(p,
                                                              parsed_buffers_);
       });
+  dispatcher_.AddRequestHandler(
+      "textDocument/prepareRename",
+      [this](const verible::lsp::PrepareRenameParams &p) {
```
hzeller, line 144: If we do the optional `lsp::Range` or `null` return here as discussed above, we might need to help out the compiler to deduce the return type by adding `-> nlohmann::json`.

verilog/tools/ls/verilog-language-server_test.cc
```diff
@@ -1442,6 +1442,138 @@ TEST_F(VerilogLanguageServerSymbolTableTest, CheckReferenceUnknownSymbol) {
...
+               "position",
+               {{"line", params.line}, {"character", params.character}},
+           },
+           {"newName", params.newName},
+       }}};
+  return renamerq.dump();
+}
+
+// Runs tests for textDocument/rangeFormatting requests
+TEST_F(VerilogLanguageServerSymbolTableTest, PrepareRenameTest) {
```
hzeller, line 1495: Some of this testing should probably go into `symbol-table-handler_test`

hzeller, line 1495: The naming should reflect what we're testing, e.g. `PrepareRenameReturnsRangeOfEditableSymbol` or something along those lines. Not all tests are named like this yet, but this is helpful when seeing all the test outputs from the `bazel test //...`

verilog/tools/ls/verilog-language-server_test.cc
```diff
@@ -1442,6 +1442,138 @@ TEST_F(VerilogLanguageServerSymbolTableTest, CheckReferenceUnknownSymbol) {
...
+  // Create sample file and make sure diagnostics do not have errors
+  RenameRequestParams params;
```
hzeller, line 1497: Given that we're here in a test that figures out prepare rename, having the parameter called `RenameRequestParams` is of course confusing; also a good reason to switch to the auto-generated protocol structs (`verible::lsp::PrepareRenameParams`).

verilog/tools/ls/verilog-language-server_test.cc
```diff
@@ -1442,6 +1442,138 @@ TEST_F(VerilogLanguageServerSymbolTableTest, CheckReferenceUnknownSymbol) {
...
+  EXPECT_EQ(response["result"]["start"]["line"], 2)
+      << "Invalid result for id:  ";
+  EXPECT_EQ(response["result"]["start"]["character"], 0)
+      << "Invalid result for id:  ";
+  EXPECT_EQ(response["result"]["end"]["line"], 2) << "Invalid result for id:  ";
+  EXPECT_EQ(response["result"]["end"]["character"], 3)
+      << "Invalid result for id:  ";
+}
+
+TEST_F(VerilogLanguageServerSymbolTableTest, RenameTest) {
```
hzeller, line 1530: Multiple filenames will actually be interesting. So maybe have a test `RenameTestSymbolSingleFile` and `RenameTestSymbolMultipleFiles`, in which a symbol is defined in one file and used in another file - and the rename properly touches all files.

verilog/tools/ls/verilog-language-server_test.cc
```diff
@@ -1442,6 +1442,138 @@ TEST_F(VerilogLanguageServerSymbolTableTest, CheckReferenceUnknownSymbol) {
...
+  ASSERT_OK(SendRequest(request));
+
+  const json response = json::parse(GetResponse());
+  std::cout << response << std::endl;
+  EXPECT_EQ(response["result"]["changes"].size(), 1)
+      << "Invalid result size for id:  ";
+  EXPECT_EQ(response["result"]["changes"][params.file].size(), 3)
+      << "Invalid result size for id:  ";
+}
+
```
hzeller, line 1576: Additional possible tests: put the cursor over the definition of a symbol in one test and over one reference of a symbol. In both cases, the definition and all references should be updated.

Also other corner cases
  * symbol defined in a package ? Making sure that the same symbolname ('foo') defined in two different packages is not cross-renaming the other package symbol `PackageA::foo` and `PackageB::foo`

verilog/tools/ls/verilog-language-server_test.cc
```diff
@@ -1442,6 +1442,138 @@ TEST_F(VerilogLanguageServerSymbolTableTest, CheckReferenceUnknownSymbol) {
...
+      << "textDocument/publishDiagnostics not received";
+  EXPECT_EQ(diagnostics["params"]["uri"], params.file)
+      << "Diagnostics for invalid file";
+
+  EXPECT_EQ(diagnostics["params"]["diagnostics"].size(), 0)
+      << "The test file has errors";
+  std::string request = PrepareRenameRequest(params);
+  ASSERT_OK(SendRequest(request));
+
+  const json response = json::parse(GetResponse());
```
hzeller, line 1520: With the new potential 'null' return of this request, we should also test that.

verilog/tools/ls/symbol-table-handler.h
```diff
@@ -65,11 +65,18 @@ class SymbolTableHandler {
       const verible::lsp::ReferenceParams &params,
       const verilog::BufferTrackerContainer &parsed_buffers);
 
+  verible::lsp::Range FindRenameLocations(
```
hzeller, line 68: These new fucntions need a test in symbol-table-handler_test.cc. These functions are implicitly tested later in the verilog-language-server_test.cc, but in the symbol-table-handler_test.cc, we should check all the potential corner cases while the verilog-language-sever_test.cc is more for the final end2end testing.

An indication about the tests you can get by running the coverage by just running the specific `symbol-table-handler_test.cc` to see how much it covers from `symbol-table.cc`:

```
MODE=coverage .github/bin/build-and-test.sh verilog/tools/ls:symbol-table-handler_test
# Then open browser and check out the coverage of symbol-table-handler.cc
```

... Our goal is typically to be > 90% for these library specific tests.

common/lsp/lsp-protocol.yaml
jbylicki: Yes it is! I have at one point misread the spec and implemented it for `Range[]`. It was fixed within the code itself, but I forgot it here. Changed it now

verilog/tools/ls/symbol-table-handler.cc
jbylicki: As with the YAML spec, it is a leftover from when it did return a `Range[]`. Changed it

verilog/tools/ls/symbol-table-handler.cc
```diff
@@ -317,6 +343,71 @@ std::vector<verible::lsp::Location> SymbolTableHandler::FindReferencesLocations(
...
+    const verilog::BufferTrackerContainer &parsed_buffers) {
+  Prepare();
+  if (files_dirty_) {
+    BuildProjectSymbolTable();
+  }
+  verible::LineColumnRange symbol =
+      GetTokenRangeAtTextDocumentPosition(params, parsed_buffers);
+  verible::lsp::Range range;
+  range = RangeFromLineColumn(symbol);
+  return range;
```
hzeller, line 357: And now, these three lines can be reduced to simply
```c++
return RangeFromLineColumn(symbol);
```

verilog/tools/ls/verilog-language-server_test.cc
```diff
@@ -1442,6 +1442,138 @@ TEST_F(VerilogLanguageServerSymbolTableTest, CheckReferenceUnknownSymbol) {
...
 
+struct RenameRequestParams {
+  int id;
+  std::string file;
+  std::string newName;
+  int line;
+  int character;
+};
+
+std::string RenameRequest(RenameRequestParams params) {
```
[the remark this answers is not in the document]
jbylicki: I have changed the methods such that they place the `(Prepare)RenameParams` into a prepared snippet of JSON. Using `to_json()` didn't make sense here, as it dumps the contents of the struct only, however the request itself also requres an ID, method and other boilerplate. Methods built into the tests already use this approach (such as `ReferencesRequest()` and I have found no logical way to construct a full request in any other manner.

verilog/tools/ls/verilog-language-server_test.cc
jbylicki: The first test: this is known to be broken and expected to be fixed by #1678 (as well as the duplication of positions with module names); I will add the test and remove duplicate finding once that PR gets merged, if that is okay.

As for the second one: added it and it works as expected.

verilog/tools/ls/verilog-language-server_test.cc
jbylicki: Both are now renamed/added

verilog/tools/ls/symbol-table-handler.h
jbylicki: The new methods in this PR are now covered in `symbol-table-handler_test`

verilog/tools/ls/symbol-table-handler.cc
```diff
@@ -317,6 +366,79 @@ std::vector<verible::lsp::Location> SymbolTableHandler::FindReferencesLocations(
...
+SymbolTableHandler::FindRenameableRangeAtCursor(
+    const verible::lsp::PrepareRenameParams &params,
+    const verilog::BufferTrackerContainer &parsed_buffers) {
+  Prepare();
+  if (files_dirty_) {
+    BuildProjectSymbolTable();
+  }
+  std::optional<verible::TokenInfo> symbol =
+      GetTokenInfoAtTextDocumentPosition(params, parsed_buffers);
+  if (symbol.has_value()) {
```
mglb, line 379:
```suggestion
  if (symbol) {
```
for consistency, as this is how you use std::optional below (`if (!location)`)

verilog/tools/ls/symbol-table-handler.cc
```diff
@@ -317,6 +366,79 @@ std::vector<verible::lsp::Location> SymbolTableHandler::FindReferencesLocations(
...
+      GetLocationFromSymbolName(*node->Key(), node->Value().file_origin);
+  if (!location) return {};
+  std::vector<verible::lsp::Location> locations;
+  locations.push_back(location.value());
+  std::vector<verible::lsp::TextEdit> textedits;
+  CollectReferences(&root, node, &locations);
+  if (locations.empty()) return {};
+  std::map<absl::string_view, std::vector<verible::lsp::TextEdit>>
+      file_edit_pairs;
+  for (auto &loc : locations) {
```
mglb, line 414:
```suggestion
  for (const auto &loc : locations) {
```

mglb (changes requested):

verilog/tools/ls/symbol-table-handler.h
```diff
@@ -88,6 +95,15 @@ class SymbolTableHandler {
       const verible::lsp::TextDocumentPositionParams &params,
       const verilog::BufferTrackerContainer &parsed_buffers);
 
+  // TODO(jbylicki): Add docstring
```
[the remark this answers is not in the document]
jbylicki: Done for both methods

mglb (approved):

hzeller:
Looks pretty good. If you rebase, the symbol table handler has a little failure, that would probably need to be addressed.

The above coverage tool claims that only 86% of the changes have coverage, are there missed test opportunities ?

hzeller (approved):
Ready to submit after rebase.

hzeller:
Rebased, then merged as #2081

PR #1963, opened 2023-07-13
Run clang-format on all files that are not currently open in PRs.

hzeller:
This only touches files that are not currently open in any pending pull requests (except #1912, which needs to be manually rebased anyway). That way merging these will be easier.

This unifies the pointer/reference alignment between all files.


fangism (approved):
Oh, alright.

Issue #1962, opened 2023-07-13
Build fails on undefined `uint32_t` in protobuf


nathanbeckmann:
Here is what I get after doing a `bazel clean --expunge`:

```
➜  verible git:(master) bazel build -c opt //...
INFO: Analyzed 1898 targets (0 packages loaded, 0 targets configured).
INFO: Found 1898 targets...
ERROR: /home/beckmann/.cache/bazel/_bazel_beckmann/d932d629abbf7de2514b0932259df566/external/com_google_protobuf/src/google/protobuf/compiler/BUILD.bazel:73:11: Compiling src/google/protobuf/compiler/subprocess.cc failed: (Exit 1): gcc failed: error executing command /usr/bin/gcc -U_FORTIFY_SOURCE -fstack-protector -Wall -Wunused-but-set-parameter -Wno-free-nonheap-object -fno-omit-frame-pointer -g0 -O2 '-D_FORTIFY_SOURCE=1' -DNDEBUG -ffunction-sections ... (remaining 63 arguments skipped)

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
In file included from bazel-out/k8-opt-exec-2B5CBBC6/bin/external/com_google_protobuf/src/google/protobuf/compiler/_virtual_includes/command_line_interface/google/protobuf/compiler/subprocess.h:47,
                 from external/com_google_protobuf/src/google/protobuf/compiler/subprocess.cc:33:
bazel-out/k8-opt-exec-2B5CBBC6/bin/external/com_google_protobuf/src/google/protobuf/_virtual_includes/protobuf_lite/google/protobuf/port.h:186:67: error: 'uint32_t' was not declared in this scope
  186 |     : absl::disjunction<std::is_same<T, int32_t>, std::is_same<T, uint32_t>,
      |                                                                   ^~~~~~~~
bazel-out/k8-opt-exec-2B5CBBC6/bin/external/com_google_protobuf/src/google/protobuf/_virtual_includes/protobuf_lite/google/protobuf/port.h:56:1: note: 'uint32_t' is defined in header '<cstdint>'; did you forget to '#include <cstdint>'?
   55 | #include "google/protobuf/port_def.inc"
  +++ |+#include <cstdint>
   56 | 
[...]

PR #1965, opened 2023-07-14
Bump protobuf and absl dependency.

hzeller:
Be on latest version for #1962 to test against.


hzeller:
Looks like it is not working quite yet with an old bazel 4. Will have to find out what is needed to make it happy to work with that version.

PR #1969, opened 2023-07-18
Slightly reduce dependency.

hzeller:
No need to depend on full protobuf, just the io part is what we're interested in.


hzeller:
... saves about 170 processes run by bazel. Small amount, but also worthwhile to be specific.

ivan444 (approved):

PR #1989, opened 2023-08-01
Simplify json rpc header parsing: remove lenient mode.

hzeller:
We had a lenient mode for header parsing coming from the early days of the implementation of the LSP protocol, however, that is not used anymore. Remove it to simplify logic.

### DIFF
--- a/common/lsp/message-stream-splitter.h
+++ b/common/lsp/message-stream-splitter.h
@@ -54,16 +54,12 @@ class MessageStreamSplitter {
   using MessageProcessFun =
       std::function<void(absl::string_view header, absl::string_view body)>;
 
-  // Optional parameters are "initial_read_buffer_size" for the initial
+  // Optional parameter is "initial_read_buffer_size" for the initial
   // internal buffer size (will be realloc'ed when needed).
-  // If "strict_crlf_header_separation" is false, also allows for simple
-  // newline as separation character in the header. Useful for manually
-  // speaking the protocol.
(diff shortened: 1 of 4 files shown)


common/lsp/message-stream-splitter.h
```diff
@@ -54,15 +54,10 @@ class MessageStreamSplitter {
...
-  // If "strict_crlf_header_separation" is false, also allows for simple
-  // newline as separation character in the header. Useful for manually
-  // speaking the protocol.
-  explicit MessageStreamSplitter(size_t initial_read_buffer_size = 4096,
-                                 bool strict_crlf_header_separation = true)
-      : read_buffer_(initial_read_buffer_size),
-        lenient_lf_separation_(!strict_crlf_header_separation) {}
+  explicit MessageStreamSplitter(size_t initial_read_buffer_size = 4096)
+      : read_buffer_(initial_read_buffer_size) {}
   MessageStreamSplitter(const MessageStreamSplitter &) = delete;
```
fangism, line 61: Just noticed that this is marked non-copyable.  
Should this also delete move constructors? or is it intended to be move-able?  assign-able?

fangism (approved):

common/lsp/message-stream-splitter.h
hzeller: Ah good point. Now that I removed the last `const` it is important to pay attention there again. I've now also deleted the assign operator with const reference.
(No need to add clutter to delete the move constructor/assign, they are implicitly delete with the const reference ones).

PR #1997, opened 2023-08-11
Switch Kythe Kzip creator from absl::Hash to SHA256

ivan444:
Switch Kythe Kzip creator from absl::Hash to new SHA256 implementation as some Kythe tools assume SHA256 in the Kzips. The SHA256 implementation is based on https://www.rfc-editor.org/rfc/rfc6234.txt -- extracted only the SHA256 part, cleaned up the code and made it closer to C++ than C.

### DIFF
--- a/common/util/BUILD
+++ b/common/util/BUILD
@@ -658,3 +658,22 @@ cc_test(
         "@com_google_googletest//:gtest_main",
     ],
 )
+
+cc_library(
+    name = "sha256",
+    srcs = ["sha256.cc"],
+    hdrs = ["sha256.h"],
+    deps = [
+        "@com_google_absl//absl/strings",
+    ],
+)
+
+cc_test(
+    name = "sha256_test",
+    srcs = ["sha256_test.cc"],
+    deps = [
+        ":sha256",
+        "@com_google_absl//absl/strings",
+        "@com_google_googletest//:gtest_main",
+    ],
+)
--- a/common/util/sha256.h
+++ b/common/util/sha256.h
@@ -1,0 +1,138 @@
+/***************** See RFC 6234 for details. *******************/
+/*
+   Copyright (c) 2011 IETF Trust and the persons identified as
+   authors of the code.  All rights reserved.
+
+   Redistribution and use in source and binary forms, with or
+   without modification, are permitted provided that the following
+   conditions are met:
+
+   - Redistributions of source code must retain the above
+     copyright notice, this list of conditions and
+     the following disclaimer.
+
+   - Redistributions in binary form must reproduce the above
+     copyright notice, this list of conditions and the following
+     disclaimer in the documentation and/or other materials provided
+     with the distribution.
+
+   - Neither the name of Internet Society, IETF or IETF Trust, nor
+     the names of specific contributors, may be used to endorse or
+     promote products derived from this software without specific
+     prior written permission.
+
+   THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND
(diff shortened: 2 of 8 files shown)


common/util/BUILD
```diff
@@ -658,3 +658,22 @@ cc_test(
...
     ],
 )
+
+cc_library(
+    name = "sha256",
+    srcs = ["sha256.cc"],
+    hdrs = ["sha256.h"],
+    deps = [
+        "@com_google_absl//absl/strings",
+    ],
```
hzeller, line 668: The code looks like some sort of MIT license. Do we need to annotate this library with some different applicable license here in the cc_library ? (the [`default_applicable_license`](https://github.com/chipsalliance/verible/blob/0b8cb4e0b1e95b2431239a859dd5f4a29d346b32/common/util/BUILD#L5) on the package-level points to [apache](https://github.com/chipsalliance/verible/blob/0b8cb4e0b1e95b2431239a859dd5f4a29d346b32/BUILD#L16-L21))

(but I am also fine with deferring this decision a little bit until we talked to the licensing experts and fix it then)

common/util/sha256.h
```diff
@@ -0,0 +1,138 @@
...
+  // NOTE: The first octet of hash is stored in the element with index 0, the
+  // last octet of hash in the element with index 27/31.
+  std::array<uint8_t, kSha256HashSize> BuildAndReset();
+
+  // Adds an array of octets as the next portion of the message.  Returns false
+  // if the accumulated message is too large (>2 Exabytes). Can be called
+  // multiple times to incrementally build the digest.
+  bool AddInput(absl::string_view message);
+
+  bool IsOverflowed() { return overflowed_; }
```
hzeller, line 82: make `const` ?
(maybe also comment when that happens, with 2 Exibytes ?)

Probably for us _2 Exibytes ought to be enough for anybody..._

common/util/sha256.h
```diff
@@ -0,0 +1,138 @@
...
+#include <array>
+#include <cstdint>
+#include <string>
+
+#include "absl/strings/string_view.h"
+
+namespace verible {
+
+// The size of SHA256 hash in bytes.
+constexpr int32_t kSha256HashSize = 32;
```
hzeller, line 62: Using `inline constexpr` in the header makes sure that there is only one copy of this constant in the binary.

common/util/sha256_test.cc
```diff
@@ -0,0 +1,116 @@
...
+}
+
+TEST(Sha256, NonAsciiDigestsAreEqual) {
+  const std::string kOpenSslSha256JaBananaDigest =
+      "787bcc7042939ad9607bc8ca87332e4178716be0f0b890cbf673884d39d8ff79";
+  EXPECT_THAT(Sha256Hex("バナナ"), Eq(kOpenSslSha256JaBananaDigest));
+}
+
+TEST(Sha256, LargeInputDigestsAreEqual) {
+  const std::string kLargeText = R"(
```
hzeller, line 62: I'd make that a `static constexpr absl::string_view` (similar with the other `std::string` constants in this test)

hzeller (commented):
Very nice, a c++-ified sha256 is very useful.
Some minor comments to address and a license attribution question that we can probably defer.

common/util/sha256.h
ivan444: Done

common/util/sha256.h
ivan444: Done

common/util/BUILD
ivan444: I wasn't sure what to do here. Forwarding the original license seemed like a reasonable thing to do.
I can check with the licensing experts.

PR #2000, opened 2023-08-15
linter: do not require 'default' case for 'unique case'

IEncinas10:
Fixes https://github.com/chipsalliance/verible/issues/1278

Currently, case-missing-default rule requires the `default` case for every case statement. This PR makes an exception for `unique case` statements as requested by the issue.

I also changed where the diagnostics point to. Previously they pointed to the leftmost token of the kCaseItemList, but now they point to the `case` keyword.

### DIFF
--- a/verilog/CST/verilog_matchers.h
+++ b/verilog/CST/verilog_matchers.h
@@ -1,4 +1,4 @@
-// Copyright 2017-2020 The Verible Authors.
+// Copyright 2017-2023 The Verible Authors.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


hzeller:
(currently on vacation, slow to respond with less internet bandwidth; will be back in a week or so for review)

IEncinas10:
Thanks for the heads up. Enjoy the holidays!

goekce:
@hzeller This patch is waiting for ~8 months. I would be glad to see this patch in verible.

hzeller:
Ah sorry everyone. This fell off my radar. Will see if I can make some time to review.

IEncinas10:
> Ah sorry everyone. This fell off my radar. Will see if I can make some time to review.

There is no hurry **at all**, thanks for all your work :)

hzeller (approved):
LGTM. Sorry for the deiay, sometimes things are busy ...

PR #2011, opened 2023-09-18
Cover sha256 code with BSD-3 license. Add default_applicable_licenses to common/tools/BUILD package.

ivan444:
Resolve the licensing mismatch after adding sha256 code by adding the BSD-3 license to the LICENSE file (but limit it to the sha256 code). The compliance check tool is passing with this change.
Moving the code to a separate package, with a separate license didn't work -- the top level LICENSE file had to be updated anyway. This is the simplest & easiest to maintain solution.

I addition, add default_applicable_licenses to common/tools/BUILD package to silence the compliance check tool.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,3 +1,4 @@
+Files: *
 
                                  Apache License
                            Version 2.0, January 2004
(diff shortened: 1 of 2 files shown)


hzeller (approved):
thanks!

PR #2012, opened 2023-09-18
Fix missing includes of standard headers.

### LINKED ISSUE #2013: Windows CI currently not working - compiler installation issue ?
Looks like the Windows CI currently has issues compiling, which does not seem to be rooted in the code itself but probably some installation issue as it seems it misses headers.

The `choco` install fetches [llvm 16.0.6](https://community.chocolatey.org/packages/llvm), which is fairly new, maybe it requires some additional dependencies ?

I currently don't have time to look into this, but if there is someone who could, I'll accept pull requests :)

hzeller:
This is adding all the standard headers missing in an attempt to see if this fixes #2013 ... it does not ( #2014 fixed it)
However, can't harm to have this more clean anyway :)

While at it, update the current expectations in the smoke tests.

### DIFF
--- a/.github/bin/smoke-test.sh
+++ b/.github/bin/smoke-test.sh
@@ -132,30 +132,30 @@ ExpectedFailCount[lint:ibex]=14
 ExpectedFailCount[project:ibex]=193
 ExpectedFailCount[preprocessor:ibex]=368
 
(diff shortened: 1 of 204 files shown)


fangism (approved):
Good cleanup

PR #2014, opened 2023-09-18
Fix Windows LLVM to version 15 until bazel 6.4

corco:
Bazel fails with LLVM 16, which is now installed on Windows CI.

Fix version 15 for now.

The issue is [resolved in bazel](https://github.com/bazelbuild/bazel/pull/19430), but not released yet.


hzeller:
Fixes #2013 - thanks @corco 

I'll wait for the Windows build to pass, then merge.

PR #2017, opened 2023-09-23
linter: make constraint-name-style configurable

IEncinas10:
Up until now, constraint-name-style rule required constraint names to end in `_c`. This makes it configurable, so that users can require constraint names to start with `c_` instead.

I'm open to sugestions about the configuration parameter's name, it doesn't convince me too much but I can't think of a better option. We could have `check_suffix` and `check_prefix` but as it doesn't make sense for them to be on at the same time I don't like it too much either.

-----------
Edit: forgot to handle this, leaving it here so I don't forget

- [X] Handle invalid regex provided by user


hzeller:
(Sorry for the long delay, I was pretty busy with other things...)

Maybe we can provide a regular expression, e.g. configuration parameter name `pattern`, and then the user can decide. And then we can set the default to `"([a-z0-9]+_)+c"` or so to enforce snake-case with trailing-`_c`' (we can keep the regex simple and don't have to worry about the beginning starting with a digit as that already has been taken care of by the lexer/parser)

Using regular expressions is probably also a good choice to configure some of the other rules - they are
We use the re2 library successfully in other parts of the project e.g. in `./common/tools/jcxxgen.cc` and `./verilog/tools/ls/autoexpand.cc`.

IEncinas10:
No worries, first things first!

I remember seeing discussion about regular expressions for some other rules in the past but thought the support wasn't there yet. I'll give this a look and definitely change it. If it ends up being as easy at it sounds I'll make an issue to eventually update every rule that could make use of this feature.

hzeller:
At the time we had the disccussions about regular expressions, the choice was `std::regex`, but that is generally a poor choice as it is slow. Since then, `RE2` had been added to the dependencies, and that is pretty performant.

verilog/analysis/checkers/constraint_name_style_rule.cc
```diff
@@ -44,50 +42,64 @@ using verible::matcher::Matcher;
...
+          "specified by a regular expression.",
+      .param = {{"pattern", kSuffix}},
   };
   return d;
 }
 
-static const Matcher& ConstraintMatcher() {
+absl::Status ConstraintNameStyleRule::Configure(
+    absl::string_view configuration) {
+  std::string pattern = kSuffix;
```
IEncinas10, line 59: I think this is redundant but the rationale is the following:

If at some point we add other parameters to the rule, but `pattern` is not provided, the `ParseNameValues` wouldn't set any value to `pattern`, so we would then need to set a default value. 

Avoid future problems by doing it already?

verilog/analysis/checkers/constraint_name_style_rule.cc
```diff
@@ -44,50 +42,64 @@ using verible::matcher::Matcher;
...
     if (!identifier_token) return;
 
     const absl::string_view constraint_name = identifier_token->text();
 
-    if (!verible::IsLowerSnakeCaseWithDigits(constraint_name) ||
-        !absl::EndsWith(constraint_name, "_c")) {
-      violations_.insert(LintViolation(*identifier_token, kMessage, context));
+    if (!RE2::FullMatch(constraint_name, *regex)) {
+      violations_.insert(
+          LintViolation(*identifier_token, FormatReason(), context));
```
IEncinas10, line 93: Didn't find many resources on RE2. Can it somehow help us provide a better linting message than just "follow the regex"? Perhaps provide an autofix?

hzeller:
Looking backwards in time to see if there are active PRs. This looks like it is still relevant though probably some rebase needed due to conflicts.

IEncinas10:
> Looking backwards in time to see if there are active PRs. This looks like it is still relevant though probably some rebase needed due to conflicts.

Updated, thanks for the reminder!

verible/verilog/analysis/checkers/constraint-name-style-rule.h
```diff
@@ -26,22 +30,36 @@
...
 
   static const LintRuleDescriptor &GetDescriptor();
 
+  absl::Status Configure(absl::string_view configuration) final;
   void HandleSymbol(const verible::Symbol &symbol,
                     const verible::SyntaxTreeContext &context) final;
 
   verible::LintRuleStatus Report() const final;
 
+  std::string Pattern() const { return regex->pattern(); }
```
hzeller, line 56: Is this public method (returning a string by value) needed anywhere ?

verible/verilog/analysis/checkers/constraint-name-style-rule.h
```diff
@@ -26,22 +30,36 @@
 namespace verilog {
 namespace analysis {
 
-// ConstraintNameStyleRule check each constraint name follows the correct naming
-// convention.
-// The constraints should be named with lower_snake_case and end with _c.
+// Lower snake case, ends with `_c`
+#define kSuffix "([a-z0-9]+_)+c"
```
hzeller, line 34: Instead of a macro, can this be a `static constexpr absl::string_view kSuffix` in the private section of ConstraintNameStyleRule instead ?

verible/verilog/analysis/checkers/constraint-name-style-rule.h
```diff
@@ -26,22 +30,36 @@
...
+//
+// The default, `kSuffix` checks that the name is written in lower_snake_case
+// and ends with `_c`
 class ConstraintNameStyleRule : public verible::SyntaxTreeLintRule {
  public:
   using rule_type = verible::SyntaxTreeLintRule;
 
   static const LintRuleDescriptor &GetDescriptor();
 
+  absl::Status Configure(absl::string_view configuration) final;
```
hzeller, line 50: `clang-tidy` points out that we now also need to include `absl/strings/string_view.h`

verible/verilog/analysis/checkers/constraint-name-style-rule.h
IEncinas10: It isn't, I used it to diagnose if the user-provided regex failed to parse but it is not needed anymore thanks to the `SetRegex` utility. Missed it, thanks!

verible/verilog/analysis/checkers/constraint-name-style-rule.h
IEncinas10: Indeed, thanks! Before `kPrefix` was here too and that "needed" to be a define to do this stupid trick https://github.com/IEncinas10/verible/blob/de368dbc5d0a46ad14763854ee105f89a65d476a/verible/verilog/analysis/checkers/constraint-name-style-rule_test.cc#L73

verible/verilog/analysis/checkers/constraint-name-style-rule_test.cc
```diff
@@ -27,8 +27,12 @@ namespace analysis {
 namespace {
 
 using verible::LintTestCase;
+using verible::RunConfiguredLintTestCases;
 using verible::RunLintTestCases;
 
+// Lower snake case, starts with `c_`
+#define kPrefix "c+(_[a-z0-9]+)+"
```
hzeller, line 34: I'd probably move this into the `VariousPrefixTests` TEST to be close to where it is needed.
(and since it is only used in one place to assemble the `pattern:` configuration, maybe it doesn't need to
be a macro at all).

hzeller: (other than that, looks great and ready to be merged)

IEncinas10: I agree. I needed this because my original intent was to offer a choice between `_c` suffix and prefix. Thanks!

hzeller (approved):
Nice.
Thanks!

PR #2045, opened 2023-12-13
Pin bazel version to version 6 in MacOS

hzeller:
By default, it comes with bazel 7, but that won't work with our version of protobuf.
We can't update protobuf, as that doesn't work with bazel 4, which we currently have as minimum requirement.


fangism (approved):
When could we bump the minimum Bazel version?

hzeller:
End of support for bazel 4 is end of January next year: https://bazel.build/release#support-matrix
So I'd consider switching to bazel 5 minimum version then.

PR #2061, opened 2023-12-20
Update Emacs config section of language server README.md

gmlarumbe:
This language server can be easily configured for Emacs also through the [`verilog-ext`](https://github.com/gmlarumbe/verilog-ext/) extension package.

It supports Emacs builtin `eglot` client as well as `lsp-mode`.


hzeller (approved):
Nice!

PR #2066, opened 2023-12-21
Enable suspicious-semicolon_rule by default

IEncinas10:
This recently added rule inspired by a [clang-tidy](https://clang.llvm.org/extra/clang-tidy/checks/bugprone/suspicious-semicolon.html) seems a good candidate for being enabled by default. Only one "possible" false positive, and could potentially save hours of staring at a screen (don't ask me how I know)


hzeller:
I'll defer approval here a bit as I first need to look through some of our script-generated code yet that produces this pattern.

hzeller:
(still on it. I remember I looked used it on a bunch of DV files and there were some pattern that they used that triggered it, but didn't have time yet to look in detail)

PR #2067, opened 2023-12-21
linter: configuration: Ignore incorrect rules

IEncinas10:
Context: https://github.com/chipsalliance/verible/issues/1994 

Already explained in the issue but I think this is way more intuitive than current behaviour. You fire up your editor and might or might not notice that the linter is going back to the default rules because your config isn't right.

If we had a fixed set of rules I guess that would be ok but it is a bit annoying when developing new rules and might also be annoying if users borrow other's `.rules.verible_lint` files with different versions of verible

### DIFF
--- a/verilog/analysis/verilog_linter_configuration.cc
+++ b/verilog/analysis/verilog_linter_configuration.cc
@@ -92,6 +92,7 @@ bool RuleBundle::ParseConfiguration(absl::string_view text, char separator,
   // Clear the vector to overwrite any existing value.
   rules.clear();
 
+  bool parsed_correctly = true;
   for (absl::string_view part :
        absl::StrSplit(text, separator, absl::SkipEmpty())) {
     if (separator == '\n') {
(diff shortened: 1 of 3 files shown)


hzeller (approved):
I like that this more graceful approach.

verilog/analysis/verilog_linter_configuration.cc
```diff
@@ -312,15 +316,12 @@ absl::Status LinterConfiguration::AppendFromFile(
...
-        std::cerr << "Warnings in parse configuration: " << error << std::endl;
-      }
-      UseRuleBundle(local_rules_bundle);
-    } else {
-      std::cerr << "Unable to fully parse configuration: " << error
-                << std::endl;
+    local_rules_bundle.ParseConfiguration(*config_or, '\n', &error);
+    // Log warnings and errors
+    if (!error.empty()) {
+      std::cerr << error;
```
hzeller, line 322: Maybe: if we get an error print out that we use the partial configuration.

IEncinas10: Thanks for the suggestion!

PR #2070, opened 2024-01-03
Add some missing includes.

hzeller:


IEncinas10 (approved):
Thanks! How did you notice this? I just quickly took some other rule as a reference, as it built I guessed it was ok. 

I'll be more careful next time, but just wondering if there is any trick other than being really thorough :)

hzeller:
Mostly eye-balling. There is a `misc-include-cleaner` in newer clang-tidy I believe, but that is too noisy at times.

PR #2072, opened 2024-01-04
Complete suspicious_semicolon_rule_test includes.

hzeller:
Found some more opportunities to add needed includes.


IEncinas10 (approved):
Thanks again :) I guess the point is to avoid relying on transitive dependencies so that if refactor happens we don't make anyone miserable, right?

Edit: I should probably review other PRs I have done because I'm sure I missed some headers 😅

PR #2089, opened 2024-01-31
Make installer work with bazel 7

hzeller:
This is work in progress, playing with what needs to be done.

One of the issues to get things to work with bazel7 is that the installer does not work, it creates this error message:

```
verible/BUILD:49:10: in (an implicit dependency) attribute of _gen_installer rule //:_install_gen: in $whitelist_function_transition attribute of _gen_installer rule //:_install_gen: package group '@@bazel_tools//tools/whitelists/function_transition_whitelist:function_transition_whitelist' is misplaced here (they are only allowed in the visibility attribute).
```

I hacked around here by adding a patch that removes that line :)

Things to work on
  * Make the choice if the patch is applied depending on the bazel version either in WORKSPACE or maybe in the installer file itself. I have no experience with Skylark (or, Python for that matter) to know what to do. Hopefully reviewers on this PR can help.
  * This is a hack of course, but once we know what needs to be done, we could upstream it. The upstream project is kind-of dormant, but it might be possible to get things in (I might have commit permission)
  * Also, given the unmaintained state of the installler, we could consider not using the installer and providing that manually.

There are other issues with bazel 7: for some reason the protocol buffers are completely borked, but that is for another day :)

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -81,7 +81,10 @@ http_archive(
     # it is benign to leave out.
     # Upstream bug https://github.com/google/bazel_rules_install/issues/31
     patch_args = ["-p1"],
-    patches = ["//bazel:installer.patch"],
+    patches = [
+            "//bazel:installer.patch",
+            #"//bazel:installer-bazel7.patch",  # TODO: if-magic for if bazel7
+    ],
     sha256 = "880217b21dbd40928bbe3bca3d97bd4de7d70d5383665ec007d7e1aac41d9739",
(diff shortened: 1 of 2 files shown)


hzeller:
Jonathan, you're the only one close to this project who I imagine knows bazel enough to help figuring these things out.

hzeller:
Yeah, kythe was the first project that I encountered that used .bazelversion, and it created major headaches back then and now.

So I think for now, we just have to tell developers to use bazel 5 or 6 and not 7, but do not write any .bazelversion as it is fundamentally broken the way it is implemented in bazel and bazelisk.

PR #2090, opened 2024-01-31
Use local win_flex_bison instead of fetching it automatically

corco:
This should resolve NixOS requiring a patch to disable win_flex_bison in order to be able to compile verible.

It does mean Windows user must install win_flex_bison on their system though.


hzeller:
Thank you!

PR #2098, opened 2024-02-03
Document a strange situation in which we find end() iterators in toke…

hzeller:
…n-views.

Needs to be explored more, but I think the view should only contain dereferencable iterators.
If CHECK() is enabled (`#if 1`), the formatter_test will fail.


hzeller:
@fangism : Maybe you can remember this part: is my assumption that a view should not contain the end() iterator from the source.

PR #2102, opened 2024-02-06
Resolve "Rejected valid signal name"

### LINKED ISSUE #2076: Rejected valid signal name
**Describe the bug**

Short summary.

**To Reproduce**

```bash
verible-verilog-lint test.sv
```

with `test.sv`

```systemverilog
module test;
  logic analog;
endmodule
```

**Actual behavior:**

Rejects valid syntax 

```
test.sv:2 9-14 syntax error at token "analog"
```

**Expected behavior**

No syntax error.
`analog` is not a reserved keyword according to LRM Annex B Table B.1

matlupi:
Close #2076: add `analog` as potential AMS-dialect word in parser according to https://github.com/chipsalliance/verible/issues/2076#issuecomment-1919763868

### DIFF
--- a/verible/verilog/parser/verilog-parser_test.cc
+++ b/verible/verilog/parser/verilog-parser_test.cc
@@ -1839,6 +1839,7 @@ static constexpr ParserTestCaseArray kModuleTests = {
     // keyword tests
     "module keyword_identifiers;\n"
     "reg branch;  // branch is a Verilog-AMS keyword\n"
+    "reg analog;  // analog is a Verilog-AMS keyword\n"
     "input from;  // from is a Verilog-AMS keyword\n"
     "wire access;  // access is a Verilog-AMS keyword\n"
     "wire exclude;  // exclude is a Verilog-AMS keyword\n"
--- a/verible/verilog/parser/verilog-parser_test.cc
+++ b/verible/verilog/parser/verilog-parser_test.cc
@@ -4749,6 +4749,18 @@ static constexpr ParserTestCaseArray kModuleMemberTests = {
     "    `MACRO();\n"  // macro as item
     "  }\n"
     "endgroup",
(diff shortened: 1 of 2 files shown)


hzeller:
This creates a shift-reduce conflict, so the grammar probably has to be massaged more.

hzeller:
This is essentially an ambiguity in the grammar where the parser might not be able to decide. One problem is, that now `TK_analog` is part of multiple rules, and these rules might both be matched. Next step would probably to check the original location of where `Tk_analog` and see if it can be replaced with `KeywordIdentifier` and still keep the semantics.

There are probably also a bunch of resources 'on the net' for shift/reduce conflicts resources.

hzeller:
There should be a file generated in `bazel-bin/verilog/parser/verilog.output` which might be helpful

fangism:
The `verilog.output` file is a "human-readable" state machine of the parser, which explains how conflicting states are reached.

fangism (approved):

hzeller:
can you fix the shift-reduce conflicts, so that it compiles ?

matlupi:
> can you fix the shift-reduce conflicts, so that it compiles ?

Fixed: 0 conflicts now (at least locally, CI is running now in this PR).

For the record, replacing `TK_analog` with `KeywordIdentifier` in `analog_construct` still gives 12 conflicts, and so does `%define lr.type ielr`. So I resolved it with token precedence instead, following the `/* to resolve exclude (... ambiguity */` block just above it.

I also added a regression tests for `analog foo <+`, which was not there.

matlupi:
@fangism @hzeller, this looks ready to merge.
Thanks for the feedback and the review!

hzeller:
Refering the review to @fangism as he has more experience with bison/yacc resolution.

hzeller:
@fangism do you have time to have a quick look ?

PR #2109, opened 2024-02-15
lint: remove false positives from uvm-macro-semicolon

IEncinas10:
Macros like `UVM_REG_DATA_WIDTH were being flagged by the rule when they shouldn't.

To the best of my knowledge, the only UVM macros that are parsed like `MacroIdentifier` and should be flagged are the ones ending with `_end`:

  uvm_component_utils_end,
  uvm_error_context_end,
  ...

More context: https://github.com/chipsalliance/verible/issues/2107


fangism (approved):
Thank you!

hzeller:
Thanks, merged!

matlupi:
Thanks!

PR #2111, opened 2024-02-17
Don't declare lint_waiver.h to be exported by two different targets.

hzeller:
It was exported by `common/analysis:command-file-lexer` and `common/analysis:lint-waiver`

Found by github.com/hzeller/bant.


hzeller:
@fangism FYI

PR #2115, opened 2024-02-22
Update bazel_rules_install. Compatible with bzlmod.

hzeller:
This is the first step to make this work on newer bazels.


corco (approved):
Good news!

corco:
I checked, we only need to update "platforms" to make it work on bazel 7, I will prepare a PR once this one is merged

PR #2116, opened 2024-02-22
Updated platforms dependency to support bazel 7

corco:
When we drop bazel 5, we will be able to either switch to bzlmod or move the options "--noenable_bzlmod" to the .bazelrc.


hzeller:
Thanks!
I think next step could just be to add a bzlmod. It should work from 5..7 (I really want to keep things going with 5 while it is supported. Users often have really old machines/set-ups they work with)

PR #2117, opened 2024-02-23
More header cleanups: add needed, removed unnneded.

hzeller:
This time, after the previous cleanup, this now could be done automatically with
```
.github/bin/run-clang-tidy-cached.cc --checks="-*,misc-include-cleaner" --fix
```

The subsequent BUILD file cleanups (DWYU - Depend on What You Use) has been done with `bant` (github.com/hzeller/bant).


hzeller:
With this one, we now have a well-layered include and dependency chain, so these checks can be enabled (bazel layering_check is already enabled, need a newer `clang-tidy` to also get the `misc-include-cleaner` checks).

Overall, this will make sure that we keep dependencies and includes of the the project in order almost automatically.

fangism (approved):

PR #2122, opened 2024-02-28
bazel: add try-import .user.bazelrc for custom user configurations

lromor:


.bazelrc
```diff
@@ -36,3 +36,5 @@ build:asan --copt -fno-omit-frame-pointer
 build:asan --linkopt -fsanitize=address
 
 build:create_static_linked_executables --linkopt=-fuse-ld=bfd --features=-supports_start_end_lib --//bazel:create_static_linked_executables
+
+try-import %workspace%/.user.bazelrc
```
hzeller, line 40: Can you add a comment, something like
```
# User-chosen additional development configuration
```

I'd probaly make it a visible file without prefix-dot: `user.bazelrc`

If you add it to the `.gitignore` it does not accidentally get submitted

hzeller (commented):
This is a good idea.

.bazelrc
hzeller: Looks like bazel itself agrees with a visible `user.bazelrc` : https://github.com/bazelbuild/bazel/blob/master/.bazelrc#L61

hzeller: Ah, also found a documentation with that suggestion https://bazel.build/configure/best-practices#bazelrc-file

hzeller:
Can you add it also to the `.gitignore` ?

hzeller:
Thanks, merged!

PR #2121, opened 2024-02-25
bazel: migrate to bazelmod and bazel 7

lromor:

PR #2128, opened 2024-03-03
Add README to Python examples and highlight that these are standalone.

hzeller:
... and not supported.

Preserve the BUILD file, but make it a BUILD.example to act as a starting point if used in a bazel project.
But really, the programs can and probably should be run stand-alone.

These are not officially supported and keeping them going with the rules_python() is a headache, so take that out of the main build.


hzeller:
This will make #2121 simpler

hzeller:
@lromor Sounds good ?

PR #2120, opened 2024-02-25
Change name of "@jsonhpp" to "@jsonhpp//:json" for WORKSPACE

hzeller:
These seem to be different otherwise in WORKSPACE and MODULE.bazel.

PR #2129, opened 2024-03-04
Change name of our json-package at the right place.

hzeller:
I didnn't realize that the name of our json package is not coming from the nlohmann json source tar, but from our own Bazel file we made ourself. So changing the name is much simpler there than providing a patch :)

Remove the patch from #2120 again and just change the name there.

For the work in #2121 though this means, that it would probably better to use our bazel file (as it uses the single-header version as well as does proper settings for exceptions)


hzeller:
@lromor this is for you to review

WORKSPACE
```diff
@@ -168,10 +168,6 @@ http_archive(
 http_archive(
     name = "jsonhpp",
     build_file = "//bazel:jsonhpp.BUILD",
```
lromor, line 170: I think you can remove both this line:

```suggestion
```

and delete the file `bazel/jsonhpp.BUILD` since the archive already contains the BUILD file with the same target.

lromor (changes requested):

WORKSPACE
lromor: My bad :) Nice!

lromor (approved):

PR #2133, opened 2024-03-07
Variable shadowing rule (Rebase of #790)

hzeller:
This is a rebase of #790 to compile with head.
    
Fixing #247

### DIFF
--- a/verilog/analysis/checkers/instance_shadow_rule.h
+++ b/verilog/analysis/checkers/instance_shadow_rule.h
@@ -16,8 +16,8 @@
 #define VERIBLE_VERILOG_ANALYSIS_CHECKERS_INSTANCE_SHADOW_RULE_H_
 
 #include <set>
-#include <string>
 
+#include "absl/strings/string_view.h"
 #include "common/analysis/lint_rule_status.h"
 #include "common/analysis/syntax_tree_lint_rule.h"
 #include "common/text/symbol.h"
--- a/verilog/analysis/checkers/instance_shadow_rule.h
+++ b/verilog/analysis/checkers/instance_shadow_rule.h
@@ -1,0 +1,57 @@
+// Copyright 2017-2020 The Verible Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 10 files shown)


verilog/analysis/checkers/instance_shadow_rule.h
```diff
@@ -0,0 +1,57 @@
...
+#include "common/analysis/lint_rule_status.h"
+#include "common/analysis/syntax_tree_lint_rule.h"
+#include "common/text/symbol.h"
+#include "common/text/syntax_tree_context.h"
+#include "verilog/analysis/descriptions.h"
+
+namespace verilog {
+namespace analysis {
+
+class InstanceShadowRule : public verible::SyntaxTreeLintRule {
```
fangism, line 30: Can you add a full description with a few examples to the class docstring?

verilog/analysis/checkers/instance_shadow_rule_test.cc
```diff
@@ -0,0 +1,174 @@
...
+       {kToken, "in"},
+       ";\n",
+       "endfunction\n"},
+      {"function automatic int foo (input bit in, input bit ",
+       {kToken, "in"},
+       ");\n",
+       "bit out;\n",
+       "endfunction\n"},
+  };
+  RunLintTestCases<VerilogAnalyzer, InstanceShadowRule>(
```
fangism, line 138: What should happen with test cases that resemble K&R style declarations, like:

```systemverilog
module foo(a, b);
input wire a;   // not a shadow, but attaches a type to the `a` port
output reg b;  // not a shadow
endmodule
```

Can you include a few cases like this?

fangism (approved):

verilog/analysis/checkers/instance_shadow_rule.h
hzeller: I've added a brief description to the class header, and refined the description in the descriptor.

verilog/analysis/checkers/instance_shadow_rule_test.cc
hzeller: It actually does complain about this situation. I added a test that confirms that, and a TODO to fix.
Given that by default, the rule is disabled and that this case of ancient code is rare (errm, should be), I'd like to defer that to a later PR.

hzeller:
Moved to https://github.com/chipsalliance/verible/pull/2230 as I accidentally deleted the old branch.

PR #2137, opened 2024-03-09
Since we use zlib, don't rely on protobuf implicitly providing it.

hzeller:
The way WORKSPACE processes dependencies, we need to define before proto.


hzeller:
FYI: @lromor

Issue #2140, opened 2024-03-10
Windows CI currently failing; wants chocolatey credentials ?


hzeller:
Currently, the CI on Windows is broken:
https://github.com/chipsalliance/verible/actions/runs/8218621248/job/22479126298#step:5:17

Looks like the package manager used there is needing some sort of credentials. Why would credentials be needed to download from the package manager ? Maybe the URL changed ?

hzeller:
I have no idea how things work on Windows, paging @corco for expertise.

hzeller:
Hold on, maybe it fixed itself ...

corco:
The info I found seems to point to a server issue with Chocolatey (or Cloudflare) or an environment issue. I did a PR to update Windows runner to 2022 (from 2019).

Otherwise I would give it a day or two and hope it solves itself.

PR #2141, opened 2024-03-10
Updated Windows to 2022 in CI

corco:
Hopefully it helps to solve #2140, if not then it will not hurt to update the Windows runner to a newer version.


hzeller:
Thanks! Let's give it a go.

PR #2142, opened 2024-03-11
linter: configuration: emit warnings user configuration overrides itself

IEncinas10:
When parsing a configuration file, if the user provides more than one configuration for the same rule, the last value will be used.

As this can be confusing, let's emit a warning whenever this happens.

See https://github.com/chipsalliance/verible/issues/2136 for context.

I'd argue that this isn't **crucial** but it is pretty cheap to do and it might save a lot of time for some users

### DIFF
--- a/verilog/analysis/verilog_linter_configuration_test.cc
+++ b/verilog/analysis/verilog_linter_configuration_test.cc
@@ -835,6 +835,33 @@ TEST(RuleBundleTest, ParseRuleBundleIgnoreExtraComma) {
   }
 }
 
+TEST(RuleBundleTest, ParseRuleBundleDontWarnIfNoConfig) {
+  constexpr absl::string_view text = "test-rule-1,\ntest-rule-1";
+  {
+    RuleBundle bundle;
+    std::string error;
+    bool success = bundle.ParseConfiguration(text, '\n', &error);
+    ASSERT_TRUE(success) << error;
+    EXPECT_THAT(error, testing::Not(testing::HasSubstr(kRepeatedFlagMessage)))
+        << error;  // don't warn about overriden config if there is no value
+    EXPECT_TRUE(bundle.rules["test-rule-1"].enabled);
+  }
+}
+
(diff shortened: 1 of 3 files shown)


hzeller:
Is it possible to add a test for this in `verilog_linter_configuration_test.cc` ?

IEncinas10:
Yeah, sorry about that. Completely forgot about the testing stuff part 😆. Thanks for pointing it out.

I also added more testing for a previous PR and discovered a ""bug"" in my proposed implementation.

verilog/analysis/verilog_linter_configuration_test.cc
```diff
@@ -771,6 +771,20 @@ TEST(RuleBundleTest, ParseRuleBundleReject) {
...
 }
 
+TEST(RuleBundleTest, ParseRuleBundleAcceptGoodRulesEvenWhenRejecting) {
+  constexpr absl::string_view text = "test-rule-unknown-rules\ntest-rule-1";
+  {
+    RuleBundle bundle;
+    std::string error;
+    bool success = bundle.ParseConfiguration(text, '\n', &error);
+    ASSERT_TRUE(!success) << error;
+    ASSERT_NE(error.find(std::string(kInvalidFlagMessage)), std::string::npos)
```
hzeller, line 781: These string comparisons to look for a particular message might be reading more natrually by using the test framework HasSubstr() comparison:

```
EXPECT_THAT(error, HasSubstr(kInvalidFlagMessage));
```

(and somewhere towards the top of the file `using testing::HasSubstr;`)

hzeller:
Looks good in general, but see the suggestion to maybe make the message finding code a bit more naturally to read. Either way would be fine, by me, but I prefer the `HasSubstr()` one.

verilog/analysis/verilog_linter_configuration_test.cc
IEncinas10: I didn't know those functions existed, thanks for the suggestion, it definitely looks better. I went with the full name rather than doing `using testing...`

hzeller (approved):
LGTM

Issue #2148, opened 2024-03-22
Generate default .rules.verible_lint from verible-verilog-lint


sconwayaus:
**Summary**
Generating a complete '.rules.verible_lint' file is a bit tedious to do by hand. It would be better to be able to get a default  '.rules.verible_lint' file out of the lint tool as a starting point.  

**Test cases and examples**
Execute 'verible-verilog-lint --print_rules_file' to print the default rules to stdout. Users can then pipe/copy&paste the output to a file for latter use ('.rules.verible_lint' for example).

Example output:
~/git/verible/bazel-bin/verilog/tools/lint$ ./verible-verilog-lint --print_rules_file
+always-comb
+always-comb-blocking
+always-ff-non-blocking=catch_modifying_assignments:false;waive_for_locals:false
-banned-declared-name-patterns
+case-missing-default
+constraint-name-style
+create-object-name-match
-disable-statement
-endif-comment
...

**Proposal**
I've implemented the above in a fork, mainly to see if I could implement the feature, with the view to start a pull request. Before I do is there is any appetite for such a feature?

**Additional context**
None

hzeller:
I like that, please send a pull request!

IEncinas10:
Closing, thanks for the PR!

PR #2150, opened 2024-03-23
Adding feature to print the lint rules in a format suitable to be used as '.rules.verible_lint' @see #2148

sconwayaus:
Adding feature to print the lint rules in a format suitable to be used as '.rules.verible_lint' as mentioned in #2148

Execute 'verible-verilog-lint --print_rules_file' to print the default rules to stdout. Users can then pipe/copy&paste the output to a file for latter use ('.rules.verible_lint' for example).

Example output:
~/git/verible/bazel-bin/verilog/tools/lint$ ./verible-verilog-lint --print_rules_file
+always-comb
+always-comb-blocking
+always-ff-non-blocking=catch_modifying_assignments:false;waive_for_locals:false
-banned-declared-name-patterns
+case-missing-default
+constraint-name-style
+create-object-name-match
-disable-statement
-endif-comment
...

### DIFF
--- a/verilog/analysis/checkers/explicit_parameter_storage_type_rule.cc
+++ b/verilog/analysis/checkers/explicit_parameter_storage_type_rule.cc
@@ -100,7 +100,7 @@ void ExplicitParameterStorageTypeRule::HandleSymbol(
 // a common type that can't be handled well in some old tools.
 absl::Status ExplicitParameterStorageTypeRule::Configure(
     absl::string_view configuration) {
-  static const std::vector<absl::string_view> allowed = {"string"};
+  static const std::vector<absl::string_view> allowed = {"", "string"};
   using verible::config::SetStringOneOf;
   std::string value;
   auto s = verible::ParseNameValues(
--- a/verilog/analysis/verilog_linter.cc
+++ b/verilog/analysis/verilog_linter.cc
@@ -383,29 +383,50 @@ void GetLintRuleDescriptionsHelpFlag(std::ostream *os,
   }
 }
 
-void GetLintRuleFile(std::ostream *os) {
-  // Set up the map.
-  auto rule_map = analysis::GetAllRuleDescriptions();
-  for (const auto &rule_id : analysis::kDefaultRuleSet) {
-    rule_map[rule_id].default_enabled = true;
-  }
+void GetLintRuleFile(std::ostream *os, const LinterConfiguration &config) {
+  // This rule bundle contains only a list of enabled rules. There
+  // are also no param's defined (an empty string), unless the user
+  // assigns them.
+  RuleBundle rule_bundle;
+  config.GetRuleBundle(&rule_bundle);
 
-  // Print all the rules
-  for (const auto &rule : rule_map) {
-    *os << (rule.second.default_enabled ? "+" : "-");
-    *os << rule.second.descriptor.name;
+  // Grab all the rule descriptions, so we can get default
+  // configuration and disabled rules
+  auto rule_descriptions = analysis::GetAllRuleDescriptions();
 
-    bool first = true;
-    for (const auto &rule_param : rule.second.descriptor.param) {
-      *os << (first ? "=" : ";");
-      *os << rule_param.name << ":";
-      *os << rule_param.default_value;
+  // Update the rule_bundle with default configration if none was
+  // provided and add disabled rules with default configuration
+  for (const auto &rule : rule_descriptions) {
+    // Form the rules default_configuration string
+    std::string default_configuration;
+    bool first_param = true;
+    for (const auto &param : rule.second.descriptor.param) {
+      if (!first_param) {
+        default_configuration.append(";");
+      }
 
-      first = false;
+      default_configuration.append(std::string(param.name));
+      default_configuration.append(":");
+      default_configuration.append(param.default_value);
+
+      first_param = false;
     }
 
-    *os << "\n";
+    if (auto found_rule = rule_bundle.rules.find(rule.first); found_rule != rule_bundle.rules.end()) {
+      // Rule is enabled, add default configuration if none exists
+      if (found_rule->second.configuration.empty()) {
+        found_rule->second.configuration = default_configuration;
+      }
+    } else {
+      // Add disbaled rule, along with its default configuration
+      rule_bundle.rules[rule.first].enabled = false;
+      rule_bundle.rules[rule.first].configuration = default_configuration;
+    }
   }
+
+  // Print the rules
+  std::string rules = rule_bundle.UnparseConfiguration('\n', false);
+  *os << rules << "\n";
 }
 
 void GetLintRuleDescriptionsMarkdown(std::ostream *os) {
--- a/verilog/analysis/verilog_linter.cc
+++ b/verilog/analysis/verilog_linter.cc
@@ -83,11 +83,11 @@ using verible::TextStructureView;
 using verible::TokenInfo;
 
 std::set<LintViolationWithStatus> GetSortedViolations(
-    const std::vector<LintRuleStatus>& statuses) {
+    const std::vector<LintRuleStatus> &statuses) {
   std::set<LintViolationWithStatus> violations;
 
-  for (const auto& status : statuses) {
-    for (const auto& violation : status.violations) {
+  for (const auto &status : statuses) {
+    for (const auto &violation : status.violations) {
       violations.insert(LintViolationWithStatus(&violation, &status));
     }
   }
@@ -99,9 +99,9 @@ std::set<LintViolationWithStatus> GetSortedViolations(
 //  0: success
 //  1: linting error (if parse_fatal == true)
 //  2..: other fatal issues such as file not found.
-int LintOneFile(std::ostream* stream, absl::string_view filename,
-                const LinterConfiguration& config,
-                verible::ViolationHandler* violation_handler, bool check_syntax,
+int LintOneFile(std::ostream *stream, absl::string_view filename,
+                const LinterConfiguration &config,
+                verible::ViolationHandler *violation_handler, bool check_syntax,
                 bool parse_fatal, bool lint_fatal, bool show_context) {
   const absl::StatusOr<std::string> content_or =
       verible::file::GetContentAsString(filename);
@@ -126,7 +126,7 @@ int LintOneFile(std::ostream* stream, absl::string_view filename,
     if (!lex_status.ok() || !parse_status.ok()) {
       const std::vector<std::string> syntax_error_messages(
           analyzer->LinterTokenErrorMessages(show_context));
-      for (const auto& message : syntax_error_messages) {
+      for (const auto &message : syntax_error_messages) {
         *stream << message << std::endl;
       }
       if (parse_fatal) {
@@ -138,7 +138,7 @@ int LintOneFile(std::ostream* stream, absl::string_view filename,
   }
 
   // Analyze the parsed structure for lint violations.
-  const auto& text_structure = analyzer->Data();
+  const auto &text_structure = analyzer->Data();
   const auto linter_result =
       VerilogLintTextStructure(filename, config, text_structure);
   if (!linter_result.ok()) {
@@ -147,10 +147,10 @@ int LintOneFile(std::ostream* stream, absl::string_view filename,
     return 2;
   }
 
-  const std::vector<LintRuleStatus>& linter_statuses = linter_result.value();
+  const std::vector<LintRuleStatus> &linter_statuses = linter_result.value();
 
   size_t total_violations = 0;
-  for (const auto& rule_status : linter_statuses) {
+  for (const auto &rule_status : linter_statuses) {
     total_violations += rule_status.violations.size();
   }
 
@@ -174,45 +174,45 @@ int LintOneFile(std::ostream* stream, absl::string_view filename,
 
 VerilogLinter::VerilogLinter()
     : lint_waiver_(
-          [](const TokenInfo& t) {
+          [](const TokenInfo &t) {
             return IsComment(verilog_tokentype(t.token_enum()));
           },
-          [](const TokenInfo& t) {
+          [](const TokenInfo &t) {
             return IsWhitespace(verilog_tokentype(t.token_enum()));
           },
           kLinterTrigger, kLinterWaiveLineCommand, kLinterWaiveStartCommand,
           kLinterWaiveStopCommand) {}
 
-absl::Status VerilogLinter::Configure(const LinterConfiguration& configuration,
+absl::Status VerilogLinter::Configure(const LinterConfiguration &configuration,
                                       absl::string_view lintee_filename) {
   if (VLOG_IS_ON(2)) {
-    for (const auto& name : configuration.ActiveRuleIds()) {
+    for (const auto &name : configuration.ActiveRuleIds()) {
       LOG(INFO) << "active rule: '" << name << '\'';
     }
   }
   auto text_rules = configuration.CreateTextStructureRules();
   if (!text_rules.ok()) return text_rules.status();
-  for (auto& rule : *text_rules) {
+  for (auto &rule : *text_rules) {
     text_structure_linter_.AddRule(std::move(rule));
   }
   auto line_rules = configuration.CreateLineRules();
   if (!line_rules.ok()) return line_rules.status();
-  for (auto& rule : *line_rules) {
+  for (auto &rule : *line_rules) {
     line_linter_.AddRule(std::move(rule));
   }
   auto token_rules = configuration.CreateTokenStreamRules();
   if (!token_rules.ok()) return token_rules.status();
-  for (auto& rule : *token_rules) {
+  for (auto &rule : *token_rules) {
     token_stream_linter_.AddRule(std::move(rule));
   }
   auto syntax_rules = configuration.CreateSyntaxTreeRules();
   if (!syntax_rules.ok()) return syntax_rules.status();
-  for (auto& rule : *syntax_rules) {
+  for (auto &rule : *syntax_rules) {
     syntax_tree_linter_.AddRule(std::move(rule));
   }
 
   absl::Status rc = absl::OkStatus();
-  for (const auto& waiver_file :
+  for (const auto &waiver_file :
        absl::StrSplit(configuration.external_waivers, ',', absl::SkipEmpty())) {
     auto content_or = verible::file::GetContentAsString(waiver_file);
     if (!content_or.ok()) continue;  // Couldn't read lint file: ignore
@@ -227,7 +227,7 @@ absl::Status VerilogLinter::Configure(const LinterConfiguration& configuration,
   return rc;
 }
 
-void VerilogLinter::Lint(const TextStructureView& text_structure,
+void VerilogLinter::Lint(const TextStructureView &text_structure,
                          absl::string_view filename) {
   // Collect all lint waivers in an initial pass.
   lint_waiver_.ProcessTokenRangesByLine(text_structure);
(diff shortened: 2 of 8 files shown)


verilog/analysis/verilog_linter.cc
```diff
@@ -383,22 +383,48 @@ void GetLintRuleDescriptionsHelpFlag(std::ostream* os,
...
+    rule_map[rule_id].default_enabled = true;
+  }
+
+  // Print all the rules
+  for (const auto &rule : rule_map) {
+    *os << (rule.second.default_enabled ? "+" : "-");
+    *os << rule.second.descriptor.name;
+
+    bool first = true;
+    for (auto it = rule.second.descriptor.param.begin();
```
hzeller, line 399: Can you make this a range-based loop ? `for (const auto &rule_param : rule.second.descriptor.param) {...}`

(this is what the clang-tidy CI check is about but for some reason the github UI is brokken, so you can't see the log output in the CI run, but by looking at the [raw log](https://pipelinesghubeus23.actions.githubusercontent.com/Vjp6wX1uV2r1O1Bj6h2lKqrBcOftZCnAHGZLuirThVBBZtBnxy/_apis/pipelines/1/runs/5563/signedlogcontent/6?urlExpires=2024-03-23T16%3A18%3A47.0220215Z&urlSigningMethod=HMACV1&urlSignature=rVLdOFSuGdtk6YYaxPGUkg2JjYTLTLUvmal%2B4cOCRVE%3D)

hzeller (commented):
Thanks! A first comment to fix the clang-tidy CI, I'll look through the details once I have my morning coffee :)

Don't worry about the smoke test issues in the CI, I'll fix them separately.

When you do little changes that don't warrant a separate commit, like changing the for-loop just amend your commit with `git commit --amend` locally, then `git push -f`

verilog/tools/lint/verilog_lint.cc
```diff
@@ -92,7 +92,10 @@ ABSL_FLAG(
     "If true, print the description of every rule formatted for the "
     "Markdown and exit immediately. Intended for the output to be written "
     "to a snippet of Markdown.");
-
+ABSL_FLAG(
+    bool, print_rules_file, false,
+    "Print the default lint values in a format that can be used to create a"
```
hzeller, line 97: there is probably a space missing at the end of the string in this line, otherwise the `.rules.verible_lint` will attach right to it and look something like `"...create a.rules.vver..."`

verilog/tools/lint/verilog_lint.cc
```diff
@@ -92,7 +92,10 @@ ABSL_FLAG(
...
+    ".rules.verible_lint file.");
```
hzeller, line 98: To be consistent what we write in `generate_markdown` and `help_rules`, we should probably also mention the exit immediately part.

verilog/analysis/verilog_linter_test.cc
```diff
@@ -439,6 +439,17 @@ TEST(VerilogLinterDocumentationTest, AllRulesMarkdown) {
...
 
+TEST(VerilogLinterDocumentationTest, PrintLintRuleFile) {
+  std::ostringstream stream;
+  verilog::GetLintRuleFile(&stream);
+  // Spot-check a few patterns, must mostly make sure generation
+  // works without any fatal errors.
+  EXPECT_TRUE(absl::StrContains(stream.str(), "+always-comb"));
+  EXPECT_TRUE(absl::StrContains(
+      stream.str(), "+module-filename=allow-dash-for-underscore:false"));
+  EXPECT_TRUE(absl::StrContains(stream.str(), "-forbid-negative-array-dim"));
```
hzeller, line 450: A good test would be to do a roundtrip
  * have a set of rules that we print out
  * use the print-out and parse that into a new set of rules. This will make sure it is parseable.
  * Test that the newly parsed set of rules matches the original set of rules, i.e. that the parsing was successful in extracting all the data.

verilog/analysis/verilog_linter.cc
```diff
@@ -383,22 +383,48 @@ void GetLintRuleDescriptionsHelpFlag(std::ostream* os,
...
+         it != rule.second.descriptor.param.end(); ++it) {
+      *os << (first ? "=" : ";");
+      *os << it->name << ":";
+      *os << it->default_value;
+
+      first = false;
+    }
+
+    *os << std::endl;
```
hzeller, line 408: Just make that "\n". `std::endl` unnecessarily flushes the stream.

verilog/analysis/verilog_linter.cc
```diff
@@ -383,22 +383,48 @@ void GetLintRuleDescriptionsHelpFlag(std::ostream* os,
...
+  for (const auto &rule : rule_map) {
+    *os << (rule.second.default_enabled ? "+" : "-");
+    *os << rule.second.descriptor.name;
+
+    bool first = true;
+    for (auto it = rule.second.descriptor.param.begin();
+         it != rule.second.descriptor.param.end(); ++it) {
+      *os << (first ? "=" : ";");
+      *os << it->name << ":";
+      *os << it->default_value;
```
hzeller, line 403: (Note: just something for possible future change)
Maybe something to consider in the future: if the user has set values on the command line with the `--rules` flag, the output would not reflect that as we only print the defaults here. There might be good reasons for both cases - default to get a starting files and the other after someone has optimized their settings on the command line and wants to print a file with all the settings

hzeller (changes requested):
Looks great!

I have one suggestion for a more thorough test so that we make sure that whatever we print will also be parseable again.

Then I think this is ready to merge.

verilog/analysis/verilog_linter.cc
sconwayaus: No problems. Done

verilog/tools/lint/verilog_lint.cc
sconwayaus: No worries, Done.

verilog/analysis/verilog_linter.cc
sconwayaus: Thanks for pointing the difference out. Done.

verilog/tools/lint/verilog_lint.cc
sconwayaus: Nice pick up. Fixed.

verilog/analysis/verilog_linter_test.cc
sconwayaus: Yep, I'll put something together.

sconwayaus: I've implemented this change. Let me know what you think.

verilog/analysis/verilog_linter.cc
sconwayaus: I ran into the use case you mentioned so I've implemented this change.

sconwayaus:
I think this is good to go again. The print will now output the current set of rules and configuration that the lint tool will use, where rules can be changed by the user through the usual means. All disabled rules also print out with their default params too.

Clang-tidy was complaining about an issue that I've since fixed, but the CI hasn't triggered again.

There are also a couple of small bug fixes that have snuck into this PR that I should point out.

1. When rules are "configured", the error message was not helpful to trace the issue, so added the rule name to the error message. @see verilog/analysis/verilog_linter_configuration.cc:304
Example: 
```
verible-verilog-lint --rules explicit-parameter-storage-type=exempt_type:invalid_value ./verilog/tools/kythe/testdata/module_parameter.sv
E0327 20:11:47.143266    2977 verilog_linter.cc:146] Fatal error: explicit-parameter-storage-type exempt_type: Value can only be one of ['', 'string']; got 'invalid_value'
``` 

2. Fixed an issue with the explicit_parameter_storage_type_rule which didn't include the default value of "" (empty string) in the allowed list of values. @see verilog/analysis/checkers/explicit_parameter_storage_type_rule.cc:103.

hzeller:
Sorry, was a bit busy this week, but will hopefully be able to review tomorrow.

verilog/analysis/verilog_linter.h
```diff
@@ -130,19 +130,23 @@ absl::Status AppendLinterConfigurationFromFile(
...
-void GetLintRuleDescriptionsHelpFlag(std::ostream*, absl::string_view);
+void GetLintRuleDescriptionsHelpFlag(std::ostream *, absl::string_view);
 
 // Outputs the descriptions for every rule, formatted for markdown.
-void GetLintRuleDescriptionsMarkdown(std::ostream*);
+void GetLintRuleDescriptionsMarkdown(std::ostream *);
+
+// Outputs the default linting rules in a format suitable to produce a
+// .rules.verible_lint file
+void GetLintRuleFile(std::ostream *os, const LinterConfiguration &config);
```
hzeller, line 149: Can you add a TODO for all these three Get*() functions that print to a stream to actually call them 'Print*()' as they are printing, not really getting.
(only TODO sufficient, we don't want to widen the scope of this change)

verilog/analysis/verilog_linter.cc
```diff
@@ -383,22 +383,68 @@ void GetLintRuleDescriptionsHelpFlag(std::ostream* os,
...
+      default_configuration.append(std::string(param.name));
+      default_configuration.append(":");
+      default_configuration.append(param.default_value);
+
+      first_param = false;
+    }
+
+    if (rule_bundle.rules.find(rule.first) != rule_bundle.rules.end()) {
+      // Rule is enabled, add default configuration if none exists
+      if (rule_bundle.rules[rule.first].configuration.empty()) {
```
hzeller, line 417: with this access with the `[]` operator, you're doing the same look-up work multiple times.

```c++
if (auto found = rule_bundle.rules.find(rule.first); found != rule_bundle.rules.end()) {
   if (found->second.configuration.empty()) {
     found->second.configuration = default_configuration;
  }  
}
```

verilog/analysis/verilog_linter_configuration.cc
```diff
@@ -262,6 +268,15 @@ std::set<analysis::LintRuleId> LinterConfiguration::ActiveRuleIds() const {
   return result;
 }
 
+void LinterConfiguration::GetRuleBundle(RuleBundle *rule_bundle) const {
+  // Clear the vector to overwrite any existing value.
+  rule_bundle->rules.clear();
+
+  for (const auto &rule_pair : configuration_) {
+    rule_bundle->rules[rule_pair.first] = rule_pair.second;
```
hzeller, line 276: couldn't this whole function just be a `rule_bundle_->rules = configuration_` ?

verilog/analysis/verilog_linter_configuration.h
```diff
@@ -82,7 +82,7 @@ struct RuleBundle {
   // typically that would be a comma or newline.
   bool ParseConfiguration(absl::string_view text, char separator,
                           std::string *error);
-  std::string UnparseConfiguration(char separator) const;
+  std::string UnparseConfiguration(char separator, bool reverse = true) const;
```
hzeller, line 85: Why is reverse needed ? Sounds like it was a leftover from some debugging ?

(also, if it is needed, it needs a documentation here).

verilog/tools/lint/verilog_lint.cc
```diff
@@ -92,7 +92,11 @@ ABSL_FLAG(
     "If true, print the description of every rule formatted for the "
     "Markdown and exit immediately. Intended for the output to be written "
     "to a snippet of Markdown.");
-
+ABSL_FLAG(
+    bool, print_rules_file, false,
+    "Print the default lint rules in a format that can be used to create a "
```
hzeller, line 97: Below, you're reading the `LinterConfigurationFromFlags()`. So that means it is not the 'default lint rules' anymore ? In that case I suggest we rephrase that as 'currrent active lint rules' or something.

verilog/analysis/checkers/explicit_parameter_storage_type_rule.cc
```diff
@@ -100,7 +100,7 @@ void ExplicitParameterStorageTypeRule::HandleSymbol(
 // a common type that can't be handled well in some old tools.
 absl::Status ExplicitParameterStorageTypeRule::Configure(
     absl::string_view configuration) {
-  static const std::vector<absl::string_view> allowed = {"string"};
+  static const std::vector<absl::string_view> allowed = {"", "string"};
```
hzeller, line 103: this change and the one in the corresponding test don't like like they are meant to be in this PR.

verilog/analysis/verilog_linter_configuration.cc
```diff
@@ -283,7 +298,12 @@ static absl::StatusOr<std::vector<std::unique_ptr<T>>> CreateRules(
     if (!setting.configuration.empty()) {
       if (absl::Status status = rule_ptr->Configure(setting.configuration);
           !status.ok()) {
-        return status;
+        std::string error_msg;
+
+        error_msg.append(std::string(rule_pair.first));
```
hzeller, line 303: `absl::StrCat()` is a good way to join strings in a readable way.
(also, this change only mildly looks like it should be part of this PR as it is not part of the rules_verible_lint printing.

verilog/analysis/verilog_linter_configuration.cc
```diff
@@ -334,8 +354,9 @@ absl::Status LinterConfiguration::AppendFromFile(
     // Log warnings and errors
     if (!error.empty()) {
       std::cerr << "Using a partial version from " << config_filename
-                << ". Found the following issues: " << error;
+                << ". Found the following issues: " << error << std::endl;
```
hzeller, line 357: Just use `<< "\n"` instead of `<< std::endl`.

verilog/analysis/verilog_linter_test.cc
```diff
@@ -439,6 +441,87 @@ TEST(VerilogLinterDocumentationTest, AllRulesMarkdown) {
   EXPECT_TRUE(absl::StrContains(stream.str(), "Enabled by default:"));
 }
 
+TEST(VerilogLinterDocumentationTest, PrintLintRuleFile) {
+  auto config_status = verilog::LinterConfigurationFromFlags("");
```
hzeller, line 445: A good naming scheme for things that return StatusOr<> is `<expectedthing>_or`

```c++
 auto config_or = ...;
```

verilog/analysis/verilog_linter_test.cc
```diff
@@ -439,6 +441,87 @@ TEST(VerilogLinterDocumentationTest, AllRulesMarkdown) {
...
+
+  // Generate Line Rule File
+  std::ostringstream stream;
+  verilog::GetLintRuleFile(&stream, config);
+
+  // Spot-check a few patterns, must mostly make sure generation
+  // works without any fatal errors.
+  // NOTE: This will break if/when the rules change so this part
+  // of the test is not ideal.
+  EXPECT_TRUE(absl::StrContains(stream.str(), "always-comb"));
```
hzeller, line 460: In tests, we can make use of the `testing::HasSubstr` for slightly more readable way of writing this (grep for this pattern in other tests):

```c++
EXPECT_THAT(stream.str(), HasSubstr("always-comb"));
```

verilog/analysis/verilog_linter_test.cc
```diff
@@ -439,6 +441,87 @@ TEST(VerilogLinterDocumentationTest, AllRulesMarkdown) {
...
+  EXPECT_TRUE(absl::StrContains(
+      stream.str(), "module-filename=allow-dash-for-underscore:false"));
+  EXPECT_TRUE(absl::StrContains(stream.str(), "-forbid-negative-array-dim"));
+
```
hzeller, line 464: If you move the `const std::string generated_default_rules_str = stream.str();` further up, directly after the `GetLintRuleFile()` call, you can use the same readable name in the other tests above as well, where `stream.str()` is used.

verilog/analysis/verilog_linter_test.cc
```diff
@@ -439,6 +441,87 @@ TEST(VerilogLinterDocumentationTest, AllRulesMarkdown) {
...
+  // Roundtrip test, first parse the rules
+  std::string generated_default_rules_str = stream.str();
+  absl::StatusOr<std::string> generated_default_rules =
+      generated_default_rules_str;
+  RuleBundle parsed_rules;
+  std::string error;
+  parsed_rules.ParseConfiguration(*generated_default_rules, '\n', &error);
+  EXPECT_TRUE(error.empty());
+
```
hzeller, line 473: at this point, you have a `RuleBundle` from parsing.
Wouldn't it just be sufficient to compare that `parsed_rules` is the same as you'd get from `config.GetRuleBundle()` ? Otherwise you're essentially re-creating the parsing logic from `RuleBundle::ParseConfiguration` here, which makes things more brittle.

hzeller (changes requested):
Looks pretty good. Most comments are to show different ways that are more common in this project, or points out things that might be not necessary.

verilog/analysis/checkers/explicit_parameter_storage_type_rule.cc
sconwayaus: This change was a by-product of this change. The lint rule print generates the following statement:

`explicit-parameter-storage-type=exempt_type:`

which ExplicitParameterStorageTypeRule::Configure rejects. So I added the rules default value of an empty string to the allowed list.

verilog/analysis/verilog_linter_configuration.h
sconwayaus: By defauit the output string was coming out with paramters in a reverse alphebetical order, so to avoid breaking things I introduced the reverse boolean so the lint rule print looks better.

FWIW, I commented out the reverse ordered print out and ran all the test cases and it only fails in the 3 tests you'd expect in verilog_linter_configration_test.cc around the UnparseConfiguration. Fixing those and there dosn't seem to be any issues.

I can add some doco, unless you want to do something else here.

verilog/tools/lint/verilog_lint.cc
sconwayaus: I'll fix that.

verilog/analysis/verilog_linter_configuration.cc
sconwayaus: There's a couple of other places in this file with std::endl streaming to std::cerr. Do you want me to fix those while I'm there?

verilog/analysis/verilog_linter_test.cc
```diff
@@ -439,6 +441,87 @@ TEST(VerilogLinterDocumentationTest, AllRulesMarkdown) {
   EXPECT_TRUE(absl::StrContains(stream.str(), "Enabled by default:"));
 }
 
+TEST(VerilogLinterDocumentationTest, PrintLintRuleFile) {
+  auto config_status = verilog::LinterConfigurationFromFlags("");
+  EXPECT_TRUE(config_status.ok());
```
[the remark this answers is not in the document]
sconwayaus: Honestly that bugged me when I wrote it. Thanks.

verilog/analysis/verilog_linter.h
sconwayaus: No problem, I'll push this over once I get you're feeback on the below.

verilog/analysis/verilog_linter.cc
sconwayaus: Thanks. I can make that change.

verilog/analysis/verilog_linter_configuration.cc
sconwayaus: My C++ is rusty (thanks for your patience). Yes it would be.

verilog/analysis/verilog_linter_test.cc
sconwayaus: No worries.

verilog/analysis/verilog_linter_test.cc
sconwayaus: I think I did it this way as there are a number of other tests in that file using the EXPECT_TRUE(sbsl::StrContains... pattern. Happy to change if you prefer.

verilog/analysis/verilog_linter_test.cc
sconwayaus: Yep, agree.

verilog/analysis/verilog_linter_test.cc
sconwayaus: As it stands, config.GetRuleBundle() dosn't inlcude the disabled rules or the default configuration for each rule.

I could pass parsed_rule_bundle back through UnparseConfiguration and compare against the generated rules string. Something like this perhaps?
 
```
  std::string generated_default_rules_str verilog::GetLintRuleFile(&stream, config).str()

  // Roundtrip test, first parse the rules
  parsed_rule_bundle.ParseConfiguration(*generated_default_rules_str, '\n', &error);

  // Convert the parsed_rule_bundle back to a string
  std::string unparsed_rule_bundle = parsed_rule_bundle.UnparseConfiguration('\n', false);

  // compare the unparsed_rule_bundle to the generated_default_rules_str
  EXPECT_TRUE(unparsed_rule_bundle == generated_default_rules_str));
}
```

verilog/analysis/verilog_linter_configuration.cc
hzeller: Let's leave the other places for some separate cleanup PR. That way it is easy to focus on one thing at a time.

verilog/analysis/verilog_linter_test.cc
hzeller: Similar here: let's do cleanup of existing improvement places to a separate PR.

verilog/analysis/verilog_linter_configuration.cc
sconwayaus: I hit this error message when I was miss configuring the rule above. The error message generated didn't link the fault back to a rule so I've prepended the rule name to help the user narrow down the issue better. The new error message should look like this:

`E0327 20:11:47.143266    2977 verilog_linter.cc:146] Fatal error: explicit-parameter-storage-type exempt_type: Value can only be one of ['', 'string']; got 'invalid_value'
`

For future reference, would you prefer this to come across as a seperate PR?

hzeller: Since it helped you discovering inadequate error reporting while working on this PR, this is good to keep in this PR. Just use `absl::StrCat()` to assemble the string.

verilog/analysis/verilog_linter_test.cc
sconwayaus: I've pushed some changes across, let me know if there's anything else I should address.

hzeller (approved):

PR #2155, opened 2024-04-10
Vscode is binary found

### LINKED ISSUE #2154: VSCode can't set binary path in windows
VSCode 1.87.2

**What activity failed**
1. Unable to set the absolute path to verible-verilog-ls.exe on Windows 10. Workaround is to put the executable on the %PATH%.
2. Unable to use HOME directory path (~/verible/verible-verilog-ls).   

**Expectation**
Able to set an path to the ls executable.  

**What actually happened**
Windows: Fails to find the executable, and starts a download.
Linux: Seems to try to start the executable but fails.

**Windows Task Log** 
> Set language server executable (C:\My_Apps\verible\verible-verilog-ls.exe) doesn't exist or cannot be accessed
> Extension will attempt to download executables (v0.0-3624-gd256d779) from GitHub Release page
> ...

**Linux Task Log**
> [Error - 10:20:47 PM] Starting client failed
> Launching server using command ~/verible/verible-verilog-ls failed.
> ...

sconwayaus:
Resolves #2154 

Fixes several issues with the VSCode extension
1. Fixes issue on windows where users are unable to set an absolute (or relative) path to the language server.
2. Fixes issue on Linux preventing users to use ~/ paths to language server.
3. Fixed issue found while testing the above where the extension on windows dosn't extract the langauge server out of the downloaded zip file.

### DIFF
--- a/verilog/tools/ls/vscode/src/download-ls.ts
+++ b/verilog/tools/ls/vscode/src/download-ls.ts
@@ -1,5 +1,5 @@
 import * as vscode from "vscode";
-import { platform, arch } from "os";
+import { homedir, platform, arch } from "os";
 import * as fs from "fs";
 import * as path from "path";
 import { IncomingMessage } from "http";
@@ -11,13 +11,19 @@ const TAG = require("../package.json").repository.tag;
 
 
 function checkIfBinaryExists(binaryPath: string) {
-  let whichCommand: string, binaryExists: boolean;
-  if (platform() == "win32") whichCommand = "where";
-  else whichCommand = "command -v";
+  let whichCommand: string;
+  let binaryExists: boolean;
+
+  if (platform() == "win32") {
+    let parsedBinPath = path.parse(binaryPath);
+    whichCommand = `where.exe ${parsedBinPath.dir}:${parsedBinPath.base}`;
+  } else {
+    whichCommand = `command -v ${binaryPath}`;
+  }
 
   binaryExists = true;
   try {
-    execSync(`${whichCommand} ${binaryPath}`, { windowsHide: true });
+    execSync(whichCommand, { windowsHide: true });
   } catch {
     binaryExists = false;
   }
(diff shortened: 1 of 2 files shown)


verilog/tools/ls/vscode/src/download-ls.ts
```diff
@@ -105,14 +123,13 @@ export async function checkAndDownloadBinaries(
       file.path = path.basename(file.path);
       return file;
     },
-    plugins: platform() === "win32" ? [] : [decompressTargz()],
+    plugins: platform() === "win32" ? [decompressUnzip()] : [decompressTargz()],
```
hzeller, line 126: we implicitly assume that the binary is zipped for windows and tar.gz'ed for the rest. Might be useful to use the uncompression method based on the file-ending (fine for now, just might be something for the future).

verilog/tools/ls/vscode/src/download-ls.ts
```diff
@@ -29,6 +36,16 @@ export async function checkAndDownloadBinaries(
   binaryPath: string,
   output: vscode.OutputChannel
 ): Promise<string> {
+
+  output.appendLine("Platform: '" + platform() + "'");
+
+  // Update home paths to an absolute path
+  if(platform() != "win32" && binaryPath.startsWith("~/")) {
```
hzeller, line 43: To be consistent with the rest of the file, leave a space after the `if`.

hzeller (approved):
Looks good. Only minor comments. If you want to address them, I am leaving this open for now, otherwise will merge as-is tomorrow (California time).

verilog/tools/ls/vscode/src/download-ls.ts
sconwayaus: I'll leave this one for now.

verilog/tools/ls/vscode/src/download-ls.ts
sconwayaus: Done

PR #2169, opened 2024-04-21
Remove explicit copy constructors with default implementation.

hzeller:
If there is a copy constructor, we also would have to define an assignment operator. Since these are default anyway, just use the default as-is.


fangism (commented):
This is fine, but I also don't mind keeping explicit declarations for classes that are not obviously pure data.  Ok for `ThisInfo` and `ThatConfiguration`.

hzeller:
This was discovered by a newer `clang` that complained that if an explicit definition of a copy constructor is there, the implicit `operator=` would be removed.

So the other alternative would be to _also_ add the `operator=`, but at that point I figured it looks more cluttered than not adding the default copy constructor in the first place.

Issue #2181, opened 2024-05-14
verible-verilog-syntax crashing


joaovam:
The verible-verilog-syntax crashes when trying to analyze the following input:

```
     module m();
         id_0var_0 (var_3), var_1, var_2; 
     endmodule
```

the crash output is the following:
```
E0514 11:38:50.098383 2290788 tree_utils.h:92] Node: Programming error: expected kInstantiationType but got kDataType
F0514 11:38:50.098421 2290788 tree_utils.h:128] Check failed: 'MatchNodeEnumOrNull(SymbolCastToNode(symbol), node_enum)' Must be non-null
*** Check failure stack trace: ***
    @     0x60601bc9a2c9  absl::lts_20240116::log_internal::LogMessage::PrepareToDie()
    @     0x60601bc9a341  absl::lts_20240116::log_internal::LogMessage::SendToLog()
    @     0x60601bc99d93  absl::lts_20240116::log_internal::LogMessage::Flush()
    @     0x60601bc9a640  absl::lts_20240116::log_internal::LogMessageFatal::~LogMessageFatal()
    @     0x60601bc989a2  absl::lts_20240116::log_internal::DieBecauseNull()
    @     0x60601bbcda51  absl::lts_20240116::log_internal::DieIfNull<>()
    @     0x60601bbcd7d5  verible::CheckSymbolAsNode<>()
    @     0x60601bc613c5  verilog::MakeInstantiationBase<>()
    @     0x60601bc3ed8c  verilog::verilog_parse()
    @     0x60601bc2f107  verilog::verilog_parse_wrapper()
    @     0x60601bb6f711  verible::BisonParserAdapter<>::Parse()
    @     0x60601bb737d7  verible::FileAnalyzer::Parse()
    @     0x60601bb6379a  verilog::VerilogAnalyzer::Analyze()
    @     0x60601bb61ce5  verilog::VerilogAnalyzer::AnalyzeAutomaticMode()
    @     0x60601bb1c930  ParseWithLanguageMode()
    @     0x60601bb1ce7f  AnalyzeOneFile()
    @     0x60601bb1de65  main
    @     0x7b154b629d90  (unknown)
*** SIGABRT received at time=1715697530 on cpu 11 ***
PC: @     0x7b154b6969fc  (unknown)  pthread_kill
    @     0x60601bb3f17b         64  absl::lts_20240116::WriteFailureInfo()
[...]

PR #2187, opened 2024-05-31
Fix bug caused by faulty production rule

rafasumi:
Fixes issue #2181.

As suggested by @hzeller, I've ran `.github/bin/smoke-test.sh` and the output was the same before and after commenting out the faulty production rule.


fangism (approved):
Failing is better than crashing!

PR #2190, opened 2024-06-06
[proper-parameter-declaration] parameter vs localparam in packages

### LINKED ISSUE #2160: [proper-parameter-declaration] parameter vs localparam in packages
**Describe the bug**
After fixing a few violations from [proper-parameter-declaration] wrt checking parameters in packages, my synthesis tool (Quartus PRO) throws a warning. It seems Quartus expects "localparam" instead of "parameter". Usually I think Quartus is wrong, but after reviewing the LRM I think Quartus is right. 

Section 6.20.4 in the LRM (2017 and older) states "parameters" in packages are treated as a synonym for a localparam, so tools should treat parameters as localparam's in a package. More intuitively, as parameters are overidable, a parameter in a package dosn't make sense in that packages are constant and not parameterisable.

I propose the rule default be changed to raise a violation when a parameter is used in a package.

```systemverilog
// Ran the style linter on this code.
package test_pkg;
  parameter string PARAM = "BAD";  // No lint violation
  localparam string LOCAL_PARAM = "GOOD";  // lint violation reported here
endpackage
```

output from verible-verilog-lint
```
./test_pkg.sv:3:21-31: Non-type localparam names must be styled with CamelCase [Style: constants] [parameter-name-style]
```

What should have happened?
Expect "parameter" in package to be reported as lint failure.

Citations to published style guides:
System Verilog LRM IEEE Std 1800™-2017, Section 6.20.4 

```
6.20.4 Local parameters (localparam)
Local parameters are identical to parameters except that they cannot directly be modified by defparam
statements (see 23.10.1) or instance parameter value assignments (see 23.10.2). Local parameters can be
assigned constant expressions (see 11.2.1) containing parameters, which in turn can be modified with
defparam statements or instance parameter value assignments.

Unlike nonlocal parameters, local parameters can be declared in a generate block, package, class body, or compilation-unit scope. In these contexts, the parameter keyword shall be a synonym for the localparam keyword.

Local parameters may be declared in a module’s parameter_port_list. Any parameter declaration appearing
in such a list between a localparam keyword and the next parameter keyword (or the end of the list, if
there is no next parameter keyword) shall be a local parameter. Any other parameter declaration in such a
list shall be a nonlocal parameter that may be overridden as described in 23.10.
```

sconwayaus:
Resolves #2160 

This PR allows the user to configure the proper-parameter-declaration rule to allow either or both parameter and localparam in packages. 

The default behaviour has changed to be more inline with the LRM, where the LRM encourages users to use localparam in packages as they cannot be overridden, so the rule defaults to only allow localparam's within packages.


sconwayaus:
The windows CI build seems to have broken with the latest runner release (windows2022 runner was updated 3rd June 2024). 

The build seems to be failing due to a number of missing standard libraries. Example below:
```
ERROR: D:/a/verible/verible/common/util/BUILD:101:11: Compiling common/util/init_command_line.cc [for tool] failed: (Exit 1): clang-cl.exe failed: error executing command (from target //common/util:init-command-line) C:\Program Files\LLVM\bin\clang-cl.exe /nologo /DCOMPILER_MSVC /DNOMINMAX /D_WIN32_WINNT=0x0601 /D_CRT_SECURE_NO_DEPRECATE /D_CRT_SECURE_NO_WARNINGS /bigobj /Zm500 /EHsc /wd4351 /wd4291 /wd4250 /wd4996 ... (remaining 21 arguments skipped)
In file included from common/util/init_command_line.cc:15:
.\common/util/init_command_line.h(18,10): fatal error: 'string' file not found
#include <string>
         ^~~~~~~~
1 error generated.
```
I've tried the following:
- Manually trigger the CI on an unmodified branch. Fails in a similar way. 
- Changing the runner to windows 2019 seems to build.
- There is some conversation happening here https://github.com/actions/runner-images/issues about issues with clang-cl but nothing exactly like what I'm observing.

@corco, maybe you have some thoughts?

hzeller:
Created #2193 to investigate.

verilog/analysis/checkers/proper_parameter_declaration_rule.cc
```diff
@@ -40,25 +42,66 @@ using verible::matcher::Matcher;
...
-          "package or in the formal parameter list of modules/classes and "
-          "every `localparam` declaration is inside a module or class.",
-  };
+          "formal parameter list of modules/classes and "
+          "every `localparam` declaration is inside a module, class or "
+          "package.",
+      .param = {
+          {
+              .name = "package_allow_parameter",
+              .default_value = "false",
```
hzeller, line 73: I'd probably not change the default in this PR.

If the default is to be changed, it should be done in a separate (1-liner) change referencing the LRM recommendations. That way new behavior can be narrowed to a small change.

hzeller (changes requested):
Looks good, but I'd be careful changing the default behavior in one go.

I've updated the CI to use 2019, so hopefully it will pass now (and if not in this PR, we can ignore a Windows test failure)

hzeller (approved):

hzeller:
Thanks!

Can you now add a follow-up Pull Request so that we can change the default to warn about using `parameter` in packages (I really like the feature, just important to change code from changing default configuration).

verilog/analysis/checkers/proper_parameter_declaration_rule.cc
```diff
@@ -79,11 +122,24 @@ void ProperParameterDeclarationRule::HandleSymbol(
       } else if (ContextIsInsideModule(context) &&
                  !ContextIsInsideFormalParameterList(context)) {
         violations_.insert(LintViolation(symbol, kParameterMessage, context));
+      } else if (ContextIsInsidePackage(context)) {
+        if (!package_allow_parameter_) {
+          violations_.insert(LintViolation(symbol, kParameterMessage, context));
+        }
```
hzeller, line 128: We have a opportunity here to provide an autofix to change `parameter` to `localparam`.
That way, in the language server it is just one click away to get fixed.

Might be a good follow-up change.

sconwayaus:
I like it. I'll take a look.

PR #2192, opened 2024-06-09
Update CI actions versions to latest

hzeller:


hzeller:
(Won't work currently, looks like the self-hosted runners also would need updating)

PR #2204, opened 2024-06-17
Add a regex configration for the signal name style rule

sconwayaus:
Adding regex configuration to the signal name style rule, part of issue #2074 .


verilog/analysis/checkers/signal_name_style_rule.cc
```diff
@@ -43,17 +48,37 @@ using verible::LintViolation;
 using verible::SyntaxTreeContext;
 using verible::matcher::Matcher;
 
-static constexpr absl::string_view kMessage =
-    "Signal names must use lower_snake_case naming convention.";
+#define STYLE_DEFAULT_REGEX "[a-z_0-9]+"
+static std::string style_default_regex = STYLE_DEFAULT_REGEX;
```
hzeller, line 52: Please don't use global `std::string` as it allocation and deallocation these is a bit icky. A `static constexpr absl::string_view` would be fine.
(also, assign that directly, no need for the `#define`. This should work below in the `.param` init).

hzeller (commented):
Are you still working it it ? (as it is marked as draft)

sconwayaus:
> Are you still working it it ? (as it is marked as draft)
I was. Did some real world testing with it during the week but I think this is ready to move forward.

hzeller:
Did you check the review comment so far ?

sconwayaus (commented):
Just realised my replies were "pending"...

verilog/analysis/checkers/signal_name_style_rule.cc
```diff
@@ -43,17 +48,37 @@ using verible::LintViolation;
...
+          "Example common regex patterns:\n"
+          "  lower_snake_case: \"[a-z_0-9]+\"\n"
+          "  UPPER_SNAKE_CASE: \"[A-Z_0-9]+\"\n"
+          "  Title_Snake_Case: \"[A-Z]+[a-z0-9]*(_[A-Z0-9]+[a-z0-9]*)*\"\n"
+          "  Sentence_snake_case: \"([A-Z0-9]+[a-z0-9]*_?)([a-z0-9]*_*)*\"\n"
+          "  camelCase: \"([a-z0-9]+[A-Z0-9]*)+\"\n"
+          "  PascalCaseRegexPattern: \"([A-Z0-9]+[a-z0-9]*)+\"\n"
+          "RE2 regular expression syntax documentation can be found at "
+          "https://github.com/google/re2/wiki/syntax\n",
+      .param = {{"style_regex", STYLE_DEFAULT_REGEX,
```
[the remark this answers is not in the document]
hzeller: You might need to explicitly convert it to a string in the param case

```
.param = {{"style_regex", std::string(style_default_regex), ...
```

Also ok is to use

```
static const char *style_default_regex = "[a-z_0-9]+";
```

The main issue is not with the `#define`, but the `std::string` as that has the global initialization sequencing issue (as you have seen if you use `std::string` and have an empty string in the default config).

verilog/analysis/checkers/signal_name_style_rule.h
```diff
@@ -15,33 +15,48 @@
...
 
+  absl::Status Configure(absl::string_view configuration) final;
+
  private:
   std::set<verible::LintViolation> violations_;
+
+  // A regex to check the style against
+  std::unique_ptr<re2::RE2> style_regex_;
+
+  std::string kMessage;
```
hzeller, line 59: Only  [real constant values](https://google.github.io/styleguide/cppguide.html#Constant_Names) should have the `kUppercase` style.

Just call it `violation_message_` or similar.

verilog/analysis/checkers/signal_name_style_rule.cc
```diff
@@ -43,17 +48,37 @@ using verible::LintViolation;
 using verible::SyntaxTreeContext;
 using verible::matcher::Matcher;
 
-static constexpr absl::string_view kMessage =
-    "Signal names must use lower_snake_case naming convention.";
+#define STYLE_DEFAULT_REGEX "[a-z_0-9]+"
+static std::string style_default_regex = STYLE_DEFAULT_REGEX;
+
+SignalNameStyleRule::SignalNameStyleRule() {
+  style_regex_ =
```
hzeller, line 55: use RAII instead of assigning values inside the constructor.

So
```c++
SignalNameStyleRule::SignalNameStyleRule() 
  : style_regex_(new re::RE2(...)), violation_message_(absl::StrCat("...")) {}
```

verilog/analysis/checkers/signal_name_style_rule.cc
sconwayaus: Done

verilog/analysis/checkers/signal_name_style_rule.cc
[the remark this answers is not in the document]
sconwayaus: Thanks for that. 
Done

verilog/analysis/checkers/signal_name_style_rule.cc
sconwayaus: No problem. Done.

verilog/analysis/checkers/signal_name_style_rule.h
sconwayaus: Done

sconwayaus:
Thanks for your patience Henner. 

I've added regex's to most of the rules listed in #2074. I'll add your feedback from this PR into those and send on through.

PR #2206, opened 2024-06-29
Fixing the Read The Docs action

mithro:
Add a bunch of files that are no longer provided by default;
 * `.readthedocs.yml` and `requirements.txt` for configuration of the readthedocs build.
 * `conf.py` file for configuration of sphinx.
 * `index.rst` file to provide a based page.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -14,6 +14,10 @@ build:
 sphinx:
   configuration:
 
+python:
+  install:
+    - requirements: doc/requirements.txt
+
 formats:
   - pdf
   - epub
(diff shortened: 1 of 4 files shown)


mithro:
@hzeller - The readthedocs action is now green but the output document appears to be broken - https://verible--2206.org.readthedocs.build/en/2206/

mithro:
Moving to the `myst-parser` seems to have fixed the docs output --> https://verible--2206.org.readthedocs.build/en/2206/

I'm unsure if this if how I organized the output in `doc/index.rst` makes sense.....

mithro:
There also seems to docs @ https://chipsalliance.github.io/verible/ -- Should the readthedocs just redirect to that instead?

mithro:
Well, that is probably as much as I'm going to do / be able to fix.

hzeller:
Cool, thank you!

PR #2210, opened 2024-07-09
Add a regex style pattern to the enum-name-style

sconwayaus:
Adding regex to the enum-name-style rule, part of issue #2074


verilog/analysis/checkers/enum_name_style_rule.cc
```diff
@@ -40,39 +44,54 @@ using verible::LintViolation;
...
+const LintRuleDescriptor &EnumNameStyleRule::GetDescriptor() {
   static const LintRuleDescriptor d{
       .name = "enum-name-style",
       .topic = "enumerations",
       .desc =
-          "Checks that `enum` names use lower_snake_case naming convention "
-          "and end with '_t' or '_e'.",
+          "Checks that enum type names follow a naming convention defined by "
+          "a RE2 regular expression.\n"
+          "Example common regex patterns:\n"
```
hzeller, line 63: Instead of repeating the same examples in various places using regexps, how about we add a little section in the lint [README.md](https://github.com/chipsalliance/verible/tree/master/verilog/tools/lint#readme) with regexp examples and just add a link here ?

(also here, the default regex, the one with the trailing _e or _t is not even mentioned)

verilog/analysis/checkers/enum_name_style_rule.cc
```diff
@@ -81,6 +100,19 @@ void EnumNameStyleRule::HandleSymbol(const verible::Symbol& symbol,
   }
 }
 
+absl::Status EnumNameStyleRule::Configure(absl::string_view configuration) {
+  using verible::config::SetRegex;
+  absl::Status s = verible::ParseNameValues(
+      configuration, {{"style_regex", SetRegex(&style_regex_)}});
+  if (!s.ok()) return s;
+
+  violation_message_ =
```
hzeller, line 109: I am wondering: instead of having a validation_message_ member that we most likely never use, yet allocate and prepare should we maybe have a `std::string CreateValidationMessage() {}` that creates it on-the-fly if needed in the actual valdation ? It also would mean that we don't have to duplicate the code creating it in the constructor and in the `Configure()` method.

verilog/analysis/checkers/enum_name_style_rule.cc
sconwayaus: No problem.

verilog/analysis/checkers/enum_name_style_rule.cc
sconwayaus: No problem.

PR #2213, opened 2024-07-09
Adding regex configuration to the parameter-name-style rule

sconwayaus:
Adding regex configuration to the parameter-name-style rule, part of issue #2074


verilog/analysis/checkers/parameter_name_style_rule.cc
```diff
@@ -92,43 +105,43 @@ void ParameterNameStyleRule::HandleSymbol(const verible::Symbol &symbol,
...
-  static const std::vector<absl::string_view> choices = {
-      "CamelCase", "ALL_CAPS"};  // same sequence as enum StyleChoicesBits
-  using verible::config::SetNamedBits;
-  return verible::ParseNameValues(
+  using verible::config::SetRegex;
+  absl::Status s = verible::ParseNameValues(
       configuration,
-      {{"localparam_style", SetNamedBits(&localparam_allowed_style_, choices)},
-       {"parameter_style", SetNamedBits(&parameter_allowed_style_, choices)}});
+      {{"localparam_style_regex", SetRegex(&localparam_style_regex_)},
```
hzeller, line 141: We shouldn't remove the old configuration names: if someone has used localparam_style and parameter_style in their configuration it will suddenly fail as it doesn't exist anymore.

Instead, we should have not only the new regex names, but also the old names, which then select a regex.
So someone can set `localparam_style_regex`, but they can also say  `localparam_style=ALL_CAPS`, which could be considered a name for a regex; So if such value has been set, we then then can set the upper case regex.

verilog/analysis/checkers/parameter_name_style_rule.cc
```diff
@@ -40,27 +38,47 @@
...
+
 using verible::LintRuleStatus;
 using verible::LintViolation;
 using verible::SyntaxTreeContext;
 using Matcher = verible::matcher::Matcher;
 
-// Register ParameterNameStyleRule.
-VERILOG_REGISTER_LINT_RULE(ParameterNameStyleRule);
+// PascalCase, may end in _[0-9]+
+static constexpr absl::string_view localparam_default_regex =
```
hzeller, line 49: Note, that all constants are always named k-prefix-with-camel-case throughout this project `kFooBar`; so this would be `kLocalparamDefaultRegex` (I think I have overlooked that in the other reviews, it should be corrected there as well).

I think we should have an indirection here with naming that reflects the meaning of the regexes and then the meaning of the default value.
So

```c++
static constexpr absl::string_view kCamelCaseRegex = ...;
static constexpr absl::string_view kUpperSnakeCaseRegex = ...;

static constexpr absl::string_view kLocalparamDefaultRegex = kCamelCaseRegex;
static constexpr absl::string_view kParameterDefaultRegex = kUpperSnakeCaseRegex;
```

Not only makes it more readable here what we mean, but also in the Configuration() below, if we use the named styles (see comment there), it is much easier to refer to it accordingly.

verilog/analysis/checkers/parameter_name_style_rule_test.cc
```diff
@@ -119,31 +119,38 @@ TEST(ParameterNameStyleRuleTest, ConfigurationStyleParameterAssignment) {
...
     const std::initializer_list<LintTestCase> kTestCases = {
         {"package a; parameter int FooBar = 1; endpackage"},
         {"package a; parameter int ", {kToken, "FOO_BAR"}, " = 1; endpackage"},
         {"module a; localparam int ", {kToken, "FooBar"}, " = 1; endmodule"},
         {"module a; localparam int FOO_BAR = 1; endmodule"},
     };
     RunConfiguredLintTestCases<VerilogAnalyzer, ParameterNameStyleRule>(
-        kTestCases, "parameter_style:CamelCase;localparam_style:ALL_CAPS");
+        kTestCases,
+        "parameter_style_regex:([A-Z0-9]+[a-z0-9]*)+(_[0-9]+)?;localparam_"
```
hzeller, line 132: This example shows how hard it is to actually use if just the defaults would be needed; so making it possible to set the `parameter_style:CamelCase` will not only be backwards compatible, but also much more usable while still keeping the possibility open to actually use the regex.

hzeller (changes requested):
I think we need to make sure to be able to use the old configuration value with the named values.

verilog/analysis/checkers/parameter_name_style_rule.cc
sconwayaus: Do you have a preference on how it should handle if both `localparam_style` and `localparam_style_regex` are set with valid values? 

Maybe have `ParameterNameStyleRule::Configure` return an `absl::InvalidArgumentError()` noting the user should use one or the other.

verilog/analysis/checkers/parameter_name_style_rule.cc
sconwayaus: No worries, I'll fix that up in the open PR's. Would you prefer a single or multiple follow up PR's to fix the others that sneaked through?

hzeller: One change that touches all is fine.

verilog/analysis/checkers/parameter_name_style_rule.cc
hzeller: Ah, yes, the OR-ing part, I didn't think of it: yes that should then assemble a `(foo)|(bar)` kindof regex.

But with OR-ing in place, we can assemble all the regex from `localparam_style` with  the `localparam_style_regex` if both of them are given in the lint configuration.

Probably good to have a test for all these situations; for that, you could have public functions that return the `const RE2 *` for each regex
```c++
const RE2 *localparam_style_regex() const { return localparam_style_regex_.get(); }
```
, so that it is easy to write a test that verifies that the returned value is `re != nullptr` and the `re->pattern()` matches the expectation after having called `Configuration()`.

sconwayaus:
Ok, I think I've addressed all issues, with the regex configuration now extending the existing configuration in a non-breaking way.

verilog/analysis/checkers/parameter_name_style_rule.cc
sconwayaus: No problem, see PR #2236

hzeller:
Oh, got distracted. Will do another review today.

hzeller:
Thanks! Sorry for the delay, sometimes life is just busy...

PR #2230, opened 2024-08-05
Add variable shadowing rule.

### LINKED ISSUE #247: Flag declarations that shadow other declared names

hzeller:
This is a rebase of #790 to compile with head. Fixes #247


hzeller:
Continuation of #2133 here (accidentally deleted the old branch)

common/analysis/lint_rule_status.cc
```diff
@@ -76,12 +79,14 @@ static TokenInfo SymbolToToken(const Symbol& root) {
 
 LintViolation::LintViolation(const Symbol& root, absl::string_view reason,
                              const SyntaxTreeContext& context,
-                             const std::vector<AutoFix>& autofixes)
+                             const std::vector<AutoFix>& autofixes,
+                             const std::vector<TokenInfo>& tokens)
```
fangism, line 83: nit: name this `related_tokens` for consistency

common/analysis/lint_rule_status.h
```diff
@@ -92,23 +92,40 @@ class AutoFix {
 struct LintViolation {
   // This construct records a token stream lint violation.
   LintViolation(const TokenInfo& token, absl::string_view reason,
-                const std::vector<AutoFix>& autofixes = {})
-      : token(token), reason(reason), context(), autofixes(autofixes) {}
+                const std::vector<AutoFix>& autofixes = {},
+                const std::vector<TokenInfo>& tokens = {})
```
fangism, line 96: same, `related_tokens`

common/analysis/lint_rule_status.h
```diff
@@ -127,6 +144,10 @@ struct LintViolation {
 
   const std::vector<AutoFix> autofixes;
 
+  // Additional tokens that are related somehow to
+  // vulnerable token
```
fangism, line 148: would "offending" token be more descriptive?

fangism (approved):
This is an awesome lint rule!  This will help steer clear of esoteric naming/scoping rules in SV.

common/analysis/lint_rule_status.h
hzeller: done.

common/analysis/lint_rule_status.cc
hzeller: done.

common/analysis/lint_rule_status.h
hzeller: sounds good. changed.

PR #2236, opened 2024-08-10
Fixing naming style for constant expressions

sconwayaus:
As discussed in #2213, fixing the naming style for a few constants.


hzeller:
Thanks!

PR #2242, opened 2024-08-24
Fix trailing whitespace.

hzeller:


hzeller:
@cdleary Looks like the Verible CI does not check for trailing whitespace in yml files...

cdleary:
> @cdleary Looks like the Verible CI does not check for trailing whitespace in yml files...

Ah sorry about that, I guess shows I don't edit yaml a lot 😬

hzeller:
It's a good thing :) Nobody should have to edit yaml more than absolutely necessary.

PR #2247, opened 2024-09-04
Rule/explicit begin

### LINKED ISSUE #2040: How to use rule for having a begin and end
**Summary**

I want to use the rule for checking if there are any if-else or for or while with single statements but doesn't have begin-end.

Which rules should I use, as I don't see them in the rules list.

**Proposal**

[optional] Suggest approach or include pseudo code for how one might implement
the rule.

**Additional context**

https://chipsalliance.github.io/verible/lint.html

sconwayaus:
This PR forks the initial work started in PR #1336 and extends the rule to check for begin-end blocks for all if, else, always, always_comb, always_latch, always_ff, for, forever, foreach, while and initial statements.

The rule is configurable allowing users to enable/disable this rule for each statement to provide some level of flexibility.

Aside from the usual unit testing, I've also spent the last 3 months using this rule in anger and it seems to be working quite well. 

Resolves #2040
Related to #1321, misses on the case statement suggestion.


verilog/tools/lint/testdata/always_comb_blocking.sv
```diff
@@ -1,4 +1,5 @@
 module always_comb_blocking;
-  always_comb
+  always_comb begin
```
hzeller, line 2: given that the default is not enabled, do we need these changes in `testdata/*.sv` at this point ?

hzeller (commented):
Very cool!
I need to look at it more in detail later this week.

hzeller:
Adding @fangism as he always enjoys looking at new rules.

verilog/tools/lint/testdata/always_comb_blocking.sv
sconwayaus: There's also an -allrules_test that all these need to pass.

hzeller (approved):
Thanks, this is really cool!

hzeller:
Looks like the automatic generation of the rule documentation picked this up https://chipsalliance.github.io/verible/lint.html#explicit-begin

PR #2260, opened 2024-09-27
Use single-header json implementation.

hzeller:
clang-tidy complains if a symbol is not in a header directly included with the [misc-include-cleaner] check. Since the multi-header does one indirection, this results in these warnings.

Using the single-header library helps. Patch upostream bzlmod package to provide that.

TODO: check with bzlmod mintainer for nlohmann_json to maybe provide a target for that.


hzeller:
@lromor FYI

PR #2263, opened 2024-09-28
Update README.md

scutwengxinqian:
vim.loop is deprecated. Use vim.uv instead.


hzeller:
Thanks! Can't verify, but looks plausible.

PR #2268, opened 2024-10-01
(experiment): distinguish singleheader json in toplevel jsonhpp BUILD

hzeller:
Sketch how this can be done in a way that

  * We can make the same layout for projects supporting both, WORKSPACE and modules. In WORKSPACE, only the toplevel BUILD file can be provided to an external project. Thus we could not make a sub-package without an awkward patch.
  * Still is easy to read as the singleheader-name reflects what it is for, but `json` is also its own separate name.
  * Two simple targets: `:json` and` :singleheader/json`

### DIFF
--- a/.github/workflows/verible-ci.yml
+++ b/.github/workflows/verible-ci.yml
@@ -127,7 +127,7 @@ jobs:
       - name: Get Bant
         run: |
           # TODO: provide this as action where we simply say with version=...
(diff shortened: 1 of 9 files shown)


hzeller:
I guess I need to teach [`bant`](https://github.com/hzeller/bant) that these are valid target labels :)

Issue #2171, opened 2024-04-30
automatically comment preprocessor conditionals


fangism:
**Summary**

This request captures features from a script that lived outside of Verible, but some developers found useful.
The script (by @jonmayergoogle ) would automatically add comments after `ifdef` `elsif` `else` `endif` directives that reflected the condition under which the section was active, and served as a readability aid.

**Test cases and examples**

```systemverilog
/* Hello, world. */
alpha
`ifdef FOO
/*
beta
`ifndef BAR
*/
luke
`ifdef BUM  // is BUM defined
wookie
`endif  // this comment gets removed
r2d2
`endif
wedge
```

would automatically become annotated like:

```systemverilog
/* Hello, world. */
alpha
`ifdef FOO  // FOO
/*
beta
`ifndef BAR
*/
luke
`ifdef BUM  // FOO && BUM
wookie
`endif  // FOO && BUM
r2d2
`endif  // FOO
wedge
```

**Proposal**

The implementation for this would be based on an unpreprocessed token stream.  No syntactic validation is needed, as this is purely a transformation based on preprocessing tokens.  The simplicity of this task means that we don't need a full syntax tree transformation/rewriting framework.

**Additional context**

cl/629042737 (Google)

kuku929:
hello, this seems interesting. In which file should I put the code for the feature?

Bhavin-umatiya:
@fangism Hello
please check any further change so u can guide me i'll try

fangism:
> @fangism Hello please check any further change so u can guide me i'll try

Got your message.  I'm a little occupied over the holidays but will try to take a look at https://github.com/chipsalliance/verible/pull/2314.

PR #2314, opened 2024-12-25
fix: #2171

Bhavin-umatiya:
fix issue: #2171

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,39 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "(gdb) Attach",
+            "type": "cppdbg",
+            "request": "attach",
+            "program": "enter program name, for example ${workspaceFolder}/a.exe",
+            "MIMode": "gdb",
+            "miDebuggerPath": "/path/to/gdb",
+            "setupCommands": [
+                {
+                    "description": "Enable pretty-printing for gdb",
+                    "text": "-enable-pretty-printing",
+                    "ignoreFailures": true
+                },
+                {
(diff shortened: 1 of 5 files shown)


verible/verilog/analysis/descriptions.h
```diff
@@ -35,14 +35,16 @@ struct LintConfigParameterDescriptor {
...
-  absl::string_view topic;  // section in style-guide
-  std::string desc;         // Detailed description.
-  std::vector<LintConfigParameterDescriptor> param;
-};
-
-}  // namespace analysis
-}  // namespace verilog
-
-#endif  // VERIBLE_VERILOG_ANALYSIS_DESCRIPTIONS_H_
+std::string format_long_description(const std::string& description) {
```
fangism, line 38: Is this new function used anywhere?  If so, you may declare it in a .h file, but define it in the corresponding .cc file.

.vscode/launch.json
fangism, line 1: This file doesn't look like it should be added in this PR.
(I'm not familiar with VSCode and how this should work.  Perhaps someone else can comment?)

verible/verilog/preprocessor/verilog-preprocess.cc
fangism, line 1: Are you missing the change to the `verilog-preprocess.h` that includes the new `condition_stack_` class member and `AnnotatePreprocessorDirectives` method?

verible/verilog/preprocessor/verilog-preprocess_test.cc
```diff
@@ -1035,3 +1035,29 @@ TEST(VerilogPreprocessTest,
 
 }  // namespace
 }  // namespace verilog
+
+TEST(PreprocessorAnnotationTest, AnnotatesConditionals) {
```
fangism, line 1039: Nice test!
I suggest adding tests to cover `ifndef` and `elsif` as well.

fangism (commented):
Great start, thank you for working on this!

hzeller:
What is the status of this ? Looks like you were on a good track, just some lines got accidentally deleted.

Issue #2323, opened 2025-01-12
A lot of code assumes that `string_view::iterator` is `const char*`. Does not work on every compiler


hzeller:
A lot of code within Verible assumes that the iterator of `string_view` just decays to `const char*`. This assumption was true while `absl::string_view` was the implementation used, and even after switching to c++17, is as well true for gcc/clang standard library std::string_view.

However, on Windows, the iterator is its own wrapper object that can not be assigned to `const char*`, so Verible would fail to compile if we used that `std::string_view` there.

Thus, for a long while, we have a [patch](https://github.com/chipsalliance/verible/blob/master/bazel/absl.patch) to tell abseil to always use the internal `absl::string_view` internal implementation on Windows and not just typedef `absl::string_view` to `std::string_view`.

However, due to that patch, our abseil-cpp in MODULES.bazel deviates from others users that depend on [Verible in the registry](https://registry.bazel.build/modules/verible). Not sure if this will be a problem, but it is an additional reason to get this fixed.

It would be good to remove these assumptions. This will probably involve using `absl::string_view::iterator` instead of `const char*` for parameters and data members where the intent is to work with iterators.

Also construction of `std::string_view`'s that use iterators as start point will be different.

```c++
   absl::string_view foo0(bar.begin(), 42);  // before
   absl::string_view foo1(&*bar.begin(). 42);  // after (somewhat ugly)
   absl::string_view foo2(bar.data(), 42);  // slightly less ugly, but only works if we know it's begin.
```

This is a collection issue to collect all the pull requests towards that goal until everything is done.

Exit criteria: we can replace all `absl::string_view` with `std::string_view` and code compiles on all platforms.

hzeller:
Done back at #2324

PR #2324, opened 2025-01-12
Don't assume string_view::iterator is const char*

hzeller:
There were a few assumptions in the code that these types are equivalent, but they are not on all platforms.

The fixes sometimes involve ugly `&*some_iterator` constructs to get to the underlying location. Some of these should be re-considered after switching to c++20.

Issues: #2323


verible/common/formatting/format-token.h
```diff
@@ -28,6 +28,13 @@
 
 namespace verible {
 
+// TODO: find something platform independent that is less ugly.
+// Or maybe revisit the place where we need this and see if they could
+// be actually represented by a const char * nullptr.
+inline absl::string_view::iterator string_view_null_iterator() {
```
hzeller, line 34: this should probably in a different file; or in no file at all, just using `absl::string_view::iterator{}` everywhere (but that would then be less explicit).

fangism: maybe place in `common/strings`?

fangism (approved):
I agree with the general direction of this clean-up.

verible/common/formatting/format-token.h
hzeller: For now, I narrowed its use to anything that anyway uses `format-token.h`, so decided to leave it there for now.
(otherwise it would be a bit awkward to have a `strings/null-iterator.h` with a single inlined one-liner function in it...)

PR #2326, opened 2025-01-14
Use std::string_view instead of absl::string_view.

hzeller:
Now that we're compatible with all systems after
PR #2324, can switch to use std::string_view now.


fangism (approved):
At long last, well done.

Issue #2372, opened 2025-03-28
Verible fails to detect invalid UTF-8 (e.g., incomplete Chinese characters) and throws exception


wxllllll:
**Describe the bug**

Verible fails to detect invalid UTF-8 sequences (e.g., truncated Chinese characters) as lexical errors.
Instead of rejecting the malformed text safely, it attempts to serialize it into JSON, causing an exception and crash.

**To Reproduce**

```
// This file includes an incomplete UTF-8 sequence (e.g., partial Chinese character)
// Simulate a scenario where a Chinese character is not fully written:
// For example, a single byte like \xE4 is written alone

module top;
  你
endmodule
```

**Actual behavior:**

Verible does not emit a proper lexical error for the malformed UTF-8 character.
During JSON-RPC notification construction, Verible crashes due to an exception from nlohmann::json.

Exception message:
json-rpc-dispatcher.cc :82 Notification error for 'textDocument/didOpen' :
[json.exception.type_error.316] invalid UTF-8 byte at index 19: 0x22

**Expected behavior**

Verible should detect the malformed UTF-8 sequence as an illegal token and report it as a proper diagnostic.
No exception or crash should occur during JSON serialization.
textDocument/didOpen  and textDocument/didChanged notifications with invalid UTF-8 in the source file should return diagnostics, not internal errors.

wxllllll:
After debugging, I found that the issue was caused by RejectChar truncating multi-byte illegal characters, which led to the exception when converting JSON to string. I have made the necessary changes, and this resolves the issue.

PR #2374, opened 2025-03-28
fix(verilog/parser):fix the truncation bug of rejectChar #2372

wxllllll:
After debugging, I found that the issue was caused by RejectChar truncating multi-byte illegal characters, which led to the exception when converting JSON to string. I have made the necessary changes, and this resolves issue https://github.com/chipsalliance/verible/issues/2372.


hzeller:
Thanks!

PR #2390, opened 2025-04-11
Re-enable Kythe xRefs publishing

ivan444:
Bump the Kythe version for the CI. Enable the xRefs action and use newer Ubuntu

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -10,23 +10,24 @@ build --enable_platform_specific_config
 build --workspace_status_command="bash bazel/build-version.sh"
 
 # Systems with gcc or clang
-common:linux   --cxxopt=-xc++ --host_cxxopt=-xc++ --cxxopt=-std=c++17 --host_cxxopt=-std=c++17 --client_env=BAZEL_CXXOPTS=-std=c++17
-common:freebsd --cxxopt=-xc++ --host_cxxopt=-xc++ --cxxopt=-std=c++17 --host_cxxopt=-std=c++17 --client_env=BAZEL_CXXOPTS=-std=c++17 --linkopt=-lm --host_linkopt=-lm
-common:openbsd --cxxopt=-xc++ --host_cxxopt=-xc++ --cxxopt=-std=c++17 --host_cxxopt=-std=c++17 --client_env=BAZEL_CXXOPTS=-std=c++17 --linkopt=-lm --host_linkopt=-lm
-common:macos   --cxxopt=-xc++ --host_cxxopt=-xc++ --cxxopt=-std=c++17 --host_cxxopt=-std=c++17 --client_env=BAZEL_CXXOPTS=-std=c++17
+common:unix    --cxxopt=-xc++ --host_cxxopt=-xc++ --cxxopt=-std=c++17 --host_cxxopt=-std=c++17 --client_env=BAZEL_CXXOPTS=-std=c++17
+common:linux   --config=unix
+common:freebsd --config=unix --linkopt=-lm --host_linkopt=-lm
+common:openbsd --config=unix --linkopt=-lm --host_linkopt=-lm
+common:macos   --config=unix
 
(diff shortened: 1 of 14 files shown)


.github/bin/install-kythe.sh
```diff
@@ -42,5 +42,9 @@ else
     wget --no-verbose -O kythe.tar.gz \
       "https://github.com/kythe/kythe/releases/download/$KYTHE_VERSION/kythe-$KYTHE_VERSION.tar.gz"
     tar -xzf kythe.tar.gz
+    # Bazel renamed @com_google_protobuf to @protobuf while Kythe is still
+    # using it in their BUILD targets. Manually rename until the fixed Kythe is
+    # released.
+    sed -i 's/@com_google_protobuf/@protobuf/' "kythe-${KYTHE_VERSION}/BUILD"
```
hzeller, line 48: there is also repo-mapping, maybe it can work with that ? https://bazel.build/rules/lib/repo/http#http_archive-repo_mapping

ivan444: I thought of that. We used it in past. But verible is now using the Bazel modules & I don't see a way how to do remapping with the modules.

hzeller:
the problem with clang-tidy is because I don't deal with the forwarding of options yet in .bazelrc, thus the `-xc++` flag is not properly seen. Let me see what I can do.

.bazelrc
```diff
@@ -25,9 +30,6 @@ common:macos --features=-supports_dynamic_linker --linkopt=-framework --linkopt=
 # Since we use designated initializers (C99 ... but >= c++20), compiler really wants the c++20 mode
 common:windows --compiler=clang-cl --cxxopt=/std:c++20 --host_cxxopt=/std:c++20 --client_env=BAZEL_CXXOPTS=/std:c++20
 
-build --cxxopt="-Wno-unknown-warning-option" --host_cxxopt="-Wno-unknown-warning-option"
-# TODO: this looks like benign where it happens but to be explored further
-build --cxxopt="-Wno-dangling-reference" --host_cxxopt="-Wno-dangling-reference"
 # Newer bisons create an unused label.
 build --cxxopt="-Wno-unused-label" --host_cxxopt="-Wno-unused-label"
```
hzeller, line 34: Looking at the current windows CI, this `-Wno-unused-label` also needs to go into our unix-specific config.

hzeller:
Hopefully with merging head, the clang-tidy issues should disappear (https://github.com/chipsalliance/verible/pull/2397)

MODULE.bazel
```diff
@@ -17,9 +17,20 @@ bazel_dep(name = "rules_bison", version = "0.3")
...
 bazel_dep(name = "googletest", version = "1.14.0.bcr.1", dev_dependency = True)
 
 # To build compilation DB and run build-cleaning
 bazel_dep(name = "bant", version = "0.2.0", dev_dependency = True)
+
+# Make it possible to use rules_python as root (needed for the CI). See
+# https://github.com/bazel-contrib/rules_python/pull/713 for more details.
+bazel_dep(name = "rules_python", version = "1.3.0")
+
+python = use_extension("@rules_python//python/extensions:python.bzl", "python")
```
hzeller, line 31: Where do we need Python, and is there a way to avoid it ?

hzeller:
I am fixing the smoke tests separately.

.bazelrc
ivan444: Yep. Added the remaining ones. The windows build is still failing, but on different issues.

MODULE.bazel
ivan444: Some rule from Kythe extractors.bzl needed it. But I rolled out the extractors.bzl and took only what's needed to extract Verible. Now it works.

hzeller:
Smoke tests fixed in #2399

PR #2411, opened 2025-05-14
Handle invalid format range in LSP adapter and add test

henryhchchc:
Prevent a potential crash when receiving an invalid range formatting parameter.


hzeller:
Thanks!

PR #2426, opened 2025-06-03
Provide an 'auto' option for --line_terminator

hzeller:
If 'auto' is chosen, output CRLF line endings if more than 50% of the input is CRLF, otherwise LF.

The other choicese, CRLF or LF, behave as before.

NB: The default changes with this change: previously, we always converted a CRLF input to LF, now the default is 'auto'


verible/verilog/formatting/formatter.cc
```diff
@@ -970,6 +973,22 @@ Status Formatter::Format(const ExecutionControl &control) {
   return absl::OkStatus();
 }
 
+// From options, extraact the line terminator style. If 'auto' was chosen,
```
IEncinas10, line 976:
```suggestion
// From options, extract the line terminator style. If 'auto' was chosen,
```

Typo :)

verible/verilog/tools/formatter/README.md
```diff
@@ -23,6 +23,9 @@ To pipe from stdin, use '-' as <file>.
       default: 2;
     --line_break_penalty (Penalty added to solution for each introduced line
       break.); default: 2;
+    --line_terminator (Line terminator. The 'auto' option chooses the output
```
IEncinas10, line 26: Missed this one on my PR. Thanks for taking care of it!

IEncinas10 (approved):
Thanks for the improvement Henner! I don't have any significant comment, LGTM :)

verible/verilog/formatting/formatter.cc
hzeller: Thanks for spotting the typo. Fixed.

hzeller:
> I still need to write a test for `GuessLineTerminator()`, but the rest can already be reviewed.

Also added that test now.
Will merge once CI finishes green.

PR #2452, opened 2025-08-29
Update bison.bzl and flex.bzl to work with Bazel 8

mneilly:
This change modifies the bison.bzl and flex.bzl files to accommodate changes in Bazel 8 that disallow use of select in the toolchain attribute. Instead of select, multiple targets are created and the resulting files (unique per target) are copied to the expected location.

If anybody is aware of a better approach feel free to chime in.


hzeller:
Don't worry about the smoke tests, I'll take care of that.

The windows build looks like it _almost_ works, but there is a mysterious `.windows`-suffix on the generated files.

bazel/flex.bzl
```diff
@@ -20,21 +20,49 @@
...
+        outs = [out + ".local"],
+        cmd = "flex --outfile=$(location " + out + ".local) $<",
+        tags = ["manual"],
+    )
+
+    # For Windows case
+    native.genrule(
+        name = name + "_windows",
+        srcs = [src],
+        outs = [out + ".windows"],
```
hzeller, line 53: ah, the suffix is not mysterious anymore. But looks like it messes with the rest of the build-system

mneilly:
I'll take a look at the windows issue tomorrow. It shouldn't be difficult to resolve (I hope).

hzeller:
I am currently working on two possible solutions
  * https://github.com/chipsalliance/verible/pull/2520 move involved with local/non-local rules. Works, but looks fragile
  * https://github.com/chipsalliance/verible/pull/2524 use bison and flex compilation directly without the toolchain; they made good strides towards compilable on Windows, just m4 is a little hold-out, hopefully fixed soon. Makes the whole *.bzl rules a lot easier.

mneilly:
FWIW - Ended up getting rid of the split generators and using tools instead of toolchains. Much simpler and easier to follow. Also updated the M4 var ref.

hzeller:
Things are fixed in the meantime, closing this PR.
(Thanks anyway!)

PR #2472, opened 2025-11-25
Add TextMate Grammar for Verilog/SystemVerilog Highlighting

clysto:
This PR adds a TextMate grammar to the Verible VS Code extension to provide proper syntax highlighting for SystemVerilog files.
The grammar is adapted from [vscode-verilog-hdl-support](https://github.com/mshr-h/vscode-verilog-hdl-support)￼, with minimal adjustments to integrate cleanly into the existing extension structure.

Before this change, the extension lacked built-in syntax highlighting and relied solely on LSP diagnostics. With the added grammar, users now get immediate and consistent token coloring without installing an additional extension.

### DIFF
--- a/verible/verilog/tools/ls/vscode/package.json
+++ b/verible/verilog/tools/ls/vscode/package.json
@@ -14,6 +14,13 @@
   },
   "icon": "images/icon.png",
   "contributes": {
+    "grammars": [
+      {
+        "language": "systemverilog",
+        "scopeName": "source.systemverilog",
+        "path": "./syntaxes/systemverilog.tmLanguage.json"
+      }
+    ],
     "languages": [
       {
         "id": "systemverilog",
--- a/verible/verilog/tools/ls/vscode/syntaxes/systemverilog.tmLanguage.json
+++ b/verible/verilog/tools/ls/vscode/syntaxes/systemverilog.tmLanguage.json
@@ -1,0 +1,1416 @@
+{
+  "name": "SystemVerilog",
+  "scopeName": "source.systemverilog",
+  "fileTypes": [
+    "v",
+    "sv"
+  ],
+  "patterns": [
+    {
+      "include": "#comments"
+    },
+    {
+      "include": "#strings"
+    },
+    {
+      "include": "#typedef-enum-struct-union"
+    },
+    {
+      "include": "#typedef"
+    },
+    {
+      "include": "#functions"
+    },
+    {
+      "include": "#keywords"
+    },
+    {
+      "include": "#tables"
+    },
+    {
+      "include": "#function-task"
+    },
+    {
+      "include": "#module-declaration"
+    },
+    {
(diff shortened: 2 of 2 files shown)


clysto:
Any suggestions?

hzeller:
I like the idea, but I am wondering how it interferes is someone already has a system verilog plugin installed ? If so, their coloring should precedence as the Verible plugin would only do that as secondary functionality.
Do you know how these things are configured ?

clysto:
Multiple extensions can contribute grammars. If several installed extensions provide a TextMate grammar for the same language (e.g., SystemVerilog), VS Code will pick one based on the language id and the extension’s contributions. There isn’t a user-visible “precedence” setting exposed in the UI—you normally see one set of highlight rules applied. Extensions can override or replace existing grammars by contributing a grammar under the same language id. ￼

If someone already has a SystemVerilog plugin with its own grammar installed, VS Code will pick one grammar to apply. There isn’t a native per-extension “precedence” setting for TextMate grammars; the editor resolves this based on the active language id/grammar contributions. If desired, users can override the grammar selection manually in the editor for their workspace.

hzeller:
mmh, this does sound it can result in surprising effects for users (in particular existing users of the Verible plug-in that I expect typically already have another plug-in installed for the syntax-highlighting).

Could there be a configuration option that we can offer in the Verible plug-in to use or not use the grammar ? The VSCode options pane allows for configuration options that are [provided in the package.json](https://github.com/chipsalliance/verible/blob/67f7038305d628fce9f6420772bf0365c0276f1e/verible/verilog/tools/ls/vscode/package.json#L31-L50) and I wonder if we can have a 'checkbox' configured in the plugin json to then toggle if the grammar is active ? (we can have it default on, but at least it would allow people to switch off things)

RedCMD:
VSCode provides no way to configure TextMate grammars
the only options are to disable/enable the extension 
and staticly via the alphabetical ordering of extension names `publisher.name`
Z is higher than A

you could add a different language id
to separate them
but then you'll have separate languages

> users of the Verible plug-in that I expect typically already have another plug-in installed for the syntax-highlighting

are the grammars not the same?

clysto:
The syntax highlighting is contributed via package.json using contributes.grammars, which is a static registration mechanism in VSCode and cannot be reliably toggled on or off via an extension setting at runtime. This is therefore a limitation of the VSCode extension model rather than a missing option on our side. More generally, it is quite common for language extensions to ship their own syntax highlighting together with the LSP server (e.g. clangd, rust-analyzer, pylance), so that highlighting stays consistent with formatting, linting, and diagnostics. In this context, we believe Verible should act as the primary SystemVerilog provider when installed, and disabling other SystemVerilog syntax plugins avoids conflicts; additionally, this also leaves room for future improvements such as richer, LSP-driven semantic highlighting based on Verible’s analysis.

PR #2487, opened 2026-01-18
Add inlay hint support for Verilog module instantiations

ya-uhs:
## Summary

Add LSP inlay hint support to display port direction (input/output/inout) and type information for named port connections in module instantiations.

### Example

For code like:
```verilog
module child(input clk, input [7:0] data, output valid);
endmodule

module top;
  child inst (
    .clk(sys_clk),
    .data(bus_data),
    .valid(is_valid)
  );
endmodule
```

The editor will display hints like:

## Changes

- Add `inlay-hint.cc` and `inlay-hint.h` implementing `GenerateInlayHints()`
- Register `textDocument/inlayHint` handler in `verilog-language-server.cc`
- Add `inlayHintProvider` capability
- Add tests for inlay hint functionality

## Testing

- Added 3 test cases in `verilog-language-server_test.cc`:
  - `InlayHintEmptyWhenModuleNotFound` - returns empty when module not in symbol table
  - `InlayHintEmptyForSimpleModule` - returns empty when no instantiations
  - `InlayHintWithModuleDefinition` - returns hints with correct direction info

All existing tests pass.


verible/verilog/tools/ls/inlay-hint.cc
```diff
@@ -0,0 +1,258 @@
...
+            ExtractTypeString(port_info.declared_type.syntax_origin);
+        return hint;
+      }
+    }
+  }
+
+  return std::nullopt;
+}
+
+// Check if a position is within the requested range.
```
hzeller: This looks like it might be a useful function for other code in the LSP as well. So if something like this does not exist yet, it is probably implemented ad-hoc somwehere.
Can you add a `TODO: maybe extract as utility function`

verible/verilog/tools/ls/inlay-hint.cc
```diff
@@ -0,0 +1,258 @@
...
+
+// Information about a port that we want to show in an inlay hint.
+struct PortHintInfo {
+  std::string_view direction;  // "input", "output", "inout"
+  std::string type_string;     // e.g., "[7:0]", "logic [15:0]"
+};
+
+// Format a port hint label from direction and type info.
+std::string FormatPortHint(const PortHintInfo &info) {
+  std::string result;
```
hzeller: This whole function looks like it could be written as

```c++
  return absl::StrJoin({info.direction, info.type_string}, " ", absl::SkipEmpty());
```

verible/verilog/tools/ls/inlay-hint.cc
```diff
@@ -0,0 +1,258 @@
...
+
+  // The module's children contain its ports and other declarations.
+  for (const auto &child : module_node->Children()) {
+    if (child.first == port_name) {
+      const SymbolInfo &port_info = child.second.Value();
+      // Check if this is actually a port
+      if (port_info.metatype == SymbolMetaType::kDataNetVariableInstance) {
+        PortHintInfo hint;
+        hint.direction = port_info.declared_type.direction;
+        hint.type_string =
```
hzeller: here, you're assigning a string view the temporary string object of ExtractTypeString(), this will be problematic.
But the string view over the syntax origin will be available longer, so you can make ExtractTypeString() return a string_view.

verible/verilog/tools/ls/inlay-hint.cc
```diff
@@ -0,0 +1,258 @@
...
+    std::string_view port_name = token.text();
+
+    // Calculate the position for the inlay hint.
+    // Position it after the port name (before the opening parenthesis).
+    std::string_view base_text = text_structure_->Contents();
+    int offset = token.right(base_text);
+    verible::LineColumn line_col = text_structure_->GetLineColAtOffset(offset);
+
+    // Check if the position is within the requested range.
+    if (!IsInRange(line_col, params_.range)) return;
```
hzeller: this is a strong filter as it narrows to things we are really interested in. I wonder, can we already weed out early in the process all symbols that are not overlapping ?
I am worried about walking the whole AST every time an inlay hint is requested.

hzeller:
Very cool, thank you. Sorry for the delay, I was busy this year and forgot when I saw the pull request message pop up.

I have a few comments, some help making things more compact, one avoids a memory bug and maybe there is also something to filter early.

PR #2488, opened 2026-01-21
Added support for token concatenation

sifferman:
Hello!

This is a stepping stone to https://github.com/chipsalliance/verible/issues/2322. Once this is merged in, I can submit a PR for `` `include `` support.

Happy to make changes!

### DIFF
--- a/verible/verilog/preprocessor/verilog-preprocess.cc
+++ b/verible/verilog/preprocessor/verilog-preprocess.cc
@@ -328,6 +328,89 @@ void VerilogPreprocess::RegisterMacroDefinition(
   // TODO(hzeller): multiline warning with 'previously defined here' location
 }
 
+// Process token concatenation (`` operator) in a token sequence.
+// This function scans for PP_TOKEN_CONCAT operators and concatenates
+// adjacent tokens by combining their text and re-lexing the result.
+// Whitespace around `` is removed per SystemVerilog standard.
+// Chained concatenations (a``b``c) are processed left-to-right.
+// Concatenated strings are stored in preprocess_data for lifetime management.
+static void ProcessTokenConcatenation(verible::TokenSequence *tokens,
+                                      VerilogPreprocessData *preprocess_data) {
+  if (tokens->empty()) return;
+
+  verible::TokenSequence result;
+  result.reserve(tokens->size());
+
+  for (size_t i = 0; i < tokens->size(); ++i) {
+    const auto &curr_token = (*tokens)[i];
+
+    // Check if next non-space token is `` operator.
+    size_t next_idx = i + 1;
+    while (next_idx < tokens->size() &&
+           (*tokens)[next_idx].token_enum() == TK_SPACE) {
+      ++next_idx;
+    }
+
+    if (next_idx < tokens->size() &&
+        (*tokens)[next_idx].token_enum() == PP_TOKEN_CONCAT) {
+      // Handle chained concatenations (a``b``c) by accumulating text.
+      std::string accumulated(curr_token.text());
+
+      size_t scan_idx = next_idx;
+      while (scan_idx < tokens->size()) {
+        if ((*tokens)[scan_idx].token_enum() != PP_TOKEN_CONCAT) break;
+
+        // Skip to token after ``, skipping whitespace.
+        size_t after_concat_idx = scan_idx + 1;
+        while (after_concat_idx < tokens->size() &&
+               (*tokens)[after_concat_idx].token_enum() == TK_SPACE) {
+          ++after_concat_idx;
+        }
+
+        if (after_concat_idx >= tokens->size()) break;
+
+        accumulated += std::string((*tokens)[after_concat_idx].text());
+
+        // Check if there's another `` after this token.
+        scan_idx = after_concat_idx + 1;
+        while (scan_idx < tokens->size() &&
+               (*tokens)[scan_idx].token_enum() == TK_SPACE) {
+          ++scan_idx;
+        }
+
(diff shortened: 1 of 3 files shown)


hzeller:
Sorry for the delay in the review - I was pretty busy this year and didn't notice this in my in-box. Hopefully more time now.

Before I look in detail, can you line out the include support ? This sounds exciting (but also something that we might want to switch on/off depending on context, e.g. the formatter would not have to read the include)

verible/verilog/preprocessor/verilog-preprocess.cc
```diff
@@ -328,6 +329,91 @@ void VerilogPreprocess::RegisterMacroDefinition(
...
+    }
+
+    // Not part of concatenation - keep token unless it's PP_TOKEN_CONCAT
+    // itself.
+    if (curr_token.token_enum() != PP_TOKEN_CONCAT) {
+      result.push_back(curr_token);
+    }
+  }
+
+  *tokens = std::move(result);
```
fangism, line 414: Since you are assigning over `tokens`, effectively discarding them, you could also `swap` them.

Alternatively, you could retain the original tokens as `const&` and return a new result.  That would give the caller the option to preserve the original tokens.  (Default non-modification, optional mutability.)

verible/verilog/preprocessor/verilog-preprocess.cc
```diff
@@ -328,6 +329,91 @@ void VerilogPreprocess::RegisterMacroDefinition(
...
+  verible::TokenSequence result;
+  result.reserve(tokens->size());
+
+  for (size_t i = 0; i < tokens->size(); ++i) {
+    const auto &curr_token = (*tokens)[i];
+
+    // Check if next non-space token is `` operator.
+    size_t next_idx = i + 1;
+    while (next_idx < tokens->size() &&
+           (*tokens)[next_idx].token_enum() == TK_SPACE) {
```
fangism, line 351: Do we know if `TK_NEWLINE` counts as a space in this context?  (I don't actually know myself.)

verible/verilog/parser/verilog-token-classifications.h provides an `IsWhitespace` function that could be used if newlines should be treated like spaces.

verible/verilog/preprocessor/verilog-preprocess.cc
```diff
@@ -328,6 +329,91 @@ void VerilogPreprocess::RegisterMacroDefinition(
...
+        // Skip to token after ``, skipping whitespace.
+        size_t after_concat_idx = scan_idx + 1;
+        while (after_concat_idx < tokens->size() &&
+               (*tokens)[after_concat_idx].token_enum() == TK_SPACE) {
+          ++after_concat_idx;
+        }
+
+        if (after_concat_idx >= tokens->size()) break;
+
+        accumulated += std::string((*tokens)[after_concat_idx].text());
```
fangism, line 373: I'm concerned about performance of repeated conversions to `std::string` in this loop in the accumulation process.  I recommend `absl::StrJoin` on a container (`vector` or SmallVector) of `string_view`.  It computes the size needed and performs a single allocation at once.

verible/verilog/preprocessor/verilog-preprocess.cc
```diff
@@ -328,6 +329,91 @@ void VerilogPreprocess::RegisterMacroDefinition(
...
+          break;
+        }
+      }
+
+      // Store the accumulated string in persistent storage.
+      preprocess_data->concatenated_strings.push_back(accumulated);
+      const std::string &concatenated =
+          preprocess_data->concatenated_strings.back();
+
+      // Re-lex the concatenated text to get the proper token type.
```
fangism, line 394: This might be a nice function to factor out, like `LexOneToken()`.

verible/verilog/preprocessor/verilog-preprocess.h
```diff
@@ -76,6 +77,11 @@ struct VerilogPreprocessData {
   verible::TokenStreamView preprocessed_token_stream;
   std::vector<TokenSequence> lexed_macros_backup;
 
+  // Storage for concatenated token text (from `` operator).
+  // These strings must persist for the lifetime of tokens that reference them.
+  // Using deque because it doesn't invalidate references when adding elements.
+  std::deque<std::string> concatenated_strings;
```
fangism, line 83: Perhaps in the future, consider retaining original source-offset ranges that can be used in diagnostics that involve concatenated tokens, e.g "FOOBAR, expanded by fizz``buzz at ...".

verible/verilog/preprocessor/verilog-preprocess_test.cc
```diff
@@ -1033,5 +1033,359 @@ TEST(VerilogPreprocessTest,
...
+    EXPECT_TRUE(equivalent.Status().ok())
+        << equivalent.Status() << " " << test.description;
+    const auto &expanded_stream = expanded.Data().GetTokenStreamView();
+    const auto &equivalent_stream = equivalent.Data().GetTokenStreamView();
+    EXPECT_GT(expanded_stream.size(), 0) << test.description;
+    EXPECT_EQ(expanded_stream.size(), equivalent_stream.size())
+        << test.description;
+  }
+}
+
```
fangism, line 1161: What happens if you concatenate `123` with `abc`?  Does that form an invalid identifier `123abc`?  Can we cover this scenario in a test?

fangism (changes requested):
This is very nice, thank you for tackling this work!

sifferman:
Just got back from vacation. Thanks for the feedback! I'll take a look within the next week or so

PR #2497, opened 2026-03-17
Modernize use of bison and flex in a more elaborate rule.

hzeller:
Issues #2435


hzeller:
meh, flex not compilable under Windows.

PR #2498, opened 2026-03-21
fix issue 2288: add --rules=+generate-label-prefix=prefix:A_ to change the style_regex

### LINKED ISSUE #2288: how to change lint rule "legacy-generate-region"?
**Summary**
//Brief description of request. Reference `[style-lint-rule-name]` if this is
//about an existing rule, or suggest new lint rule name.
hello:
in rule file https://chipsalliance.github.io/verible/lint.html#generate-label-prefix , there is no parameter to user change the  `style_regex`.

May I ask whether modification is not supported or I have not found a way to modify

ddppt-yy:
in issues https://github.com/chipsalliance/verible/issues/2288, 
I proposed making the generate labels parameterized. Presumably, everyone was too busy to optimize this issue, so I have attempted to fix the problem.

In this issue, you also agreed that this was a good idea.

Currently, **g_** remains the default option, but users can customize it. 
For example:
`verible-verilog-lint --rules=+generate-label-prefix=prefix:A_ ./sv.sv`

Fixes #2288

### DIFF
--- a/verible/verilog/analysis/checkers/BUILD
+++ b/verible/verilog/analysis/checkers/BUILD
@@ -816,6 +816,7 @@ cc_library(
         "//verible/common/analysis:syntax-tree-lint-rule",
         "//verible/common/analysis/matcher",
         "//verible/common/analysis/matcher:bound-symbol-manager",
+        "//verible/common/text:config-utils",
         "//verible/common/text:symbol",
         "//verible/common/text:syntax-tree-context",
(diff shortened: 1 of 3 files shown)


hzeller:
Thanks for your contribution.
Can you have a look at the test failures ?

ddppt-yy:
> Thanks for your contribution. Can you have a look at the test failures ?

ok i will do it.

Issue #2481, opened 2025-12-18
assignment blocks exceeds the column limit,alignment will fail


LinYu369:
**vscode parameter use**

```json
  "verible.arguments": [
    "--column_limit=400",
    "--rules=+line-length=length:600,-parameter-name-style,-explicit-parameter-storage-type,-always-comb,-case-missing-default,-unpacked-dimensions-range-ordering,-no-tabs,-no-trailing-spaces,-posix-eof",
    "--assignment_statement_alignment=align",
    "--class_member_variable_alignment=align",
    "--named_parameter_alignment=align",
    "--named_port_alignment=align",
    "--port_declarations_alignment=align",
    "--module_net_variable_alignment=align",
    "--try_wrap_long_lines=true",
    "--variables_in_outline=false"
  ],
```

**Actual output**

```systemverilog

module eth_top (
    input wire clk,   // clk
    input wire rst_n  // rst
);
  wire [  7:0] rx_crc;  // crc checksum
  wire         ins_rx_crc_ok;  // crc checksum ok
  wire         ins_machine_code_ok;
  wire         data_machine_code_ok;
  wire         ins_initialize_ctrl;
  wire         ins_asyn_rest_ctrl;
  wire         ins_set_ip_ctrl;
  wire         ins_set_mac_ctrl;
  reg  [127:0] ins_rx_data_buff;
  reg          udp_get_ins_state;
  reg          udp_rx_ram_rd_data;

  //parameter define
  parameter FPGA_VERSION = 32'h01_00_00_02;  // fpga ver
  parameter INSTRUCTION_MESSAGE_LENGTH = 16;
  parameter DATA_MESSAGE_DATA_LENGTH = 1024;
  parameter MACHINE_CODE = 16'h01;

  parameter UDP_INS_INITIALIZE = 16'h0000;
  parameter UDP_INS_ASYN_REST = 16'h0001;
  parameter UDP_INS_SET_IP = 16'h0002;
  parameter UDP_INS_SET_MAC = 16'h0003;


  always @(posedge clk or negedge rst_n) begin
    if (!rst_n) begin
      ins_rx_data_buff <= 128'd0;
    end else if (udp_get_ins_state == 1) begin
      ins_rx_data_buff <= {ins_rx_data_buff[119:0], udp_rx_ram_rd_data};
    end else begin
      ins_rx_data_buff <= ins_rx_data_buff;
    end
  end

[...]

Issue #2021, opened 2023-09-29
Assign alignment breaks when line length would go over column limit


shareefj:
**Test case**

```systemverilog
module test ();
  assign short_sig              = 1'b1;
  assign a_really_long_sig_name = 1'b1;
  assign another_long_sig       = short_sig || a_really_long_sig_name;
  // assign mumbo_jumbo = short_sig || a_really_long_sig_name || yet_another_sig;
endmodule
```

I'm using VSCode and trying to use the Verible language server extension. I've got the following arguments setup in the settings for the extension:

```
--assignment_statement_alignment=align
--module_net_variable_alignment=align
--named_port_alignment=align
--column_limit=88
--try_wrap_long_lines=true
```

It looks like the formatter will align the assign statement as long as indenting them won't push them over the column width.  If I uncomment the line in the example above, then the formatter squashes all the alignment:

```systemverilog
module test ();
  assign short_sig = 1'b1;
  assign a_really_long_sig_name = 1'b1;
  assign another_long_sig = short_sig || a_really_long_sig_name;
  assign mumbo_jumbo = short_sig || a_really_long_sig_name || yet_another_sig;
endmodule
```

Is there any way to force the alignment here and accept the line length violation or force the tool to wrap the line?  It seems to wrap under some circumstances but not this.  This behaviour is hugely annoying and I'd rather just have a few lines extend over the line limit than all the formatting break.

It would also be good if when the tool wrapped long lines, it actually matched the alignment with what has gone before rather than just indent it by 4 spaces.  I'm not sure if that is a user config issue or has been reported elsewhere in the issues...

fangism:
Acknowledged and confirmed, the alignment logic is conservative at the moment, it that it won't expand past the column limit.  Extending this behavior remains to be specified: how to handle overflow/wrap.  Thanks for starting with some suggestions.

PR #2500, opened 2026-03-27
Fix module-level alignment breaking after non-alignable constructs

dannyoler:
## Summary

- `AlignModuleItems` was the only alignment function that did not split partitions by blank lines before classifying groups
- When a non-alignable construct (e.g. module instantiation, macro call) appeared in the module body, `GetConsecutiveModuleItemGroups` returned `kNoMatch`, permanently breaking alignment for all subsequent `assign` statements and declarations in that scope
- Added `GetModuleItemGroupsBetweenBlankLines` which first partitions by blank lines, then runs `GetConsecutiveModuleItemGroups` within each sub-range — matching the pattern used by every other alignment function

### Before

```systemverilog
    assign short_a            = 1;  // aligned
    assign really_long_name_b = 2;  // aligned
    inst #() u (.Z(z));             // breaks alignment permanently
    assign c = 3;                   // NOT aligned
    assign longer_d = 4;            // NOT aligned
```

### After

```systemverilog
    assign short_a            = 1;  // aligned (group 1)
    assign really_long_name_b = 2;  // aligned (group 1)
    inst #() u (.Z(z));             // ends group 1, does not affect group 2
    assign c        = 3;            // aligned (group 2)
    assign longer_d = 4;            // aligned (group 2)
```

Blank lines continue to separate alignment groups as expected. Also removes the resolved `TODO(b/161814377)` comment.

## Test plan

- [ ] Verified on a 7400-line production RTL file with module instantiations, `DFF` macros, and hundreds of `assign` statements
- [ ] Small test case confirms: instantiations don't break alignment, blank lines do break alignment
- [ ] Existing verible tests should pass (alignment functions for ports, structs, enums, etc. already use blank-line splitting)

Related issues: #2021, #2481

### DIFF
--- a/verible/verilog/formatting/align.cc
+++ b/verible/verilog/formatting/align.cc
@@ -1487,12 +1487,27 @@ static std::vector<AlignablePartitionGroup> AlignActualNamedPorts(
       &IgnoreWithinActualNamedPortPartitionGroup, full_range, vstyle);
 }
 
+// First partition by blank lines, then within each blank-line-separated
+// group, identify consecutive alignable items (declarations and continuous
+// assignments).  This ensures that blank lines break alignment groups
+// while non-alignable constructs (e.g. module instantiations) within a
+// blank-line group do not prevent subsequent items from being aligned.
+static std::vector<TaggedTokenPartitionRange>
+GetModuleItemGroupsBetweenBlankLines(const TokenPartitionRange &partitions) {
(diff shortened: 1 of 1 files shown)


hzeller:
Thanks for your contribution. Sorry for the delay, was swamped with other projects and missed the mail from github about the change.

Can you have a look at the tests and make them all pass ?
(you probably also have to rebase)

PR #2508, opened 2026-04-26
Add graphviz and tsort output formats to project file-deps

Johannes-Larsson:
Add a new argument `--output_format` with options `human` (default), `tsort`, and `dot` to verible-verilog-project. It only affects `file-deps`.

The various formats can surely be implemented in a neater way in dependencies.cc.

### DIFF
--- a/verible/verilog/tools/project/project-tool.cc
+++ b/verible/verilog/tools/project/project-tool.cc
@@ -255,7 +255,7 @@ static absl::Status ShowFileDependencies(const SubcommandArgsRange &args,
 
   // Print.
   if (config.output_format == "human") {
-    deps.PrintGraph(outs);
+    deps.PrintHumanGraph(outs);
   } else if (config.output_format == "tsort") {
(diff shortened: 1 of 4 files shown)


verible/verilog/tools/project/project-tool.cc
```diff
@@ -55,6 +55,14 @@ e.g --include_dir_paths directory1,directory2
 if "A.sv" exists in both "directory1" and "directory2" the one in "directory1" is the one we will use.
 )");
 
+ABSL_FLAG(std::string, output_format, "human",
+          R"(Output format for file-deps.
+Available options:
```
hzeller, line 60: Note: the value could also be an enum, and then use the AbslUnparse machinery to convert a string to an enum. See other places in the Verible code.

I don't want to delay the review furher, so this can be a follow-up PR.

hzeller (approved):
Thanks!
Sorry for the delay, was busy with other projects and didn't see the github PR message.

PR #2514, opened 2026-06-03
Make formatter line-wrap search ordering deterministic

mmoskal:
Make Verible line-wrap search deterministic when multiple queued states have equal cost and terminal column.

`SearchLineWraps()` uses `std::priority_queue<SearchState>`. Previously, equivalent states had no final ordering, so heap behavior could select different wrapping paths across standard-library implementations. Official macOS and Linux release artifacts format the same input differently.

This change adds FIFO insertion order as the final tie-breaker and includes a reduced regression test.

## Reproduction

The issue reproduces with official `v0.0-4051-g9fdb4057` and `v0.0-4053-g89d4d98a` release artifacts.

Save as `repro.sv`:

```systemverilog
module m;
  function f();
    return first_call_with_a_long_name_to_force_platform_dependent_wrapping_xxxxxxxxxxxxxx(a, b, c, d) && second_call_with_a_long_name_to_keep_the_rhs_wrapped_xxxxxxxxxxxxxxxxxxxxxx();
  endfunction
endmodule
```

Run:

```bash
verible-verilog-format repro.sv
```

The official Linux x86-64 artifact produces:

```systemverilog
module m;
  function f();
    return first_call_with_a_long_name_to_force_platform_dependent_wrapping_xxxxxxxxxxxxxx(a, b, c,
                                                                                           d) &&
        second_call_with_a_long_name_to_keep_the_rhs_wrapped_xxxxxxxxxxxxxxxxxxxxxx();
  endfunction
endmodule
```

The official macOS artifact produces:

```systemverilog
module m;
  function f();
    return first_call_with_a_long_name_to_force_platform_dependent_wrapping_xxxxxxxxxxxxxx(a, b,
                                                                                           c, d) &&
        second_call_with_a_long_name_to_keep_the_rhs_wrapped_xxxxxxxxxxxxxxxxxxxxxx();
  endfunction
endmodule
```

Each artifact is idempotent on its own output and rewrites the other artifact's output back to its preferred layout. `--show_equally_optimal_wrappings` prints no diagnostics for this fixture.

## Fix

Assign each queued search state a monotonically increasing sequence number. When the existing comparator cannot distinguish two states, prefer the state inserted first.

This preserves the existing cost and terminal-column priorities while making equivalent-state ordering deterministic.

## Validation

Built the patched formatter from source on macOS arm64 and Linux x86-64. Both now produce identical output for the reduced fixture.

```text
SHA-256:
1a3775db233cc9ef45deb98eeacf9e465d19b6d2fc119552ae0140b15ce4baa5
```

Passed on both platforms:

```bash
bazel test \
  //verible/common/formatting:all \
  //verible/verilog/formatting:all
```

Also ran:

```bash
clang-format --dry-run -Werror \
  verible/common/formatting/line-wrap-searcher.cc \
  verible/verilog/formatting/formatter-tuning_test.cc
```

### DIFF
--- a/verible/common/formatting/line-wrap-searcher.cc
+++ b/verible/common/formatting/line-wrap-searcher.cc
@@ -14,6 +14,7 @@
 
 #include "verible/common/formatting/line-wrap-searcher.h"
 
+#include <cstddef>
 #include <memory>
 #include <ostream>
 #include <queue>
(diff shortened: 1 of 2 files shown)


hzeller (approved):
Thanks for the contribution and fixing the subtle platform dependency.

PR #2520, opened 2026-06-14
Make biso/flex work with local and BCR tools work with bazel 8

### LINKED ISSUE #2435: Flex and Bison toolchain switching stopped working with bazel 8

hzeller:
We can't select on a toolchain anymore, so break up the build rules into a local and non-local genrule and choose outputs from these.

Fixes #2435

### DIFF
--- a/bazel/bison.bzl
+++ b/bazel/bison.bzl
@@ -26,21 +26,56 @@ def genyacc(
         extra_outs = []):
     """Build rule for generating C or C++ sources with Bison.
     """
+    all_outs = [header_out, source_out] + extra_outs
+
+    def replace_outs(options, prefix):
+        res = []
+        for opt in options:
+            for out in all_outs:
+                opt = opt.replace("$(location %s)" % out, "$(location %s%s)" % (prefix, out))
+            res.append(opt)
+        return res
+
+    native.genrule(
+        name = name + "_toolchain",
+        srcs = [src],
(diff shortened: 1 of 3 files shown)


hzeller:
This is somewhat ugly, but it helps bring us over the post-bazel 7 hump.
Ideally we can have the solution in #2524 at some point, but that currently still has issues to compile on Windows.

hzeller:
Closing in favor of #2563

PR #2525, opened 2026-06-18
Fix --alignment_group_boundary string parsing

sjalloq:
The initial implementation used to find separator comments was too strict: it required the entire comment body to be 4+ repeats of a single character. This failed to identify dividers that also contain text, e.g. "// ---- Registers ----".

The fix scans for a run of 4 or more consecutive identical "divider" characters (any non-alphanumeric, non-whitespace character) anywhere in the comment, ignoring surrounding text.


hzeller (approved):
neat!

PR #2526, opened 2026-06-19
Fix indent of `let`

wsnyder:
Fixes previous mis-indent of let, as there was no code I could find to handle it.

Previous output:

```
-  let OFF = 4;
-  let EXPECT = 16;
-  let UNIQUE = 32;
+  let OFF = 4; let EXPECT = 16; let UNIQUE = 32;
```


wsnyder:
Closing as I don't have a good debug path presently.

PR #2527, opened 2026-06-24
Fix unique0 case missing default

### LINKED ISSUE #2413: unique0 support missing
**Describe the bug**
`[Style: case-statements] [case-missing-default]`

`unique0` should allow for missing defaults in case statements, the linter currently reports these are errors.

```systemverilog
logic [2:0] foo;
logic bar;

always_ff begin
  unique0 case(foo)
    3'b001: bar <= 1'b1;
  endcase
end
```

Expected no error, got:
```
Explicitly define a default case for every case statement or add `unique` qualifier to the case statement. [Style: case-statements] [case-missing-default]
```

False-positive? False-negative? Crash? Wrong diagnostic location?
False-positive

What should have happened?
No linter error, see [here](https://dvcon-proceedings.org/wp-content/uploads/systemverilog-2009-enhancements-priority-unique-unique.pdf) for example usage of `unique0`

MobiusRoamer:
Fixes #2413 .

The `case-missing-default` rule currently allows `unique case` statements without a `default` item, but still reports error under `unique0 case` statements while it should not --`unique0 case` is also allowed to omit a matching branch. This updates the rule to treat `unique0 case` like `unique case` when deciding whether a `default` item is required.

Also I have added regression coverage for both the matcher and the lint rule.

Tests:
- bazel test --cache_test_results=no --test_output=errors //verible/verilog/analysis/checkers:case-missing-default-rule_test
- bazel test --cache_test_results=no --test_output=errors //verible/verilog/CST:verilog-matchers_test

### DIFF
--- a/verible/verilog/CST/verilog-matchers.h
+++ b/verible/verilog/CST/verilog-matchers.h
@@ -437,6 +437,8 @@ inline constexpr auto HasDefaultCase =
 //     ...
 inline constexpr auto HasUniqueQualifier =
     verible::matcher::MakePathMatcher(L(TK_unique));
+inline constexpr auto HasUnique0Qualifier =
+    verible::matcher::MakePathMatcher(L(TK_unique0));
 // Clean up macros
 #undef N
 #undef L
--- a/verible/verilog/CST/verilog-matchers_test.cc
+++ b/verible/verilog/CST/verilog-matchers_test.cc
@@ -827,5 +827,55 @@ TEST(VerilogMatchers, HasUniqueQualifierTests) {
     verible::matcher::RunRawMatcherTestCase<VerilogAnalyzer>(test);
   }
 }
+
+// Tests for HasUnique0Qualifier matching.
+TEST(VerilogMatchers, HasUnique0QualifierTests) {
+  const RawMatcherTestCase tests[] = {
+      {HasUnique0Qualifier(), "", 0},
+      {HasUnique0Qualifier(),
+       R"(
+       function automatic int foo (input in);
+         case (in)
+           default: return 0;
+         endcase
+       endfunction
+       )",
+       0},
+      {HasUnique0Qualifier(),
+       R"(
+       function automatic int foo (input in);
+         unique0 case (in)
+           1: return 0;
+         endcase
+       endfunction
+       )",
+       1},
+      {HasUnique0Qualifier(),
+       R"(
+       function automatic int foo (input in);
+         unique case (in)
+           1: return 0;
+         endcase
+       endfunction
+       )",
+       0},
+      {HasUnique0Qualifier(),
+       R"(
+       function automatic int foo (input in);
+         unique0 if (in) begin
+           return 0;
+         end
+         else if(!in) begin
+           return 1;
+         end
+       endfunction
+       )",
+       1},
+  };
+  for (const auto &test : tests) {
+    verible::matcher::RunRawMatcherTestCase<VerilogAnalyzer>(test);
+  }
+}
+
 }  // namespace
 }  // namespace verilog
--- a/verible/verilog/CST/verilog-matchers_test.cc
+++ b/verible/verilog/CST/verilog-matchers_test.cc
@@ -822,6 +822,15 @@ TEST(VerilogMatchers, HasUniqueQualifierTests) {
        endfunction
        )",
        1},
+      {HasUniqueQualifier(),
+       R"(
+       function automatic int foo (input in);
+         unique0 case (in)
+           1: return 0;
+         endcase
+       endfunction
+       )",
+       0},
   };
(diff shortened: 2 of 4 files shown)


aishasalim:
Hi @MobiusRoamer — I independently hit #2413 and landed on the same fix before spotting your PR, so I've closed mine (#2550). Same matcher, same approach.

Before closing I'd written some extra test coverage that isn't in this PR. Offering it here in case you'd like to fold it in — no obligation.

In `case-missing-default-rule_test.cc`, extending beyond plain `case` to the other two forms, since the rule applies to all three:

```systemverilog
// unique0 casex / casez
function automatic int foo (input in);
  unique0 casex (in)
    1: return 0;
  endcase
endfunction

function automatic int foo (input in);
  unique0 casez (in)
    1: return 0;
  endcase
endfunction
```

A qualifier combined with an explicit default:

```systemverilog
function automatic int foo (input in);
  unique0 case (in)
    1: return 0;
    default: return 1;
  endcase
endfunction
```

And the nesting case — a qualified outer case should not exempt an unqualified inner one. This is the one most likely to catch a regression:

```systemverilog
function automatic int foo (input in);
  unique0 case (in)
    1: begin;
      case (in)        // <- still expected to violate
        1: return 1;
      endcase
    end
  endcase
endfunction
```

In `verilog-matchers_test.cc` you test that `HasUnique0Qualifier` doesn't match `unique`; the reverse is worth pinning too, since the whole fix rests on these being distinct tokens:

```cpp
{HasUniqueQualifier(),
 R"(
 function automatic int foo (input in);
   unique0 case (in)
     1: return 0;
   endcase
 endfunction
 )",
 0},
```

Unrelated to your change: the current MacOsBuild failures are a zlib absolute-path-inclusion problem in the Bazel cache, which @hzeller appears to be addressing in #2549.

MobiusRoamer:
> Hi @MobiusRoamer — I independently hit #2413 and landed on the same fix before spotting your PR, so I've closed mine (#2550). Same matcher, same approach.
> 
> Before closing I'd written some extra test coverage that isn't in this PR. Offering it here in case you'd like to fold it in — no obligation.
> 
> In `case-missing-default-rule_test.cc`, extending beyond plain `case` to the other two forms, since the rule applies to all three:
> 
> ```systemverilog
> // unique0 casex / casez
> function automatic int foo (input in);
>   unique0 casex (in)
>     1: return 0;
>   endcase
> endfunction
> 
> function automatic int foo (input in);
>   unique0 casez (in)
>     1: return 0;
>   endcase
> endfunction
> ```
> 
> A qualifier combined with an explicit default:
> 
> ```systemverilog
> function automatic int foo (input in);
>   unique0 case (in)
>     1: return 0;
>     default: return 1;
>   endcase
> endfunction
> ```
> 
> And the nesting case — a qualified outer case should not exempt an unqualified inner one. This is the one most likely to catch a regression:
> 
> ```systemverilog
> function automatic int foo (input in);
>   unique0 case (in)
>     1: begin;
>       case (in)        // <- still expected to violate
>         1: return 1;
>       endcase
>     end
>   endcase
> endfunction
> ```
> 
> In `verilog-matchers_test.cc` you test that `HasUnique0Qualifier` doesn't match `unique`; the reverse is worth pinning too, since the whole fix rests on these being distinct tokens:
> 
> ```c++
> {HasUniqueQualifier(),
>  R"(
>  function automatic int foo (input in);
>    unique0 case (in)
>      1: return 0;
>    endcase
>  endfunction
>  )",
>  0},
> ```
> 
> Unrelated to your change: the current MacOsBuild failures are a zlib absolute-path-inclusion problem in the Bazel cache, which @hzeller appears to be addressing in #2549.

Thanks for sharing improved coverage tests! 
That makes sense -- I agree the rule should not just cover plain `case`, also should look for reverse matching, I will definitely take a loot at the tests from #2550 and fold relevant ones into this PR. 

Also sorry if I confused u -- I didnt comment on the issue beforehand to claim, it was my first go at an issue/pr, so wasn't fully sure about the workflow. Will go back and do that too.

hzeller:
Just ping this PR once you're ready with the additional testing.

MobiusRoamer:
> Just ping this PR once you're ready with the additional testing.

Thanks for checking on this. I added suggested extra coverage for the reverse matcher case, qualified `casez`, `casex`, and a nested-case regression where a qualified outer case should not exempt an unqualified inner one. Thanks again to @aishasalim on collaborating. 

Open to further changes.

hzeller:
Thanks for your contribution! Sorry I was a bit busy and thus reviews are a bit slow...

MobiusRoamer:
> Thanks for your contribution! Sorry I was a bit busy and thus reviews are a bit slow...

Not at all, thx for the review!

PR #2530, opened 2026-06-28
Add --parameter_declaration_alignment for body-level param/localparam

skyhgzsh:
Previously, body-level parameter/localparam declarations in module, generate, package, and interface bodies received no alignment treatment. This flag fills that gap, separate from --formal_parameters_alignment which now exclusively controls #(...) header parameter lists. Class body params are intentionally excluded (already handled by class_member_variable_alignment).

Reuses ParameterDeclarationColumnSchemaScanner with a new kBodyParameterDeclaration subtype. kPackageItemList is moved from kAlwaysExpand to kTabularAlignment to route packages through AlignModuleItems, which also enables net/variable and assignment alignment in packages as a side benefit.

Also fix: struct_union_members_alignment was missing from ApplyToAllAlignmentPolicies; added for consistency.


hzeller (approved):
Thanks for your contribution! Sorry it took so long, I was tied up in other projects.

I like that you made these choices configurable via the format-style, which will allow folks to choose the style they prefer.

PR #2531, opened 2026-07-05
patch: reject non-positive hunk start line to avoid OOB read

EylonKrause:
### Problem
`Hunk::VerifyAgainstOriginalLines` can read a `std::vector` out of bounds on a malformed patch.

### Root cause
`common/strings/patch.cc` guards `original_lines[line_number - 1]` (line 237) only against `line_number` being too large; there is no lower-bound check. `line_number` comes from `header_.old_range.start`, parsed by `HunkIndices::Parse` with `absl::SimpleAtoi` (which accepts `0` and negatives), and `Hunk::IsValid` only checks that the non-added line count equals `old_range.count`. So a hunk header like `@@ -0,1 +1,1 @@` followed by a context line parses successfully with `line_number == 0`, and `original_lines[0 - 1]` indexes at `(size_t)-1` — an out-of-bounds read; the subsequent `.Text()` comparison then dereferences a wild `string_view`. Reachable via `verible-patch-tool apply` on a crafted patch file.

### Fix
Extend the existing bounds error to also reject `line_number < 1`.

### Verification
Static reasoning; the existing `LineNumberOutOfBounds` test exercises only the upper bound. (Verible builds with Bazel, which I don't have set up locally — happy to add a lower-bound test case.)

### DIFF
--- a/verible/common/strings/patch.cc
+++ b/verible/common/strings/patch.cc
@@ -229,7 +229,8 @@ absl::Status Hunk::VerifyAgainstOriginalLines(
   int line_number = header_.old_range.start;  // 1-indexed
   for (const MarkedLine &line : lines_) {
     if (line.IsAdded()) continue;  // ignore added lines
-    if (line_number > static_cast<int>(original_lines.size())) {
+    if (line_number < 1 ||
+        line_number > static_cast<int>(original_lines.size())) {
       return absl::OutOfRangeError(absl::StrCat(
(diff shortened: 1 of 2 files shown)


EylonKrause:

hzeller (approved):
Thanks for your contribution, and improving the robustness!

Sorry for the delay in the review, I was a bit busy the last couple of weeks.

PR #2532, opened 2026-07-05
preprocessor: stop macro-argument scan on an unexpected/EOF token

EylonKrause:
### Problem
`VerilogPreprocess::ConsumeAndParseMacroCall` hangs (infinite loop) on an unterminated macro call.

### Root cause
The `while (parameters_size > 0)` argument-scan loop (`verilog/preprocessor/verilog-preprocess.cc:252`) only handles three token kinds: `MacroArg`, `,`, and `)`. An unterminated call such as `` `FOO( `` at end of input leaves the current token as the EOF token, which matches none of the three branches — so neither `token_iter` nor `parameters_size` advances and the loop spins forever. This is reached from untrusted/truncated `.sv` via `AnalyzeAutomaticPreprocessFallback` (`expand_macros=true`) in both the linter and the LSP.

### Fix
Break out of the loop on any other token; the existing trailing loop already back-fills the remaining parameters with default `TokenInfo` (the same terminal state an early `)` produces), so the fix routes into a path the tests already exercise.

### Related (not fixed here)
On the macro-*expansion* paths (`ExpandMacro`/`ExpandText`), the lexed sequence is built without a trailing EOF sentinel, so `GenerateBypassWhiteSpaces` can dereference an exhausted generator (crash) on similar input. That needs a separate structural fix (append an EOF sentinel / add an end-guard) — happy to follow up in a second PR.

### Verification
Static reasoning; the `` `MACRO3 () `` test already exercises the all-defaults back-fill path this fix routes into. (Bazel build not set up locally.)

### DIFF
--- a/verible/verilog/preprocessor/verilog-preprocess.cc
+++ b/verible/verilog/preprocessor/verilog-preprocess.cc
@@ -270,10 +270,14 @@ absl::Status VerilogPreprocess::ConsumeAndParseMacroCall(
     }
     // Any other token -- in particular the EOF token from an unterminated
     // macro call -- would otherwise leave token_iter and parameters_size
-    // unchanged and spin this loop forever. Stop scanning; the loop below
-    // back-fills the remaining parameters with default TokenInfo (the same
-    // terminal state produced by an early ')').
-    break;
+    // unchanged and spin this loop forever. An early ')' (handled above) is the
+    // legal way to pass fewer arguments than parameters; anything else here is
+    // a malformed call, so reject it rather than silently accepting it. The
+    // caller records the returned message as a preprocessor error.
+    return absl::InvalidArgumentError(absl::StrCat(
+        "unexpected token while scanning arguments of macro call `",
+        macro_name_str,
+        ": expected ',' or ')', but got: ", (*token_iter)->ToString()));
   }
   if (parameters_size > 0) {
     while (parameters_size--) {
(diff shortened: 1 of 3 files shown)


EylonKrause:

verible/verilog/preprocessor/verilog-preprocess.cc
```diff
@@ -268,6 +268,12 @@ absl::Status VerilogPreprocess::ConsumeAndParseMacroCall(
     if ((*token_iter)->text() == ")") {
       break;
     }
+    // Any other token -- in particular the EOF token from an unterminated
+    // macro call -- would otherwise leave token_iter and parameters_size
+    // unchanged and spin this loop forever. Stop scanning; the loop below
+    // back-fills the remaining parameters with default TokenInfo (the same
+    // terminal state produced by an early ')').
+    break;
```
hzeller: I wonder if instead of silently accepting the issue, we should may return `absl::InvalidArgumentError()` ?

EylonKrause:
Good idea — done. `ConsumeAndParseMacroCall` now returns `absl::InvalidArgumentError` when the argument scan hits an unexpected token (in particular the EOF of an unterminated call), and the caller (`HandleMacroIdentifier`) records it as a preprocessor error at the call site, so a malformed call surfaces a real diagnostic instead of being silently back-filled. An early `)` remains the legal way to pass fewer arguments than parameters. The test now asserts the error is reported.

(`ConsumeAndParseMacroCall` is `static`, so it can't push to `preprocess_data_.errors` itself; the caller records the returned message — mirroring the existing undefined-macro path.)

EylonKrause:
The smoke-test failure here is the same one as on #2535: comparing the master and branch builds of the preprocessor over all 1031 sv-tests files, exactly one file changes status — `tests/chapter-22/22.5.1--define-expansion_18.sv`, a `should_fail` test (macro defined with arguments, invoked without parentheses). Master silently accepts it; this branch now rejects it, as 22.5.1 requires. Bumped `ExpectedFailCount[preprocessor:sv-tests]` 128 -> 129. Note both this PR and #2535 make the same bump, so whichever lands second will need a trivial rebase of that line.

hzeller (approved):
Thanks

hzeller:
So the test that is supposed to receive error messages is actually not doing that at head, making the CI break after the merge. Can you have a look what is happening ?

hzeller:
Can you have a look @EylonKrause ?

PR #2534, opened 2026-07-06
port-name-suffix: handle `ref` direction and all-underscore names

EylonKrause:
### Problem
`PortNameSuffixRule` crashes the linter on two kinds of valid SystemVerilog:

1. A `ref` port direction throws `std::out_of_range` (uncaught → linter abort).
2. An all-underscore port name (e.g. `_`) dereferences an empty vector (UB).

### Details
**1. `ref` direction** — `IsSuffixCorrect` does `suffixes.at(direction)`, where `suffixes` only has keys `{input, output, inout}`. The grammar's `port_direction : dir | TK_ref` allows `ref`, so a port like `module m(ref logic data_x); endmodule` reaches `suffixes.at("ref")`, which throws. `HandleSymbol` runs without a surrounding try/catch, so the linter terminates. (The comment claiming the direction "is guaranteed" to be a known key is incorrect.)

**2. all-underscore name** — `name_parts = absl::StrSplit(name, '_', absl::SkipEmpty())`. For a name that is only underscores (`_`, `__`), every token is empty and `SkipEmpty` drops them all, leaving `name_parts` empty. The `if (name_parts.size() < 2)` block reports a violation but does **not** return, so `name_parts.back()` on the next line dereferences an empty vector (UB).

### Fix
1. Look the direction up with `find()` and treat an unknown direction (e.g. `ref`) as "correct" — no violation — matching `Violation()`, which already ignores non-input/output/inout directions.
2. `return` after the `size() < 2` violation (a name with fewer than two underscore-delimited parts has no suffix to check). This also removes a redundant duplicate `Violation` insert for no-underscore names.

### Tests
`port-name-suffix-rule_test.cc`: a `ref` port is accepted (no violation, no crash); an all-underscore `_` port reports a single suffix violation instead of crashing.

### DIFF
--- a/verible/verilog/analysis/checkers/port-name-suffix-rule.cc
+++ b/verible/verilog/analysis/checkers/port-name-suffix-rule.cc
@@ -90,10 +90,14 @@ bool PortNameSuffixRule::IsSuffixCorrect(std::string_view suffix,
        {"output", {"o", "no", "po"}},
        {"inout", {"io", "nio", "pio"}}};
 
-  // At this point it is guaranteed that the direction will be set to
-  // one of the expected values (used as keys in the map above).
-  // Therefore checking the suffix like this is safe
-  return suffixes.at(direction).count(suffix) == 1;
+  // `direction` is usually one of the map keys, but the grammar also permits a
+  // `ref` port direction, which has no suffix convention. Look the direction up
+  // safely and treat an unknown direction as "correct" (no violation),
+  // consistent with Violation() which also ignores non-input/output/inout
+  // directions. Using std::map::at() here would throw on e.g. "ref".
+  const auto it = suffixes.find(direction);
+  if (it == suffixes.end()) return true;
+  return it->second.count(suffix) == 1;
 }
 
 void PortNameSuffixRule::HandleSymbol(const Symbol &symbol,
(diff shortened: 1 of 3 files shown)


EylonKrause:

verible/verilog/analysis/checkers/port-name-suffix-rule.cc
```diff
@@ -90,10 +90,14 @@ bool PortNameSuffixRule::IsSuffixCorrect(std::string_view suffix,
...
-  // At this point it is guaranteed that the direction will be set to
-  // one of the expected values (used as keys in the map above).
-  // Therefore checking the suffix like this is safe
-  return suffixes.at(direction).count(suffix) == 1;
+  // `direction` is usually one of the map keys, but the grammar also permits a
+  // `ref` port direction, which has no suffix convention. Look the direction up
+  // safely and treat an unknown direction as "correct" (no violation),
+  // consistent with Violation() which also ignores non-input/output/inout
+  // directions. Using std::map::at() here would throw on e.g. "ref".
+  const auto it = suffixes.find(direction);
```
hzeller: We have the `FindWithDefault()` in [`verible/common/util/container-util.h`](https://github.com/chipsalliance/verible/blob/70eba4dda1bab8e21012d9a4508cd3a64bd7a47e/verible/common/util/container-util.h#L70-L77)

Maybe use that instead, not only for a more compact writing but also a more visible 'we're looking for something that might not be there but we have a fallback' intent.

```
return FindWithDefault(suffixes, direction, 1) == 1;
```

hzeller (commented):
Thanks for your contribution!
LGTM, but I have one comment about using a map lookup function that might be nicer in this case.

EylonKrause:
Thanks! Done — switched to `FindWithDefault`. Since `suffixes` maps each direction to a *set* of valid suffixes (rather than a scalar), I used it with an empty-set fallback so the "look for something that might not be there, with a fallback" intent is explicit, while keeping the behavior that a direction with no convention (e.g. `ref`) is treated as correct:

```cpp
static const std::set<std::string_view> kNoConvention;
const std::set<std::string_view> &valid =
    verible::container::FindWithDefault(suffixes, direction, kNoConvention);
return valid.empty() || valid.count(suffix) == 1;
```

fangism (approved):
thank you

PR #2535, opened 2026-07-09
[preprocessor] Fix crash/hang on truncated callable macro invocation

EylonKrause:
### Problem
The preprocessor crashes (SIGSEGV) or hangs on parseable SystemVerilog when a **callable macro invocation is truncated at end-of-stream**. Minimal reproducers (all with `--expand_macros`, e.g. via `verible-verilog-preprocessor preprocess`):

```systemverilog
`define A(x) hello `A
`A(1)
```
```systemverilog
`define A(x) x
`A            // top-level, no arguments
```
An included file ending in a callable macro reference (e.g. `foo \`A`) crashes the same way, and a `(` with no matching `)` (`` `define A(x) hello `C( ``) spins in an infinite loop.

### Root cause
`GenerateBypassWhiteSpaces` (verilog-preprocess.cc:63) pulls tokens from a `StreamIteratorGenerator` and dereferences the result — `**iterator` — with **no `end` guard**:
```cpp
auto iterator = generator();
while (verilog::VerilogLexer::KeepSyntaxTreeTokens(**iterator) == 0) {
  iterator = generator();
}
```
`MakeConstIteratorStreamer` returns the view's `end()` iterator once exhausted (token-stream-adapter.h), so `**iterator` on `end()` is a double past-the-end dereference. The function relies on the invariant that every stream ends in a kept EOF token (`KeepSyntaxTreeTokens(EOF)` is true), which stops the loop and lets callers see `isEOF()`. The top-level analyzer stream satisfies this (it retains EOF), but the streams re-lexed for macro expansion do **not**: `ExpandText`, `ExpandMacro`, `HandleInclude`, and the standalone tool all build their token sequence with a loop that stops *before* the EOF (`!lexer.GetLastToken().isEOF()`). A truncated callable-macro invocation at the end of such a stream reaches `ConsumeAndParseMacroCall`, which calls `GenerateBypassWhiteSpaces`, whose `generator()` returns `end()` → crash. The sibling streamers (`MakeTokenStreamer`) already guard this by returning an `EOFToken` sentinel; `GenerateBypassWhiteSpaces` does not, because it only has the `std::function`, not the `end` iterator.

### Fix
Restore the kept-EOF invariant on every re-lexed stream, so `GenerateBypassWhiteSpaces` stops at EOF and callers handle it gracefully (returning `InvalidArgumentError`) instead of dereferencing past the end:

- **`ExpandText` / `ExpandMacro` / `HandleInclude` / the standalone tool**: append the EOF token as an end sentinel before building the stream view.
- **`ExpandText` / `ExpandMacro` token-pulling loops**: `break` on the EOF sentinel so it is not forwarded into the expanded output.
- **`HandleInclude`**: skip the sentinel EOF when splicing the child stream into the parent (the child `ScanStream` forwards it as a pass-through token; without this it would land in the middle of the parent stream).
- **`ConsumeAndParseMacroCall` argument loop**: `break` on EOF. With the sentinel in place a `(` without a matching `)` would otherwise spin forever; this is the one caller that did not already handle a mid-scan EOF.
- **`ConsumeAndParseMacroCall`**: record the "illegal to call a callable macro without ()" error in `preprocess_data_.errors` (it was returned as a `Status` but never surfaced, so truncated input was silently accepted). This required making the method non-static (its only caller is non-static).

### Testing
Built `verible-verilog-preprocessor` and confirmed all four crash variants and the open-paren hang now exit cleanly. The full `//verible/verilog/preprocessor/...` suite (including the existing "Nested callable macros" test) and `verilog-analyzer_test` pass. Added `TruncatedCallableMacroDoesNotCrash` covering the no-`(` (now a diagnostic) and open-`(` (now terminates) cases.

Out of scope (separate, pre-existing bug, not addressed here): unbounded macro **self-recursion** such as `` `define A(x) `A(x) `` overflows the stack with no depth guard — worth a follow-up.

---

### DIFF
--- a/.github/bin/smoke-test.sh
+++ b/.github/bin/smoke-test.sh
@@ -131,7 +131,7 @@ ExpectedFailCount[preprocessor:opentitan]=3017
 ExpectedFailCount[syntax:sv-tests]=74
 ExpectedFailCount[lint:sv-tests]=73
 ExpectedFailCount[project:sv-tests]=176
-ExpectedFailCount[preprocessor:sv-tests]=128
+ExpectedFailCount[preprocessor:sv-tests]=129
 
 ExpectedFailCount[syntax:caliptra-rtl]=41
 ExpectedFailCount[lint:caliptra-rtl]=40
--- a/verible/verilog/preprocessor/verilog-preprocess.cc
+++ b/verible/verilog/preprocessor/verilog-preprocess.cc
@@ -245,11 +245,15 @@ absl::Status VerilogPreprocess::ConsumeAndParseMacroCall(
   if ((*token_iter)->text() == "(") {
     token_iter = GenerateBypassWhiteSpaces(generator);  // skip the "("
   } else {
+    preprocess_data_.errors.emplace_back(
+        **token_iter,
+        "Error it is illegal to call a callable macro without ().");
     return absl::InvalidArgumentError(
         "Error it is illegal to call a callable macro without ().");
   }
 
   while (parameters_size > 0) {
+    if ((*token_iter)->isEOF()) break;  // truncated call; stop scanning args
     if ((*token_iter)->token_enum() == MacroArg) {
       macro_call->positional_arguments.emplace_back(**token_iter);
       token_iter = GenerateBypassWhiteSpaces(generator);
@@ -342,6 +346,10 @@ absl::Status VerilogPreprocess::ExpandText(
        lexer.DoNextToken()) {
     lexed_sequence.push_back(lexer.GetLastToken());
   }
+  // Retain the EOF token as an end sentinel so a truncated callable-macro
+  // invocation stops at EOF in GenerateBypassWhiteSpaces instead of
+  // dereferencing past the end of the stream view.
+  lexed_sequence.push_back(lexer.GetLastToken());
   verible::TokenStreamView lexed_streamview;
   // Initializing the lexed token stream view.
   InitTokenStreamView(lexed_sequence, &lexed_streamview);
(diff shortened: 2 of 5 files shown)


hzeller:
The preprocessing test is failing the CI. Can you rebase (as there was some unrelated mac failure issue), and have a look at the failing test ?

EylonKrause:
Rebased onto current master. The "preprocessing test failing" was the stale branch (it was ~4k commits behind) — after the rebase the whole `verilog-preprocess_test` suite, including this PR's `TruncatedCallableMacroDoesNotCrash`, passes locally, and it also picks up the unrelated macOS fix you mentioned.

One heads-up on overlap: this PR and #2532 both touch `ConsumeAndParseMacroCall`'s end-of-stream handling, via different routes — #2535 makes the method non-static so it records the diagnostic itself, while #2532 keeps it static and has the caller record it. Whichever you'd rather land first, I'll rebase the other on top (or fold them together) so they don't collide — just say the word.

EylonKrause:
Found it, and it was mine: the preprocessor tool now appends the lexer's EOF token to the stream as an end sentinel (so a truncated callable-macro call stops at EOF instead of running off the end of the stream view), and the tool's print loop was echoing that sentinel, which shows up as one extra trailing newline — exactly the diff `verilog-preprocessor_test` was seeing. The print loop now skips the EOF token. Rebased onto current master; `verilog-preprocessor_test` and `verilog-preprocess_test` both pass locally.

hzeller:
If you run `.github/bin/smoke-test.sh`, it looks like there is one more issue now in the sv-test running into one more issue with the preprocessing. Can you figure out why ?

EylonKrause:
Ran the smoke test against sv-tests with the master and branch builds of the preprocessor over all 1031 files: exactly one file changes status, `tests/chapter-22/22.5.1--define-expansion_18.sv` (master exit 0, branch exit 1). That file is a `should_fail` test — its header says "for a macro with arguments, the parentheses are always required in the macro call, even if all the arguments have defaults" (22.5.1), and it invokes `MACRO3` with no parentheses. Master silently accepted it; the branch now reports "it is illegal to call a callable macro without ()" and exits non-zero, which is the required behaviour. So the extra preprocessor failure is a correct rejection, not a regression — I bumped `ExpectedFailCount[preprocessor:sv-tests]` from 128 to 129 in the smoke test.

hzeller (approved):
Thanks!

PR #2541, opened 2026-07-17
Fix formatter non-convergence with EOL comment before else if

### LINKED ISSUE #2540: Re-formatted text does not match formatted text; formatting failed to converge!  Please file a bug.
Verible 4071 verible-verilog-format fails the convergence check on an if / else if chain when an EOL comment sits between end and the next else if. The first format pass and the second disagree on whether the following assignment stays on one line.

Reproducer

**Test case**

```systemverilog
module m;
  always_comb begin
    case (state)
      STATE_A: begin
        if (cond_aaaa) next_state_value = STATE_B;
        else if (cond_bbbb) begin
          next_state_value = STATE_B;
        end
        // xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
        else if (cond_cccc) next_state_value = STATE_C;
      end
    endcase
  end
endmodule
```

**Actual output**

```bash
```

kbrunham-intel:
## Summary

Fixes [#2540](https://github.com/chipsalliance/verible/issues/2540): formatter non-convergence when an EOL comment sits between `end` and a following `else if`.

When `end` carried a trailing EOL comment, `PushEndIntoElsePartition` merged that `end`+comment into the else-if header partition. `fit-else-expand` then treated the header as wider than the eventual printed line, so the first format kept the else-if assignment on one line and the second wrap split it — `format(format(x)) != format(x)`.

**Change:** skip merging `end` into the else partition when the `end` leaf contains an EOL comment (else must start on the next line anyway when preceded by a comment).

## Test plan

- [x] Added `FormatterEndToEndTest.EndElseIfWithEOLCommentConverges` covering comment-on-own-line and comment-already-on-`end`
- [x] `bazel test //verible/verilog/formatting:formatter_test`
- [ ] CI green on this PR


hzeller:
Thanks for your contribution - Sorry for the delay in the review, I was a bit busy the last couple of weeks.

Looks good. Can you rebase, as there were other changes in the formatter_test that now require a merge ?

hzeller (approved):
Thanks

PR #2543, opened 2026-07-17
Fix formatter non-convergence for continuation EOL comments

### LINKED ISSUE #2542: Verible-verilog format fails when a wrapped assign has a multiline comment
verible-verilog-format fails the convergence check when a wrapped assign has a multi-line EOL comment after the statement. The first format pass column-aligns the continuation comment far to the right; re-formatting left-aligns it. So format(format(x)) != format(x).

This is not the same as #2540 (end / EOL comment / else if).

**Test case**

```systemverilog
module m;
  assign status_ur = !(status_sc || status_ca ||
    status_crs);      // Completions with a Reserved Completion
                      // Status value are treated as UR
endmodule
```

**Actual output**

```bash
```

kbrunham-intel:
## Summary
- Fixes non-convergence when a wrapped `assign` is followed by column-aligned continuation EOL comments (`format(format(x)) != format(x)`).
- `CalculateEolCommentColumn` now treats newlines in preserved leading whitespace like emit does (reset column after the last `\n`) instead of counting them as width.
- Adds regression test `ContinuationCommentAfterWrappedAssignConverges`.

Fixes #2542

## Test plan
- [x] `bazel test -c opt //verible/verilog/formatting:formatter_test`
- [x] Reproduce from #2542 formats and converges
- [ ] CI on this PR is green


verible/verilog/formatting/formatter.cc
```diff
@@ -777,12 +763,39 @@ class ContinuationCommentAligner {
     }
     column += front.token->text().length();
 
-    for (const auto &ftoken : verible::make_range(line.Tokens().begin() + 1,
-                                                  line.Tokens().end() - 1)) {
-      AdjustColumnUsingTokenSpacing(ftoken, &column);
-      column += ftoken.token->text().length();
+    for (size_t i = 1; i < tokens.size(); ++i) {
```
hzeller, line 797: since we're using tokens here first, I'd probably move the `const auto &tokens = line.Tokens();` right in front of the for loop.

verible/verilog/formatting/formatter.cc
```diff
@@ -777,12 +763,39 @@ class ContinuationCommentAligner {
...
+      const auto &ftoken = tokens[i];
+      switch (ftoken.before.action) {
```
hzeller: The previous implementation, that put the logic into its own function (`AdjustColumnsUsingTokenSpacing()`), was nicely separating that dense logic into the function call. Do you think we can do something here as well again ?

hzeller:
Sorry, due to other changes, the formatter_test.cc now needs merging. Can you rebase ?
I have some initial comments.

verible/verilog/formatting/formatter.cc
kbrunham-intel: I've created a new helper function here to help with readability.

verible/verilog/formatting/formatter.cc
kbrunham-intel: Done.

kbrunham-intel:
Hi @hzeller. I've updated the PR and provided a change to incorporate your feedback. Please approve and merge the change if it is ready, or let me know what other feedback I should look to incorporate.

hzeller (approved):
thanks!

PR #2545, opened 2026-07-17
Fix formatter lexical mismatch for macros at end of line

### LINKED ISSUE #2544: Formatter: lexical mismatch (MacroIdentifier vs MacroIdItem) when wrapping puts a macro at end of line
verible-verilog-format can rewrite a $bits(...)'(...) cast so a `MACRO ends a line. Re-lexing that output classifies the macro as MacroIdItem instead of MacroIdentifier, so verification reports a lexical difference even though the text looks equivalent.

**Test case**

```systemverilog
module m;
  assign result_value = $bits(result_value)'( compare_bytes(input_data[DATA_WIDTH_INT-1:0], input_datak[STROBE_WIDTH_INT-1:0], `TOKEN_BYTE) );
endmodule
```

**Actual output**

```bash
Formatted output is lexically different from the input.    Please file a bug.  Details:
Mismatched token enums.  got: (MacroIdentifier) (#296: "`TOKEN_BYTE") vs. (MacroIdItem) (#298: "`TOKEN_BYTE")
```

kbrunham-intel:
## Summary
- Fixes `Formatted output is lexically different from the input` when wrapping leaves a bare `` `MACRO `` at end of line.
- The lexer then emits `MacroIdItem` instead of `MacroIdentifier`; `FormatEquivalent` now treats spelling-equal macros as equivalent (same idea as existing `MacroCallCloseToEndLine` vs `)`).
- Adds equivalence and formatter regression tests.

Fixes #2544

## Test plan
- [x] `bazel test -c opt //verible/verilog/analysis:verilog-equivalence_test //verible/verilog/formatting:formatter_test`
- [x] Reproduce from #2544 formats without lexical mismatch
- [ ] CI on this PR is green


hzeller (approved):
Thanks!
Can you rebase, so that we can merge cleanrly ? (there were other conflicting changes in the formatter test in the meanwhile).

verible/verilog/analysis/verilog-equivalence.cc
```diff
@@ -167,11 +167,21 @@ DiffStatus LexicallyEquivalent(
   DiffStatus diff_status = DiffStatus::kEquivalent;
   auto recursive_comparator = [&](const TokenSequence::const_iterator l,
                                   const TokenSequence::const_iterator r) {
+    // Some token enums differ only by surrounding whitespace (e.g. whether a
+    // macro or ')' ends a line). Treat those pairs as matching enums when the
+    // spelling is unchanged so FormatEquivalent tolerates re-wrapping.
+    const bool whitespace_dependent_enum_match =
```
hzeller: here, you're combining some parts that in part are later repeated in line 271.

So I'd suggest to break that out and give a good name (... naming is hard so here just a placeholder name in this review comment :) )

```
const bool some_good_name_for_this_condition = 
     ((l->token_enum() == verilog_tokentype::MacroIdentifier ||
           l->token_enum() == verilog_tokentype::MacroIdItem) &&
          (r->token_enum() == verilog_tokentype::MacroIdentifier ||
           r->token_enum() == verilog_tokentype::MacroIdItem)  && l-.>Ltext() == r->text()
```

... and then compose `whitespace_dependent_enum` out of this
```
const bool whitespace_dependent_enum = ((l->token_enum() == verilog_tokentype::MacroCallCloseToEndLin .... 
    || some_good_name_for_this_condition
```

(`whitespace_dependent_enum` can probably also have a slightly better name; I'd expect it should contain the word `macro`...)

verible/verilog/analysis/verilog-equivalence.cc
```diff
@@ -258,6 +268,16 @@ DiffStatus FormatEquivalent(std::string_view left, std::string_view right,
              (r.text() == ")"))) {
           return true;
         }
+        // MacroIdentifier vs MacroIdItem depends only on whether the macro
+        // ends the line (see POST_MACRO_ID in verilog.lex). Spelling-equal
+        // macros are format-equivalent across that reclassification.
+        if ((l.token_enum() == verilog_tokentype::MacroIdentifier ||
```
hzeller: ... and then here also use that `some_good_name_for_this_condition`

hzeller (commented):
Thanks!
Mostly the current suggestions are about better naming and re-using an expression (and also giving it a name).

verible/verilog/analysis/verilog-equivalence.cc
kbrunham-intel: I've proposed a fix in the update code review.

verible/verilog/analysis/verilog-equivalence.cc
kbrunham-intel: I've proposed a fix in the update code review.

kbrunham-intel:
@hzeller : Please review this updated code and merge if acceptable. My goal was to adopt your feedback with this new common function for checking the macros. I don't have write access, so I can't merge even though this PR is approved.

hzeller:
Thanks, this looks good.
We really need to de-tangle the formatter_test.cc so that we have less conflicts. Can you rebase, so that we can resolve this ?

kbrunham-intel:
@hzeller : This merge conflict is ready to go now.

hzeller (approved):

hzeller:
Detangling and breaking up in different files according to some criterialis hard as one would as it is hard to find a classification that always makes sense. In the past, I have worked around that to add my test to formatter related things in random places in the formatter_test, not necessarily at the end, so that it is easy to not accidentally run into a collision with other pull requests that also do formatting changes.

Breaking the formatting examples into many small files and then ingest all of these pre/post-formatted files could be a solution, but then this makes it really hard to see in an overview.

Roping in @fangism if he has thought about it how to deal with > 20'000 lines of [formatter_test](https://github.com/chipsalliance/verible/blob/master/verible/verilog/formatting/formatter_test.cc)

PR #2548, opened 2026-07-17
Fix formatter non-convergence for multi-line macro sums

### LINKED ISSUE #2547: Formatter: non-convergence wrapping a multi-line sum of long macros in a localparam
verible-verilog-format fails to converge when a localparam is initialized to a sum of several long `MACRO identifiers written across lines. The first format pass puts each + and macro on its own line; the second pass joins the final + onto the last macro, so format(format(x)) != format(x).

**Test case**

```systemverilog
module m;
  localparam N =
      `MACRO_GEN3_SCRAMBLE_LFSR_REGOUT
      + `MACRO_GEN3_SCRAMBLE_REGIN
      + `MACRO_GEN3_SCRAMBLE_REGOUT
      ;
endmodule
```

**Actual output**

```bash
```

kbrunham-intel:
## Summary

- Fixes #2547: non-convergence when a `localparam` is initialized to a multi-line sum of long `` `MACRO `` identifiers (`format(format(x)) != format(x)`).
- Infix `+` partitions under `kParamDeclaration` are always attached to the following operand, so partition shape no longer depends on original newlines.
- Treats spelling-equal `MacroIdentifier` / `MacroIdItem` (and `MacroCallCloseToEndLine` / `)`) as format-equivalent via shared helper `TokensAreWhitespaceDependentFormatEquivalent()`, so verification tolerates macros moving relative to `;`.
- Adds regression test `LongMacroSumLocalparamConverges`.

Fixes #2547

## Overlap with #2545

This PR includes the `MacroIdentifier`/`MacroIdItem` `FormatEquivalent` change also proposed in #2545. If #2545 merges first, those equivalence-file hunks should be dropped from this PR (or this PR rebased) to avoid duplicate commits. The tree-unwrapper infix-operator attachment is unique to this fix.

## Test plan

- [x] Rebased on current `master` (merge conflict in `formatter_test.cc` resolved)
- [x] `bazel build -c opt //...` and `bazel test -c opt //...` (608/608)
- [x] Reproduce from #2547 formats and converges
- [ ] CI on this PR is green


kbrunham-intel:
**Note on overlap with #2545:** this PR also includes treating spelling-equal `MacroIdentifier` / `MacroIdItem` as format-equivalent (needed so moving `;` onto the last macro in the sum does not fail lexical verification). That change is the same as in #2545.

If #2545 lands first, please drop or rebase away the `verilog-equivalence.cc` / `verilog-equivalence_test.cc` hunks here—the remaining tree-unwrapper change (`AttachInfixOperatorsToFollowingOperands` for `kParamDeclaration`) is the part unique to #2547.

hzeller:
I've changed the Mac CI, can you rebase and see if that resolves that CI failure ?

verible/verilog/analysis/verilog-equivalence.cc
```diff
@@ -167,11 +167,21 @@ DiffStatus LexicallyEquivalent(
   DiffStatus diff_status = DiffStatus::kEquivalent;
   auto recursive_comparator = [&](const TokenSequence::const_iterator l,
                                   const TokenSequence::const_iterator r) {
+    // Some token enums differ only by surrounding whitespace (e.g. whether a
+    // macro or ')' ends a line). Treat those pairs as matching enums when the
+    // spelling is unchanged so FormatEquivalent tolerates re-wrapping.
+    const bool whitespace_dependent_enum_match =
```
hzeller: (similar to the other PR).
Maybe we should have some predicate function getting left/right const Token references and return the context-dependent result ? TODO: find good name for that predicate function)

hzeller:
Just had a quick look, but maybe we should rebase first so that we can get the formatter test unconflicted.

verible/verilog/analysis/verilog-equivalence.cc
kbrunham-intel: I've created a new predicate function to help here.

kbrunham-intel:
HI @hzeller  Please review this update and let me know if there is any more feedback.

hzeller (approved):
Thanks!

hzeller (approved):
Thanks!

PR #2550, opened 2026-07-19
Exempt unique0 case statements from case-missing-default

### LINKED ISSUE #2413: unique0 support missing

aishasalim:
The case-missing-default rule suppressed its violation only for the `unique` qualifier, so `unique0 case` was reported as missing a default even though the qualifier explicitly permits zero matching case items.

TK_unique0 was already handled by the lexer and parser; only the HasUniqueQualifier matcher was missing it. Add a separate HasUnique0Qualifier matcher rather than widening HasUniqueQualifier, so the matcher name keeps matching the token it looks for, and check both in the rule. Update the violation message and rule description, which previously mentioned only `unique`.

Adds rule-level tests for `unique` and `unique0` across case/casex/casez, including nesting and unqualified control cases; the rule had no qualifier tests before. Also tests that neither matcher matches the other's token.

Fixes #2413

### DIFF
--- a/verible/verilog/CST/verilog-matchers.h
+++ b/verible/verilog/CST/verilog-matchers.h
@@ -437,6 +437,17 @@ inline constexpr auto HasDefaultCase =
 //     ...
 inline constexpr auto HasUniqueQualifier =
     verible::matcher::MakePathMatcher(L(TK_unique));
+
+// Matches the `unique0` qualifier, e.g.
+//
+//   unique0 case (in)
+//     1: return 0;
+//   endcase
+//
+// Unlike `unique`, `unique0` permits zero matching case items, so it is
+// tracked separately from HasUniqueQualifier.
+inline constexpr auto HasUnique0Qualifier =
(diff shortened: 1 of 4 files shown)


aishasalim:
Closing as a duplicate of #2527 (which predates this). I've offered test coverage there — nothing further needed here.

PR #2557, opened 2026-08-02
Moderately update external deps.

hzeller:


hzeller:
Darn. Protobuf really should fix their warnings.

PR #2563, opened 2026-08-07
Updated flex/bison toolchain for bazel 8+ compat

corco:
This is an alternative to #2520 to make bison/flex rules works on bazel 8 and beyond.

Bazel 7 allowed to use select on a toolchain, which is not allowed in bazel 8.

The command line option `//bazel:use_local_flex_bison_enabled` works on my machine as is. From test in #2460 it needs the extra_env to work on Windows in CI environment.

Disclaimer: this code is from an AI agent


hzeller (approved):
Thanks, this looks better than what I came up with.

PR #2564, opened 2026-08-07
Update code for msvc compatibility

corco:
This allow forward compatibility with MSVC on Windows, which supports should come once the project is migrated to bazel 8/9.

Disclaimer: The fix to `tree-operations.h` and `tree-utils_test.cc` was done by AI.


hzeller (approved):
Thanks!

PR #2572, opened 2026-08-11
Fix formatting of `interface class`

wsnyder:
This fixes the formatting of `interface class` to properly be similar to `class`.


hzeller (approved):
Thanks!

PR #2573, opened 2026-08-12
Fix define name concatenation

wsnyder:
This fixes a bug where

    `define FOO```BAR

was misformatting to 

    `define FOO ```BAR

which breaks the meaning.

Assisted with Claude, but human rewrote and reviewed.


hzeller (approved):
Thanks!

PR #2574, opened 2026-08-12
kythe: Add --output_path to write extracted facts to a file

### LINKED ISSUE #736: [kythe] Output to file capability
## Input test case
Any sv file.

As far as I can tell, the `verible-verilog-kythe-extractor` tool cannot directly write its output into a file.
The issue is that it force having a potentially huge buffer in memory. 

It would be nice to have the capability to output the result of the kythe extraction directly to an output file without having to rely on stdout capture.

aishasalim:
Fixes #736.

`verible-verilog-kythe-extractor` could only write Kythe facts to stdout, so every caller had to capture the stream. This adds `--output_path` to name a file to write to instead.

```
verible-verilog-kythe-extractor --file_list_path files.txt --output_path facts.json
```

An empty value or `-` keeps the current stdout behavior, so nothing changes for existing callers.

### One note on the issue

The issue says stdout-only output "force[s] having a potentially huge buffer in memory". That part is no longer true, and may not have been true when it was filed. `KytheFactsPrinter::PrintJsonStream` already emits one fact at a time to the stream, and the proto mode already streamed to a file descriptor. So this PR does not fix a memory problem; it only closes the missing-flag gap that the issue title asks for.

### Implementation

All four print modes now share one `std::ostream`. To let the proto mode join them, `KytheProtoOutput` takes a `std::ostream` and wraps it in an `OstreamOutputStream`, instead of taking a file descriptor and wrapping it in a `FileOutputStream`. The old `int fd` constructor had a single caller, so I removed it rather than leave it unused. Happy to keep it if you prefer.

The output file is opened in binary mode, because `--print_kythe_facts=proto` is not text. The tool logs an error and exits 1 when it cannot open the file.

This also removes the `_WIN32` / `STDOUT_FILENO` block, which is now unused.

### Testing

New cases in `verilog_kythe_extractor_test.sh`:

- Each of `json`, `json_debug` and `proto` writes a non-empty file through `--output_path`, and `cmp` confirms the bytes match what the same run writes to stdout. This covers the binary proto stream.
- `--output_path -` still writes to stdout.
- An unwritable path exits 1 and reports the failure.

`bazel test //verible/verilog/tools/kythe/...` passes, 5/5.

### DIFF
--- a/verible/verilog/tools/kythe/verilog-kythe-extractor.cc
+++ b/verible/verilog/tools/kythe/verilog-kythe-extractor.cc
@@ -37,6 +37,11 @@
 #include "verible/verilog/tools/kythe/kythe-facts.h"
 #include "verible/verilog/tools/kythe/kythe-proto-output.h"
 
+#ifdef _WIN32
+#include <fcntl.h>
+#include <io.h>
+#endif
+
 // for --print_kythe_facts flag
 enum class PrintMode {
   kJSON,
@@ -174,6 +179,13 @@ static std::vector<absl::Status> ExtractTranslationUnits(
 }  // namespace verilog
 
 int main(int argc, char **argv) {
+#ifdef _WIN32
+  // Windows messes with newlines by default. Fix this here, so that stdout
+  // carries the same bytes as --output_path, and so that the proto entries
+  // stay binary.
+  _setmode(_fileno(stdout), _O_BINARY);
+#endif
+
   const auto usage =
       absl::StrCat("usage: ", argv[0], " [options] --file_list_path FILE\n", R"(
 Extracts kythe indexing facts from the given SystemVerilog source files.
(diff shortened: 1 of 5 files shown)


verible/verilog/tools/kythe/verilog-kythe-extractor.cc
```diff
@@ -109,14 +106,18 @@ if "A.sv" exists in both "directory1" and "directory2" the one in "directory1" i
 ABSL_FLAG(std::string, verilog_project_name, "",
           "Verilog project name to use as Kythe corpus. Optional");
 
+ABSL_FLAG(std::string, output_path, "",
```
hzeller, line 114: To be backward compatible, maybe make the default `"-"` ?

hzeller (commented):
Can you have a look at the windows build issue ? Probably some CRLF vs LF sitution.

aishasalim:
You were right, it was CRLF.

Windows opens stdout in text mode, so the stdout capture in the test had `\r\n` line endings while the `--output_path` file, which is opened in binary mode, had `\n`. The byte comparison then failed on the first mode it tried (`json`).

Fixed by setting stdout to binary mode in `main()`, the same way `verible-verilog-format` and `verible-verilog-obfuscate` already do. Kythe facts are data rather than text, so the output should be identical on every platform.

That also repairs `--print_kythe_facts=proto` to stdout on Windows: the old `FileOutputStream(STDOUT_FILENO)` path wrote through a text-mode fd, so it injected `\r` bytes into the proto stream.

hzeller:
Thanks!

PR #2575, opened 2026-08-13
Fix `input real` and similar module declarations

wsnyder:
This fixes syntax errors with

    module foo;
       input real sig;

and similar other types, strings etc.

Assisted by claude but manually improved & reviewed.


hzeller:
Do we also need to check that the formatter properly deals with these ?

wsnyder:
It does format properly so I didn't add a test, but will if you prefer.

hzeller:
Yes, I think it will keep things robust as there are sometimes formatting decisions made depending on tokens or node count, so that way we ensure things don't break on modifications in the future.
A simple test with formatting of these constructs would be sufficient.

hzeller (approved):
Thanks!

PR #2578, opened 2026-08-18
Fix formatter dropping function ports when packed dims contain $clog2()

### LINKED ISSUE #886: Formatting long function header removes part of code
**Test case**

```systemverilog
package foo;
  function some_large_return_type [$clog2(some_large_contant_name)-1:0] f_some_long_function( input int parameter_1, input int parameter_2);
    return 1;
  endfunction
endpackage

```

Formatting with default options gives:

**Actual output**

```systemverilog
package foo;
  function some_large_return_type [$clog2(some_large_contant_name)-1:0] f_some_long_function(
    return 1;
  endfunction
endpackage

```

There are no errors produced, but the arguments to the function have been removed.
It only happens when the function return type contains the `$clog2()` call.

This behaviour is seen in version 0.0.1320 and is not present in 0.0.1261. 

```systemverilog
package foo;
  function some_large_return_type [$clog2(some_large_contant_name)-1:0] f_some_long_function(
      input int parameter_1,
      input int paramter_2);
    return 1;
  endfunction
endpackage

```

kbrunham-intel:
## Summary
- Function headers whose return packed dimensions contain a system call such as `$clog2(...)` used to lose their port list: `ReshapeFittingSubpartitions` expected `[header, args, trailer]` and treated extra `$clog2` fragments as the args, discarding the real ports.
- Collapse those extra header fragments before reshaping so the ports are kept and wrapped normally.
- Adds formatter regression tests for the original #886 sample, `$bits()`, multi-arg `$clog2()`, and `extern` prototypes.

Fixes #886

## Test plan
- [x] `bazel test -c opt //verible/common/formatting:token-partition-tree_test //verible/verilog/formatting:formatter_test`
- [x] Original #886 sample keeps both function arguments
- [ ] CI on this PR is green

### DIFF
--- a/verible/common/formatting/token-partition-tree.cc
+++ b/verible/common/formatting/token-partition-tree.cc
@@ -783,6 +783,41 @@ static AppendFittingSubpartitionsResult AppendFittingSubpartitions(
 //
 // When "subpartitions" group has kAlwaysExpand policy, line break is forced
 // between each subpartition from the group.
+// kAppendFittingSubPartitions expects [header, args] or [header, args,
+// trailer]. Packed dimensions with $clog2(...) (issue #886) can split the
+// header into extra sibling leaves; if those are left in place, the port
+// list is treated as a trailer and dropped.
+// Only merge leading *leaf* fragments. Nested argument lists (non-leaves)
+// are flattened only when another non-leaf (the real port list) follows.
+static int CountNonLeafChildren(const TokenPartitionTree &node) {
+  int n = 0;
+  for (const auto &child : node.Children()) {
+    if (!is_leaf(child)) ++n;
+  }
+  return n;
+}
+
+static void CollapseHeaderFragmentsBeforeArgs(TokenPartitionTree *node) {
+  while (node->Children().size() > 2) {
+    auto &children = node->Children();
+    auto &first = children[0];
+    auto &second = children[1];
+    // Merge extra header leaves only when a nested argument list still
+    // follows.  All-leaf trees are the flattened one-argument form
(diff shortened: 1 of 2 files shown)


kbrunham-intel:
Hi @hzeller. Please consider this change which fixes issue 886.

kbrunham-intel:
I plan to merge #2602 first

kbrunham-intel:
@hzeller this is ready for merge

verible/common/formatting/token-partition-tree.cc
```diff
@@ -783,6 +783,40 @@ static AppendFittingSubpartitionsResult AppendFittingSubpartitions(
 //
 // When "subpartitions" group has kAlwaysExpand policy, line break is forced
 // between each subpartition from the group.
+// kAppendFittingSubPartitions expects [header, args] or [header, args,
+// trailer]. Packed dimensions with $clog2(...) (issue #886) can split the
+// header into extra sibling leaves; if those are left in place, the port
+// list is treated as a trailer and dropped.
+// Only merge leading *leaf* fragments. Nested argument lists (non-leaves)
+// are flattened only when another non-leaf (the real port list) follows.
+static int CountNonLeafChildren(const TokenPartitionTree &node) {
```
hzeller, line 792: could this be formulated with `std::count_if` ?

kbrunham-intel: I've proposed an updated implementation. Let me know your thoughts.

kbrunham-intel:

hzeller:
Thanks!

PR #2585, opened 2026-08-26
formatter: avoid crash on attribute instance in for-loop header

EylonKrause:
## Problem

`verible-verilog-format` aborts with a `CHECK` failure (SIGABRT, exit 134)
when an attribute instance `(* ... *)` appears in a for-loop header.
`TreeUnwrapper::ReshapeTokenPartitions`' `kForSpec` case assumes the header
always produces two semicolon-leading partitions:

```cpp
CHECK(iter1 != children.end());
...
CHECK(iter2 != children.end());
```

When an attribute instance appears in the init/condition, the partition
structure differs and a semicolon-leading partition is missing, firing the
CHECK and aborting the process on otherwise valid input.

### Minimal repro

```systemverilog
module m; initial for(int i=0(* a *);;) ; endmodule
module m; initial for(int i=0(* a *);i<4;i++) x=i; endmodule
```

Before this change both abort:

```
F tree-unwrapper.cc:2996] Check failed: iter2 != children.end()
... SIGABRT (exit 134)
```

## Fix

Replace the two unconditional `CHECK`s with graceful early-outs: when a
required semicolon-leading partition is not found, leave the partitions
unreshaped instead of aborting. Normal for-loops (both partitions present)
reshape exactly as before.

After the change the same inputs format without aborting, e.g.:

```systemverilog
module m;
  initial for (int i = 0 (* a *); i < 4; i++) x = i;
endmodule
```

## Testing

- `bazel test -c opt //verible/verilog/formatting:tree-unwrapper_test //verible/verilog/formatting:formatter_test` — both pass.
- Added a formatter regression test for the attribute-in-for-header input.
- Verified a normal for-loop (`for (int i=0; i<4; i++)`) formats unchanged.

### DIFF
--- a/verible/verilog/formatting/formatter_test.cc
+++ b/verible/verilog/formatting/formatter_test.cc
@@ -4707,6 +4707,15 @@ static constexpr FormatterTestCase kFormatterTestCases[] = {
      "  for (int i = 0; i < f(m); i--) begin\n"
      "  end\n"
      "endfunction\n"},
+    {// for loop with an attribute instance in the initializer.
+     // Regression: this used to abort with a CHECK failure while reshaping
+     // the kForSpec partitions when an attribute appears in the header.
+     "module m; initial for(int i=0(* a *);i<4;i++) x=i; endmodule",
+     "module m;\n"
+     "  initial\n"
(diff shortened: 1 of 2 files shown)


hzeller (approved):
thanks!

PR #2586, opened 2026-08-27
verilog: make generate-label-prefix configurable via style_regex

### LINKED ISSUE #2288: how to change lint rule "legacy-generate-region"?

Talha-Dmr:
Fixes #2288

Allow users to configure generate-label-prefix via style_regex (default: (g_|gen_).*). Previously hardcoded to g_/gen_. Uses RE2 FullMatch and config-utils SetRegex like enum-name-style. Default preserves existing behavior, now users can set e.g. --rules_config=generate-label-prefix:style_regex=foo_.*

Test: existing generate-label-prefix-rule_test still passes, new regex tested manually. BUILD deps updated for re2/config-utils.

### DIFF
--- a/verible/verilog/analysis/checkers/BUILD
+++ b/verible/verilog/analysis/checkers/BUILD
@@ -817,6 +817,7 @@ cc_library(
         "//verible/common/analysis:syntax-tree-lint-rule",
         "//verible/common/analysis/matcher",
         "//verible/common/analysis/matcher:bound-symbol-manager",
+        "//verible/common/text:config-utils",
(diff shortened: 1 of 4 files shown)


hzeller (approved):
Nice, thank you!

Issue #2568, opened 2026-08-10
[BUG] Add error on binary use of unary reduction operator ~| and ~&


MrCookieeeee:
**Verible verision**
commit: 7807ee1

**Describe the bug**

Verible accepts the expression `var1 ~| var2` and `var1 ~& var2` without reporting a syntax error.
In SystemVerilog, ~| is a unary reduction NOR operator, not a valid binary infix operator.
Therefore, var1 ~| var2 should be rejected, but Verilator appears to parse and accept it.

**To Reproduce**

```systemverilog
module bug_reduction_nor_binary (
    input  logic [3:0] var1,
    input  logic [3:0] var2,
    output logic [3:0] out
);

    // BUG: ~& and ~| is a unary reduction operator, not a binary infix operator
    assign out = var1 ~& var2;
    assign out = var1 ~| var2;
    
endmodule
```
***command***
`verible-verilog-syntax minimal.sv`

**Actual behavior:**

Nothing happen, no warnings or errors.

**Expected behavior**

decline this syntax

raspbeep:
Hi, I’ve reproduced this issue and traced it to the parser grammar. I have a small fix with regression tests for the invalid binary uses of ~& and ~|, while keeping the valid unary reduction forms intact.

Could I be assigned to this issue? I’d be happy to submit a PR.

hzeller:
It looks like an oversight that Verilog does not support it :)
Anyway, given that most tools reject the syntax, not allowing it in Verible is probably also good. @fangism do you remember if there was a tool that accepted this as binary operands and thus it was supported in Verible ? Otherwise removing it like in #2590 is probably a good idea.

(we should have a concept of 'soft reject'. So that we don't get a hard syntax error that stops at that point but continues from there to be able to catch more issues. For another time)

PR #2590, opened 2026-08-31
Fix binary use of reduction NAND and NOR

raspbeep:
Addresses #2568 by removing the binary operator grammar rules for `~&` and `~|`, which are valid only as unary reduction operators. Adds regression tests to verify that their invalid binary usage is rejected.

### DIFF
--- a/.github/bin/smoke-test.sh
+++ b/.github/bin/smoke-test.sh
@@ -171,9 +171,9 @@ ExpectedFailCount[lint:black-parrot]=155
 ExpectedFailCount[project:black-parrot]=172
 ExpectedFailCount[preprocessor:black-parrot]=173
 
-ExpectedFailCount[syntax:ivtest]=116
-ExpectedFailCount[lint:ivtest]=116
-ExpectedFailCount[project:ivtest]=145
+ExpectedFailCount[syntax:ivtest]=123
+ExpectedFailCount[lint:ivtest]=123
+ExpectedFailCount[project:ivtest]=152
 ExpectedFailCount[preprocessor:ivtest]=26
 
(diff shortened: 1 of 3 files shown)


raspbeep:
Updated the ivtest smoke-test expectations to account for the 7 files that intentionally use Icarus Verilog’s non-standard binary `~&` / `~|` operators. These files were accepted before this fix and are now correctly rejected, so the parser-dependent failure counts increase by 7.

.github/bin/smoke-test.sh
```diff
@@ -171,9 +171,9 @@ ExpectedFailCount[lint:black-parrot]=155
 ExpectedFailCount[project:black-parrot]=172
 ExpectedFailCount[preprocessor:black-parrot]=173
 
-ExpectedFailCount[syntax:ivtest]=116
-ExpectedFailCount[lint:ivtest]=116
-ExpectedFailCount[project:ivtest]=145
+ExpectedFailCount[syntax:ivtest]=123
```
hzeller, line 174: I wonder if the test that is now failing was using that as a negative example or if iverilog indeed supports this as binary operator, did you dig into the example ?

hzeller:
I do like to support a super-set of Verilog if there is any verilog parser that supports this particular feature (but I don't remember what original was supporting it, maybe @fangism knows).

The problem with taking it out of the grammar is, that it then will be a hard failure and prevent the parser to recover after that. Ideally we would like to introduce some 'soft' failure where we parse this, but fail with an error but then continue parsing to find potentially other issues. I need to think about the details, so will leave this PR open for now while that is settled.

PR #2591, opened 2026-09-01
Migrate from the deprecated ABSL_ATTRIBUTE_UNUSED attribute to C++17 [[maybe_unused]]

fangism:
Imported automatically from internal Piper robot CL 972730562.

Please review and merge these changes on GitHub.

### DIFF
--- a/verible/verilog/analysis/symbol-table_test.cc
+++ b/verible/verilog/analysis/symbol-table_test.cc
@@ -73,8 +73,7 @@ static std::ostream &operator<<(std::ostream &stream,
   const auto found_##dest(map.find(key)); /* iterator */ \
   ASSERT_NE(found_##dest, map.end())                     \
       << "No element at \"" << key << "\" in " #map;     \
-  const auto &dest ABSL_ATTRIBUTE_UNUSED(                \
-      found_##dest->second); /* mapped_type */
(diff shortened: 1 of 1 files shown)


hzeller:
Can you run clang-format on this (`.github/bin/run-format.sh`) ?

I think the default Google configuration changed to have pointers and references always left-hugged instead of auto-determine (the style guide changed to promote the wrong way to format C pointers/references), and this is why this Rosie CL probably got auto-formatted this way.

I wonder if we can get the settings somewhere in METADATA to make sure things don't get messed up ?
Well, you, I can't :)
https://github.com/chipsalliance/verible/blob/master/.clang-format

fangism:
> Can you run clang-format on this (`.github/bin/run-format.sh`) ?
> 
> I think the default Google configuration changed to have pointers and references always left-hugged instead of auto-determine (the style guide changed to promote the wrong way to format C pointers/references), and this is why this Rosie CL probably got auto-formatted this way.
> 
> I wonder if we can get the settings somewhere in METADATA to make sure things don't get messed up ? Well, you, I can't :) https://github.com/chipsalliance/verible/blob/master/.clang-format

working on scripts to automate the transformation on my end...

hzeller:
thanks!

PR #2595, opened 2026-09-08
Add --class_parameter_space (#2571)

### LINKED ISSUE #2571: Disable space before class name parameters
@aishasalim, #850 #2551 changed the formatting of

`someclass#(.params)` to `someclass #(params)`

Can you please make a PR to add a flag to make this change optional?  I attempted an upgrade, and disagree with its attempt to change this by default in my projects.

1. The data type includes the parameters so the space here makes it harder to read IMO.
2. In C++ templated classes one typically says `someclass<param>` not `someclass <param>`
3. IEEE in 12 of 12 examples omits the space e.g. ` typedef Base#(32) MyBaseT;`
4. UVM in 65 of 111 examples omits the space.  (So small majority - but at least shows it's opinion versus common style to have the space).
. 
Thanks

wsnyder:
Fixes #2571.

Adds --typedef_parameter_space to make `typedef classname #(...)` indent optional and `typedef classname#(...)` the default.

Claude was used to assist this PR, all changes were reviewed by hand and cleaned up.

### DIFF
--- a/verible/verilog/formatting/format-style-init.cc
+++ b/verible/verilog/formatting/format-style-init.cc
@@ -140,6 +140,13 @@ ABSL_FLAG(bool, compact_indexing_and_selections, true,
           "Use compact binary expressions inside indexing / bit selection "
           "operators");
 
+ABSL_FLAG(bool, class_parameter_space, false,
+          "If true, keep/insert a space before '#' in a class"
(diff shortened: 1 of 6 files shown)


wsnyder:
Renamed to `class_parameter_space` as think that better describes it, and might later use it other places too.

hzeller (approved):
Thanks. That should've been configurable to begin with.

PR #2601, opened 2026-09-10
formatter: Preserve operator spacing in ternary expressions inside subscript brackets

### LINKED ISSUE #2597: formatter: literals within bracket[]-accesses are emitted without whitespace causing formatter validation failures
**Test case**

```systemverilog
module foo();
   assign a = d[ b > 1'h0 ? 1'h0 : c ];
endmodule
```

This breaks with a message

```
verible-fmt.sv: Error lex/parsing-ing formatted output.  Please file a bug.
```
The problem is, that it smushes together the hex literals without any space and then the parser starts stumbling with a syntax error `First error: token: "'h" at 2:24-25:`

```systemverilog
module foo ();
  assign a = d[b>1'h0?1'h0 : c];
endmodule
```
This should preserve the whitespace.

It seems that this is a problem just within brackets `[]` as the following formats nicely including whitepace.

```systemverilog
module foo ();
  assign a = b > 1'h0 ? 1'h0 : c;
endmodule
```

hzeller:
`InRangeLikeContext()` previously matched all tokens nested inside subscript
dimension nodes (`kDimensionScalar`), causing compact indexing rules to strip spaces
around operators even within ternary condition expressions (`kConditionExpression`).

This produced invalid syntax like `some_array[b>1'h0?1'h0 : c]`.

Exclude `kConditionExpression` in `InRangeLikeContext()` so operators inside
ternary expressions within subscript brackets retain normal spacing rules.
    
Verified #2597 with new test case `FormatterEndToEndTest.TernaryInsideSubscriptExpression`

Fixes #2597


hzeller:
Disclaimer: AI tools were involved in analyzing the problem

hzeller:
I wait for #2602 to finish to then put this in the right test file.

PR #2602, opened 2026-09-11
Split formatter_test.cc into concern-based test files

kbrunham-intel:
## Summary
- Split the ~21k-line `formatter_test.cc` into thematic `cc_test` targets (macros, modules, class/package, function/task, align, wrap/style, disable, issue regressions) plus a thin core smoke/`VerifyFormatting` file.
- Add shared `formatter-test-utils.h` (`FormatterTestCase`, `RunFormatterTestCases*`) so files do not duplicate the FormatVerilog loop.
- Document that **new GitHub-issue formatter regressions go in `formatter_issue_regression_test.cc` only**, so open PRs stop colliding at one EOF hotspot.

Fixes #2579

## Test plan
- [x] `bazel test -c opt //verible/verilog/formatting:formatter_test //verible/verilog/formatting:formatter_macro_test //verible/verilog/formatting:formatter_module_test //verible/verilog/formatting:formatter_class_package_test //verible/verilog/formatting:formatter_function_task_test //verible/verilog/formatting:formatter_align_test //verible/verilog/formatting:formatter_wrap_style_test //verible/verilog/formatting:formatter_disable_test //verible/verilog/formatting:formatter_issue_regression_test //verible/verilog/formatting:formatter-tuning_test`
- [ ] CI on this PR is green
- [ ] After merge, rebase open formatter PRs (`fix/2352`, `fix/2539`, `fix/886`, `fix/2547`) and place new TESTs in `formatter_issue_regression_test.cc`

PR #2603, opened 2026-09-12
Fix prefix and event trigger operator missing spaces

wsnyder:
Fixes missing space in e.g. `while (coyote)++super_genius;`

Assisted with Claude, manually updated and reviewed.


wsnyder:
Likewise `#1 ->event_trigger` which hit same code spots.

wsnyder:
@hzeller it appears that master has failing tests (https://github.com/chipsalliance/verible/actions/runs/34693186126).  When master is fixed I'll merge that in and rerun tests.

hzeller:
I've fixed that, so after rebase, this should work.

However, after that I want first #2602 to land (breaking up formatter test so that we have less likely conflicts during pull requests and also more manageable file size), as it affects formatter_test.cc, and you then need to rebase again to fix in the new location (if we merged now, then 2602 would have a hard time to keep up)

hzeller:
alright, #2602 is merged. When rebasing you need to put the test into one of the new formatter test files.
Sorry for the disruption, but that test became huge, so needed to be split.

hzeller:
Thanks!

PR #2604, opened 2026-09-13
Temporarily disable test.

hzeller:
A test introucded in #2532 is not triggering as expected. Disabling tempoarily.


hzeller:
CC: @EylonKrause

PR #2606, opened 2026-09-14
Fix timescale CRLF // comments leaving CR in the token

kbrunham-intel:
## Summary
- Adds a failing-then-fixed regression for https://github.com/chipsalliance/verible/issues/2605: CRLF `//` comments after `` `timescale `` were emitted as `\r\r\n` and failed `FormatEquivalent`.
- `TIMESCALE_DIRECTIVE` now peels off a two-character CRLF terminator the same way `IN_EOL_COMMENT` already does (see #2371).
- Lexer coverage for next-line and same-line timescale comments with CRLF.

## Test plan
- [x] `FormatterEndToEndTest.TimescaleCrlfEolComment` failed on the test-only commit (`// hello\r` vs `// hello`)
- [x] `bazel test //verible/verilog/formatting:formatter_test` after the lexer fix
- [x] `bazel test //verible/verilog/parser:verilog-lexer_test` after the lexer fix


hzeller:
oh, one more regression test conflict

kbrunham-intel:

verible/verilog/parser/verilog.lex
```diff
@@ -990,7 +990,13 @@ zi_zp { UpdateLocation(); return TK_zi_zp; }
     return TK_COMMENT_BLOCK;
   }
   {EndOfLineComment} {
-    yyless(yyleng-1);  /* return \n to input stream */
+    // Match IN_EOL_COMMENT: CRLF is a two-character terminator. Using
```
hzeller, line 993: this is the second time we have to do this. Can you file an issue, referring to your two pull requests to investigate a systematic approach looking at all these and to potentially discover other issues ?

```
grep "n to input" verible/verilog/parser/verilog.lex
```

Then we can also put this repeated code-snippet into a function

kbrunham-intel: #2617 has been created to track this

hzeller: Thanks!

PR #2608, opened 2026-09-14
Fix CRLF macro definitions failing formatter convergence

### LINKED ISSUE #2607: formatter: CRLF \define lines fail to converge (blank line oscillates
**Suggested title:** formatter: CRLF `define lines fail to converge (blank line oscillates)

---

**Test case**

```systemverilog
`define A x.y
module m;
endmodule
```

Save the file with **CRLF** line endings (`\r\n`). The same text with LF-only (`\n`) formats cleanly.

```bash
verible-verilog-format --nofailsafe_success define_crlf.sv
```

Defaults otherwise (`--line_terminator=auto`).

Found after [#2605](https://github.com/chipsalliance/verible/issues/2605): `baseline_a55_tb.sv` now passes the timescale `//` lexical check and then fails here. One `` `define `` is enough; adjacent defines in that file hit the same path.

**Actual output**

Exit code 1. Formatting does not converge: pass 1 inserts a blank line after the define; pass 2 removes it.

```
`define A x.y

module m;
endmodule
```

Diagnostic:

```
...
-`define A x.y

+`define A x.y
 module m;
 endmodule
```

Stdout is the original file (fail-safe fallback). The oscillating formatted text is dumped after `problematic formatter output is`.

Keep a single CRLF after the define (same as LF input already produces):

```systemverilog
`define A x.y
module m;
endmodule
```

Bytes:

```
`define A x.y\r\n
module m;\r\n
endmodule\r\n
```

This is a line-ending / convergence correctness issue, not a style-guide preference. Distinct from #2605 (`timescale` + `//` leaving `\r` in the comment token).

kbrunham-intel:
## Summary
- add a formatter regression for CRLF `define lines that fail convergence
- return both bytes of a CRLF terminator from `PP_CONSUME_BODY`
- add lexer coverage ensuring the macro body excludes `\r`

Fixes #2607.

## Test plan
- [x] Confirmed `FormatterEndToEndTest.DefineCrlfConverges` fails on the test-only commit with `\r\r\n`
- [x] `bazel test //verible/verilog/formatting:formatter_test //verible/verilog/parser:verilog-lexer_test --test_output=errors`
- [x] `bazel build //verible/verilog/tools/formatter:verible-verilog-format`


hzeller:
Can you check the clang-tidy report ? Some include is missing.

kbrunham-intel:
I can't rerun the clangtidy workflow and don't believe the current failure is due to this change.

kbrunham-intel:

hzeller:
thanks!

PR #2609, opened 2026-09-17
Only depend on necessary dependencies.

hzeller:
Using bant, optimize the dependencies to what
is needed.

```
source <(bant dwyu ...)
```


hzeller:
FYI @kbrunham-intel this is a simple way to automatically add the dependencies.

PR #2618, opened 2026-09-20
Use designated initializers where possible.

hzeller:
Finally, in c++20, we can now use features from C99...


hzeller:
@fangism wdyt ?

PR #2620, opened 2026-09-22
formatter: do not fuse +/- inside [] into --/++

### LINKED ISSUE #2359: Verible's formatter fails to lex/parse input
**Test case**

```systemverilog
module module_0 (  )  ; 
    assign  id_9  [  -1  -  -1  ]  =  0 ;  
endmodule
```

Command used: `verible-verilog-format example.v  --inplace`

**Actual output**

```systemverilog
example.v: Error lex/parsing-ing formatted output.  Please file a bug.
First error: token: "--" at 3:17-18:; problematic formatter output is
module module_0 ();
  assign id_9[-1--1] = 0;
endmodule

<<EOF>>
```

**Version used**
```
<unknown-git-version>
Commit  2025-02-06
Built   2025-02-20T11:40:59Z
```

kbrunham-intel:
## Summary
- Compact indexing inside `[]` was emitting `[-1--1]`, which the lexer treats as decrement `--` (issue #2359).
- Require a space between consecutive `+`/`-` operators so the formatted output stays lexically valid (`[-1- -1]`).

Fixes #2359

## Test plan
- [x] `bazel test -c opt //verible/verilog/formatting:formatter_issue_regression_test`
- [x] `bazel test -c opt //verible/verilog/formatting:token-annotator_test`
- [x] `bazel test -c opt //verible/verilog/formatting:formatter_module_test`


verible/verilog/formatting/formatter_issue_regression_test.cc
```diff
@@ -522,6 +522,21 @@ TEST(FormatterEndToEndTest, PortListCommentWithLineContinuationDoesNotAbort) {
...
+
+// Regression for https://github.com/chipsalliance/verible/issues/2359:
+// Unary minus after a binary minus inside [] must not fuse into `--`.
+TEST(FormatterEndToEndTest, UnaryMinusInsideIndexDoesNotFuse) {
+  static constexpr FormatterTestCase kTestCases[] = {
+      {"module module_0 ();\n"
+       "  assign id_9[-1 - -1] = 0;\n"
+       "endmodule\n",
+       "module module_0 ();\n"
+       "  assign id_9[-1- -1] = 0;\n"
```
hzeller: the `-` directly after the first literal looks a bit funny, but I guess that is due to our usual 'compactness' inside [], so is as expected. Nothing to do here.

hzeller:
Have you seen if we already have another test that verifies that in regular expressions outside `[]` things are properly separated ? (I suspect they are as we add spaces around operators anyway, but might be worthwhile adding somewhere if not already to make sure that that case is not getting messed up at some point.

```
assign foo = -1 - -1;
```

kbrunham-intel:
Hi @hzeller I don't think we already have that case. The closest existing tests are parameter int foo = - -1  and parameter int a = --b - --c (-- tokens), not binary minus then unary minus.

Outside [], binary operators already get spaces, so -1 - -1 stays -1 - -1. I added that as a second case next to the [] regression so compact indexing cannot start eating those spaces later.

PR #2621, opened 2026-09-22
formatter: alias -i to --inplace; document failsafe exit

### LINKED ISSUE #2257: Alias -i to --inplace
Several other formatters, e.g. `clangformat`, `yapf`, `gersemi`, use the `-i` option to mean format in place.

Please consider aliasing `--inplace` to `-i`, for compatibly with those, to allow an use-without-reading-the-docs experience, thanks.

kbrunham-intel:
## Summary
- Alias `-i` to `--inplace` for clang-format / yapf familiarity (issue #2257).
- Document that `--failsafe_success` defaults to true (exit 0 even on format errors). CI should use `--failsafe_success=false` if a formatter bug must fail the job.

Fixes #2257

## Test plan
- [x] `verible-verilog-format -i file.sv` is accepted (no unknown-flag error)
- [x] `-i` works with multiple files (same as `--inplace`)


verible/verilog/tools/formatter/verilog-format.cc
```diff
@@ -86,9 +86,9 @@ static std::string AbslUnparseFlag(LineRanges /* unused */) {
                        absl::StreamFormatter());
 }
 
-// TODO(fangism): Provide -i alias, as it is canonical to many formatters
 ABSL_FLAG(bool, inplace, false,
           "If true, overwrite the input file on successful conditions.");
+ABSL_FLAG(bool, i, false, "Alias for --inplace.");
```
hzeller, line 91: IIRC correctly, the ABSL flags are a bit annoying in that even single-letter flags need the double-dash.
So if that is the case, we might need to manually 'fish out' the `-i` from the argv in main.

hzeller:
Can you add a test that verifies the `-i` is working ? `verible/verilog/tools/formatter/format_inplace_test.sh` looks like  a good place.
(In particular we want to make sure that indeed `-i` is working and does not require `--i`)

hzeller:
You agent wrote in the 'test plan' the various things to check, but it did not, in fact, add these tests in this PR...

kbrunham-intel:
I forgot about format_inplace_test.sh and manually tested. Let me add tests to format_inplace_test.sh.

kbrunham-intel:
ABSL accepts both -i and --i for a one-letter flag, so I'm leveraging that here. 

I've created a new test to ensure running the formatter with -i works as expected.

PR #2622, opened 2026-09-22
formatter: converge port lists with a blank line

### LINKED ISSUE #2182: Verible formatter fails to handle comma-at-beginning-of-line style and/or inline comments
v0.0-3454-g7fdc2876
Commit	2023-12-20
Built	2023-12-20T15:18:36Z

**Test case**

```systemverilog
module my_module (
        input   logic                           clk_pl_100 // a comment
    , input   logic                          aresetn // another comment

    , output a
    , output b
    );
endmodule
```

**Actual output**

```
Already formatted, no change.
```

OR

```
```

This fails to format the file unless any of the following is done:
1) I remove an underscore from `clk_pl_100`
2) I remove the blank line in the input list
3) I remove the inline comment `// another comment`

It's very unusual -- something is clearly tripping up verible in this simplified example.

kbrunham-intel:
## Summary
- Port lists that mix a blank-line group break with leading commas and EOL comments no longer hang-indent the first port by the width of a missing `, `.
- That extra indent made `infer` alignment choose flush-left on the second pass, so format(format(x)) != format(x) (exit 1 with `--failsafe_success=false`).

## Test plan
- [x] `formatter_issue_regression_test` `PortListBlankLineWithCommentsConverges` (issue sample)
- [x] `align_test`, `formatter_module_test`, `formatter_align_test`, `formatter_wrap_style_test`
- [x] CLI `--failsafe_success=false` on the issue snippet converges

Fixes #2182


verible/verilog/formatting/formatter_align_test.cc
```diff
@@ -177,7 +177,7 @@ static constexpr FormatterTestCase kAlignFormatterTestCases[] = {
      "endmodule\n",
      "module m;\n"
      "  foo bar (\n"
-     "        .a  (a)    //comment1\n"
+     "      .a  (a)    //comment1\n"
```
hzeller, line 180: All these changes in this file that now mis-align the enums and ports don't look like they should happen or are intended.

kbrunham-intel: Hi @hzeller This is the same change as the bugfix .

The extra two spaces on the first row were the unused leading-, column: later rows are , .aaa / , second, so the aligner padded the first row (no comma) so .a / first lined up under the identifier after the comma. 

This file already had a lot of leading-comma + comment cases, and their expected output encoded that difference. After the fix in thos PR those goldens failed, so I updated them to the new indent. 

If the old behavior was the expectation, let me revisit this change, but looking at the new changes to the formatting this seems correct.

